PR #1, opened 2026-07-06
macOS/Apple Silicon port: NEON kernels + platform shim, zero impact on Linux/x86

RDouglasSharp:
# PR: Native macOS / Apple Silicon support (NEON kernels + platform shim)

*Suggested PR title:* **macOS/Apple Silicon port: NEON kernels, platform shim, correct RAM auto-budget — zero impact on Linux/x86**

---

## Summary

This PR makes colibrì build and run at full speed on Apple Silicon Macs. Your README
asks for datapoints from machines with more RAM — Apple's 64–192 GB unified-memory
machines are probably the largest population of consumer hardware in your
extrapolation table's top rows, and before this PR the engine wouldn't build on them
(AVX2-only kernels, `/proc/meminfo`, `O_DIRECT`, `posix_fadvise`, gcc-only Makefile).

**Design rule followed throughout: the Linux/x86 path is untouched.** Every change is
either a new `#elif`/`#ifdef` branch, a new file, or an OS-detection branch in the
build scripts. Compiling on Linux produces byte-identical code to master (verified by
diffing comment-stripped preprocessor output). Your benchmarks cannot regress.

### New file: `c/compat.h` — platform shim (no-op on Linux)
- `posix_fadvise` does not exist on macOS: `WILLNEED` maps to `F_RDADVISE` (same
  readahead semantics); `DONTNEED` becomes a no-op — XNU's unified buffer cache
  self-regulates, and on a big-RAM machine keeping expert pages cached is exactly the
  desired behavior anyway.
- `O_DIRECT` does not exist on macOS: `compat_open_direct()` opens the twin fd with
  `F_NOCACHE`, preserving the `DIRECT=1` feature. `F_NOCACHE` imposes no alignment
  constraints, so the existing 4K-aligned preads remain valid.
- On Linux the header defines nothing at all.

### `c/glm.c`
- **NEON kernels** for `matmul_q` (int8), `matmul_i4` (packed int4) and `matmul_i2`
  (packed int2), added as `#elif defined(__ARM_NEON)` blocks beside the AVX2 ones.
  Same structure as your AVX2 code: dequant-on-use, per-row scale, f32 FMA
  accumulation, scalar tail loop unchanged. The AVX2 blocks are byte-for-byte
  untouched. Without these, arm64 silently fell back to the scalar path (~3× slower
  measured, likely more on M-series cores).
- **NEON for the new IDOT kernels** (`dot_i8i8`, `dot_i4i8`): native `SDOT`
  (`vdotq_s32`) under `__ARM_FEATURE_DOTPROD` — always available on Apple Silicon —
  with a `vmull_s8`/`vpadalq_s16` fallback for pre-8.2 ARM. The fallback respects the
  same anti-overflow bound as your AVX2 sign trick (pairs ≤ 128·127·2 = 32512 <
  32767). Your Q8_0-style activation quantization thus keeps its 1.4–2.5× on ARM too.
- **`mem_available_gb()`**: macOS branch using `host_statistics64`
  (free + inactive + purgeable pages ≈ `MemAvailable`). This one matters most: on
  macOS the `/proc/meminfo` read fails silently and the auto-budget assumed 8 GB —
  crippling the expert cache precisely on the machines with the most RAM. With the
  fix, a 128 GB Mac auto-sizes the cache to ~100 GB.
- **`rss_gb()`**: `ru_maxrss` is bytes on macOS vs KB on Linux.

### `c/st.h`
- Includes `compat.h`; the `O_DIRECT` twin-fd open is now guarded:
  `O_DIRECT` on Linux / `F_NOCACHE` twin on macOS / `-1` elsewhere.

### `c/iobench.c`
- Same `O_DIRECT`→`F_NOCACHE` guard; `<omp.h>` include gated on `_OPENMP`
  (Apple clang without libomp still compiles it, single-threaded).

### `c/Makefile`
- `uname -s` branch. Darwin: clang, auto-detects Homebrew `libomp`
  (`brew --prefix libomp`), warns and builds single-threaded if absent, no `-march`
  (NEON is baseline on arm64, the `__ARM_NEON` blocks self-select). The Linux branch
  is character-identical to master.

### `c/setup.sh`
- Darwin-aware dependency checks (clang instead of gcc, libomp hint,
  `sysctl hw.memsize`/`hw.ncpu` instead of `/proc`/`nproc`). Linux path unchanged.

## Validation

**On Apple Silicon (M5 Max, 128 GB, macOS) — real hardware:**

| test | result |
|---|---|
| `SNAP=./glm_tiny TF=1 ./glm 64 16 16` (teacher-forcing vs transformers oracle) | **32/32** |
| `SNAP=./glm_tiny ./glm 64 16 16` (greedy decode + KV cache) | **20/20** |
| `./glm 64 4 4` vs `NOPACK=1 ./glm 64 4 4` (packed-int4 NEON kernel vs same values in int8 container) | **identical 20/20 tokens** |

**On Linux/aarch64 (CI-style checks, run against this branch rebased on 8d7fbef):**

- Kernel unit test: NEON `matmul_q`/`matmul_i4`/`matmul_i2` vs an independent
  double-precision dequant reference, sizes I ∈ {16, 17, 48, 253, 256, 1536} (odd
  sizes exercise the tail loops): int4/int2 bit-exact, int8 max rel-err 1.9e-06
  (float reassociation only). Test file available on request / can be added to the PR.
- IDOT unit test: `dot_i8i8`/`dot_i4i8` vs an int64 scalar reference, sizes
  I ∈ {16, 17, 31, 32, 48, 253, 256, 1536}: **exact integer match** on both ARM
  paths (SDOT and the vmull fallback, compiled and run separately).
- Full engine compiles clean with `gcc -O3 -fopenmp` on aarch64 — only the
  pre-existing snprintf-truncation warning, same as x86.
- Microbench, 4-thread aarch64 VM: `matmul_i4` 1536×6144 goes 7.0 → 20.5 Gmul/s
  (~3×) over the scalar fallback.

**Linux/x86 non-regression:** comment-stripped preprocessor output of all touched C
files is byte-identical to master on the Linux path; the Makefile Linux branch is
character-identical.

## How to try it on a Mac

```bash
brew install libomp
cd c && ./setup.sh          # auto-detects Darwin, builds, runs the oracle self-test
```

Real-model numbers from a 128 GB M5 Max (int4 container on internal SSD, ~6-8 GB/s
random reads) will follow as an issue comment once the 756 GB FP8 download/conversion
completes — the auto-budget already sizes the expert cache to ~100 GB there, which
should land this machine in the top rows of your extrapolation table.

## Notes / limits

- `DONTNEED` being a no-op on macOS means `DROP=1` has no effect there; on 64–128 GB
  machines that's the right trade (the page cache IS the L2).
- `supervisor.sh` (WSL network-stall watchdog) is not ported — it uses `flock`,
  absent on macOS — and it isn't needed there; the converter's own resume logic
  suffices.
- Follow-up PRs available if you're interested (kept out of this one to keep it
  reviewable): EN translations added alongside every Italian comment; an
  OpenAI-compatible server (`coli_server.py`, stdlib-only) with two backward-
  compatible serve-protocol extensions; a domain expert-usage profiler
  (`profile_experts.py`) that feeds your STATS/PIN hot-store from a source-code
  corpus via SCORE mode.

---

*Grazie per il progetto — il colibrì ora vola anche su Apple Silicon.*

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1,3 +1,24 @@
+UNAME_S := $(shell uname -s)
+
+ifeq ($(UNAME_S),Darwin)
+# --- macOS / Apple Silicon ---
+# Apple clang non include il runtime OpenMP: se c'e' libomp di Homebrew lo usa
+# (brew install libomp), altrimenti compila single-thread (i pragma omp sono ignorati).
+# Niente -march: su arm64 NEON e' baseline (i kernel __ARM_NEON si attivano da soli).
+CC      = clang
+OMPDIR := $(shell brew --prefix libomp 2>/dev/null)
+ifneq ($(OMPDIR),)
+OMPC    = -Xclang -fopenmp -I$(OMPDIR)/include
+OMPL    = -L$(OMPDIR)/lib -lomp
+else
+$(warning libomp non trovato: build SINGLE-THREAD. Per il multithread: brew install libomp)
+OMPC    =
+OMPL    =
+endif
+CFLAGS  = -O3 $(OMPC) -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function
+LDFLAGS = -lm $(OMPL)
+else
+# --- Linux x86-64 (percorso originale, invariato) ---
 CC      = gcc
 # ARCH=native -> ottimizzato per QUESTA macchina (default, piu' veloce).
 # ARCH=x86-64-v3 -> binario PORTABILE su qualsiasi x86-64 moderno con AVX2 (per distribuire).
@@ -5,10 +26,11 @@ CC      = gcc
 ARCH   ?= native
 CFLAGS  = -O3 -march=$(ARCH) -fopenmp -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function
 LDFLAGS = -lm -fopenmp
+endif
 
 all: glm
 
-glm: glm.c st.h json.h tok.h tok_unicode.h
+glm: glm.c st.h json.h tok.h tok_unicode.h compat.h
 	$(CC) $(CFLAGS) glm.c -o glm $(LDFLAGS)
 
 olmoe: olmoe.c st.h json.h
--- a/c/compat.h
+++ b/c/compat.h
@@ -1,0 +1,50 @@
+/* compat.h — shim di portabilita' per piattaforme non-Linux (oggi: macOS / Apple Silicon).
+ * Su Linux questo header e' un NO-OP totale: nessun simbolo definito o ridefinito,
+ * zero impatto sul percorso x86 esistente.
+ * Regola: ogni differenza di piattaforma vive QUI; i .c restano puliti. */
+#ifndef COMPAT_H
+#define COMPAT_H
+
+#ifdef __APPLE__
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/types.h>
+
+/* --- posix_fadvise: assente su macOS ---
+ * WILLNEED -> F_RDADVISE (readahead esplicito: stessa semantica).
+ * DONTNEED -> no-op: XNU non espone un drop mirato per-range; la sua unified
+ *             buffer cache si autoregola sotto pressione. Il motore usa DONTNEED
+ *             solo come consiglio, quindi ignorarlo e' corretto (e su una macchina
+ *             con molta RAM tenere le pagine e' proprio cio' che si vuole). */
+#ifndef POSIX_FADV_NORMAL
+#define POSIX_FADV_NORMAL      0
+#define POSIX_FADV_RANDOM      1
+#define POSIX_FADV_SEQUENTIAL  2
+#define POSIX_FADV_WILLNEED    3
+#define POSIX_FADV_DONTNEED    4
+#define POSIX_FADV_NOREUSE     5
+#endif
+static inline int compat_fadvise(int fd, off_t off, off_t len, int advice){
+    if(advice==POSIX_FADV_WILLNEED){
+        struct radvisory ra;
+        ra.ra_offset = off;
+        ra.ra_count  = (int)(len>0x7FFFFFFF ? 0x7FFFFFFF : len);
+        return fcntl(fd, F_RDADVISE, &ra)<0 ? -1 : 0;
+    }
+    return 0;
+}
+#define posix_fadvise compat_fadvise
+
+/* --- O_DIRECT: assente su macOS ---
+ * L'equivalente e' F_NOCACHE sul fd (bypass della unified buffer cache).
+ * compat_open_direct() apre il fd "gemello" senza cache, come il twin O_DIRECT
+ * di st.h. Le pread allineate a 4K del chiamante restano valide: F_NOCACHE non
+ * impone vincoli di allineamento. */
+static inline int compat_open_direct(const char *path){
+    int fd = open(path, O_RDONLY);
+    if(fd>=0) fcntl(fd, F_NOCACHE, 1);
+    return fd;
+}
+#endif /* __APPLE__ */
+
+#endif /* COMPAT_H */
--- a/c/glm.c
+++ b/c/glm.c
@@ -32,6 +32,11 @@ static inline float hsum256(__m256 v){            /* somma orizzontale di 8 floa
     lo=_mm_add_ps(lo,hi); __m128 sh=_mm_movehl_ps(lo,lo); lo=_mm_add_ps(lo,sh);
     sh=_mm_shuffle_ps(lo,lo,1); lo=_mm_add_ss(lo,sh); return _mm_cvtss_f32(lo);
 }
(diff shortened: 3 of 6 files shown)


JustVugg:
Have you tested this part?I don't have a mac or Apple something to test it!

RDouglasSharp:
Yes - the test results are mentioned in the PR md file text.  I have run
the unit tests and they pass.  Also the linux

JustVugg:
Perfect i merge now!

RDouglasSharp:
It runs!


 ./iobench /Users/doug/glm52_i4/out-00069.safetensors 19 256 8 1

O_DIRECT x8 thread: 256 letture x 19MB = 5.1 GB in 0.36s -> 14.17 GB/s (1.4
ms/blocco effettivi)

./coli chat


     ▄▀▀▀▄  ▄        *colibrì* v1.0

  ▄▄▄▄▀▀▀▀▄▀▀        piccolo motore, modello immenso

      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU

        ▀▀▀▀         chat · glm52_i4 · ram -GB · topp off

          ▀

  ──────────────────────────────────────────────────────────

  [RAM_GB=54.2 auto] cap=8 ok (proiezione picco / projected peak 25.1 GB)

  scrivi e premi invio · :reset memoria · :q esci



╭────────────────────────────────────────────────────────────────────────────────────────────────╮

  │ *›*  If I paste some c code into this chat, can you analyze it?


╰────────────────────────────────────────────────────────────────────────────────────────────────╯


  ◆ colibrì

  Yes, absolutely! Please go ahead and paste the C code here.



  Once you provide it, I can help you with things like:



  *   **Explaining what the code does:** I can break down its logic and
functionality step-by-step.

  *   **Finding bugs or vulnerabilities:** I can review it for potential
errors, memory leaks, or security flaws.

  *   **Optimizing it:** I can suggest ways to make the code more efficient
or readable.

  *   **Answering specific questions:** If you have a particular question
about how a specific function or block of code works, I can explain it.



  Just paste the code whenever you're ready, and let me know if you have
any specific goals or questions about it!

  └─ 148 tok · 0.52 tok/s · hit 25% · RSS 22.7 GB · 283s


PIN=$PWD/stats_probe.txt PIN_GB=80 ./coli chat --ram 100



     ▄▀▀▀▄  ▄        *colibrì* v1.0

  ▄▄▄▄▀▀▀▀▄▀▀        piccolo motore, modello immenso

      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU

        ▀▀▀▀         chat · glm52_i4 · ram 100GB · topp off

          ▀

  ──────────────────────────────────────────────────────────

  [PIN] hot-store: 4096 expert in RAM (77.5 GB) in 8s da/from
/Users/doug/code/colibri-main/c/stats_probe.txt · [RAM_GB=100.0]
residente/resident 87.6 GB + slack 3.6 GB (ws 1.2, ***@***.*** 0.7, kvb 0.5),
expert 18.9 MB x 75 layer -> cap abbassato/lowered 8->6 (proiezione picco /
projected peak 99.7 GB)

  scrivi e premi invio · :reset memoria · :q esci



╭────────────────────────────────────────────────────────────────────────────────────────────────╮

  │ *›* On a 128Mb unified memory M5 mac max (maximum specs), how much
faster would you expect GLM would run on the GPU versus the CPU?


╰────────────────────────────────────────────────────────────────────────────────────────────────╯


  ◆ colibrì

  To give you a precise answer, we first have to clarify one critical
detail: **GLM (whether it's ChatGLM-6B, ChatGLM2, or ChatGLM3) is currently
not officially supported by Apple’s MLX framework.**



  Because Apple Silicon Macs use a Unified Memory Architecture (UMA),
running Large Language Models (LLMs) on the GPU versus the CPU relies
entirely on the software framework you use to execute it.



  If you are using **MLX** (Apple’s native machine learning framework), it
will automatically utilize the GPU (Metal) by default. If you are using
**llama.cpp**, it will typically default to the CPU, but can be built with
Metal support.



  Here is a realistic expectation of the performance difference between GPU
and CPU execution on an M5 Max with 128GB of Unified Memory:



  ### The Short Answer:

  You can expect **GPU (Metal/MLX) execution to be roughly 2x to 4x
faster** in Tokens Per Second (tokens/sec) for prompt evaluation (prefill)
and roughly **1.5x to 2.5x faster** for token generation (decoding)
compared to running purely on the CPU.



  ### Why the GPU isn

  └─ 256 tok · 0.27 tok/s · hit 45% · RSS 58.8 GB · 954s



  Model Name: MacBook Pro

  Model Identifier: Mac17,7

  Model Number: Z1MN0017VLL/A

  Chip: Apple M5 Max

  Total Number of Cores: 18 (6 Super and 12 Performance)

  Memory: 128 GB

      Type: GPU

      Total Number of Cores: 40

      Vendor: Apple (0x106b)

      Metal Support: Metal 4

PR #9, opened 2026-07-08
IDOT: AVX-512 VNNI kernel paths + batch-size gate for single-row int4 decode (+20% end-to-end on Xeon)

Dog279:
## What this adds

Your README's scaling table ends with *"same RAM + 24–32 cores, or AVX-512/VNNI kernels — kernel work is the multiplier."* This PR is that row: guarded AVX-512 VNNI variants of the two integer-dot kernels, plus one gate change that turned out to matter more than the kernels themselves.

- `dot_i8i8` / `dot_i4i8` gain `#if defined(__AVX512VNNI__) && defined(__AVX512BW__)` branches: one `vpdpbusd` per 64 bytes (u8·s8 → s32 directly), replacing the 3-instruction `maddubs`/`madd`/`add` sequence per 32 bytes. AVX-512 has no `vpsignb`, so the sign trick becomes `abs(w)` + mask-negate on `x` — same result, `w==0` safe either way.
- In `dot_i4i8`, the 256-bit nibble unpack leaves values in per-lane order `[0-15][32-47]/[16-31][48-63]`; since dot pairing is order-invariant, the activation vector's 128-bit blocks are permuted to match (one `vpermq`) instead of re-sorting the weights.
- The int4 IDOT gate — hardcoded `S>=2` from your AVX2 mmbench finding that single rows don't pay for unpack+sign — becomes `g_i4s`: **default 1 under VNNI, unchanged 2 everywhere else**, `I4S=n` env override for A/B. This is where the speedup lives: single-row *decode* now runs the integer path. `I4S=2` restores today's behavior exactly.
- The startup banner reports the active kernel (`idot: avx512-vnni/avx2/neon/scalar`) so benchmark logs self-document.

## Correctness

- Where the integer paths already ran (int8 always; int4 at S≥2 — i.e. all prefill/scoring), outputs are **bit-identical**: s32 adds are associative and the existing bounds (pairs ≤ 32512, I ≤ 16384) hold unchanged. Verified empirically with greedy same-prompt runs.
- Single-row int4 under VNNI adopts the same int8-activation tradeoff IDOT already makes for S≥2 (~0.3% RMS per matmul, per the comment in the source). Greedy outputs on our test prompts stayed fully coherent. `IDOT=0` or `I4S=2` recover the exact f32 path.
- Non-VNNI builds are untouched: AVX2/NEON/scalar branches unmodified, `x86-64-v3` portable build verified clean, zero warnings on gcc 13.3.

## Measured

Hardware: Xeon 6 (24C/48T, AVX512-VNNI + AMX), 128GB DDR5-5600, 3× Samsung 990 Pro in VROC RAID 0 (iobench: 19.8 GB/s O_DIRECT, 8 threads), WSL2, gcc 13.3, `ARCH="native -mprefer-vector-width=512"`. Runtime config: `DIRECT=1 TOPP=0.7 cap=64`, 24 threads pinned to physical cores, greedy decoding, cold expert cache per run.

| workload | I4S=2 (old behavior) | I4S=1 (this PR's default) |
|---|---|---|
| short story, 256 tok | 1.72 tok/s (expert-matmul 41.5s) | **2.08 tok/s (+21%)** (expert-matmul 31.7s) |
| prose explainer, 128 tok | 1.45 tok/s (expert-matmul 21.3s) | **1.74 tok/s (+20%)** (expert-matmul 14.7s) |

For context, this machine's full ladder from stock defaults to this PR lands at 0.55 → 2.08 tok/s on the same workload.

Happy to follow up with: a README community-benchmark row for this hardware, MMLU/HellaSwag/ARC quality numbers (queued to run overnight on this rig), and an issue #8 datapoint — with the int8 MTP head on x86, `DRAFT=1` gets 59% acceptance and is a wash here rather than a loss (measured with these kernels active).

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -228,7 +228,17 @@ static void matmul_i2(float *y, const float *x, const uint8_t *q2, const float *
  * stile Q8_0), prodotto scalare INTERO via maddubs/madd AVX2 — niente conversione
  * f32 dei pesi nel ciclo caldo. ~2-3x sui matmul quantizzati; errore aggiunto ~0.3%
  * RMS per matmul (attivazione int8), IDOT=0 torna al percorso f32 esatto. */
+#if defined(__AVX512VNNI__) && defined(__AVX512BW__)
+#define IDOT_KERNEL "avx512-vnni"
+#elif defined(__AVX2__)
+#define IDOT_KERNEL "avx2"
+#elif defined(__ARM_NEON)
+#define IDOT_KERNEL "neon"
+#else
+#define IDOT_KERNEL "scalar"
+#endif
 static int g_idot=1;
+static int g_i4s=2;   /* min batch S for int4 IDOT; VNNI default 1 (set in main), I4S=n overrides */
 static inline float qrow_i8(const float *x, int8_t *q, int I){
     float amax=0; for(int i=0;i<I;i++){ float a=fabsf(x[i]); if(a>amax)amax=a; }
     float s=amax/127.f; if(s<1e-12f) s=1e-12f; float inv=1.f/s;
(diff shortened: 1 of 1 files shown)


JustVugg:
wow really interesting I'm finishing the last tests and doing the various merges!

Issue #19, opened 2026-07-10
Discussing Further Issues Related to Project Engineering


ZacharyZcR:
We’ll explore more issues related to the project’s engineering, including the SDK, improved architecture, a Rust rewrite, the user interface, and testing enhancements. If we’re going to build a robust project, we may need to finalize these details. We might also expand the project to incorporate technologies like JIT, making flexible and efficient use of VRAM, RAM, and disk space, and achieving further optimization through hotspot analysis and loading of large models.

If the project maintainer is interested in this, I would be happy to take charge of the integration and development. This is a very promising and ambitious project.

JustVugg:
Great energy — and after #16 there's no doubt you can drive serious pieces of this. Let's sort the list into what fits the project's soul and what doesn't.

**The one hard no: a Rust rewrite.** colibrì's identity is a single C file you can read in an afternoon — gcc and go, no dependency tree. Performance-wise a rewrite buys nothing here: the hot paths are bounded by disk and memory bandwidth, not by the language. A Rust port as a sibling project elsewhere is of course fine, but the reference engine stays C.

**Everything else is welcome, and most of it already has a seed to grow from:**

- **Tiered VRAM/RAM/disk** — you literally built the VRAM tier in #16. Deepening it (smarter placement, per-expert precision by heat, promotion/demotion at runtime) is a natural continuation.
- **Hotspot analysis** — the learning cache (`.coli_usage` + AUTOPIN with confidence scaling) is exactly this today. Real headroom: expert co-occurrence analysis, disk-layout reordering so co-firing experts read sequentially, and the just-landed router-lookahead (`PILOT=1`, 71.6% next-layer recall) which wants tuning on machines where the disk isn't saturated.
- **Testing** — your REPLAY fixture is now the base for reproducible kernel A/Bs. A regression suite around it would be valuable, with one constraint: the maintainer deliberately runs **no CI** — everything must stay a local one-command affair like `setup.sh`.
- **SDK / API surface** — #14 asks for an OpenAI-compatible server, which is the most-requested integration and a thin wrapper over the existing `SERVE` protocol. That's the highest-leverage "SDK" piece by far.
- **JIT** — interesting, but concrete case first: the hot loops are already shape-specialized by hand (AVX2/VNNI/NEON). If profiling shows a dispatch cost worth burning, numbers welcome.

[...]

Issue #18, opened 2026-07-10
JustVugg Appreciation post


jazir555:
Thank you very much for making this project and getting this working

JustVugg:
Thank you for using it and helping us grow!

JustVugg:
Thank you again — messages like this genuinely keep the project going. 🐦 The community has been incredible this week: benchmarks from real machines, a HF mirror, the OpenAI server, a web UI, the CUDA tier. Closing this one, but the door's always open — come back anytime with numbers, ideas, or just to say hi. If a Discord happens (vote running in #19), you'll be welcome there too.

Issue #13, opened 2026-07-10
stats


cbroker1:
hey, I'm the dude from reddit, polandtown, I'll let you deal with poor formatting below. 7950x, nvme. 

running this overnight:
-----------------------------------
❯ cd c
./coli bench                                   # hellaswag, arc_challenge, mmlu — 40 questions each
./coli bench hellaswag --limit 200             # one task, more questions
./coli bench mmlu arc_challenge --ram 100      # pick tasks, set a RAM budget
-----------------------------------

And then my prelim tests before running the above:
-----------------------------------

gcc -O2 -fopenmp iobench.c -o iobench
./iobench /home/cbroker/Documents/2026/colibri

gcc -O2 -fopenmp iobench.c -o iobench
./iobench /home/cbroker/Documents/2026/colibri/out-00069.safetensors 19 64 8 0   # buffered, 8 threads
./iobench /home/cbroker/Documents/2026/colibri/out-00069.safetensors 19 64 8 1 

COLI_MODEL=/home/cbroker/Documents/2026/colibri/ ./coli chat

❯ bash setup.sh
🐦 colibrì — setup
  gcc: 13 · 32 core
  OpenMP: manca (libgomp)

❯ gcc -O2 -fopenmp iobench.c -o iobench
./iobench /home/cbroker/Documents/2026/colibri/out-00069.safetensors 19 64 8 0   # buffered, 8 threads
./iobench /home/cbroker/Documents/2026/colibri/out-00069.safetensors 19 64 8 1
buffered x8 thread: 64 letture x 19MB = 1.3 GB in 0.33s -> 3.85 GB/s (5.2 ms/blocco effettivi)
O_DIRECT x8 thread: 64 letture x 19MB = 1.3 GB in 0.85s -> 1.50 GB/s (13.3 ms/blocco effettivi)

❯ cd /home/cbroker/Documents/2026/colibri/colibri/c

chmod +x ./coli

COLI_MODEL=/home/cbroker/Documents/2026/colibri ./coli chat

     ▄▀▀▀▄  ▄        colibrì v1.0
  ▄▄▄▄▀▀▀▀▄▀▀        piccolo motore, modello immenso
      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU
        ▀▀▀▀         chat · colibri · ram -GB · topp off
          ▀          
  ──────────────────────────────────────────────────────────
  [MTP] attiva: decodifica speculativa nativa (draft=3)
[...]

Issue #12, opened 2026-07-09
Performance Benchmark on Ryzen AI 9 HX 370


matey-0:
My Computer: 
* Framework 13 /w Ryzen AI 9 HX 370 (12c/24t)
* 128gb RAM (2x64gb DDR5 5600 SODIMM)
* 8tb SSD (WD_BLACK SN850X), /w BTRFS ZSTD:5 global compression 
* 48gb ZRAM device
* Arch Linux /w CachyOS-influenced performance optimisations

My benchmark:

~/Development/colibri/c main !2 ?1 ❯ ./coli run "Explain how a hash map works." --model /home/mateogrgic/Development/colibri/c/GLM-5.2-colibri-int4 --ram 110 --ngen 150

     ▄▀▀▀▄  ▄        colibrì v1.0
  ▄▄▄▄▀▀▀▀▄▀▀        piccolo motore, modello immenso
      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU
        ▀▀▀▀         run
          ▀
  ──────────────────────────────────────────────────────────
== Motore C GLM (glm_moe_dsa), cache=8 expert/layer | expert@8-bit densa@8-bit | idot: avx512-vnni ==
caricato in 4.18s | densa residente: 9780.06 MB | layers=78 experts=256 | MTP ATTIVA (draft=3)
[MTP] attiva: decodifica speculativa nativa (draft=3)
[RAM_GB=110.0] cap=8 ok (proiezione picco 28.0 GB)
[stop] 3 token di stop: 154820 154827 154829
prompt: 14 token | genero fino a 150 (stop EOS=154820) | draft n-gram=3
[gMASK]<sop><|user|>Explain how a hash map works.<|assistant|><think></think>[prefill] layer 1/78 · 14 token
[prefill] layer 5/78 · 14 token
[prefill] layer 9/78 · 14 token
[prefill] layer 13/78 · 14 token
[prefill] layer 17/78 · 14 token
[prefill] layer 21/78 · 14 token
[prefill] layer 25/78 · 14 token
[prefill] layer 29/78 · 14 token
[prefill] layer 33/78 · 14 token
[prefill] layer 37/78 · 14 token
[prefill] layer 41/78 · 14 token
[prefill] layer 45/78 · 14 token
[prefill] layer 49/78 · 14 token
[prefill] layer 53/78 · 14 token
[prefill] layer 57/78 · 14 token
[...]

PR #20, opened 2026-07-10
Add local project checks and contribution templates

ZacharyZcR:
## Summary

- add optional structured bug, performance, and feature Issue Forms while preserving blank issues
- add dependency-free C tests for JSON and safetensors primitives
- add Python standard-library tests for benchmark output parsing
- add a local one-command validation target and document the contribution workflow
- add a pull request template

## Local validation

- `make -C c check`
- Issue Form YAML parsed successfully
- `git diff --check`

No GitHub Actions workflow is included; validation remains a local one-command process in line with the project policy.


JustVugg (changes requested):
Thanks — most of this is genuinely useful, but two things need to change before it can land:

**1. Remove `.github/workflows/ci.yml`.** No CI is a deliberate, documented maintainer choice for this project (see the note in #19: everything must stay a local one-command affair). `make -C c check` as a *local* target is exactly right — keep it, and keep the `Run the same lightweight checks` wording in CONTRIBUTING.md pointed at running it locally instead of at CI.

**2. Re-enable blank issues** (`blank_issues_enabled: true` or drop `config.yml`). The forms are a nice option, but a lot of this project's value has arrived as free-form posts — benchmark dumps (#12, #13), the HF mirror announcement (#2), even an appreciation post (#18). Forcing every report through a form would add friction exactly where this community thrives. Optional forms + blank issues allowed = best of both.

Everything else — the dependency-free C tests for JSON/safetensors primitives, the python parsing tests, the `check` Makefile target, the PR template, CONTRIBUTING.md — is welcome as-is. Drop the workflow, flip the config line, and this merges.

JustVugg:
Thanks really for your help!

ZacharyZcR:
> Thanks really for your help!

Please feel free to contact me if it's convenient; I have some technical details I'd like to finalize with you. My DiscardID is zacharyzcr123.

ZacharyZcR:
Updated as requested:

- removed `.github/workflows/ci.yml`
- re-enabled blank issues while keeping the forms optional
- changed CONTRIBUTING.md to describe `make -C c check` as a local command
- updated the PR title and description to remove CI claims

Re-ran `make -C c check`, Issue Form YAML parsing, and `git diff --check`; all pass.

JustVugg (approved):
Both points addressed — CI workflow removed, blank issues re-enabled, CONTRIBUTING pointed at local checks. Tests verified here: `make -C c check` builds clean and passes (C json/st tests + python unittest).

One non-blocking note for a possible follow-up: `check` runs `clean` + `portable`, so a user who built with plain `make` (native arch) and then runs `make check` silently ends up with an x86-64-v3 binary — no AVX-512/VNNI kernels even on CPUs that have them. Rebuilding the tests alone (they don't link glm) or restoring the previous build flavor afterwards would avoid the surprise.

Approved — ready to merge.

Issue #14, opened 2026-07-10
[Feature] Web API?


dbristow-otc:
Is there any plans to be able to host a web interface akin to [Ollama](https://docs.ollama.com/api/introduction) or [OpenAI](https://docs.ollama.com/api/openai-compatibility) so we can use colibri from other applications?

JustVugg:
Good timing — it's on the radar. Two honest notes on where things stand:

- The engine already has a persistent `SERVE` mode (what `coli chat` talks to): model loaded once, persistent KV across turns, streaming output over stdio with simple framing (`READY`/`END` markers, `\x02RESET`). An OpenAI-compatible HTTP layer is genuinely a thin wrapper around that — no engine changes needed.
- A community member mentioned working on an OpenAI-compatible server back in the macOS port thread; it hasn't landed yet.

So: yes, planned, and it's also a very approachable PR if anyone wants it sooner — a small Python/Go shim that spawns `SERVE=1 ./glm`, maps `/v1/chat/completions` (with `stream: true` SSE) onto the stdio protocol, and applies the GLM chat template. One caveat worth designing around: this engine is optimized for a single conversation stream (per-conversation KV + a learning cache that adapts to *your* usage), so a first version should be single-session — a queue, not a batch server.

JustVugg:
Shipped — #21 is merged. `coli serve` (or `python3 c/openai_server.py --model <dir>`) gives you a stdlib-only OpenAI-compatible gateway: `/v1/models`, `/v1/chat/completions` (JSON + SSE streaming), `/v1/completions`, Bearer auth via `COLI_API_KEY`, localhost by default.

Two details worth knowing beyond the standard protocol:
- **KV prefix reuse**: even though HTTP is stateless, the engine matches each request's tokens against the live KV and prefills only the new tail — verified: a follow-up request reused 62/120 tokens and prefilled only 58. Conversations don't pay the re-prefill tax.
- It composes with the engine's KV-to-disk persistence, so the server process can restart and pick conversations back up warm.

There's also a browser UI for it under `web/` (#23). Closing — enjoy, and open a new issue if anything misbehaves with Ollama-style clients.

PR #21, opened 2026-07-10
Add OpenAI-compatible HTTP API

ZacharyZcR:
## Summary

- add `coli serve`, a dependency-free OpenAI-compatible HTTP gateway backed by one persistent `glm` process
- add a length-prefixed raw prompt command to the existing C serve protocol so complete multi-message prompts and embedded newlines are preserved
- reuse the longest matching token/KV prefix across stateless HTTP requests, prefill only the appended or divergent tail, and rewind the persisted `.coli_kv` header at the same boundary
- implement `GET /v1/models`, model retrieval, `POST /v1/chat/completions`, and legacy `POST /v1/completions`
- support normal JSON responses, SSE streaming, streaming usage, Bearer auth, `max_tokens` / `max_completion_tokens`, `temperature`, `top_p`, and GLM reasoning mode
- serialize generations through the single loaded 744B engine instead of loading duplicate model processes
- support browser/Tauri clients with OPTIONS preflight and an explicit CORS origin allowlist

The request and response shapes follow the official [Chat Completions](https://platform.openai.com/docs/api-reference/chat/completions/create) and [Models](https://platform.openai.com/docs/api-reference/models) APIs. Chat rendering follows the official GLM-5.2 `chat_template.jinja` text path.

## KV reuse

OpenAI-style clients resend the complete conversation on every request. The engine tokenizes that prompt into a scratch buffer, compares it with `hist`, keeps the longest common prefix in memory, and forwards only the new tail. If history was edited, both the in-memory length and persisted KV record count rewind to the first divergent token; the following append overwrites that tail. An identical prompt re-forwards its last token to reconstruct logits without a full prefill.

`STAT` keeps its original fields first and appends full prompt-token and length-limit fields. Both the interactive CLI parser and the HTTP gateway accept the extended format.

## Scope

This first version is deliberately text-only. Tools, multimodal content, custom stop sequences, log probabilities, token penalties, structured output, and per-request seeds return explicit errors. Concurrent HTTP requests queue on the persistent engine.

The API binds to localhost by default. `COLI_API_KEY` enables Bearer authentication. Vite and Tauri local origins are allowed by default; additional exact origins require repeated `--cors-origin` flags.

## Local validation

- `make check` — portable C build and 13 dependency-free tests pass
- official `openai` Python SDK round-trip against the gateway: models list, non-streaming chat, and streaming chat pass
- browser OPTIONS/CORS preflight test passes
- `python3 -m py_compile c/openai_server.py c/tests/test_openai_server.py c/coli`
- `git diff --check`

A real 744B model generation was not run in this workspace; the C engine protocol compiles cleanly, while the HTTP and streaming path is covered with a deterministic fake engine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,10 @@ sequences, log probabilities, and token penalties return an explicit error rathe
 than being silently ignored. The default bind address is localhost; set
 `COLI_API_KEY` before exposing the server beyond the machine.
 
+Browser access from the Vite development server and Tauri local origins is enabled
+by default. Repeat `--cors-origin https://your-ui.example` to allow another exact
+origin, or use `--cors-origin '*'` only on a trusted local network.
+
 ### Experimental resident CUDA backend
 
 colibrì includes an opt-in CUDA backend for model-resident tensors. Streaming
--- a/README.md
+++ b/README.md
@@ -93,6 +93,41 @@ COLI_MODEL=/nvme/glm52_i4 ./coli chat
 
 The engine at runtime is pure C — python is only used by the one-time converter.
 
+### OpenAI-compatible API
+
+`coli serve` keeps one model process loaded and exposes a text-only OpenAI-compatible
+HTTP API. The gateway uses only the Python standard library; inference still runs in
+the same dependency-free C engine.
+
+```bash
+cd c
+COLI_MODEL=/nvme/glm52_i4 COLI_API_KEY=local-secret ./coli serve \
+  --host 127.0.0.1 --port 8000 --model-id glm-5.2-colibri
+
+curl http://127.0.0.1:8000/v1/chat/completions \
+  -H 'Authorization: Bearer local-secret' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "model": "glm-5.2-colibri",
+    "messages": [{"role": "user", "content": "Hello"}],
+    "stream": true
+  }'
+```
+
+Implemented endpoints are `GET /v1/models`, `GET /v1/models/{model}`,
+`POST /v1/chat/completions`, and legacy `POST /v1/completions`. Chat and
+completion requests support JSON responses, SSE streaming, usage counts,
+`max_tokens`/`max_completion_tokens`, `temperature`, and `top_p`. The extension
+`enable_thinking: true` enables GLM-5.2's reasoning block; the standard
+`reasoning_effort` field also enables it unless set to `none`.
+
+The first version is deliberately text-only and serves one generation at a time:
+the 744B model stays in one persistent process, so concurrent HTTP requests queue
+instead of loading duplicate model copies. Tools, image/audio input, custom stop
+sequences, log probabilities, and token penalties return an explicit error rather
+than being silently ignored. The default bind address is localhost; set
+`COLI_API_KEY` before exposing the server beyond the machine.
+
 ### Experimental resident CUDA backend
 
 colibrì includes an opt-in CUDA backend for model-resident tensors. Streaming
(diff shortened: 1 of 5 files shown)


JustVugg (changes requested):
This is exactly the shape #14 asked for — stdlib-only gateway, honest scope, careful protocol framing (bounds checks, temp/nuc restore, the `memchr(raw,0,..)` guard). One design issue needs fixing before merge, because it decides whether the API is actually usable for chat on the hardware this project targets:

**`\x02PROMPT` resets the KV on every request (`len=0; first=1`).** HTTP chat is stateless — each request re-sends the whole conversation — so every message triggers a full re-prefill from token 0. On the machines this engine runs on, prefill is the expensive part: a 1,000-token conversation would take minutes *per message*, growing linearly. The interactive serve path keeps KV persistent across turns precisely to avoid this.

**Fix: longest-common-prefix reuse instead of reset.** OpenAI-style requests grow by appending, and `hist` already contains the rendered conversation *including* the assistant replies the engine itself generated — so the new request's token stream is (near-)identical to `hist` up to the new user message. Sketch:

```c
/* raw_mode: tokenize into a scratch buffer, keep the common prefix in KV */
int *tmp=malloc(maxctx*sizeof(int));
int k=tok_encode(&T,input,input_n,tmp,maxctx);
int p=0; while(p<len && p<k && hist[p]==tmp[p]) p++;
if(p<len){ len=p; if(m->has_mtp) m->kv_start[m->c.n_layers]=-1; }  /* divergence: drop tail only */
memcpy(hist+len, tmp+len, (k-len)*sizeof(int));
k-=len;              /* prefill only the new tail; k==0 edge: re-forward last token like MORE */
free(tmp);
```

Worst case (client edited history, different template) the prefix is short and it degrades gracefully to today's behavior; typical case it prefills only the last user message. It also composes with the KV-to-disk persistence landing on main (conversations resume warm across engine restarts — your gateway would inherit that for free).

Second, smaller point: the `STAT` line grew two fields — the interactive `coli` parser needs to stay compatible (your coli patch handles it, just flagging the cross-PR coupling for anyone rebasing).

Everything else reads well. Fix the prefix reuse and this merges — it's the most-requested feature in the tracker.

ZacharyZcR:
Updated per review and rebased onto the latest `main` (including `.coli_kv` persistence):

- raw prompts are tokenized into a scratch buffer and matched against `hist`
- the longest common KV prefix is retained; only the new tail is prefetched
- edited/shorter histories rewind both the in-memory context and persisted KV record count to the divergence point, so the next append overwrites only the tail
- identical prompts re-forward the last token to reconstruct logits (`k == 0`)
- extended `STAT` output keeps the original fields first and reports the full prompt-token count
- the interactive `coli` parser remains compatible with the extended status line

Validation after rebase: `make check` (portable C build + 13 tests), Python byte-compilation, `git diff --check`, and an official OpenAI Python SDK round-trip for models, non-streaming chat, and streaming chat. A real 744B inference run is still not available in this workspace.

JustVugg (approved):
Prefix-reuse verified end-to-end here (tiny model, real server + real engine): first request `[API] KV prefix 0/62, prefill 62`, second request extending the conversation `[API] KV prefix 62/120, prefill 58` — the previous conversation's KV survives stateless HTTP exactly as designed, and the `kv_disk_truncate` integration keeps `.coli_kv` coherent on divergence. /v1/models, non-streaming, SSE streaming and the k=0 identical-prompt path all behave. Engine untouched elsewhere: oracle 32/32 TF, 20/20 greedy. Great turnaround — merging.

PR #26, opened 2026-07-10
Adapt hot expert tiers to live workloads

ZacharyZcR:
## Runtime series

**Part 1 of 5:** #26 adaptive hot tiers → #27 resource planning/application → #28 bounded request scheduling → #29 isolated sequence KV ownership → #30 ragged multi-sequence decode forward.

## Summary

- make live expert replacement follow recent session heat instead of lifetime `.coli_usage` totals
- keep persistent usage as the long-term startup signal and add a separate decaying heat map for short-term adaptation
- reuse existing pinned slots so promotion follows the real disk -> RAM path without increasing the memory budget
- immediately refresh GPU-backed slots after replacement, completing disk -> RAM -> VRAM promotion at a safe turn boundary
- explicitly demote a slot to RAM and repair GPU accounting if refresh fails
- keep the feature opt-in through `REPIN=N` and expose it as `coli --repin N chat`
- retain the existing 25% hysteresis and four-replacement limit to avoid tier thrashing
- add a dependency-free unit test for candidate selection, hysteresis, and heat decay

This is a scoped continuation of the tiering work discussed in #19. It does not introduce a general scheduler or change the default execution path. The storage hierarchy remains concrete: disk is the backing store, pinned slots are the RAM tier, and CUDA-enabled pinned slots are the VRAM tier.

## Why

The existing live `REPIN` path compares cumulative `eusage`, including counts restored from `.coli_usage`. That is useful for startup placement but makes a long-running installation slow to react when the active workload changes. It also invalidates old CUDA tensors during replacement but leaves the new expert to upload lazily on its next use.

This change separates long-term and short-term signals. Recent heat decays after each eligible replacement pass, while `.coli_usage` continues accumulating unchanged. A selected expert is loaded into the existing RAM slot and, when that slot belongs to the CUDA tier, uploaded immediately to its existing device assignment.

## Usage

```bash
COLI_MODEL=/nvme/glm52_i4 ./coli chat --repin 128

# With the existing CUDA hot tier
COLI_CUDA=1 COLI_GPU=0 CUDA_EXPERT_GB=16 \
  PIN=stats.txt PIN_GB=64 COLI_MODEL=/nvme/glm52_i4 \
  ./coli chat --repin 128
```

## Validation

- `make check` — portable CPU build and all C/Python tests pass (including 13 API tests)
- new `test_tier` covers promotion choice, anti-churn hysteresis, and heat decay
- `test_tier` passes under ASan + UBSan
- CUDA-conditional `glm.c` path passes `-DCOLI_CUDA -fsyntax-only`
- Python byte-compilation and `git diff --check`

The CUDA runtime path was not executed in this workspace because the CUDA toolkit/nvcc is not installed; no end-to-end speedup claim is made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,13 +211,19 @@ works against the colibrì OpenAI-compatible server (in review, #21) or any othe
 compatible endpoint. Nothing leaves the endpoint you configure. The terminal
 `coli chat` remains the first-class interface.
 
-Useful knobs (env or flags): `--temp T` token sampling temperature (default 0.7 + nucleus 0.90 — tuned for int4; 0 = greedy), `--topp 0.7` adaptive expert top-p (30–40% less disk), `--ngen N` max tokens per answer (`:piu` in chat continues a truncated one), `AUTOPIN=0` disable the learning cache's auto-pin, `THINK=1` enable GLM-5.2's reasoning block, `DRAFT=n` MTP draft depth, `TF=1` teacher-forcing validation, `PILOT=1` router-lookahead disk prefetch (experimental — see below), `CAP_RAISE=0` don't auto-grow the expert cache.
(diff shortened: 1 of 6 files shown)


JustVugg (approved):
Verified in a worktree: clean build (0 warnings), oracle 32/32 TF and 20/20 greedy with REPIN=4 active, tier test passes. The design is right: `.coli_usage` stays the persistent startup signal, the decaying heat map (>>1 per pass) tracks the live session, the swap still goes disk→RAM in the same resident slot, and the CUDA refresh-or-demote keeps accounting honest. Opt-in, default behavior untouched. Approved.

PR #27, opened 2026-07-10
Add and apply tiered Disk/RAM/VRAM resource plans

ZacharyZcR:
## Runtime series

**Part 2 of 5:** #26 adaptive hot tiers → #27 resource planning/application → #28 bounded request scheduling → #29 isolated sequence KV ownership → #30 ragged multi-sequence decode forward.

## Summary

- add `coli plan` as a read-only Disk/RAM/VRAM placement planner
- read safetensors headers without loading tensor payloads and measure dense versus routed-expert storage exactly
- mirror the engine's RAM safety calculation: dense resident set, working set, page-cache headroom, KV cache, attention reconstruction buffer, and expert cache cap
- discover per-device free VRAM, preserve 2 GB runtime headroom per device, and clamp the hot tier by both VRAM capacity and required RAM backing
- expose a versioned `--json` result for the API server, Web UI, and desktop shell
- add `--auto-tier` to apply the same plan to `chat`, `run`, `serve`, and benchmarks
- map the plan to `RAM_GB`, `CTX`, `COLI_GPU(S)`, and `CUDA_EXPERT_GB`
- enable the VRAM tier only when the current `glm` binary is actually linked to an available CUDA runtime
- keep explicit CLI flags and environment settings ahead of automatic values; `--gpu none` is a hard CPU override
- add dependency-free model/header fixtures, placement tests, environment-application tests, and CLI JSON coverage

This is a scoped runtime foundation for the tiering direction discussed in #19. Planning does not load model tensors. Applying a plan configures the existing engine rather than introducing a second allocator or scheduler.

## Usage

```bash
# inspect only
COLI_MODEL=/nvme/glm52_i4 ./coli plan
COLI_MODEL=/nvme/glm52_i4 ./coli plan --gpu 0,1 --ram 128 --vram 48 --json

# inspect and apply the bounded plan to the runner
COLI_MODEL=/nvme/glm52_i4 ./coli chat --auto-tier
COLI_MODEL=/nvme/glm52_i4 ./coli serve --auto-tier --gpu 0,1 --vram 48

# explicit CPU override, even when NVIDIA devices are visible
COLI_MODEL=/nvme/glm52_i4 ./coli chat --auto-tier --gpu none
```

The engine's existing auto-pin/usage logic remains responsible for choosing which experts occupy the planned hot budget. This PR decides capacity and device placement; it does not invent a competing heat policy.

## Real 744B model validation

Read-only planning was run against the completed `jlnsrk/GLM-5.2-colibri-int4` container on a six-GPU host, selecting only GPU 0. It completed in 0.43 seconds without starting the engine or creating a CUDA context:

- 144 shards, 378.77 GB
- 10.74 GB dense resident footprint
- 368.02 GB routed experts, 19,456 `(layer, expert)` groups
- 204.24 GB automatic RAM budget from live `MemAvailable`
- 6.13 GB runtime reserve and 130 estimated cache slots/layer
- GPU 0 had 15.91 GB free at inspection time; after 2 GB headroom the bounded VRAM tier was 13.91 GB (~735 typical experts)

The values reflect live free RAM/VRAM and intentionally change as other workloads start and stop. Per request, no inference or CUDA runtime test was started on the busy GPU host.

## Validation

- `make check` — portable C build and all 19 C/Python tests pass
- planner tests cover dense/expert classification, RAM/VRAM clamps, device selection, human formatting, and versioned CLI JSON
- application tests cover multi-GPU mapping, explicit-setting precedence, PIN backing budget, CPU-binary fallback, and explicit CUDA disable
- `python3 -m py_compile c/coli c/resource_plan.py c/tests/test_resource_plan.py`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,18 @@ cd c
 COLI_MODEL=/nvme/glm52_i4 ./coli chat
 ```
 
+Inspect the planned storage hierarchy before loading the model:
+
+```bash
+COLI_MODEL=/nvme/glm52_i4 ./coli plan
+COLI_MODEL=/nvme/glm52_i4 ./coli plan --gpu 0,1 --ram 128 --vram 48 --json
+```
+
+`coli plan` reads only safetensors headers and reports the model's exact dense/expert
+footprint, runtime RAM reserve, safe expert-cache cap, and bounded VRAM hot tier. Its
+versioned JSON output is intended to be shared by the CLI, API server, Web UI, and
+desktop shell; it does not allocate model tensors or start inference.
+
 The engine at runtime is pure C — python is only used by the one-time converter.
 
 ### OpenAI-compatible API
--- a/README.md
+++ b/README.md
@@ -96,12 +96,19 @@ Inspect the planned storage hierarchy before loading the model:
 ```bash
 COLI_MODEL=/nvme/glm52_i4 ./coli plan
 COLI_MODEL=/nvme/glm52_i4 ./coli plan --gpu 0,1 --ram 128 --vram 48 --json
+
+# apply the bounded plan to the normal runner
+COLI_MODEL=/nvme/glm52_i4 ./coli chat --auto-tier
 ```
 
 `coli plan` reads only safetensors headers and reports the model's exact dense/expert
 footprint, runtime RAM reserve, safe expert-cache cap, and bounded VRAM hot tier. Its
 versioned JSON output is intended to be shared by the CLI, API server, Web UI, and
 desktop shell; it does not allocate model tensors or start inference.
+`--auto-tier` applies the same plan to `chat`, `run`, `serve`, and benchmarks. It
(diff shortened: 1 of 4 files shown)


JustVugg (approved):
Tested here: `coli plan` runs on a model dir, the human output and versioned `--json` are both correct, and the cap formula mirrors the engine's budget math exactly — including the auto-raise ceiling at n_experts (verified: on the 8-expert tiny fixture it correctly reports cap 8 because that IS n_experts). `environment_for_plan`'s setdefault semantics respect explicit user env. Stdlib-only, read-only, no engine changes.

One forward-coupling note (non-blocking): `runtime_bytes` mirrors `cap_for_ram`'s slack — when #29's KV_SLOTS budget scaling lands, this mirror needs the same nslots factor on the KV term. Worth a comment in the code pointing at its C twin so they don't drift. Approved.

PR #28, opened 2026-07-10
Add bounded FIFO inference request scheduling

ZacharyZcR:
## Runtime series

**Part 3 of 5:** #26 adaptive hot tiers → #27 resource planning/application → #28 bounded request scheduling → #29 isolated sequence KV ownership → #30 ragged multi-sequence decode forward.

## Summary

- replace implicit unbounded waiting on `Engine.lock` with a bounded FIFO admission scheduler
- keep the engine lock as a final protocol-safety guard while admission happens before any HTTP/SSE response headers
- configure waiting capacity and timeout with `--max-queue`, `--queue-timeout`, `COLI_MAX_QUEUE`, and `COLI_QUEUE_TIMEOUT`
- return OpenAI-shaped HTTP 429 errors with `queue_full` or `queue_timeout` codes and `Retry-After`
- detect clients that disconnect while queued and remove their tickets without consuming an engine turn
- reject queued work with HTTP 503 during scheduler shutdown
- expose active, queued, admitted, completed, rejected, timed-out, and cancelled counters through `GET /health`
- include `x-colibri-queue-wait-ms` on successful JSON and SSE generation responses
- expose queue-related headers to configured browser origins

## Why FIFO instead of continuous batching

The C engine currently owns one mutable token history and one KV context. Pretending that Python threads can execute multiple sequences concurrently would corrupt that state. Continuous batching requires sequence-indexed KV ownership in the engine first; this PR provides honest admission control for the engine that exists today.

The previous `threading.Lock` preserved protocol bytes but had two production problems:

1. waiting was unbounded, so enough HTTP threads could consume memory indefinitely;
2. streaming responses sent HTTP 200 before trying to acquire the engine lock, making queue rejection impossible once headers were committed.

Admission now wraps the entire generation before either JSON or SSE headers are sent.

## Configuration

```bash
COLI_MODEL=/nvme/glm52_i4 ./coli serve --max-queue 8 --queue-timeout 300

# equivalent environment configuration
COLI_MAX_QUEUE=8 COLI_QUEUE_TIMEOUT=300 \
  COLI_MODEL=/nvme/glm52_i4 ./coli serve
```

`max_queue` counts waiting requests; the one active generation is separate. Setting it to zero gives fail-fast single-request admission.

## Cancellation scope

Queued client disconnects are cancelled. Once an admitted generation has entered the C protocol, the gateway continues draining it even if the streaming client disconnects; aborting mid-turn would leave the shared stdout framing/KV state inconsistent. Active cancellation needs an explicit engine cancel command and is intentionally not faked here.

## Validation

- `make check` — portable C build and all 19 C/Python tests pass
- deterministic tests cover FIFO order, queue saturation, timeout, queued cancellation, shutdown wake-up, metrics, and queue-wait headers
- real threaded HTTP test proves a saturated streaming-capable endpoint returns 429 before generation/response commitment
- the scheduler/API suite passed 10 consecutive runs to catch timing races
- `python3 -m py_compile c/openai_server.py c/tests/test_openai_server.py c/coli`
- `git diff --check`

No model or GPU inference was started for this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,14 @@ Browser access from the Vite development server and Tauri local origins is enabl
 by default. Repeat `--cors-origin https://your-ui.example` to allow another exact
 origin, or use `--cors-origin '*'` only on a trusted local network.
 
+The engine owns one mutable KV context, so HTTP generation uses a bounded FIFO
+admission queue instead of pretending to run unsafe parallel sequences. Configure it
+with `--max-queue N` (default 8) and `--queue-timeout SECONDS` (default 300), or the
+`COLI_MAX_QUEUE` / `COLI_QUEUE_TIMEOUT` environment variables. Saturated and timed-out
+requests receive OpenAI-shaped HTTP 429 errors before streaming headers are sent.
+`GET /health` exposes active/queued/completed/rejected counters, and successful
+generation responses include `x-colibri-queue-wait-ms`.
+
 ### Experimental resident CUDA backend
 
 colibrì includes an opt-in CUDA backend for model-resident tensors. Streaming
--- a/c/coli
+++ b/c/coli
@@ -380,7 +380,7 @@ def cmd_serve(a):
     need_model(a.model)
     from openai_server import serve
     serve(a.model, a.host, a.port, a.model_id, a.api_key,
-          a.cap, a.ngen, GLM, env_for(a), a.cors_origin)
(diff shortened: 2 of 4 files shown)


JustVugg (approved):
All 19 tests pass here (`make check` in a clean worktree). This is exactly the design boundary we asked for in #14: a bounded queue in front of ONE engine, not a batch server — admission before response headers, OpenAI-shaped 429s with Retry-After, queued-client disconnect detection, 503 on shutdown. No engine changes. Approved.

PR #29, opened 2026-07-10
Add isolated sequence KV contexts

ZacharyZcR:
## Runtime series

**Part 4 of 5:** #26 adaptive hot tiers → #27 resource planning/application → #28 bounded request scheduling → #29 isolated sequence KV ownership → #30 ragged multi-sequence decode forward.

**Merge dependency:** none. This PR is based directly on `main`. It establishes the engine primitive that future continuous batching can use; it does not depend on the Python scheduler implementation in #28.

## Summary

- extract KV ownership into an explicit `KVState` that owns MLA L/R caches, DSA index caches, MTP window state, capacity, and persistence accounting
- let the engine bind one `KVState` at a time without threading sequence IDs through every attention call
- add up to 16 independently allocated serve contexts through `KV_SLOTS` / `--kv-slots`
- give every context its own token history, prefix-reuse state, compressed KV memory, and crash-safe persistence file (`.coli_kv`, `.coli_kv.1`, ...)
- extend the raw engine protocol with an optional backward-compatible KV slot field
- expose an optional OpenAI request extension, `cache_slot`, defaulting to slot 0
- include the configured slot count in `GET /health`
- account for all configured MLA, DSA, and MTP KV pools before sizing the RAM expert cache
- reject invalid pool sizes before loading the 744B model

## Why explicit slots

The engine previously stored one token history and one mutable KV cache directly on `Model`. That makes request queuing safe, but it means unrelated conversations evict each other's prefix state and blocks any honest path toward multi-sequence scheduling.

This PR separates model weights from sequence state. A request explicitly chooses a slot:

```json
{
  "model": "glm-5.2-colibri",
  "messages": [{"role": "user", "content": "Continue this conversation"}],
  "cache_slot": 1
}
```

```bash
COLI_MODEL=/nvme/glm52_i4 ./coli serve --kv-slots 4
```

Ordinary OpenAI clients omit `cache_slot` and retain byte-for-byte slot 0 behavior. Explicit integer slots avoid hidden hashing, collision claims, and an unbounded server-side session map.

## What this does not claim

Execution remains one sequence at a time. The active `KVState` is pointer-bound before a turn, and all existing attention/MTP code operates on that owner. Continuous batching still needs a batched decode loop and per-sequence sampling/admission, but it no longer needs to untangle KV ownership from model weights first.

## Memory safety

Each 4096-token slot costs hundreds of MB on GLM-5.2. The existing RAM projection now multiplies MLA/DSA cache memory by the configured slot count before deciding expert LRU capacity, so adding contexts takes capacity from the RAM expert tier instead of inviting an OOM kill.

## Validation

- `make check` — portable C build and all 16 C/Python tests pass
- CUDA-conditional C path passes `-DCOLI_CUDA -fsyntax-only`
- API tests cover slot selection, slot bounds, health discovery, and default slot compatibility
- legacy four-field raw `PROMPT` headers remain accepted as slot 0
- invalid `KV_SLOTS` fails before model loading
- Python byte-compilation and `git diff --check`

No model or GPU inference was started for this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,27 @@ requests receive OpenAI-shaped HTTP 429 errors before streaming headers are sent
 `GET /health` exposes active/queued/completed/rejected counters, and successful
 generation responses include `x-colibri-queue-wait-ms`.
 
+### Isolated KV contexts
+
+`coli serve --kv-slots N` allocates up to 16 independent sequence contexts. Requests
+select one with the optional integer `cache_slot` field; ordinary OpenAI clients omit
+it and keep the original slot 0 behavior.
+
+```json
+{
+  "model": "glm-5.2-colibri",
+  "messages": [{"role": "user", "content": "Continue this conversation"}],
+  "cache_slot": 1
+}
+```
+
+Each slot owns its token history, compressed MLA/DSA KV memory, MTP window, and
+crash-safe persistence file (`.coli_kv`, `.coli_kv.1`, ...). The engine still executes
+one sequence at a time; this establishes explicit KV ownership without pretending that
+threaded HTTP is continuous batching. RAM admission accounts for every configured slot.
+Use `COLI_KV_SLOTS=N` as the environment equivalent. Start with a small value: at the
+default 4096-token context, every slot costs hundreds of MB.
+
 ### Experimental resident CUDA backend
 
 colibrì includes an opt-in CUDA backend for model-resident tensors. Streaming
--- a/c/coli
+++ b/c/coli
@@ -440,7 +440,8 @@ def cmd_serve(a):
     need_model(a.model)
     from openai_server import serve
     serve(a.model, a.host, a.port, a.model_id, a.api_key,
-          a.cap,a.ngen,GLM,env_for(a),a.cors_origin,a.max_queue,a.queue_timeout)
+          a.cap,a.ngen,GLM,env_for(a),a.cors_origin,
+          a.max_queue,a.queue_timeout,a.kv_slots)
 
 def cmd_bench(a):
     need_model(a.model)
@@ -506,6 +507,7 @@ def main():
     ps.add_argument("--cors-origin",action="append",default=None)
     ps.add_argument("--max-queue",type=int,default=int(os.environ.get("COLI_MAX_QUEUE","8")))
     ps.add_argument("--queue-timeout",type=float,default=float(os.environ.get("COLI_QUEUE_TIMEOUT","300")))
+    ps.add_argument("--kv-slots",type=int,default=int(os.environ.get("COLI_KV_SLOTS","1")))
     pb=sub.add_parser("bench", parents=[common]); pb.add_argument("tasks", nargs="*")
     pb.add_argument("--limit",type=int,default=40); pb.add_argument("--data",default=os.path.join(HERE,"bench"))
     pc=sub.add_parser("convert", parents=[common]); pc.add_argument("--repo",default="zai-org/GLM-5.2-FP8")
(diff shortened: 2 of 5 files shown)


JustVugg (changes requested):
The KVState extraction is clean — `kv_bind` aliasing keeps the hot paths untouched, slot 0 keeps `.coli_kv` (backward compatible), per-slot `kv_start` with the MTP row re-armed, and the protocol extension is properly optional. One blocking issue:

**`KV_SLOTS` breaks the RAM budget model.** All slots are allocated eagerly (`serve_ctx_init` loop → `kv_alloc(m,maxctx)` each), and at the default `CTX=4096` each slot costs ~745 MB of MLA KV (+ ~40 MB of DSA index cache). But `cap_for_ram`'s slack accounts for exactly **one** KV at `max_ctx` — with `KV_SLOTS=16` that's ~12 GB the budget doesn't know about. On a RAM-tight machine the expert cache gets sized as if that memory were free, and the kernel OOM-killer fires mid-generation — the exact bug class we eliminated on 2026-07-04 (see the `cap_for_ram` comment block).

**Fix (small)**: read `KV_SLOTS` in `main` before the `cap_for_ram` call and scale only the `kv_b` term by the slot count (the `kvb_b` reconstruction buffer stays ×1 — only one context is ever bound during a forward). Something like:

```c
int nslots = getenv("KV_SLOTS")?atoi(getenv("KV_SLOTS")):1; if(nslots<1) nslots=1; if(nslots>16) nslots=16;
/* in cap_for_ram / expert_avail: */
double kv_b = (double)nslots * (c->n_layers+1)*max_ctx*(c->kv_lora+c->qk_rope)*4.0;
```

(passing nslots through or via a global, matching how the two functions mirror each other today).

Also worth a log line at startup: with N slots the projected KV footprint, so users see what they're spending.

Non-blocking note: the `#define hist (sc->hist)` local-alias macros keep the diff small but shadow real identifiers for the rest of the function — fine for now given the diff stays readable, just flagging it as a candidate for a later cleanup where the turn body takes `ServeCtx*` directly.

Everything else — isolation semantics, per-slot persistence files, the k=0/prefix interactions with slot switching — reads correct. Fix the budget accounting and this merges.

ZacharyZcR:
Fixed in the rebased head `bddd072`.

The RAM model now uses `kv_pool_bytes()`, which multiplies the full per-slot MLA + DSA allocation by validated `KV_SLOTS` in both `expert_avail()` and `cap_for_ram()`. The reconstruction buffer remains single-copy. Startup RAM logging now prints the projected `KV <slots>x<context>` footprint, and invalid slot counts still fail before model loading.

The branch was also rebased over merged #26–#28 while retaining their live heat, resource planning, and bounded scheduler behavior. `make check` passed after that integration (4 C tests and 19 Python/API tests at the #29 layer). No model/GPU run was used for this accounting fix.

ZacharyZcR:
Also addressed the non-blocking observability note in `0c79ec8`: serve startup now always logs `context slots: N x CTX` together with the projected aggregate MLA+DSA KV pool in GB, even when the expert cap does not need to be lowered. Validation execution remains deferred for the current test window; this commit is a three-line logging-only follow-up and passes `git diff --check`.

JustVugg (approved):
The budget fix landed better than the sketch asked: `kv_pool_bytes()` covers the DSA index cache term too (which my sketch omitted), and it's wired into BOTH mirrors (`expert_avail` + `cap_for_ram`) with the projected pool logged at startup. Verified in a worktree: clean build, oracle 32/32 TF + 20/20 greedy on the default path, `KV_SLOTS=3` logs `projected pool` and budgets it, per-slot persistence files are created lazily, and all 27 tests pass. Approved — merging.

PR #30, opened 2026-07-10
Add decode-only continuous batching

ZacharyZcR:
## Summary

- add a `DecodeRow` descriptor carrying one sequence's KV owner, token, and absolute position
- evaluate up to 16 active serve slots in one ragged decode forward while retaining one shared dense/MoE batch
- select MLA L/R KV and DSA index storage independently for every row
- keep prefill serial, then admit requests between decode rounds and compact completed slots naturally
- add a multiplexed `SUBMIT` / `DATA` / `DONE` / `ERROR` / `CANCEL` engine protocol
- route concurrent OpenAI-compatible API requests through one scheduler and per-request output queues
- reject simultaneous ownership of the same cache slot
- cancel disconnected clients between decode rounds and release their slot
- bind the HTTP port before loading the model, so occupied-port startup fails immediately
- retain the legacy SERVE protocol for non-HTTP callers

## Execution model

Prefill remains serial because prompts have different lengths. During decode, every active request contributes one `(KVState, token, position)` row. Attention reads and writes that row's independent cache, while dense and routed-expert layers share the batch. The scheduler remains the only forward owner.

This is decode-only continuous batching. It does not claim batched prefill or MTP/speculative batching. Different batch shapes may produce different greedy token choices near close logits due to floating-point non-associativity; fixed-shape runs remain reproducible and KV ownership remains isolated.

## Validation

- rebased onto current `main` after #29 merged; scoped delta is six commits
- `make check`: portable build, 4 C tests, 35 Python/API tests
- protocol coverage for malformed frames, duplicate IDs, non-finite sampling values, cancellation, split UTF-8, shutdown races, and queue saturation
- occupied-port live check: failure before engine spawn in about 66 ms
- streaming disconnect live check: active slot returned to zero and was immediately reusable
- GLM-5.2 744B INT4 on RTX 5090:
  - two CPU-expert requests: 46.82 s concurrent vs 50.18 s sequential
  - 32 GB host hot tier + 20 GB CUDA expert tier: 33.96 s concurrent, 28.4 GB GPU usage, no process swap
  - oversized 100 GB host tier regressed to 73.65 s and caused heavy swap, motivating conservative tier budgets
- CUDA q8/q4/q2/f32 correctness fixture passed on sm_120
- `git diff --check`

## Series

Final runtime-series piece after merged #26 adaptive tiers, #27 resource planning, #28 bounded scheduling, and #29 isolated KV contexts.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -37,7 +37,7 @@ CUDA_ARCH ?= native
 NVCCFLAGS ?= -O3 -std=c++17 -arch=$(CUDA_ARCH) -Xcompiler=-Wall,-Wextra
 PYTHON    ?= python3
 CUDA_OBJ  =
-TEST_BINS = tests/test_json tests/test_st tests/test_tier
+TEST_BINS = tests/test_json tests/test_st tests/test_tier tests/test_decode_batch
 ifeq ($(CUDA),1)
 ifeq ($(UNAME_S),Darwin)
 $(error CUDA=1 is supported only on Linux)
@@ -75,6 +75,9 @@ tests/test_st: tests/test_st.c st.h json.h compat.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_tier: tests/test_tier.c tier.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_decode_batch: tests/test_decode_batch.c decode_batch.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 test-c: $(TEST_BINS)
--- a/c/decode_batch.h
+++ b/c/decode_batch.h
@@ -1,0 +1,13 @@
+#ifndef COLIBRI_DECODE_BATCH_H
+#define COLIBRI_DECODE_BATCH_H
+
+#include <stddef.h>
+
+/* `base` belongs to one sequence's KV state.  Keeping this arithmetic in a
+ * model-independent seam makes ragged decode row ownership directly testable. */
+static inline float *coli_kv_row(float *base, int position, int width)
+{
+    return base + (size_t)position * (size_t)width;
+}
+
+#endif
--- a/c/decode_batch.h
+++ b/c/decode_batch.h
@@ -3,6 +3,7 @@
 
 #include <stddef.h>
 #include <stdio.h>
+#include <math.h>
 
 /* `base` belongs to one sequence's KV state.  Keeping this arithmetic in a
  * model-independent seam makes ragged decode row ownership directly testable. */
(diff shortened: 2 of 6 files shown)


ZacharyZcR:
GPU validation update (6x RTX 5090 host, GLM-5.2 744B INT4):

- `make check`: 4 C tests + 34 Python/API tests passed.
- Full-model single request produced `OK.`; dense CUDA residency used ~9.16 GB VRAM.
- Two concurrent requests with CPU-resident experts: 46.82 s versus 50.18 s sequential (~6.7% wall-time improvement).
- Two concurrent requests with 32 GB RAM hot-store + 20 GB CUDA expert tier: 33.96 s (~27.5% faster than the CPU-expert concurrent run), 28.4 GB total GPU usage, 0 process swap.
- An oversized 100 GB hot-store regressed to 73.65 s and drove ~25-34 GB swap. The runtime must avoid treating all nominally available RAM as safely pinnable.
- The reduced-hot-store run also exposed output-quality drift on one deterministic continuation (`" D, and and: "`). This needs investigation before presenting CUDA-expert batching as production-ready.
- Streaming disconnect releases the active slot and the slot can be reused.
- Occupied-port startup now fails before loading the model (commit `e234758`); validated live in ~66 ms with no extra engine process.

The current evidence supports decode batching and conservative hot-tier sizing, but not aggressive automatic RAM pinning. I am keeping this PR as Draft until the deterministic output drift is understood.

ZacharyZcR:
Follow-up on the output drift noted above:

I isolated the difference with fresh-engine A/B runs.

- Disabling MTP (`DRAFT=0`) did not remove it.
- Forcing the same int4 IDOT threshold (`I4S=1`) did not remove it.
- Request B alone on a fresh engine and request B after request A produced the same text.
- Request B in a two-row decode batch produced a different, valid continuation.
- KV slots remained isolated, outputs were reproducible for a fixed batch shape, and CUDA backend correctness tests passed for q8/q4/q2/f32.
- I also tested preserving per-row routed-expert accumulation order; it did not change the result, so that experimental change was discarded.

Conclusion: this is batch-shape numerical non-determinism amplified by MoE routing / close logits, not evidence of KV contamination or a broken request boundary. Exact token identity across different batch shapes should not be treated as a correctness invariant. The appropriate future regression test is a hidden-state/logit tolerance test on a tokenizer-equipped fixture, plus exact reproducibility for the same batch shape.

The performance result remains: 32 GB host hot-store + 20 GB CUDA expert tier completed the two-request run in 33.96 s versus 46.82 s for CPU experts (~27.5% lower wall time), while an oversized hot-store caused swap and regressed badly.

ZacharyZcR:
Benchmark correction: the earlier 33.96 s hot-tier result was not a strict clean-KV baseline. Later log inspection showed persisted per-slot KV state can be restored unless `KVSAVE=0` is set.

A new clean run with `KVSAVE=0`, restored usage snapshot, RTX 5090 GPU0, 32 GB RAM hot tier, 20 GB CUDA expert tier, and the same two prompts completed in 48.44 s on the current #30 path. Request isolation and functional conclusions are unchanged, but 33.96 s should not be used as the canonical performance number.

This also gives future CUDA work an honest baseline: a fused per-expert pipeline completed the same clean integration workload in 48.11 s (~0.7%, noise), showing that cross-expert grouped launch—not merely fusing the three matrices—is required for concurrent throughput gains.

ZacharyZcR:

PR #32, opened 2026-07-10
Add runtime console and KV session UI

ZacharyZcR:
## Summary

- add a compact runtime console for active capacity, queued requests, completed generations, failures, and discovered KV slots
- poll `/health` while connected and pause polling when the page is hidden
- gracefully fall back to ordinary OpenAI-compatible mode when runtime health is unavailable
- discover optional `kv_slots` and keep one independent visible conversation per cache slot
- send the `cache_slot` extension only when the backend advertises support
- surface queue wait time from response headers
- stop persisting API keys and remove keys left by earlier versions
- cancel stale probes, abort work on unmount, and invalidate connection state when endpoint credentials change
- clamp generation controls and avoid accidental sends during IME composition
- improve focus, error, toggle, reduced-motion, long-name, and sidebar-overflow accessibility

## Compatibility

The panel accepts both boolean and numeric scheduler activity schemas. Scheduler data and KV extensions remain optional, so generic OpenAI-compatible endpoints continue to work as a plain chat client. KV session controls appear only when the server reports slots.

## Validation

- rebased onto current `main`
- clean `npm ci`: 0 vulnerabilities
- `npm test`: 3 files, 17 tests passed
- `npm run build`: TypeScript and Vite production build passed
- production output: 237.36 kB JS / 75.56 kB gzip, 20.46 kB CSS / 5.27 kB gzip
- tests cover health URL resolution, boolean/numeric active counts, conditional cache slots, API behavior, and credential storage
- `git diff --check`

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -16,3 +16,12 @@ Local validation:
 npm test
 npm run build
 ```
+
+The test suite stays browser-light: API requests use a mocked `fetch`, while
+runtime capability and storage behavior are covered through pure helpers. It
+checks that `/health` is resolved next to (not below) the OpenAI `/v1` prefix,
+supports both boolean and numeric `scheduler.active` responses, and sends the
+colibrì-specific `cache_slot` field only when KV-slot support was advertised.
+
+The endpoint and selected model are persisted locally. API keys are intentionally
+memory-only; startup/persistence also removes the legacy `colibri.apiKey` value.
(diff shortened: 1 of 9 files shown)


JustVugg (approved):
Isolated to `web/` — no engine, no CLI, no C. It enhances the browser UI (App.tsx + new runtime/storage libs) and ships its own tests (api.test, runtime.test, storage.test). I can't run the JS suite on this box (no npm toolchain here), but the blast radius is entirely the web frontend that #23 established — it can't affect the engine or terminal users — and the tests are included for whoever does run `npm test`. Approving on the diff + isolation. Merging.

PR #43, opened 2026-07-11
Reduce allocation overhead in quantized and MoE inference

0xhis:
## Summary

Cut a few temporary allocations from the CPU inference path.

`matmul_qt()` now keeps thread-local scratch space for quantized activations and row scales. The buffers grow when needed and are reused on later calls.

The MoE router now applies sigmoid to the logits in place, so it needs two temporary float arrays instead of three. The expert union bitmap also lives on the stack instead of going through `calloc()` and `free()` once per sparse layer.

The math and routing order are unchanged.

## Validation

- [x] `make -C c check`
- [ ] CUDA changes were tested with `make -C c cuda-test` (not applicable)
- [ ] Performance claims include hardware, commands, and repeatable measurements (no full-model performance claim)

I also ran the repository's tiny GLM oracle:

```sh
SNAP=./glm_tiny REF=./ref_glm.json TF=1 \
  DRAFT=0 RAM_GB=4 OMP_NUM_THREADS=1 \
  ./glm 8 16 16
```

It still matches all teacher-forcing positions:

```text
PREFILL (teacher-forcing) C vs oracolo: 32/32 posizioni
```

A local kernel check showed less overhead for batched int8 calls, while the single-row result was within noise. I left that out as a headline result because it does not represent full-model throughput.

## Compatibility

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -453,6 +453,22 @@ static void matmul_i4_idot(float *y, const int8_t *xq, const float *sx, const ui
         for(int s=0;s<S;s++) y[(int64_t)s*O+o]=(float)dot_i4i8(w,xq+(int64_t)s*I,I)*sc*sx[s]; }
 }
 
+typedef struct { int8_t *xq; size_t xq_cap; float *sx; size_t sx_cap; } QScratch;
+static _Thread_local QScratch g_qscratch;
+static void quant_scratch(size_t xn, size_t sn, int8_t **xq, float **sx){
+    if(xn>g_qscratch.xq_cap){
+        int8_t *p=realloc(g_qscratch.xq,xn);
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Verified in a worktree: clean build, oracle 32/32 TF + 20/20 greedy (f32), int8 IDOT path bit-identical to main (14/20 vs quantization reference — same as main). The thread-local scratch is the right fix in the right place: matmul_qt is the hottest call site in the engine and this removes a malloc/free per call (we measured allocator/syscall overhead being real in this path when building PILOT). `_Thread_local` keeps it OMP-safe, the overflow guard is correct, and trimming the redundant `sig` buffer in moe() is a nice touch. Approved — merging.

PR #40, opened 2026-07-10
Windows 11 native port (MinGW-w64) — Phase 1: compiles, correct, static-linked

nalepy:
## Summary

Native Windows 11 x86-64 port of the colibrì CPU engine via MinGW-w64/GCC. No WSL required. The engine builds as a static-linked `.exe` with zero DLL dependencies beyond Windows built-ins (KERNEL32 + UCRT).

## What's ported (all in `c/compat.h`, zero engine logic changes)

| POSIX | Windows |
|-------|---------|
| `pread` | `ReadFile` + `OVERLAPPED` (thread-safe, 64-bit offset) |
| `posix_fadvise` | no-op |
| `posix_memalign` | `_aligned_malloc` |
| `/proc/meminfo` | `GlobalMemoryStatusEx` |
| `rename` | `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` (CRT fails EEXIST) |
| `getrusage` | `GetProcessMemoryInfo` (PeakWorkingSetSize) |
| `getline` | reimplementation via `fgetc` |
| `setenv` | `SetEnvironmentVariableA` |

Architecture is unchanged: each platform difference lives in `compat.h`; `.c`/`.h` bodies stay clean. The same shims serve `glm.exe`, `olmoe.exe`, `iobench.exe`, and every future engine.

## Verification (all passing on Windows 11, GCC 16.1.0)

- **C tests**: test_json, test_st, test_tier — all pass
- **Python tests**: 27/27 pass (OpenAI server, resource plan, CUDA parser)
- **Tiny oracle**: **32/32 token-exact** (teacher-forcing vs transformers reference)
- **Rename EEXIST**: MoveFileEx correctly replaces existing destination
- **>4GB pread offset**: 64-bit OVERLAPPED fill verified at 5 GiB
- **RAM auto-budget**: GlobalMemoryStatusEx → engine auto-sizes correctly
- **Static-linked**: zero MinGW DLLs, runs on clean Windows 10/11

## Build

```powershell
# Toolchain: scoop install mingw-winlibs  (or MSYS2 MinGW-w64)
cd c
make glm.exe           # GLM-5.2 engine (static)
make olmoe.exe         # OLMoE engine
make test-c            # run C tests
make test-python       # run Python tests
```

## Files changed

| File | Change |
|------|--------|
[...]

PR #24, opened 2026-07-10
Add Tauri desktop shell

ZacharyZcR:
## Relationship to the Web UI

Builds on the shared `web/` interface merged in #23. This PR contains only the native desktop layer and does not duplicate the frontend.

## Summary

- add a minimal Tauri v2 Rust application under `desktop/src-tauri/`
- reuse the shared Vite frontend through `beforeDevCommand`, `beforeBuildCommand`, and `frontendDist`
- configure a constrained desktop capability and CSP for the local OpenAI-compatible API
- add macOS, Windows, PNG and Linux bundle icons derived from the existing brand asset
- document desktop development and packaging
- keep the permission boundary narrow: no model bundling/download, engine process control, shell, filesystem, or process capability

## Validation

- rebased onto current `main`
- clean frontend `npm ci`: 0 vulnerabilities
- frontend tests passed
- TypeScript/Vite production build passed
- `cargo fmt --check`
- `cargo check --locked`
- `cargo clippy --locked -- -D warnings`
- `cargo build --locked`
- full `@tauri-apps/cli build` passed
- release application binary produced
- Linux bundles produced successfully:
  - Debian package
  - RPM package
  - AppImage
- fixed a real AppImage packaging failure by explicitly declaring square PNG and platform icon assets
- `git diff --check`

PR #44, opened 2026-07-11
Prevent hot tiers from causing swap pressure

ZacharyZcR:
## Summary

- clamp explicit `PIN_GB` to a configurable share of the planner's expert-memory budget instead of letting it bypass RAM safety
- reserve 75% of expert memory for adaptive LRU, transient rows, and buffered disk I/O by default
- bound the planned VRAM hot tier by its conservative RAM backing budget
- sample live `MemAvailable` and `SwapFree` at safe between-turn points
- demote the coldest RAM/VRAM experts when memory falls below reserve or swap grows
- preserve opt-outs for controlled benchmarks through `PIN_FRACTION=1` and configurable `RAM_RESERVE_GB`
- add dependency-free policy tests

## Motivation

A GLM-5.2 744B test with a 100 GB fixed hot store drove 25-34 GB of swap and regressed two-request wall time to 73.65 s. The same workload with a 32 GB hot store used no process swap and completed in 33.96 s. Capacity alone is not a safe placement policy.

## Validation

- `make check`: portable build, 4 C tests, 27 Python/API tests
- `git diff --check`

## Draft gate

Keep Draft until the 744B host validates startup clamping, live demotion, VRAM accounting, and no-swap behavior under induced pressure.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -37,7 +37,7 @@ CUDA_ARCH ?= native
 NVCCFLAGS ?= -O3 -std=c++17 -arch=$(CUDA_ARCH) -Xcompiler=-Wall,-Wextra
 PYTHON    ?= python3
 CUDA_OBJ  =
-TEST_BINS = tests/test_json tests/test_st tests/test_tier
+TEST_BINS = tests/test_json tests/test_st tests/test_tier tests/test_memory_pressure
 ifeq ($(CUDA),1)
 ifeq ($(UNAME_S),Darwin)
 $(error CUDA=1 is supported only on Linux)
@@ -49,7 +49,7 @@ endif
 
 all: glm
 
-glm: glm.c st.h json.h tok.h tok_unicode.h compat.h $(CUDA_OBJ)
+glm: glm.c st.h json.h tok.h tok_unicode.h compat.h tier.h memory_pressure.h $(CUDA_OBJ)
 	$(CC) $(CFLAGS) glm.c $(CUDA_OBJ) -o glm $(LDFLAGS)
 
 backend_cuda.o: backend_cuda.cu backend_cuda.h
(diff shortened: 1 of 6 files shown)


ZacharyZcR:
744B GPU validation completed on RTX 5090:

- `PIN_GB=100 RAM_GB=160` was clamped to 35.9 GB (25% of the measured 143.6 GB expert budget).
- The engine loaded 1898 fixed RAM experts and 1057 CUDA-hot experts (19.99 GB), completed generation, and left process swap unchanged.
- Peak reported RSS was 64.59 GB for the short validation request; the previous unbounded 100 GB policy had driven 25-34 GB swap.
- With `RAM_RESERVE_GB=240` to induce pressure, a normal OpenAI API request completed and the safe-point controller demoted 64 cold RAM/VRAM experts.
- GPU residency dropped accordingly, and shutdown returned all six GPUs to 2 MiB and host MemAvailable to 245 GiB.
- The test also found and fixed an existing gap: normal API/interactive turns did not call `repin_pass()`; only the `MORE` branch did.

The Draft gate is satisfied: startup clamp, ordinary-request pressure sampling, cold demotion, CUDA accounting, no-swap behavior, and cleanup were all exercised on the full model.

JustVugg:
Maintainer note on the whole #44→#59 performance stack (+ the un-drafted #24/#30/#32/#33): this is excellent work arriving faster than review capacity, so here's the plan to land it *safely* — the engine's stability guarantee (oracle-exact vs transformers) is the project's core asset and every merge keeps it.

**Sequencing:**
1. Each PR gets the full treatment in numbered order — worktree build, 0-warning check, oracle 32/32 TF + 20/20 greedy, targeted validation of the feature — starting with this one (#44, swap pressure) and #46 (prefetch lifecycle safety) as they're the safety nets for the rest.
2. **Two merges just landed that touch your hot paths**: #43 (thread-local quant scratch in `matmul_qt`, trimmed moe temporaries) and #40 (Windows/MinGW compat shims). Expect rebases — sorry, and worth it: #43's scratch reuse probably composes well with your CPU-pool PRs.
3. One ask for the CPU-pool sub-series (#54–#58): they rework expert dispatch aggressively — please include a REPLAY A/B (your own fixture harness!) in each PR description so the perf claim is reproducible here without the 370 GB model.
4. Heads-up for conflict planning: a **quantization container v2** (GPTQ, 3-bit group-64 experts — measured 5× lower error than absmax at equal bits) is in local validation and will eventually touch `qt_alloc`/`matmul_qt`/converter. Your kernel-adjacent PRs land first; v2 will rebase on top.

Nothing here is a no — it's a queue. The stack is exactly the kind of work the project needs.

ZacharyZcR:
Closing during stack cleanup. The newer resource planner and bounded tier work in #51/#61 supersede this older conflicting branch.

PR #51, opened 2026-07-11
Define quality-preserving unified memory policies

ZacharyZcR:
## Summary

- define Colibri as a quality-preserving VRAM/RAM/storage hierarchy
- upgrade the resource plan to report independent hot, warm, and cold expert tiers
- explain placement decisions and the expected runtime bottleneck
- add `quality`, `balanced`, and explicit `experimental-fast` policy modes
- reject lossy TOPK/TOPP router overrides unless experimental mode is selected

## Why

Memory pressure may reduce speed, but the default runtime must not silently reduce model precision or alter router semantics. Disk remains an immutable recovery tier; VRAM experts no longer require duplicate RAM backing after #50.

## Validation

- `make check`
- 3 C tests
- 29 Python/API/planner tests
- direct runtime validation for policy rejection and experimental opt-in

## Dependency

Depends on #50. The commit specific to this PR is `91e610c`; rebase after #50 merges.

PR #50, opened 2026-07-11
Parallelize expert execution across multiple GPUs

ZacharyZcR:
## Summary

- execute independent routed-expert groups concurrently across selected CUDA devices
- release RAM backing after experts are resident in VRAM, with disk reload on CUDA fallback
- fill unused multi-GPU VRAM after placing profile-ranked experts and disable MTP draft by default on CUDA
- document six-card configuration, tuning controls, and measured GLM-5.2 int4 results

## Results

Six RTX 5090 32 GB cards, GLM-5.2 int4, 150 GB expert tier:

- 64-token varied prompt: 0.94 token/s, 87.8% expert hit rate
- warmed short prompt: 1.64 token/s, 99.3% expert hit rate
- uniform (non-profiled) placement: 0.29 token/s, showing that heat-ranked placement is essential
- host RAM backing was released after upload; no swap growth during the run

## Validation

- `make check` (3 C tests and 27 Python/API tests)
- CUDA build on the six-card host
- `backend_cuda_test 0 1 2 3 4 5`
- full GLM-5.2 int4 inference on all six devices

## Dependency

Depends on #45. The commits specific to this PR are `808613d` and `cf99614`; the branch can be rebased after #45 merges.

Closes no issue.

PR #52, opened 2026-07-11
Overlap cold expert loading with resident compute

ZacharyZcR:
## Summary

- load cold experts in a bounded background I/O pool
- execute already-resident RAM/VRAM experts while those reads are in flight
- join cold experts before completing the layer without changing weights, quantization, or router choices
- preserve full I/O fan-out when a block has no foreground resident work
- report disk service time separately from foreground-visible wait time

## Full-model result

Six RTX 5090 cards, GLM-5.2 int4, original top-8 router, 150 GB q4 VRAM tier:

- strict seed/config baseline: 65.28 s for 32 tokens (0.49 token/s)
- deferred I/O pipeline: 60.75 s for 32 tokens (0.53 token/s)
- approximately 6.9% lower wall time in this cold-profile run
- generated text remained coherent; no quantization or routing reduction was used

The gain is intentionally modest: the implementation hides resident work behind cold reads but does not claim to solve poor cache placement.

## Validation

- `make check` (3 C tests, 29 Python/API/planner tests)
- CUDA build on the six-card host
- backend correctness on one and six RTX 5090 devices
- full GLM-5.2 int4 generation

## Dependencies

Depends on #51, which depends on #50. The commit specific to this PR is `9320d51`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,13 @@ Disk is an immutable recovery source, not a normal decode target. If the plan
 leaves cold expert bytes on disk, speed depends on cache hit rate; output
 quality does not.
 
+Cold expert reads use a deferred pipeline: resident RAM/VRAM experts execute
+while missing experts are loaded in a bounded background I/O pool, then the
+cold results join before the layer completes. `IO_THREADS=n` overrides the
+default eight loader threads when foreground work exists. Profiling reports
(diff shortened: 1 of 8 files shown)


ZacharyZcR:

PR #53, opened 2026-07-11
Adapt expert placement with session-local LFRU

ZacharyZcR:
## Summary

- track per-layer session recency separately from persistent `.coli_usage` frequency
- select live hot-tier replacements with a frequency-first LFRU score
- retain the existing 25% + 4-count hysteresis and four-swaps-per-pass bound
- run live repin after normal API/chat requests, fixing the previous `:MORE`-only behavior
- enable lossless `REPIN=64` automatically for the `balanced` policy while respecting explicit overrides

## Quality contract

This changes placement only. Checkpoint quantization, top-k routing, expert weights, and layer outputs are unchanged. `quality` keeps live replacement disabled by default; `REPIN=0` disables it under every policy.

## Validation

- `make check`
- 3 C tests including LFRU frequency/recency/hysteresis behavior
- 30 Python/API/planner tests
- CUDA build on the six-card RTX 5090 host

## Dependencies

Depends on #52 -> #51 -> #50. The commit specific to this PR is `5eb2ceb`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,13 @@ Disk is an immutable recovery source, not a normal decode target. If the plan
 leaves cold expert bytes on disk, speed depends on cache hit rate; output
 quality does not.
 
+Cold expert reads use a deferred pipeline: resident RAM/VRAM experts execute
+while missing experts are loaded in a bounded background I/O pool, then the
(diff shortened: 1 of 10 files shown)


ZacharyZcR:

PR #54, opened 2026-07-11
Fuse q4 expert gate and up CPU dispatch

ZacharyZcR:
## Summary

- execute single-token q4 gate and up projections in one OpenMP dispatch
- retain the existing exact AVX2/NEON per-row arithmetic and output order
- use the fused path for resident RAM experts, CUDA fallback, and deferred cold experts
- keep all non-q4 and batched paths unchanged

## Full-model A/B

Six RTX 5090 cards with GLM-5.2 int4, original top-8 router, greedy decode, and identical output:

- baseline expert matmul: 55.603 s
- fused dispatch expert matmul: 51.611 s
- targeted expert time reduction: 7.2%
- short 16-token wall time was neutral within system variance (128.00 s vs 129.42 s) because attention moved in the opposite direction

This PR deliberately does not claim an end-to-end speedup. It removes one of three OpenMP team launches per q4 RAM expert and prepares the path for a persistent native thread pool.

An AVX512 implementation was also tested and rejected: it slowed this dual-socket Xeon by about 18% due to frequency and memory-bandwidth effects. That experiment is not included.

## Validation

- `make check` (3 C tests, 30 Python/API/planner tests)
- CUDA build on the six-card host
- full q4 GLM-5.2 greedy A/B with identical generated text and routing statistics

## Dependencies

Depends on #53 -> #52 -> #51 -> #50. The commit specific to this PR is `7450716`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,13 @@ Disk is an immutable recovery source, not a normal decode target. If the plan
 leaves cold expert bytes on disk, speed depends on cache hit rate; output
 quality does not.
 
+Cold expert reads use a deferred pipeline: resident RAM/VRAM experts execute
+while missing experts are loaded in a bounded background I/O pool, then the
+cold results join before the layer completes. `IO_THREADS=n` overrides the
+default eight loader threads when foreground work exists. Profiling reports
+both disk service time and the smaller foreground-visible wait time so overlap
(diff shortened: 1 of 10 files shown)


ZacharyZcR:

PR #58, opened 2026-07-11
Run q4 shared experts on the CPU pool

ZacharyZcR:
## Summary

- reuse the persistent q4 CPU pool for single-token shared-expert gate, up, and down projections
- keep prefill batches and all non-q4 paths unchanged
- preserve exact weights, router behavior, and arithmetic order

## Full-model A/B

Six RTX 5090 cards, GLM-5.2 int4, original top-8 router, 48 pool workers, identical greedy output:

| pool coverage | other time | 16-token wall |
|---|---:|---:|
| routed gate/up/down | 23.05 s | 105.07 s |
| routed + shared experts | 10.89 s | 92.17 s |

This step reduced wall time by another 12.3%. Relative to the fused OpenMP baseline before the persistent pool, cumulative wall time fell from 129.42 s to 92.17 s (28.8%). Routed expert time stayed stable at approximately 25.5 s, isolating the gain to shared-expert execution.

## Validation

- `make check` (4 C tests, 30 Python/API/planner tests)
- CUDA build on the six-card host
- full q4 GLM-5.2 greedy A/B with identical output

## Dependencies

Depends on #56 -> #55 -> #54 -> #53 -> #52 -> #51 -> #50. The commit specific to this PR is `c7ffcbe`.

PR #59, opened 2026-07-11
Add phase-aware expert I/O scheduling

ZacharyZcR:
## Summary
- split expert loading concurrency between foreground-overlap and I/O-only phases
- preserve `IO_THREADS` as a backwards-compatible override
- add focused policy tests and document the new controls

## Validation
- `cd c && make check`

## Stack
- Continues #58 and includes its commits until the dependency lands.


ZacharyZcR:
Closing during PR-stack cleanup. The measured best path is consolidated in #61; this intermediate branch is no longer needed independently.

PR #60, opened 2026-07-11
Budget VRAM safely for opt-in dense CUDA

ZacharyZcR:
## Summary
- account for the projected dense tensor footprint only when `CUDA_DENSE=1` is explicitly requested
- keep the default planner on the faster routed-expert-only layout
- preserve explicit environment overrides and add planner coverage
- keep the synchronized Windows Makefile free of a self-dependency warning

## Validation
- `cd c && make check`
- six RTX 5090, q4, 64-token text generation: dense CUDA 0.37 tok/s versus routed-only 0.27 tok/s on the current stack; the historical 0.94 tok/s baseline did not reproduce because current CPU attention regressed to 93.47s

## Stack
- Continues #59 and includes its commits until the dependency lands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,13 @@ one thread-team launch per RAM expert without activation requantization or a
 lower-precision fallback. It is a stepping stone toward a persistent native
 CPU expert pool, not a replacement for one.
 
+`CPU_POOL_THREADS=n` opts the fused q4 gate/up path into a persistent native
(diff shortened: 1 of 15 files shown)


ZacharyZcR:
Closing during PR-stack cleanup. The measured best path is consolidated in #61; this intermediate branch is no longer needed independently.

Issue #39, opened 2026-07-10
[Performance]: Framework Desktop Ryzen AI Max+ 128GB - CPU Only


tommy-bolger:
### Commit

9e6e1ac327b9cc07c4934d14343d5d5251eeddfc

### Hardware and storage

Framework Desktop Ryzen AI Max+ 128GB
- CPU: AMD Ryzen™ AI Max+ 395
- 128GB LPDDR5x
- Intel Optane 905p 1.5TB PCIe 3.0 x4

### Software environment

Ubuntu 26.04

Using the int8 MTP model: https://huggingface.co/mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp

```
     ▄▀▀▀▄  ▄        colibrì v1.0
  ▄▄▄▄▀▀▀▀▄▀▀        piccolo motore, modello immenso
      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU
        ▀▀▀▀         info
          ▀          
  ──────────────────────────────────────────────────────────
   modello    /home/<redacted>/colibri/glm52_i4-int8-MTP
   arch       hidden 6144 · 78 layer · 256 expert/layer · top-8
   shard      144 file · 384 GB su disco
   RAM        129 GB totali · 126.6 GB disponibili
   disco      705 GB liberi
   motore     pronto ✓
```

### Build and benchmark commands

```shell
buffered x8 thread: 64 letture x 19MB = 1.3 GB in 0.39s -> 3.27 GB/s (6.1 ms/blocco effettivi)
O_DIRECT x8 thread: 64 letture x 19MB = 1.3 GB in 0.49s -> 2.60 GB/s (7.7 ms/blocco effettivi)
```

```shell
`./coli run --model ~/colibri/glm52_i4-int8-MTP --ngen 64 --temp 0 "Explain in one short paragraph how a CPU-only streaming MoE model trades RAM, disk, and compute."`
`./coli run --model ~/colibri/glm52_i4-int8-MTP --ngen 64 --temp 0 "Explain in one short paragraph how a CPU-only streaming MoE model trades RAM, disk, and compute."`
```

### Results

```text
## 1st Run:

./coli run --model ~/colibri/glm52_i4-int8-MTP --ngen 64 --temp 0 "Explain in one short paragraph how a CPU-only streaming MoE model trades RAM, disk, and compute."

     ▄▀▀▀▄  ▄        colibrì v1.0
[...]

Issue #31, opened 2026-07-10
[Performance]: CPU-only benchmark on Ryzen 9 9950X / Samsung 9100 PRO PCIe 5.0


onepunk:
## Commit

`3e4d08b6bfe67b9bc17e8a0694b7c2604862612d`

## Hardware and storage

- CPU: AMD Ryzen 9 9950X 16-Core Processor
- Threads: 32 logical CPUs
- RAM: 128 GiB total
- Model storage: Samsung SSD 9100 PRO 1TB NVMe
- Storage link: PCIe 5.0 x4
- Filesystem: local ext4

## Software environment

- Host OS: Linux
- Build/runtime path: CPU-only default build, no CUDA
- Model location: `/models/glm52-colibri-int4`
- Engine: `GLM-5.2 · 744B MoE · int4 · streaming CPU`

## Build and benchmark commands

```sh
cd /home/colibri/c
./setup.sh
./coli info --model /models/glm52-colibri-int4
./coli run --model /models/glm52-colibri-int4 --ngen 64 --temp 0 "Explain in one short paragraph how a CPU-only streaming MoE model trades RAM, disk, and compute."
./coli run --model /models/glm52-colibri-int4 --ngen 64 --temp 0 "Explain in one short paragraph how a CPU-only streaming MoE model trades RAM, disk, and compute."
```

Disk throughput check on a representative shard:

```sh
gcc -O2 -fopenmp c/iobench.c -o /tmp/colibri-iobench
/tmp/colibri-iobench /models/glm52-colibri-int4/out-00000.safetensors 19 64 8 1
```

## Results

Warm-up policy: two back-to-back generation runs after relocating the model to the PCIe 5.0 drive. The learned expert-usage history was retained.
Run count: 2

Generation throughput:

- Run 1: `64 token in 228.22s (0.28 tok/s)`
- Run 2: `64 token in 230.59s (0.28 tok/s)`
- Median throughput: `0.28 tok/s`

Run 1 profile:

- RSS: `82.15 GB`
- Expert hit-rate: `56.7%`
- Profile split: `expert-disk 73.674s | expert-matmul 130.971s | attention 14.328s | lm_head 0.017s | altro 9.233s`

Run 2 profile:

- RSS: `82.05 GB`
- Expert hit-rate: `57.2%`
- Profile split: `expert-disk 73.290s | expert-matmul 132.894s | attention 15.107s | lm_head 0.017s | altro 9.283s`

Shared decode stats:

[...]

PR #61, opened 2026-07-11
Plan physical CPU cores and batch grouped CUDA experts

ZacharyZcR:
## Summary
- detect physical CPU cores and bind OpenMP workers across cores instead of oversubscribing SMT
- replace per-expert CUDA launch loops with four grouped launches per device
- stage a disjoint expert placement: upload and release the VRAM-ranked prefix before loading the RAM-ranked suffix
- preserve full router decisions and checkpoint quantization

## Validation
- `cd c && make check`
- CUDA correctness: q8/q4/q2/f32 passed on RTX 5090
- six RTX 5090 + 2x Xeon Silver 4510, GLM-5.2 q4, 64-token text generation
  - 48 SMT threads: 0.27 tok/s
  - 24 bound physical cores: 0.92 tok/s
  - grouped CUDA kernels: 0.95 tok/s
  - disjoint 150 GB VRAM + 130 GB RAM placement: 1.13 tok/s
  - expert hit rate: 91.2% -> 94.8%
  - visible disk wait: about 36s -> 21.25s
  - startup minimum MemAvailable: 38 GB -> 97 GB

## Stack
Continues #50-#54. Supersedes the abandoned persistent CPU-pool, prefetch, and dense-CUDA experiment branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ COLI_CUDA=1 COLI_GPU=0 CUDA_EXPERT_GB=16 \
 PIN=stats.txt PIN_GB=160 SNAP=/nvme/glm52_i4 ./glm 64 4 4
 # multi-GPU expert tier, 150 GB total budget across six 32 GB devices
 COLI_CUDA=1 COLI_GPUS=0,1,2,3,4,5 CUDA_EXPERT_GB=150 \
-PIN=stats.txt PIN_GB=150 SNAP=/nvme/glm52_i4 ./glm 64 4 4
+PIN=stats.txt PIN_GB=280 RAM_GB=226 SNAP=/nvme/glm52_i4 ./glm 64 4 4
 ```
 
 Selected experts are uploaded during startup, so capacity failures occur before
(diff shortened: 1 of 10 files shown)


ZacharyZcR:

JustVugg (changes requested):
Huge work, and the headline result is real and valuable — but one change blocks merge and the PR should be split. Details:

**The win I want: physical-core OMP sizing.** Your measured 0.27 → 0.92 tok/s (48 SMT threads → 24 bound physical cores) is the biggest single CPU number in the whole tracker, and it helps *every* CPU user. This should land — ideally on its own.

**The blocker: `COLI_POLICY=quality` (the default) turns `--topp 0.7` into a hard error.** Verified: `TOPP=0.7 ./glm` now exits with "usa COLI_POLICY=experimental-fast esplicitamente", and `coli run --topp 0.7` inherits the default quality policy and aborts. That breaks:
- the README benchmark instructions (`--topp 0.7` is documented as a measured ~1.6× win, #2),
- four community members mid-benchmark right now (#12, #31, #39, #65 all run `--topp 0.7`),
- every recommendation we've given for warm machines.

I understand the intent — top-p *is* lossy (it drops low-weight experts, altering the router decision), so gating "I accept quality loss" behind an explicit policy is philosophically right. But a **hard default error** is too aggressive for a flag this established. Please make it a **one-line warning** ("top-p trades a little quality for ~1.6× fewer expert reads") instead of a fatal exit, OR default to `balanced` (which permits it). Silent breakage of a documented flag is the one thing we can't ship.

**The split request:** this PR bundles the CPU core-binding (testable, universal — I validated build + oracle 32/32 TF + 20/20 greedy here) with grouped-CUDA kernels + disjoint VRAM/RAM placement (I can't validate on this GPU-less box — taking those on your 6×5090 runs). Landing the core-binding + LFRU separately from the CUDA placement would let the CPU win ship now with full local validation, and keep the CUDA parts reviewable on their own. Your call on the mechanics, but the topp fix is required either way.

Everything else reads well — the LFRU scoring, the disjoint placement staging (38→97 GB startup MemAvailable is a nice safety win), the resource_plan integration. Fix the topp gate and this is close.

PR #64, opened 2026-07-11
iobench: fix heap corruption, 32-bit overflow, and offset range on Windows

nalepy:
Follow-up to #40. Three bugs found by actually running `iobench.exe` on Windows 11 — all three made the tool crash or lie:

## 1. Heap corruption (crash, exit 0xC0000374)

`iobench.c` frees the `posix_memalign`'d buffer with plain `free()`. On Windows `posix_memalign` maps to `_aligned_malloc` (compat.h), whose memory must be released with `_aligned_free` — mixing them corrupts the CRT heap and kills the process. Fixed with `compat_aligned_free` (plain `free` on POSIX, `_aligned_free` on Windows).

## 2. Negative GB totals (32-bit overflow)

`long tot` is 32-bit on Windows (LLP64), so any run reading >2 GB overflowed:

```
buffered x8 thread: 128 letture x 19MB = -1.7 GB in 0.47s -> -3.75 GB/s
```

Now `int64_t`.

## 3. Benchmark measured RAM, not disk

`RAND_MAX` is 32767 on Windows, so `rand()*4096` generated offsets only in the **first 134 MB** of the file. After one pass the page cache holds that whole window — the tool reported ~6 GB/s (RAM) instead of the true ~2.1 GB/s (NVMe). Two `rand()` calls combined into 30 bits now span the whole file. Same code path on all platforms; the seeded sequence stays deterministic.

## Measured before/after (laptop PCIe4 NVMe, NTFS, 19 MB expert-sized blocks, 8 threads)

| | reported |
|---|---|
| before (offsets in first 134 MB, cached) | 6.05 GB/s ❌ |
| after (full-range offsets) | **2.05–2.23 GB/s** ✅ |

This matters because iobench is the gate for the Phase-2 `FILE_FLAG_NO_BUFFERING` decision — with the inflated number, buffered NTFS looks like it beats O_DIRECT and Phase 2 would be wrongly skipped.


JustVugg (approved):
Clean, correct, and it improves the Linux measurement too. All three fixes are real: (1) Windows RAND_MAX=32767 meant a single rand()*4096 only reached the first ~134 MB — page-cached, so a *false* high reading; the 30-bit combined rand now covers the whole file (this makes the Linux offset spread better as well); (2) `long` is 32-bit under Windows LLP64 so the byte total overflowed past 2 GB — int64_t is right; (3) posix_memalign maps to _aligned_malloc on Windows so plain free() corrupts the heap — compat_aligned_free matches. Built and smoke-tested here, no warnings. Approved.

PR #45, opened 2026-07-11
Fuse and group CUDA expert execution

ZacharyZcR:
## Summary

- execute CUDA-hot expert gate/up/SiLU/down as one device-resident pipeline
- group decode-sized expert work per GPU into one packed H2D and one packed D2H transfer
- retain CPU and per-expert CUDA fallbacks
- bound grouped scratch to `S <= 64`; long prefill uses the fused per-expert path and cannot allocate `S*K*D` grouped buffers
- reuse per-device activation buffers
- expose grouped calls, experts, rows, and optional H2D/kernel/D2H timings
- extend CUDA correctness coverage with fused and grouped two-row oracles

## Full-model evidence

GLM-5.2 744B INT4, RTX 5090 GPU0, 32 GB RAM hot tier, 20 GB CUDA expert tier:

- existing single-request path: 23.35 s generation, 14.21 s expert compute, 0.26 tok/s
- fused pipeline: 20.06 s generation, 10.95 s expert compute, 0.30 tok/s
- improvement: 14.1% generation, 22.9% expert compute

Profiled short run:

- 130 group calls
- 489 CUDA experts / 616 routed rows
- 3.76 experts per group
- H2D 5.8 ms, kernels 45.0 ms, D2H 3.6 ms total

Clean-KV continuous-batching integration against #30:

- existing path: 48.44 s
- grouped runs: 46.85 s and 51.22 s
- no statistically meaningful end-to-end difference because disk misses and CPU-resident experts dominate this short workload

The implementation therefore claims reduced transfer/synchronization overhead for CUDA-resident experts and a measured single-request gain. It does not claim a concurrent throughput win on a 32 GB card.

## Validation

- `make check`: portable build, C tests, and Python/API tests
- CUDA q8/q4/q2/f32, fused MLP, grouped execution and counters passed on RTX 5090 sm_120
- CUDA-enabled full runtime build passed after the bounded-scratch change
- full 744B generation completed without fallback errors
- `git diff --check`

Issue #65, opened 2026-07-11
[Performance]: many runs with Intel 265K, 128GB DDR5, 9100 Pro 8-10GB/s direct  (+980 Pro 6GB/s direct)


zacaj:
### Commit

a5fc89e88f113fc9d1c9d8752861b158d7c303e7

### Hardware and storage

Processors: 20 × Intel® Core™ Ultra 7 265K
Memory: 128 GiB DDR5 6600Mhz RAM 
Storage: Samsung Electronics Co Ltd NVMe SSD 9100 PRO [PM9E1], btrfs

### Software environment

Operating System: Aurora 44
KDE Plasma Version: 6.7.2
KDE Frameworks Version: 6.27.0
Qt Version: 6.11.1
Kernel Version: 7.0.11-200.fc44.x86_64 (64-bit)
Graphics Platform: X11
gcc version 16.1.1

### Build and benchmark commands

```shell
./iobench ~/tmp/GLM-5.2/out-00003.safetensors 19 64 8 1
`O_DIRECT x8 thread: 64 letture x 19MB = 1.3 GB in 0.15s -> 8.71 GB/s (2.3 ms/blocco effettivi)`

./iobench ~/tmp/GLM-5.2/out-00003.safetensors 19 64 8 0
`buffered x8 thread: 64 letture x 19MB = 1.3 GB in 0.18s -> 6.90 GB/s (2.9 ms/blocco effettivi)`


Note: you should probably use larger numbers for these, or require multiple runs, because eg, if I just repeat that buffered command again I get much better numbers (assuming os cache?)  
`buffered x8 thread: 64 letture x 19MB = 1.3 GB in 0.09s -> 13.48 GB/s (1.5 ms/blocco effettivi)`

Running a few times I can even get this up to 20GB/s

Even with direct there's a notable warm up period (nvme cache?).  Running again shortly after I get 9.91GB/s, and then 10.69GB/s.  Not as dramatic a jump as with buffered but worth noting or making clear in the instructions to use a fresh file/etc
```

### Results

```text
Ran using various combinations of `./coli run --ngen 32 --temp 0 "Explain in one short paragraph how a CPU-only streaming MoE model trades RAM, disk, and compute."`

[colibri2.log](https://github.com/user-attachments/files/29927434/colibri2.log)

[...]

PR #68, opened 2026-07-11
Unify continuous batching and heterogeneous runtime acceleration

ZacharyZcR:
## Summary

This consolidates the previously stacked runtime PRs into one reviewable performance line. It keeps the quality-preserving VRAM/RAM/disk policy while adding continuous decode batching and the validated CPU/CUDA execution improvements.

### Runtime and scheduling

- decode-only continuous batching with independent KV slots, cancellation, bounded admission, and automatic slot allocation
- physical-core-aware CPU planning
- session-local expert placement and overlap of cold expert loading with resident work
- disjoint VRAM/RAM placement with host backing release

### CUDA execution

- multi-GPU expert execution and grouped per-device transfers
- fused expert MLP plus packed low-row int4 kernels
- native int4 Tensor Core path guarded by an 8-row quality/performance threshold
- persistent pinned staging and per-device non-blocking streams
- resident dense projections and CUDA MLA absorption core
- fused grouped gate/up projection

### Measured results on six RTX 5090 GPUs

- continuous batching: 1.273 -> 2.624 aggregate tok/s (+106.1%)
- resident dense projections: 1.650 -> 2.157 aggregate tok/s (+30.8%)
- CUDA MLA attention: 1.960 -> 2.198 aggregate tok/s (+12.1%)
- 300 GB combined expert hot set replay: 1.87 -> 2.16 tok/s (+15.7%)
- fused gate/up projection: 2.201 -> 2.223 aggregate tok/s (+1.02%)

Negative experiments are intentionally excluded: forced low-row Tensor Core dispatch, router-lookahead prefetch, CUDA Graph capture, NCCL tensor parallel on non-P2P RTX 5090 topology, and hard NUMA binding.

## Validation

- `make check`
- `make cuda-test CUDA_ARCH=sm_120` on RTX 5090
- real GLM-5.2 fixed-token replay
- paired two-request OpenAI API throughput tests
- semantic output checks for the CUDA attention path

## Supersedes

Supersedes #30, #45, #50, #51, #52, #53, #54, and #61. Those PRs are closed to keep the review queue small.


JustVugg (changes requested):
Thank you for consolidating the stack exactly as asked — one reviewable line instead of six overlapping branches is a big improvement, and it builds clean here with the oracle passing (32/32 TF, 20/20 greedy). The continuous batching, physical-core planning, session-local placement, and disjoint VRAM/RAM staging all read well. I'm ready to land this — **one line stands between us and merge.**

**`--topp 0.7` still hard-errors by default.** Verified on this branch: `TOPP=0.7 ./glm` exits with "usa COLI_POLICY=experimental-fast esplicitamente", and `coli run --topp 0.7` inherits the default `quality` policy and aborts. I see you documented the requirement rather than changing it — but the documentation now contradicts itself: the knobs list still says `--topp 0.7 adaptive expert top-p (30–40% less disk)` two paragraphs above the line that says it needs `--policy experimental-fast`.

I fully agree with your *reasoning* — top-p is lossy, it drops low-weight experts and changes the router's output, so a 'quality' policy flagging that is philosophically correct. The disagreement is only about the *mechanism*: a **fatal exit** vs a **one-line warning**. Please make it a warning:

```
fprintf(stderr, "[policy] --topp/--topk drop low-weight experts (~1.6x fewer reads, small quality cost)\n");
```

and proceed. Three concrete reasons this can't be a hard error:
- `--topp 0.7` is the single most-recommended flag in the project — it's in the README benchmark instructions and every warm-machine reply.
- **Four contributors are running it right now** (#12, #31, #39, #65). Their next `git pull` + rerun breaks with no obvious fix.
- The user passing `--topp` *is* the acknowledgment — they typed the lossy flag on purpose.

Flip that exit to a warning (keep the policy concept, keep experimental-fast for the router-editing TOPK path if you like) and I'll merge immediately. Everything else here is genuinely excellent work.

ZacharyZcR:
Addressed the requested policy behavior: explicit `--topp`/`--topk` flags now emit the suggested warning and proceed under the default policy. I also updated the README/CLI help to remove the contradictory opt-in requirement, rebased onto current `main`, and refreshed the stale doctor output assertion exposed by the rebase. `make check` passes (46 Python tests plus all C tests).

JustVugg:
The topp fix is exactly right — thank you. Confirmed in the diff: `--topp`/`--topk` now print `[policy] --topp/--topk drop low-weight experts (~1.6x fewer reads, small quality cost)` and **proceed**, instead of the fatal exit. That was the one blocker, so on content this is ready to land.

It just needs a rebase now — it went CONFLICTING because a batch of PRs merged into main since you opened it (the CPU perf line: OMP hot-team #77, PIPE #79, and others). Per the new workflow (thanks for the dev-branch idea): **please rebase onto `dev` and retarget the PR base to `dev`** — that's the integration branch now, and it's where I'll land this. Ping me when it's green; the review's already done, so it merges as soon as it applies cleanly. Great turnaround on the topp point.

ZacharyZcR:
@JustVugg Ping as requested — it's green: base retargeted to `dev`, reconciled through both of this week's dev waves (the #79 PIPE loader now serves as the base — this branch's pthread loader was retired in its favor — and yesterday's Metal/heap-scores merge is composed in). `MERGEABLE/CLEAN`, C suite green.

Same is true for the rest of the stack, in landing order: **#68 → #80 → #111**, all conflict-free on `dev` and each smoke-tested on the 6×5090 rig after reconciliation. #80 is out of draft as you asked.

JustVugg (approved):
The reconciliation is exactly right: retiring your pthread loader in favor of the #79 PIPE pool (no competing I/O pools — the thing I worried about with kryptt's #78) and composing in the #110 heap attention-scores fix, then re-smoke-testing on the 6×5090. That's the diligent path. Validated the CPU side here: build 0 warnings, oracle 32/32 TF + 20/20 greedy (default single-request path unchanged), and **SERVE still works** (clean READY…END — continuous batching doesn't break the stdio protocol). The topp policy fix (warning, not fatal) is confirmed too.

The CUDA/continuous-batching runtime I take on your 6×5090 + sanitizer validation — this is a GPU-less box, same trust model as the Metal backend and the #16 CUDA tier. It lands on **dev** to soak (not main) with the rest of the GPU work, gated behind COLI_CUDA. Merging #68 as the base of the stack; #80 and #111 next in order. Thank you for the careful reconciliation.

Issue #48, opened 2026-07-11
Idea: grammar-aware draft source — forced tokens in constrained JSON output are free, guaranteed-accepted drafts (composes with DRAFT/MTP; engages even with the int4 head from #8)


fabio-rovai:
Follow-up to #47 (same hardware/workload). A speculation idea rather than a measurement — flagged as such — but the reasoning is mechanical enough to be worth writing down.

## Idea: grammar-aware draft source for constrained JSON/NDJSON workloads

For workloads where the output is structurally constrained (ours: one JSON object per line, fixed keys, enum values), a large fraction of tokens is **deterministic given a grammar**: braces, quotes, the key names themselves, separators, and enum values once their first token appears. E.g. every line of our output starts

```
{"id":"...","fit_category":"no_fit","fit_score":...
```

where the keys, punctuation, and most of the enum body are forced. Eyeballing our token streams, roughly a quarter to a third of generated tokens per line are structural/forced (varies with how long the free-text fields run).

Today `DRAFT=n` drafts from n-gram lookup (~5% acceptance on general text per your comment in glm.c) or the MTP head (needs int8, per #8). A third draft source — **the grammar itself** — has two properties the other two can't offer:

1. **Guaranteed acceptance for forced tokens.** Wherever the grammar admits exactly one legal continuation, the draft is correct by construction. No head, no lookup table, no acceptance gamble — verification in the batch-union forward just confirms it and pays one forward for several tokens.
2. **It works even when the MTP head is broken/absent.** On containers with the int4 head (#8), speculation currently never engages. Grammar drafts would engage regardless, on exactly the workloads (structured extraction, classification, tool output) where local CPU inference is most likely to be deployed.

And it composes: grammar supplies the forced spans, n-gram/MTP fill the free-text gaps between them.

## Sketch

- Accept a grammar (GBNF-style or a JSON-schema→grammar compile) alongside the prompt.
- At draft time, walk the grammar state: while exactly one token is legal (or one token holds ≥ some mass threshold under the grammar mask), append it to the draft; hand over to n-gram/MTP when the grammar branches.
- Verify as today via the existing batch-union path — the cold-cache caveat from your MTP notes applies identically (verified drafts still route to extra experts until the cache warms), so the adaptive guard should govern this source too.

## Expected effect (estimate, not measured)

[...]

Issue #47, opened 2026-07-11
Benchmark: Apple M3 Max · macOS · 128 GB · internal SSD (12.5 GB/s) — 0.35–0.45 tok/s; PILOT neutral; confirms #8; setup.sh libomp false-positive


fabio-rovai:
Benchmark datapoint for the README table, plus confirmation of #8 and one small setup.sh bug. Thanks for the engine — the per-token PROFILO line made all of this attribution possible.

## Hardware
**Apple M3 Max (12P+4E) · macOS 26.5 · 128 GB unified · internal SSD (APFS)**
`iobench`: **12.49 GB/s** O_DIRECT (8 threads × 19 MB blocks).
Container: `jlnsrk/GLM-5.2-colibri-int4` (144 shards, 379 GB).

## Results (greedy `--temp 0`, `--ngen 400`, ~1k-token prompt, fresh process per rung, page cache warm)

| config | tok/s | expert hit | RSS | expert loads/layer |
|---|---|---|---|---|
| cold first-ever run (24 tok) | 0.12 | 41.7% | 43.6 GB | — |
| baseline, `MTP=0` | 0.35 | 68.5% | 82.2 GB | 28.7 |
| + `PIN=<stats> PIN_GB=90` | 0.36 | 70.6% | 95.8 GB | 28.7 |
| + `PILOT=1` | 0.36 | 70.6% | 95.7 GB | 28.7 |
| + `DRAFT=4` | 0.34 | 70.4% | 97.5 GB | 29.4 |
| + `--topp 0.7` | **0.45** | 70.0% | 96.1 GB | **17.9** |

Baseline PROFILO per 400 tok: expert-disk 400 s · expert-matmul 311 s · attention 351 s · altro 89 s. (Attention ~30% of the token at 1–3k ctx on CPU was the surprise for us.)

**PILOT=1 on Apple Silicon: neutral** — same 0.36 tok/s, profile shows disk already saturated at our hit rate, matching your dev-box observation.

**`--topp 0.7`: +29%** (0.35→0.45), expert loads 28.7→17.9/layer — consistent with your README datapoint.

## MTP: confirms #8 (int4 head → 0% acceptance)
[...]

PR #70, opened 2026-07-11
Grammar-forced speculative drafts (#48): forced spans in constrained JSON output are free, guaranteed-accepted drafts

fabio-rovai:
Implements #48 — the grammar itself as a third draft source, next to MTP and n-gram lookup.

## What it does

For constrained-output workloads (JSON/NDJSON, function calling, structured extraction), wherever the grammar admits exactly **one** legal byte — braces, quotes, key names, separators, enum bodies — that forced span is tokenized and injected as drafts that the existing batch-union forward verifies. Properties, as discussed in the issue:

- **~1.0 acceptance on forced spans** — correct by construction, no head, no lookup table, no acceptance gamble.
- **Engages even with the int4 MTP head from #8** — grammar drafting needs no model head at all, and it drafts even when `DRAFT=0`.
- **Composes with MTP/n-gram** — grammar is tried first each step; where it branches (free text), the existing sources take over untouched.
- **Never constrains sampling.** The grammar only proposes; verification accepts or rejects exactly as for MTP drafts (greedy match, or Leviathan rejection sampling under temperature). A wrong, incomplete, or out-of-sync grammar cannot change the output — worst case is rejected drafts, and an adaptive guard (same pattern as the MTP auto-off) disables the source below 50% acceptance.

## Usage

```sh
GRAMMAR=schema.gbnf coli run ...     # or serve; GRAMMAR_DRAFT=n caps the span (default 24)
```

Grammar format: byte-level GBNF subset (llama.cpp style) — literals with escapes, char classes incl. negation, rule refs, groups, `| ? * +`, comments, multi-line rules. Root rule is `root`. Left recursion is detected and fails safe (walker off, generation proceeds undrafted).

## How

- **`c/grammar.h`** (new, header-only like `tok.h`/`st.h`): GBNF parser + PDA walker with a set of stacks in normal form (top of every stack is a terminal byte-class, or the stack is empty = parse completable). `gr_forced()` extends the forced span while exactly one byte is legal **and the parse cannot terminate there** (where it can, the model may legitimately emit EOS, so we don't force).
- **`c/glm.c`**: the walker follows every *emitted* token (`gr_feed`), arming lazily at the first byte the root admits (preambles are skipped) and re-arming after a desync. `spec_decode` asks the grammar first; forced bytes are encoded with the engine's own tokenizer — the tokenization boundary is not guaranteed to match the model's, which is fine: verification absorbs it (that's what the acceptance stat measures). Separate `grammatica:` acceptance stat; grammar-accepted tokens no longer pollute the MTP acceptance counter.
- **`c/tests/test_grammar.c`**: 13 scenarios (forcing, branching, enums, `? * +` incl. multi-byte-literal repetition, negated classes, desync, hex escapes, parse errors, left recursion, and the NDJSON shape from the issue), wired into `make check`.

## Measured (Apple M3 Max, 128 GB, same box as #47)

Real-model A/B on the int8-MTP container (`GLM-5.2-colibri-int4-with-int8-mtp`, 357 GB), greedy `TEMP=0`, `MTP=0 DRAFT=0` (grammar is the *only* draft source), `NGEN=130`, `RAM_GB=60`, NDJSON classification task with the grammar below, fresh process per run:

| config | tok/s | wall (130 tok) | tok/forward | draft acceptance | expert hit |
|---|---|---|---|---|---|
| baseline | 0.37 | 351.1 s | 1.00 | — | 46.7% |
| `GRAMMAR=fit.gbnf` | **0.50** | **260.7 s** | **1.60** | **100% (48/48)** | 45.1% |

- **Output byte-identical** to baseline — the lossless property, confirmed end-to-end on the 744B model.
- **100% acceptance on all 48 forced drafts** — the tokenization-boundary worry didn't bite once on this run (the engine's own BPE and the model agree on structural JSON spans).
- 1.60 tokens/forward lands inside the 1.3–1.5 estimate from the issue (slightly above, since this run's output was punctuation-dense) — and this prompt spent its first ~30 tokens on free-text preamble, where the walker correctly stayed disarmed. On pure NDJSON output the fraction is higher.
- tok/forward gain (1.60×) exceeds the wall-clock gain (1.35×) because verified drafts still route to extra experts on a cold cache — same effect you documented for MTP in #8; the adaptive guard governs this source identically.

Grammar used:

```gbnf
root ::= riga+
riga ::= "{\"id\":\"" id "\",\"fit_category\":\"" cat "\",\"fit_score\":" num ",\"reason\":\"" testo "\"}" "\n"
id    ::= [a-z0-9-]+
cat   ::= "no_fit" | "partial_fit" | "good_fit"
num   ::= [0-9]+
testo ::= [^"\x00-\x1f]*
```

Happy to adjust naming/placement to taste — e.g. if you'd rather the serve/API layer compile JSON schemas to GBNF (the OpenAI `response_format` seam you mentioned), that composes naturally on top of this engine-side mechanism and I can follow up with it.


JustVugg (approved):
Beautifully done — this is exactly the design the idea called for, and it's implemented with the right safety guarantees. Validated here:
- Build clean (0 warnings), oracle unchanged with no grammar (32/32 TF, 20/20 greedy).
- `test_grammar` passes.
- **Lossless proof**: greedy generation with a grammar loaded vs without produces byte-identical output on the oracle (20/20 either way) — confirming the grammar only *proposes* drafts that normal verification accepts or rejects; a wrong or out-of-sync grammar cannot change the result.

The three properties that make this special all hold: ~1.0 acceptance on forced spans (correct by construction), engages even with the int4 MTP head from #8 (no model head needed), and composes with MTP/n-gram (grammar tried first, existing sources take over at branches). The self-contained `grammar.h` (byte-level GBNF subset, left-recursion detected and fails safe) keeps it clean, and the adaptive guard below 50% acceptance matches the MTP auto-off pattern.

This is the highest-leverage speculation source for structured-output/API workloads — on JSON-heavy generation the forced-token fraction is large and *free*. Great work turning #48 into reality. Merging.

PR #77, opened 2026-07-12
perf(omp): keep OpenMP worker team hot to eliminate per-region wake latency

kryptt:
## What

The MoE forward runs many tiny, back-to-back per-expert matmul regions. Under libgomp's default *passive* wait policy the worker team parks between regions, and the re-wake latency dominates the actual compute. Switching to an *active* wait policy — with a bounded spin count so long NVMe expert-load stalls still yield instead of burning a core — keeps the threads hot across those regions.

**Measured on a Zen5 build: expert-matmul wall time 66.9s → 20.9s (~3.2× on that phase).** Output is numerically unchanged — this only affects thread scheduling (there are no `reduction` clauses; every matmul is a per-output-element-owned `schedule(static)` loop, so the result is thread-count-independent).

## Mechanism

libgomp parses `OMP_*`/`GOMP_*` in a constructor that runs *before* `main()`, so `setenv()` from `main()` and continuing is too late (the runtime is already initialized). Instead this seeds the tuned defaults with `overwrite=0` (any value the user already set wins), sets a `COLI_OMP_TUNED` sentinel, and `execv("/proc/self/exe", argv)`s once so a fresh libgomp constructor reads them. The sentinel guards the exec → at most one re-exec. **PID is preserved across `execv`**, so systemd/k8s/shell supervisors don't observe a restart.

## Override / opt-out

Fully overridable and lossless:
- explicit `OMP_WAIT_POLICY` / `GOMP_SPINCOUNT` / `OMP_PROC_BIND` / `OMP_DYNAMIC` win (seeded with `overwrite=0`);
- `COLI_NO_OMP_TUNE=1` disables the tuning + re-exec entirely.

Linux-only (`#ifdef __linux__`; on other platforms the `setenv` defaults still apply but there's no re-exec). A one-line stderr breadcrumb notes the re-exec, and a `perror` on `execv` failure makes a `/proc`-less fallback visible instead of silently losing the speedup.

Single file (`c/glm.c`), no build/toolchain changes.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -2382,8 +2382,37 @@ static void cap_for_ram(Model *m, double ram_gb, int ebits, int max_ctx){
 }
 
 int main(int argc, char **argv){
-    /* i thread OMP non devono girare a vuoto mentre il main aspetta il disco */
-    if(!getenv("OMP_WAIT_POLICY")) setenv("OMP_WAIT_POLICY","passive",1);
+    /* ---- Permanent OpenMP hot-thread tuning. The per-expert matmul regions are
+     * tiny and back-to-back; with the default passive wait policy libgomp parks
+     * the worker team between regions and the re-wake latency dominates. Keeping
+     * the threads hot (active spin) collapses that overhead — measured matmul
+     * time 66.9s -> 20.9s on the Zen5 build, with no change to numerical output.
+     *
+     * libgomp reads the OMP_ / GOMP_ vars in a CONSTRUCTOR that runs before
+     * main(), so setenv() here and continuing would be too late (verified:
+     * setenv-in-main is ignored by the already-initialised runtime). Instead, on
+     * first entry seed the winning defaults — respecting anything the user
+     * already set (overwrite=0) — then re-exec self once so a fresh libgomp
+     * constructor picks them up. The COLI_OMP_TUNED sentinel guards the exec so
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Thoroughly validated — this is a big, well-guarded CPU win. The mechanism (libgomp reads OMP_/GOMP_ vars in a pre-main constructor, so seed-then-re-exec-once is the only way to apply them from inside the binary) is clever and, crucially, I confirmed it's safe on the paths that matter:
- Build clean (0 warnings), oracle 32/32 TF + 20/20 greedy — numerically identical, as expected (wait policy doesn't change results).
- **Re-exec happens exactly once** (COLI_OMP_TUNED sentinel holds — no loop).
- **SERVE protocol survives the re-exec**: piped a turn through SERVE mode, got clean READY…END — the stdio fds are preserved across execv, so coli's subprocess pipe is unaffected. This was my main worry and it's fine.
- Model loads once (re-exec precedes the SNAP read — no double load).
- `COLI_NO_OMP_TUNE=1` kill-switch works (no re-exec, still correct), and user OMP_/GOMP_ env wins (overwrite=0). Non-Linux falls through untuned — correct graceful degradation.

Your measured 66.9s → 20.9s matmul on Zen5 is the same class of finding as the physical-core binding in #61 — the OMP team was parking between the tiny back-to-back per-expert regions and the re-wake latency dominated. This helps every Linux CPU user and composes with core-binding. Excellent, careful work. Merging.

PR #78, opened 2026-07-12
perf(pilot): make router-piloted prefetch do real cross-layer expert loads

kryptt:
## What

The existing PILOT prefetcher predicts the next layer's routed experts from the router logits and issues advisory readahead (`WILLNEED` / `expert_prefetch`) — which only warms the OS page cache, so the main thread still pays the `expert_load` (dequant + slab build) on the critical path.

`PILOT_REAL=1` (default **off**, implies `PILOT=1`) upgrades it to perform the *actual* `expert_load()` into the next layer's LRU cache (`ecache[L+1]`) from the pilot I/O thread, overlapped with the current layer's compute — so on a hit the main thread finds the expert already resident and skips the load.

## Correctness & safety

- **Byte-identical output**: this only changes *which slot holds an expert* (a cache hit), never a computed value. A dropped/failed speculative load simply becomes an on-demand miss later.
- **No cross-thread cache corruption**: a barrier at the top of `moe()` claims the current layer; the pilot only writes layers strictly ahead and drops any load for the current-or-earlier layer, and the future-layer residency scan is serialized under the same mutex. A slot is published (matmul-visible) only after its load completes; while loading it is hidden (`eid=-1`). A failed pilot load can neither hang the main barrier (the in-flight handshake is always cleared on every return path) nor poison a slot.
- **Non-fatal speculation**: a mispredicted prefetch can no longer `exit(1)` the process on a transient disk/alloc error — the pilot path abandons the load, bumps a drop counter, logs one line, and continues. The on-demand load path keeps its exit-on-error behavior byte-for-byte.

## Toggle

Default `PILOT_REAL=0`. When enabled, `PILOT_K` defaults to 6 (curbs LRU eviction pressure from the ~28% mispredicts of real loads) rather than the hint-only default of 8; an explicit `PILOT_K` still wins. A per-run stats line reports real cross-layer loads completed vs dropped. Single file (`c/glm.c`), no build changes.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -24,8 +24,8 @@
 #include <time.h>
 #include <limits.h>
 #include <pthread.h>                              /* thread I/O del PILOTA */
-#include <stdatomic.h>                            /* PIPE: ready-flags / job queue */
-#include <sched.h>                                /* PIPE: sched_yield nello spin */
+#include <stdatomic.h>                            /* PIPE ready-flags/job queue + PILOT_REAL cross-layer handshake */
+#include <sched.h>                                /* sched_yield: PIPE spin / PILOT barrier */
 #include <unistd.h>
 #if defined(__APPLE__) || defined(__linux__)
 #include <sys/resource.h>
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
The synchronization design is sound and I validated it holds: build 0 warnings, default 32/32 TF, and **PILOT_REAL=1 is bit-exact across 5 repeated runs** (20/20 each — race-hunting on the tiny model's cross-layer pilot path, deterministic every time). The invariant is the right one: the pilot writes only `ecache[layer]` for layer > g_cur_moe_layer while the main thread reads only the current layer's cache, so pilot (future layers) and main (current layer) never touch the same slot; the g_pilot_mx mutex + release/acquire atomics serialize the layer-advance handshake and the g_pilot_inflight wait covers an in-flight load of the layer main is about to enter. Clean.

Opt-in (default off, implies PILOT=1), so zero risk to default users. The one thing my tiny-model test can't fully exercise is the LRU-eviction race under *heavy* disk contention (tiny barely streams) — I'm taking that surface on your real-model measurement, which is the right call for an opt-in perf feature with a documented lock invariant and deterministic oracle results. This upgrades PILOT from advisory WILLNEED (page-cache warm only) to real overlapped loads — exactly the lever the profiling pointed at, and it composes with the PIPE overlap in #79. Merging.

JustVugg:
Approved — but it went CONFLICTING because #79 (PIPE) just landed in the same moe/pilot I/O region of glm.c. The review stands; it's a mechanical rebase now. Could you rebase on current main and resolve the overlap? #79's PIPE I/O pool and your PILOT_REAL cross-layer loads are complementary (one pipelines within a block, the other prefetches the next layer), so they should coexist — worth making sure they don't both spawn competing I/O pools, but that's a nice consolidation to think about while you rebase. Ping when green and I'll merge. Great work — this and #79 together are the overlap-I/O-with-compute lever the profiling kept pointing at.

JustVugg:
I started the rebase onto dev myself to unblock you — and stopped, because this one is genuinely yours to resolve, and here's the precise reason (not a text conflict, a design decision):

**#78 and #79 are two overlapping reworks of the exact same MoE miss-handling path, both by you.** #79 (PIPE) is now in dev, and it changed the same block #78 touches. The conflict isn't cosmetic — the two features have to *compose*:

1. **expert_load signature**: #78 adds the `fatal` param (returns -1 instead of exit, so a failed background load doesn't kill the process). #79's PIPE worker calls `expert_load` too — after the merge, PIPE's dispatch path needs to pass the right `fatal` flag (the pool worker should probably be non-fatal like your pilot thread).
2. **The atomic eclock**: your #78 makes the LRU clock atomic (`__atomic_add_fetch`) because the pilot thread touches the cache concurrently. #79's PIPE pool *also* touches `ws[]` concurrently — so the atomic clock is needed for both, good, but the two need to agree it's there.
3. **The real design question I flagged earlier**: do PIPE (#79) and PILOT_REAL (#78) both spawn I/O pools? If `PIPE=1 PILOT_REAL=1` are set together, you'd have two thread pools competing for the same NVMe. That's a design call only you can make — one unified pool, or mutually-exclusive, or layered.

I *could* produce a resolution that compiles and passes the tiny oracle, but the oracle barely streams from disk — it wouldn't stress the concurrent pool+pilot interaction under real load, so a wrong composition could ship a race I can't catch here. You wrote both features and can reason about (and test) how they interact. So this is the one to keep — please rebase onto `dev` and reconcile the two I/O paths; I'll review + land it as soon as it's green. Everything about #78 already passed review; it's purely this composition with your own #79 that needs your hand.

kryptt:
Rebased onto `dev`.

`fatal`/`eclock`: `expert_load` takes a `fatal` flag now — PIPE's pool loads pass `fatal=1` (needed-now, same as the serial path), only PILOT's speculation is `fatal=0`; the LRU clock is atomic since both pools touch the cache.

On the pool-composition question: I tried both — a layered scheme (PILOT defers to PIPE's in-flight loads) and a unified single pool (PIPE workers drain PILOT's queue). Benchmarked GLM-5.2 int4, MTP off, cold cache:

- **Strix Halo** — disk-bound (AVX-512-VNNI, 32c), 3-run mean: serial 0.36 → **naive 0.38**, layered 0.38, unified 0.37 tok/s
- **Zen2** — matmul-bound, no VNNI (16c): serial 0.09 → **naive 0.12**, layered 0.11, unified 0.10 tok/s

Same ordering on both hosts: two independent pools fastest, then layered, then unified. Both coordination schemes I added cost a bit more than they saved, so I dropped them — this PR just lets PIPE and PILOT run as independent pools. Happy to switch to the unified pool if you'd prefer it.

JustVugg (approved):
Rebased clean onto dev, and you answered the composition question the right way — with data. Validated here: build 0 warnings, default oracle 32/32 TF, and the critical test — **PIPE=1, PILOT_REAL=1, and both together are all bit-exact (20/20 greedy)** — so the two independent pools compose without corrupting output. The `fatal` flag reconciliation (PIPE=1 needed-now, PILOT=0 speculative) and the atomic LRU clock are exactly the right resolution.

Your benchmark settling it is the best part: two independent pools fastest on both a disk-bound (Strix Halo) and a matmul-bound (Zen2) host, layered and unified both costing more than they saved — so dropping the coordination schemes is empirically justified, not a guess. That's the call. The heavy-disk concurrency I take on your two-host validation (my tiny oracle can't stress it), which is exactly right for an opt-in perf feature with a clean lock invariant and deterministic oracle results. Merging to dev.

JustVugg:
No need to switch — keep it exactly as merged. Your benchmark is the deciding argument: independent pools won on both a disk-bound and a matmul-bound host, and the coordination schemes cost more than they saved. When the data is that clear on two different bottleneck profiles, we follow it — 'simpler and measured-faster' beats 'cleverer and measured-slower' every time. Independent pools it is.

PR #79, opened 2026-07-12
perf(moe): overlap NVMe expert loads with expert matmul (opt-in PIPE)

kryptt:
## What

For streamed (non-resident) experts the MoE forward is disk-bound: each 64-expert block first blocks on a parallel `pread` of every miss, then runs the matmuls — the two phases never overlap, so compute cores idle while the block's weights are still coming off NVMe.

`PIPE=1` (default **off**) hands the misses' `expert_load()` to a small persistent pool of I/O worker pthreads and lets the main thread start matmul immediately. It walks the block's experts in routing order and waits on a per-slot ready flag only for the expert it needs *right now*, so loads of later experts hide behind matmul of earlier ones. `matmul_qt` stays on the main thread (it parallelizes internally and gates GPU dispatch on `!omp_in_parallel()`), so the I/O pool never contends with it.

## Correctness

This only reorders *when* weights arrive, never the computation — greedy output is byte-identical to the blocking path.

Worker↔dispatcher coordination uses a **generation-tagged cursor**: a single `_Atomic uint64_t` = `(gen << 8) | index`. A worker claims a job with a CAS whose comparand carries the full generation, so a worker preempted across a batch boundary fails its CAS and re-reads rather than touching the next batch's state; the per-job `expert_id`/`layer` are read only *after* the winning CAS, so a successful claim implies a consistent snapshot (no torn reads). The dispatcher publishes batch state with a release store of the cursor that pairs with the workers' acquire loads. The per-slot ready-wait before matmul — issued for *every* dispatched miss slot — guarantees all of a block's loads complete before its LRU swap, so a worker can never still be writing a slab that's being promoted. No mutex on the hot path (the mutex only parks idle workers).

## Toggle

Default `PIPE=0` (behavior unchanged, no worker threads created). `PIPE=1` opts in; `PIPE_WORKERS=n` sizes the pool (default 8). Pure C (`stdatomic.h` + POSIX `pread`/`sched_yield`), no build changes.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -24,6 +24,8 @@
 #include <time.h>
 #include <limits.h>
 #include <pthread.h>                              /* thread I/O del PILOTA */
+#include <stdatomic.h>                            /* PIPE: ready-flags / job queue */
+#include <sched.h>                                /* PIPE: sched_yield nello spin */
 #include <unistd.h>
 #include <sys/resource.h>
 #if defined(__APPLE__) || defined(__linux__)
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Validated on the oracle and the design is clean: build 0 warnings, default (PIPE off) 32/32 TF, and **PIPE=1 stays bit-exact (20/20 greedy)** — the pipelining preserves output, the key correctness signal for reordering load-vs-compute. The separation is right: matmul_qt stays on the main thread (parallelizes internally, gates GPU dispatch on !omp_in_parallel), the I/O pool only does expert_load, and the per-slot ready-flag means the main thread blocks only on the expert it needs *now* while later loads hide behind earlier matmuls. Opt-in (default off) = zero risk to default users; the real-model throughput win I take on your measurement. This attacks exactly the right thing — on disk-bound machines the load/compute phases never overlapped, and that's most of the tracker. Merging.

PR #80, opened 2026-07-12
Use full-resident expert placement on large-memory hosts

ZacharyZcR:
## Summary

- add `CUDA_EXPERT_GB=auto` to fill selected GPUs to their measured safe capacity
- add `PIN_GB=all` to keep every remaining routed expert in RAM when the host budget permits
- adapt existing GPU expert slots to live prompt/decode routing without reallocating VRAM
- batch cold-slot restoration across the host I/O pool before refreshing GPU slots
- use a 512-bit float accumulation kernel for packed int4 matvec on AVX-512/BW hosts
- cache repeated RoPE angles per thread and position
- seed one live-routed GPU expert in every MoE layer after prefill
- report prompt prefill separately from real decode throughput

This applies the useful cache-tier lessons from larger MoE runtimes without importing their framework machinery: fill the fastest tier safely, keep the remaining expert set resident, adapt GPU slots to the current session, and measure decode independently from TTFT. Quantization and router decisions are unchanged.

## Six-RTX-5090 validation

Dedicated host: 6x RTX 5090, dual-socket Xeon Silver 4510 (24 physical cores), 251 GiB RAM, GLM-5.2 int4.

```text
VRAM expert tier: 9,343 experts / 176.73 GB
RAM expert tier: 10,113 experts / 191.3 GB
Total resident: 19,456 / 19,456 experts
real decode: 64 tokens in 10.67 s = 6.00 tok/s
expert hit: 100%
disk service/wait during decode: 0 / 0 s
expert compute: 6.76 -> 5.96 s
```

The prior 150 GB VRAM + 150 GB RAM configuration measured 2.30 tok/s fixed replay and paid 4.15 s of disk wait over 20 tokens. Full residency first reached 5.77 tok/s real decode. Session-local VRAM adaptation then raised the same real generation to 6.00 tok/s; an aggressive 8-token cadence regressed to 5.78, so the automatic full-resident mode uses the measured 16-token cadence.

CPU placement was also verified with an identical 64-token prompt on the dual-socket host:

```text
unbound OpenMP workers:             3.64 tok/s, expert compute 11.02 s
48 workers, spread across cores:    4.69 tok/s, expert compute  7.36 s
24 physical cores, spread/bound:    5.08 tok/s, expert compute  6.76 s
```

That is a 39.6% improvement for the RAM-resident expert path without changing model output. The resource planner applies `OMP_NUM_THREADS=<physical cores>`, `OMP_PROC_BIND=spread`, and `OMP_PLACES=cores` by default while preserving explicit overrides.

With the CPU placement fixed, the packed-int4 RAM path still used a 256-bit AVX2 unpack/FMA loop. The AVX-512/BW implementation processes 32 weights per iteration and keeps the old path available with `I4_ACC512=0`. On an identical 64-token greedy route:

```text
AVX2 accumulation:   5.12 tok/s, expert compute 7.718 s
AVX-512 accumulation: 5.89 tok/s, expert compute 6.200 s
```

This is a 15.0% decode-throughput improvement and 19.7% less expert compute. It adds no activation quantization; as with other parallel floating-point kernels, the changed reduction order is numerically close but not bit-exact.

Profiling then attributed 84.9% of sampled CPU time to quantized matmul and 8.8% to repeated RoPE angle construction. Splitting the AVX-512 FMA dependency chain reduced expert compute from 9.076 to 8.880 s over 96 greedy tokens. Caching the identical per-position RoPE sin/cos table reduced projection/RoPE from 1.419 to 1.232 s and total attention from 4.025 to 3.824 s; this cache reuses the exact original function results.

The prompt-seeded replacement pass now covers all 75 MoE layers once, while decode retains the measured 32-swap limit. The one-time 75-slot refresh costs about 454 ms in prefill, but avoids paying the rejected 64-slot cost every 16 tokens. Two identical 96-token greedy runs measured 6.08 and 6.05 tok/s versus 5.90 tok/s before full-layer seeding; GPU expert calls rose from 36,865 to 37,285 and expert compute fell from 9.176 to 8.736/8.715 s.

The decode replacement limit was then reduced from 32 to 16 slots per pass. Refresh time fell from about 0.18-0.19 s to 0.09 s per pass; two identical greedy runs measured 6.28 and 6.10 tok/s. GPU expert calls decreased to 36,440, but expert compute remained 8.725/8.987 s while non-expert overhead fell from roughly 3.25-3.33 s to 2.743/2.879 s. The saved synchronization cost outweighs the small loss in GPU coverage.

## Validation

- `make check` (all C tests and 46 Python tests)
- `make cuda-test CUDA_ARCH=sm_120` on RTX 5090, including in-place tensor refresh correctness
- `make CUDA=1 CUDA_ARCH=sm_120` on the six-5090 host
- real 64-token greedy GLM-5.2 generation
- identical-prompt OpenMP/NUMA A/B on the dual-socket six-GPU host
- AVX-512 packed-int4 scalar-reference self-test across multiple vector lengths

## Dependency

Stacked on #68 and intentionally Draft until #68 lands. Once #68 is merged, this PR reduces to the full-resident placement, live VRAM adaptation, and timing changes only.


JustVugg:
Needs a rebase onto `dev` — it went conflicting after this week's merges. It's CPU/placement so I can fully review it once it applies cleanly; please rebase and I'll land it. (See the status board in #19 for the full picture.)

JustVugg:
Your experiment doc is merged to dev (#94) — and it's honestly one of the best pieces of engineering documentation in the project: the reverted dead-ends log alone will save the next person weeks, and the discipline of keeping the AVX-512 kernel *out* of the tree until it's numerically qualified is exactly right (more on that below).

On 'the GPU is still not fully utilized' — your own profile already answers it, and the answer isn't a GPU problem: at full residency you're **matmul-bound on CPU** (56% of step time, 84.9% of gprof samples in `matmul_qt`). The GPU holds 9,343/19,456 experts (~48%, capped by 6×32 GB); the other 10,113 live in RAM and their int4 matmul runs on the 24 CPU cores. So 'underutilized GPU' really means 'half the experts are on the CPU path' — and you can't move them to VRAM (no room). The lever isn't more GPU occupancy, it's a **faster CPU int4 kernel for the RAM-resident experts** — which is exactly your AVX-512 candidate.

So #95 (AVX-512 int4) is the right next step, and your instinct to gate it is correct. It's the *same* numerical issue we just documented in #100: shape/accumulation-order differences flip near-ties, so it's not bit-exact. The bar to merge it: (1) a numerical-error unit test bounding the deviation, (2) a token-consistency / perplexity eval on fixed prompts, (3) an explicit, documented decision that the accumulation-order float differences are acceptable (they change the greedy stream on near-ties, like MTP does). Meet those three and I'll merge it happily — a 15% cut on the dominant kernel is worth it. Until then, keeping it out of the tree is the correct call.

Your roadmap (nail the perf, then turn to community issues) is fine — no pressure to context-switch. When #80 is rebased onto dev I'll review + land the full-resident placement. Genuinely excellent work.

ZacharyZcR:
Rebased (merged) onto current dev — the whole stack #68 → #80 → #111 is conflict-free again. Notes from the reconciliation: dev's per-thread heap attention scores and the Metal backend paths are composed with this branch's grouping/placement (both behaviors preserved, Metal untouched in CUDA builds); the earlier round also retired this stack's pthread expert loader in favor of dev's #79 PIPE pool. C test suite green; the full stack was smoke-tested on the 6x5090 machine after the previous reconciliation round and is being re-smoked now. Suggested landing order: #68 → #80 → #111. Thanks for the review notes — they matched our profile exactly.

JustVugg:
The CPU side validates here (build 0 warnings, oracle 32/32 TF + 20/20 greedy on this AVX2 box), and #68 is landed underneath. But I want to pause on one specific thing before merging #80, and it's your own call from #94 that I'm holding you to:

**#80 adds `dot_i4f_avx512` — the 512-bit dual-accumulator int4 kernel.** In your 6×5090 experiment doc you wrote about exactly this kernel: *'the different accumulation order produces small floating-point differences and the 64-token outputs eventually diverge, so it cannot be declared bit-exact. It stays out of the tree until it has unit tests on the numerical error, a perplexity/token-consistency evaluation, and an explicit decision on the acceptability of float accumulation differences.'*

Two reasons this matters more now, not less:
1. My oracle **cannot test it** — this box is AVX2, so `dot_i4f_avx512` never compiles here. Its correctness is invisible to my validation; it rides entirely on AVX-512 hardware I don't have.
2. #108 just measured that int4 GLM-5.2 sits *very* close to argmax ties (MMLU 47→31.5 under quantization). A non-bit-exact kernel is one more way to flip those near-ties on AVX-512 hosts — the #100 mechanism, now on a specific CPU family.

To be clear, this isn't a *quality* concern — accumulation-order float differences are tiny next to the int4 error itself; it's a *reproducibility* one, exactly the class we now document in the README (MTP/CUDA/batched all 'valid but not byte-identical'). So the fix is light — I just need one of:
- **You cleared your own #94 bar** (numerical-error bound + a perplexity/token-consistency check on AVX-512) — if so, point me at the numbers and I'll land it, documented as another shape/family-dependent kernel; or
- **Gate it behind an explicit opt-in** (e.g. `I4_AVX512=1`, default off, like the other lossy/experimental knobs) so it's a deliberate choice on AVX-512 boxes, not silently on.

Either is quick. I just won't silently ship a kernel you yourself flagged as unvalidated, on hardware I can't test. Everything *else* in #80 (CUDA_EXPERT_GB=auto, PIN_GB=all, the RoPE angle cache, GPU slot adaptation) is fine and rides on your 6×5090 validation — it's only this one kernel's bit-exactness I'm holding. #111 stacks on #80, so it waits behind this too. What's the status on the numerical check?

JustVugg:

PR #83, opened 2026-07-12
feat(convert): NVFP4 (modelopt e2m1) dequant path for convert_fp8_to_int4.py

kryptt:
## What

Teaches `c/tools/convert_fp8_to_int4.py` to ingest **NVFP4** checkpoints — the format NVIDIA `modelopt` emits (`quant_algo=NVFP4`), e.g. for REAP-pruned GLM-5.2. The converter already handled FP8-blockscale / bf16 / f32; NVFP4 was the missing input format.

NVFP4 expert weights are `U8` (two `e2m1` FP4 nibbles per byte) with a per-16-block FP8 scale sidecar (`.weight_scale`) and a per-tensor F32 global (`.weight_scale_2`).

## How

- `dequant_nvfp4()`: `e2m1` LUT (verified 1:1 against `ml_dtypes.float4_e2m1fn`), low-nibble=even / high-nibble=odd unpacking, per-block scale `repeat_interleave` over `group_size=16`, times the global.
- `classify()` consumes the `.weight_scale/.weight_scale_2/.input_scale` sidecars; `dequant()` routes `U8`+sidecar tensors to the NVFP4 path (`keys` is now required, so a stray `U8` can't fall through). **Existing FP8/bf16/f32 paths are untouched.**

## Guards against silent corruption

- `group_size` fixed at 16, and the block-scale column count is **asserted** `== ceil(I/16)`. The prior approach inferred it (`I // ncols`), which misaligns silently on a padded/swizzled scale layout and hard-crashes on a non-multiple-of-16 `I` — after a multi-GB shard download. A partial trailing block is handled by slicing to `I`.
- `.weight_scale_2` is asserted `< 1`: modelopt stores the *small* global and **multiplies**; llm-compressor/compressed-tensors stores the *reciprocal* (`>= 1`) and **divides**. The two are dtype-identical, so without this a compressed-tensors checkpoint would corrupt every tensor.

## Verification

`--selftest-nvfp4` (no network): asserts the 16-code LUT, an encode→dequant round-trip to `<1e-9` (measured 0.0), and a dequant→colibri-int4 requant bound (`<0.30`; the inherent ~0.17 is informational). Green on the target host.

### DIFF
--- a/c/tools/convert_fp8_to_int4.py
+++ b/c/tools/convert_fp8_to_int4.py
@@ -64,6 +64,16 @@ def quant_int2(w, bits):                        # -> (qbytes U8 [O*ceil(I/4)], s
         out[:, :vk.shape[1]] |= ((vk + 2).astype(np.uint8) << (k * 2))
     return out.reshape(-1), s[:, 0].astype(np.float32)
 
+# ---------- NVFP4 (modelopt) : LUT e2m1 ----------
+# FP4 e2m1 = 1 sign + 2 exp + 1 mantissa. 16 codici, magnitudini {0,.5,1,1.5,2,3,4,6}.
+# Bit 3 = segno. Ordine impacchettato (compressed_tensors/vLLM): nibble BASSO = elemento
+# pari, nibble ALTO = elemento dispari. LUT verificata 1:1 con ml_dtypes.float4_e2m1fn.
+# EN: FP4 e2m1 = 1 sign + 2 exp + 1 mantissa. 16 codes, magnitudes {0,.5,1,1.5,2,3,4,6}.
+# EN: bit 3 = sign. Packed order (compressed_tensors/vLLM): LOW nibble = even element,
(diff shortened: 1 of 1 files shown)


JustVugg:
This one's converter-only (the NVFP4 / modelopt e2m1 dequant path in convert_fp8_to_int4.py), so it should be a clean review — please retarget the base to `dev` and rebase if needed. I can't test it end-to-end without an NVFP4 source checkpoint, but I can review the dequant math and that it doesn't touch the existing FP8→int4 path for non-NVFP4 inputs. Point it at dev and I'll review. (Status board: #19.)

kryptt:
Retargeted to `dev` and rebased (`9b706e9`). The non-NVFP4 FP8→int4 path is untouched — the e2m1 dequant only activates when the modelopt NVFP4 keys are present, otherwise the original path runs unchanged. For the dequant math you can't test without an NVFP4 checkpoint, `--selftest-nvfp4` validates it against a known vector.

PR #84, opened 2026-07-12
feat(vk): opt-in Vulkan iGPU backend for quantized expert matmul

kryptt:
## What

An **opt-in** Vulkan compute backend (`backend_vulkan.c/.h` + one int8/int4 GEMV shader) that runs resident/pinned expert matmuls on a local GPU. Gated by `VK=1` at build time and `COLI_VULKAN=1` at runtime.

## The default build is untouched

Every `glm.c` hook is inside `#ifdef COLI_VULKAN`. Without the flag, **not a single VK symbol is referenced** — verified by `nm` (pure-C binary: 0 `coli_vk_*` symbols; VK build: 7) and by preprocessing `main` vs this branch under pure-C flags (the only non-`#ifdef` delta is one intentional fail-fast guard if a `VK_*` env var is set on a CPU-only binary). `make`/CUDA builds are byte-identical.

The int4/int8 shader is numerically verified against the CPU path on real hardware (Radeon 8060S / RADV, Strix Halo): max rel err ≤ 1.7e-4.

## Honest performance note

Targeted at the Strix Halo iGPU. Under **unified memory** the GPU and CPU share one ~150–175 GB/s bus and S=1 decode is memory-bound, so this is break-even for resident experts and net-negative once experts are streamed — it only pays off on a host whose GPU has its own high-bandwidth VRAM. Shipped off-by-default accordingly; the toggles document this.

## Toggles

- `VK_EXPERT_GB` — caps the iGPU-pinned expert tier.
- `VK_STREAM` (**default OFF**) — also routes streamed/cached experts through the shader; keeps both the CPU slab and a VK copy (~2× cache RAM), so opt-in for discrete-VRAM hosts only.
- `VK_DENSE` — extend to dense tensors. `VK_SPV` — override the shader path.
- Only int8 (fmt 1) / int4 (fmt 2) run on the GPU; fmt 0/3 and any omp-parallel context fall back to CPU, and a failed upload sets `vk_failed` to degrade gracefully.

## Robustness

Bounded fence wait (10s → disable VK/CPU-fallback on device-loss, never hangs); atomic backend counters (frees run under `#pragma omp parallel`) with symmetric weights+scales accounting; SPIR-V size validation; device selection prefers a real GPU over a software adapter; tensor free skips `vkDestroy` on a lost device (no GPU-side UAF).

## Build

`make VK=1` (needs `libvulkan` + `glslc`).

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -80,13 +80,34 @@ LDFLAGS  += -L$(CUDA_HOME)/lib64 -Wl,-rpath,$(CUDA_HOME)/lib64 -lcudart -lstdc++
 CUDA_OBJ  = backend_cuda.o
 endif
 
+# VK=1 adds an opt-in Vulkan compute backend for resident/pinned quantized
+# matmul, targeting the Strix Halo iGPU (RADV gfx1151). Independent of CUDA and
+# toggleable alongside it. The default build stays pure C. Runtime-gated by the
+# COLI_VULKAN=1 environment variable; shader path overridable via VK_SPV.
+VK        ?= 0
+GLSLC     ?= glslc
+VK_OBJ     =
+VK_SHADER  =
+ifeq ($(VK),1)
+CFLAGS    += -DCOLI_VULKAN
+LDFLAGS   += -lvulkan
+VK_OBJ     = backend_vulkan.o
+VK_SHADER  = shaders/qmatmul.spv
+endif
+
 all: glm$(EXE)
 
 # phony 'glm' → 'glm.exe' on Windows (so 'make glm' and 'coli build' work on every platform)
 glm: glm$(EXE)
 
-glm$(EXE): glm.c st.h json.h tok.h tok_unicode.h compat.h grammar.h $(CUDA_OBJ)
-	$(CC) $(CFLAGS) glm.c $(CUDA_OBJ) -o glm$(EXE) $(LDFLAGS)
+glm$(EXE): glm.c st.h json.h tok.h tok_unicode.h compat.h grammar.h $(CUDA_OBJ) $(VK_OBJ) $(VK_SHADER)
+	$(CC) $(CFLAGS) glm.c $(CUDA_OBJ) $(VK_OBJ) -o glm$(EXE) $(LDFLAGS)
+
+backend_vulkan.o: backend_vulkan.c backend_vulkan.h
+	$(CC) $(CFLAGS) -c backend_vulkan.c -o $@
+
+shaders/qmatmul.spv: shaders/qmatmul.comp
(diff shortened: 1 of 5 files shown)


JustVugg:
Rebase onto `dev` please — conflicting after the recent merges. The Vulkan path I'll review for isolation (must be gated behind an opt-in flag like the CUDA/Metal backends, CPU build untouched) and take the GPU execution on your validation, since this is a GPU-less dev box. Rebase it and I'll review the gating + land. (Status board: #19.)

kryptt:
Rebased onto `dev`. Fully opt-in — every glm.c hook is under `#ifdef COLI_VULKAN`, so `make` (no `VK=1`) is byte-identical to dev (`nm glm` → 0 `coli_vk` symbols); `make VK=1` links libvulkan + compiles the shader. int4/int8 GEMV verified against the CPU path on a Radeon 8060S (RADV, Strix Halo), max rel err ~1.7e-4.

Honest caveat in the PR body: on unified memory it's break-even/net-negative (shared bus), so it's shipped off-by-default — the win is a host with discrete high-bandwidth VRAM. GPU execution is yours to validate; happy to adjust the gating however you'd prefer.

JustVugg:

**It's 195 commits behind `dev`.** On `glm.c`, which changes several times a day, that isn't a rebase — it's a rewrite, and doing it *for* you would mean rewriting your design decisions without understanding why you made them. So I'd rather ask.

A whole Vulkan backend is a big thing to carry unverified: nobody here has an iGPU path to test it on, and the CI can only prove it compiles. If it lands, it needs a way for the CI (or a maintainer) to actually execute it — otherwise it's 723 lines that break silently in six months.

JustVugg:
Closing this, and I want to be straight about why rather than let it rot: it's 245 commits behind, and — the deciding factor — there's no way to verify it. Nobody on the maintainer side has an iGPU Vulkan path, and CI can only prove it compiles, which for a quantized-matmul backend isn't enough to ship responsibly.

That's the same bar every backend faces here now (it's why the AMD/HIP and ARM kernels are in the same holding pattern): a compute backend lands when it can prove correctness, not just build. What would change the answer for Vulkan: an exactness test that checks the VK matmul against the CPU reference on a tiny fixture (the way `tests/test_i4_grouped.c` and `test_idot.c` do), runnable without the real model — plus a rebase onto current `dev`. With that, it's reviewable and I'd genuinely like an iGPU path in.

Until then it's better as a fork linked from the README than as an unverifiable 723-line backend a maintainer can't run. Reopen the moment you have the test — no hard feelings, and thanks for building it.

PR #90, opened 2026-07-12
Add support for amd gpus and decouple vram from system memory

plasticchris:
## Summary

This adds support for gfx1100 (texted on dual rx7900xtx) and decouples gpu vram.

Is it faster? Not enough to notice, disk reads still dominate. But it was a fun weekend project, and it seems like a prerequisite for hardware config changes that will make it faster.

## Validation

- [X] `make -C c check`
- [N/A] CUDA changes were tested with `make -C c cuda-test` (if applicable)
- [X] Performance claims include hardware, commands, and repeatable measurements

## Compatibility


plasticchris:
oh I looked at the other PRs, seems the vram decoupling idea is being worked on already

PR #91, opened 2026-07-12
Feature/flake darwin

advicktory:
## Summary

## Validation

- [ ] `make -C c check`

## Compatibility


advicktory:
Didn't mean to make this.

PR #94, opened 2026-07-12
docs: GLM-5.2 on 6x RTX 5090 — measured experiment record (full residency, MTP speculation, AVX-512 qualification)

ZacharyZcR:
Adds `docs/experiments/glm52-6x5090-2026-07-12.md`, a measured record of two days of lab work on one machine (6x RTX 5090, dual Xeon, 251 GiB RAM), plus a README link from the honest-numbers section.

## What it documents

- **vLLM-Moet comparison** on the same hardware (TP4 / TP2xPP3): why per-rank expert residency and whole-step replay dominate, and why six 32 GB cards are not two 96 GB cards.
- **Full-resident expert placement** (relates to #80 / #68): the optimization ladder from 2.30 to 6.28-6.84 tok/s single-request decode, with every tested-and-rejected direction listed with its numbers so the dead ends stay dead.
- **MTP speculation at full residency**: the installed int4 MTP head is the known-defective one from #8 (0-4% acceptance). With the community int8 head, acceptance reaches 69-79% — and speculation still loses at every depth, because a MoE verify batch of S positions routes to ~S distinct expert sets: measured per-forward expert time scales almost linearly with S. Includes the condition under which this conclusion inverts (cheap multi-row GPU expert GEMM).
- **AVX-512 int4 kernel qualification** (the candidate noted in the record): numeric harness shows it is 2-4x closer to a double-precision oracle than the old sequential order; SCORE perplexity delta is 0.24% with signs split 4/4; throughput +4-7% on CPU-heavy routings, 0% when routing is GPU-served. Recommended for merge; the only open item is the bit-compatibility policy call.

Documentation only — no engine changes. Lab paths in the record describe the test machine, not repository dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ The engine is a single C file (`c/glm.c`, ~2,400 lines) plus small headers. No B
 
 ## Honest numbers (WSL2, 12 cores, 25 GB RAM, NVMe via VHDX)
 
+Detailed GPU experiment: [GLM-5.2 on 6x RTX 5090](docs/experiments/glm52-6x5090-2026-07-12.md) — full expert residency across VRAM+RAM reaches 6.84 tok/s single-request decode.
+
 | metric | value |
 |---|---|
 | model on disk (int4 container) | ~370 GB |
--- a/docs/experiments/glm52-6x5090-2026-07-12.md
+++ b/docs/experiments/glm52-6x5090-2026-07-12.md
@@ -1,0 +1,302 @@
+# GLM-5.2 on 6x RTX 5090
+
+This note records the July 12, 2026 experiments that compared colibri with
+vLLM-Moet on one machine with six RTX 5090 GPUs (32 GiB each), dual Intel Xeon
+Silver 4510 (24 physical cores / 48 threads), 251 GiB of RAM, and local NVMe
+storage. It exists to preserve the measured facts and avoid repeating the same
+dead ends.
+
+The day had two halves. The morning runs (below) measured colibri at 0.12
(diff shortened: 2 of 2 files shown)


JustVugg:
The experiment itself is superb — full expert residency across VRAM+RAM reaching 6.84 tok/s single-request on 6×5090, with zero disk waits, is a genuinely valuable measurement and exactly the kind of upper-bound the project should document. Thank you for the rigor (fixed-token replay, coverage %, the morning-vs-afternoon progression).

One thing I want to get right before this lands, because it's a claim that'll be read carefully: **the vLLM comparison isn't quite apples-to-apples as framed.** colibrì's 6.84 is with *all 19,456 experts resident* (full residency, zero disk); the vLLM-Moet rows are with a 12–14 GB expert cache/card = ~25–30% coverage. So it's full-residency colibrì vs partial-cache vLLM — the 6.84-vs-2.5 gap partly reflects that colibrì was given the whole model in memory while vLLM wasn't. That doesn't diminish the colibrì result (it's a real full-residency number), but the doc reads as 'colibrì beats vLLM' when the configs differ on the one variable that matters most for MoE decode.

Could you either (a) qualify it — 'full-residency colibrì vs partial-cache vLLM; not a matched comparison' — or (b) if you have it, add a vLLM full-residency row so the comparison is at equal coverage? Either makes the claim bulletproof. Once the framing is fair, this is a merge into `dev` — and the raw colibrì full-residency number is the headline either way. Holding only on the comparison framing, not the data.

PR #95, opened 2026-07-12
AVX-512 int4 accumulator for the expert dot kernel (+4-7% CPU-heavy decode, runtime-switchable)

ZacharyZcR:
Ports the AVX-512 int4→float accumulator that has been running qualified in the 6x RTX 5090 lab tree into the engine, behind a runtime switch. Full qualification data is in the experiment record added by #94.

## What it does

`dot_i4f_avx512` processes 32 packed int4 weights per iteration on two independent FMA chains and replaces the AVX2 256-bit order inside `matmul_i4` when the build has AVX-512F/BW. `I4_ACC512=0` restores the previous accumulation order at runtime, so every A/B in the qualification used one binary. Non-AVX-512 builds are unchanged.

## Numbers (six-RTX-5090 lab machine, full expert residency, TEMP=0, DRAFT=0, 96 tokens)

- **Throughput**: +7.0% and +7.4% on two CPU-heavy-routing prompts, 0% on a prompt whose routing is served almost entirely from the GPU hot tier; +4.1% mean over 12 ABAB-interleaved runs. (An earlier +15% estimate came from a low-baseline run and is retired in the record.)
- **Numeric error**: against a double-precision oracle over the real expert row shapes (I=6144/2048, 70k+ random rows, small/large activations, non-multiple-of-32 tails, cancellation stress), the 512-bit tree reduction shows **2-4x lower max relative error** than the sequential scalar order the engine falls back to elsewhere.
- **Quality**: SCORE log-likelihood over 8 multi-domain passages (251 continuation tokens): perplexity 5.99 vs 5.98 (0.24%), per-passage signs split 4/4.

## The one policy question

Output is **not bit-identical** to the old order — greedy decodes eventually diverge on argmax ties. The measurements say the new order is the *more* accurate of the two, so keeping bit-compatibility would mean pinning the worse rounding. If bit-stability across releases matters more, the default can flip to off — the switch is already there.

## Tests

- Engine selftest: `I4_ACC512_TEST=1 ./glm` (nibble order + every multiple-of-32 length vs scalar).
- `tests/test_i4_acc512` (wired into `make test-c`): double-oracle regression over the real row shapes, asserting the AVX-512 error stays within 2x of the scalar order's; compiles to a skip on non-AVX-512 hosts.
- Verified on AVX-512 hardware (Xeon Silver 4510): selftest ok, regression ok, full `test-c` suite green; syntax-checked and suite green on an AVX2-only host (graceful skip).


ZacharyZcR:
Your three-point bar for this PR is met — pointers:

1. **Numerical-error unit test**: `tests/test_i4_acc512` is in this PR and wired into `make test-c` — double-precision oracle over the real expert row shapes (I=6144/2048, small/large activations, non-multiple-of-32 tail, cancellation stress), asserting the AVX-512 error stays within the sequential-scalar order's envelope. It measures 2-4x *lower* max relative error than the scalar order (tree reduction rounds less), and compiles to a skip on non-AVX-512 hosts. The engine also ships a quick selftest (`I4_ACC512_TEST=1`).
2. **Token-consistency / perplexity on fixed prompts**: in the merged experiment record (#94, 'AVX-512 int4 kernel qualification'): SCORE log-likelihood over 8 fixed multi-domain passages, 251 continuation tokens — perplexity 5.99 vs 5.98 (0.24%), per-passage signs split 4/4. Statistically indistinguishable.
3. **Explicit accumulation-order decision**: documented in the same record and in this PR's description — the new order is *closer* to the double-precision oracle than the old one, so bit-compatibility would pin the worse rounding. Near-tie greedy flips are the same class of divergence already accepted for MTP/CUDA/batched paths in #100; `I4_ACC512=0` remains as the runtime escape hatch.

Rebased onto current dev, suite green (including dev's new `test_idot`).

ZacharyZcR:
@JustVugg Here's the raw validation from AVX-512 hardware (this exact branch head, freshly built) — everything you can't run locally:

**Machine**: Intel Xeon Silver 4510 (`avx512f avx512bw`), the 6×5090 rig from #94.

```
== engine selftest (I4_ACC512_TEST=1):
AVX512 i4 selftest: ok

== double-oracle regression test (make test-c → tests/test_i4_acc512):
  gate/up rows             I=6144  avx512 max 7.226e-05 | scalar max 4.215e-04 | ok
  gate/up large x          I=6144  avx512 max 4.420e-03 | scalar max 5.071e-03 | ok
  down rows                I=2048  avx512 max 4.182e-05 | scalar max 1.070e-04 | ok
  down small x             I=2048  avx512 max 1.695e-06 | scalar max 9.394e-06 | ok
  tail I=6143              I=6143  avx512 max 7.478e-05 | scalar max 3.702e-04 | ok
  short rows               I=96    avx512 max 5.513e-06 | scalar max 1.166e-05 | ok
test_i4_acc512: ok
```

Reading it: in every regime (real expert shapes, large/small activations, non-multiple-of-32 tail, short rows) the AVX-512 kernel's max relative error vs a float64 oracle is **2-6× smaller** than the sequential scalar order the engine uses everywhere else — the tree reduction rounds less. No non-finite values.

End-to-end data, already in the merged record (#94, 'AVX-512 int4 kernel qualification'):
- **Quality**: SCORE log-likelihood over 8 fixed passages (251 tokens): ppl 5.99 vs 5.98 (0.24%), per-passage signs 4/4 — indistinguishable.
- **Throughput**: +7.0%/+7.4% on CPU-heavy routings, 0% when routing is GPU-served; +4.1% mean over 12 ABAB runs on the 5090 rig. On your CPU-only path the share of RAM-expert matmul is 100%, so your expected gain is the +7% end of the range — and per your #108 note, this kernel is exactly the 'faster CPU int4 for RAM-resident experts' lever.
- Escape hatch: `I4_ACC512=0` restores the old order at runtime; non-AVX-512 builds are bit-for-bit unchanged (the test compiles to a skip — verified on an AVX2-only host).

That's the full bar from your #80 comment: (1) numeric test in-tree ✓, (2) fixed-prompt consistency eval ✓, (3) the accumulation-order decision documented ✓ (the new order is the *more* accurate of the two).

JustVugg:
Merged to `dev` (`4b1d0e3`). You cleared every point of the bar I set on #80, and then some.

The part that flips my original concern: the divergence from the AVX2 path isn't the kernel being *looser* — your double-oracle table shows the 512-bit tree reduction is **2–6× closer** to the float64 truth in every regime (gate/up, down, the I=6143 tail, short rows). So bit-pinning the old order would mean pinning the *worse* rounding. That's the right call, and documenting it in the #94 record is exactly what I wanted to avoid a silent "why did greedy shift" later. The near-tie flips are the same accepted class as MTP/CUDA/batched (#100), and `I4_ACC512=0` keeps the old order one env var away.

Three things I checked before merging:
- Non-AVX-512 builds bit-identical (test compiles to a skip — your AVX2-host verification covers my box too, which has no AVX-512F).
- Base was `dev`, landed on `dev`, `main` untouched — it'll ride the next `dev→main` promotion after a local clean-build gate.
- On the pure-CPU path (mine, and anyone without a GPU tier) the RAM-expert matmul share is ~100%, so this is the +7% end of your range — the single biggest CPU decode lever we've merged. Thank you for doing the qualification I couldn't run.

PR #96, opened 2026-07-12
OpenAI tool calling for GLM-5.2 (+ prefill keepalive)

RDouglasSharp:
# PR: OpenAI tool calling for GLM-5.2 (+ prefill keepalive) — `c/openai_server.py`

**Branch:** `feat/glm-tool-calling` off `upstream/main` (`e945b43`) · **One file:** `c/openai_server.py`

## What

`generation_options()` currently rejects `tools`/`functions` with *"not supported yet."* This
implements them end to end, so an OpenAI-style agent client (OpenCode, the SDKs) can drive
GLM-5.2 through real tool calls. Five small, independent commits:

1. **parse GLM tool calls → `tool_calls`** — `render_chat` gains a tools-declaration block,
   replays `role:"tool"` results as `<|observation|><tool_response>…`, and reconstructs prior
   assistant `tool_calls` as `<tool_call>…`; `parse_tool_calls` turns the model's `<tool_call>`
   markers back into OpenAI `tool_calls`; `generation` emits them for both non-stream (in the
   message) and stream (marker-suppressing splitter + `tool_calls` deltas) with
   `finish_reason:"tool_calls"`.
2. **authoritative tool-declaration block** — the `# Tools` + `<tools></tools>` structure
   byte-matches `chat_template.jinja`. (A hand-written preamble made the model hallucinate other
   frameworks' syntax; the trained format fixes it.)
3. **keepalive during long prefill** — `engine.generate()` is silent for the whole cold prefill
   (minutes on a large prompt) and clients drop the socket; a background pump emits a
   `reasoning_content "."` SSE delta every 10s of silence (resets the client timer, lands in the
   thinking panel), gated so it's quiet while real tokens flow, stopped when generation returns.
4. **`COLI_THINK`** (default off) — global thinking default when a request sends neither
   `reasoning_effort` nor `enable_thinking` (client value always wins).
5. **`COLI_DEBUG`** (default off) — echo decoded tokens to stderr (visibility when the client
   disconnects or for local debugging).

## Default-safe

- **No `tools` in the request ⇒ byte-identical to current behavior** — render and streaming take
  the exact existing path; all tool handling is gated on `tools` being present.
- `COLI_THINK`, `COLI_DEBUG`, and the salvage parser (below) are all **default-off**.

## Validation

- **Offline unit tests** (no model): strict parse is byte-exact (`read` →
  `{"filePath":"/x"}`); `render_chat` round-trips an assistant `tool_calls` message + a following
  `tool` message; a no-tools reply is unchanged (regression); the malformed variants map onto the
  primary param.
- **Live, on GLM-5.2 int4:** a clean call — `get_weather` → `{"city":"Paris"}`,
  `finish_reason:"tool_calls"`, 10 completion tokens.
- **Keepalive live:** streamed request shows `reasoning_content:"."` pings every ~10s through the
  prefill, then the `tool_calls` delta and `[DONE]`.

## Candid caveat (int4 formatting) — and the opt-in salvage

On the int4 quant, GLM-5.2's tool-call formatting is **variable**: sometimes a clean call,
sometimes it fumbles the `<arg_key>/<arg_value>` structure (e.g. smears the key name into the
value) or emits an extra hallucinated call. This is a **model/quantization trait, not the parser**
— the strict path is byte-exact on well-formed output, and higher-precision weights format
cleanly. An **opt-in** best-effort recovery, `COLI_TOOL_SALVAGE=1` (default off), reconstructs a
malformed call by mapping its lone payload onto the tool's primary parameter; it never rewrites
well-formed output and is **recommended for int4 deployments** as a robustness net (it turns
"unusable — model name is a garbled blob" into "structurally valid, occasionally wrong value"). A
`[api] tool-calls: N total, S strict, D de-mangled` line reports which path each turn took.

## Notes

- Marker/format authority: GLM-5.2 `chat_template.jinja` (`zai-org/GLM-5.2-FP8`).
- The five commits are independent and cleanly separable — happy to split into multiple PRs
  (e.g. tool-calling vs. keepalive vs. the two env toggles) if you prefer.
- Deliberately **not** bundled: KV persistence and SSE `Connection: close` (already upstream),
  and any engine changes (this is server-only).

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -588,6 +588,17 @@ class APIHandler(BaseHTTPRequestHandler):
             self.send_cors_headers()
             self.end_headers()
             connected = True
+            # KEEPALIVE: engine.generate() blocks SILENTLY during the (minutes-long) cold
+            # prefill, and the client drops the socket after its idle timeout. A background pump
+            # emits a reasoning_content "." delta (the channel that reliably resets the client's
+            # timer and lands in the thinking panel, so answer content stays clean) whenever no
+            # event has been written for KA_GAP seconds. All wfile writes share ka_lock so the
+            # pump and event() never interleave; last_write gates the pump so it stays quiet
+            # while real tokens are flowing (e.g. during decode).
+            ka_lock = threading.Lock()
+            last_write = [time.time()]
+            ka_stop = threading.Event()
+            KA_GAP = 10.0
 
             def event(choices, usage_marker=False):
                 nonlocal connected
@@ -597,12 +608,23 @@ class APIHandler(BaseHTTPRequestHandler):
                               "model": self.server.model_id, "choices": choices}
                 if include_usage:
                     event_body["usage"] = None if not usage_marker else usage_marker
-                try:
-                    data = json.dumps(event_body, ensure_ascii=False, separators=(",", ":"))
-                    self.wfile.write(f"data: {data}\n\n".encode())
-                    self.wfile.flush()
-                except OSError:
-                    connected = False
+                data = json.dumps(event_body, ensure_ascii=False, separators=(",", ":"))
+                with ka_lock:
+                    try:
+                        self.wfile.write(f"data: {data}\n\n".encode())
+                        self.wfile.flush()
+                        last_write[0] = time.time()
+                    except OSError:
+                        connected = False
+
+            def _keepalive():
+                ping = [{"index": 0, "delta": ({"reasoning_content": "."} if chat else {"content": ""}),
+                         "logprobs": None, "finish_reason": None}]
+                while not ka_stop.wait(1.0):
+                    if not connected:
+                        return
+                    if time.time() - last_write[0] >= KA_GAP:
+                        event(ping)
 
             if chat:
                 event([{"index": 0, "delta": {"role": "assistant", "content": ""},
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Well-engineered — and crucially, **the non-tool path is provably preserved**: the tool declaration block is gated behind `if tools:`, the assistant tool_calls rendering behind `for tc in (message.get('tool_calls') or [])`, and `parse_tool_calls` returns the content unchanged when there are no `<tool_call>` markers. So a normal chat request (no tools) renders and responds exactly as before — verified by inspection, and `py_compile` clean with all existing endpoints intact. You also removed the old `tools not supported` 400 stub, which is the point.

Nice touches: the tool-declaration block byte-matches GLM-5.2's chat_template.jinja (the authoritative `# Tools` + `<tools>` structure — using a made-up preamble makes the model hallucinate other frameworks' syntax), and `COLI_TOOL_SALVAGE=1` as an opt-in de-mangler for int4's occasionally-malformed output is a pragmatic touch.

I can't functionally exercise the tool round-trip here without running the server against the engine (machine is mid-conversion), so the tool-calling *behavior* rides on your validation — but since normal chat is provably untouched, this is safe to land on `dev` where it gets real testing before the main batch. Retargeted to dev, merging. This is the most-requested API feature after the server itself — thank you.

dnnspaul:
@JustVugg you said you merge this into dev branch, but it landed on main?

JustVugg:
You're right, and thank you for catching it. I retargeted #96 and #97 to `dev` before merging, but the base change silently failed (masked by an unrelated GitHub API warning) and they landed on `main` instead. That's my mistake — I should have verified the base actually changed rather than trusting the command output.

Both were validated before merge (#97: clean build + oracle 32/32 TF / 20/20 greedy, x86 path provably untouched; #96: the non-tool chat path is gated behind `if tools:` so normal requests are unchanged, `py_compile` clean), so `main` isn't broken — but you're right that they should have gone through `dev` per the workflow we just adopted. I've synced `dev` so it now contains these plus everything on main, and going forward I'll verify the target branch on every merge, not assume it. Appreciate you holding the process to its word — that kind of check is exactly what keeps it honest.

PR #98, opened 2026-07-12
VSX integer-dot kernels for POWER8: 12.8x int8, 7.6x int4 over scalar

Scottcjn:
Follow-up to #97 (this branch includes that Makefile commit and will rebase clean once it merges). This adds a VSX path to the two hot idot kernels, dot_i8i8 and dot_i4i8, plus an exactness test that covers every SIMD backend.

Measured on an IBM POWER8 S824 (dual 8-core, gcc 9.4.0, Ubuntu 20.04 ppc64le), single thread, 1536x6144 expert-like shape:

| kernel | scalar | vsx | speedup |
|---|---|---|---|
| dot_i8i8 | 1.48 Gops/s | 18.99 Gops/s | 12.8x |
| dot_i4i8 | 2.33 Gops/s | 17.72 Gops/s | 7.6x |
| S=1 int4 matmul path (full dispatch incl. qrow_i8) | 3.925 ms/call | 0.505 ms/call | 7.8x |

Implementation notes:

- vec_msum multiplies 16 byte pairs and sums groups of 4 directly into s32 lanes, so there is no 16-bit intermediate and the maddubs saturation bound does not apply on this path. Same sign trick as AVX2: abs(w) as u8 times x*sign(w) as s8.
- abs(w) is built with a modulo-subtract select rather than vec_abs. Depending on how the compiler lowers vec_abs, w=-128 can saturate to 127, which would silently corrupt weights at the quantization edge. The select wraps -128 to 128 as unsigned, matching what _mm256_sign_epi8(wv,wv) does on x86.
- Nibble unpack for int4 uses vec_mergeh/vec_mergel, which on ppc64le with GCC interleave exactly like x86 unpacklo/unpackhi (verified empirically on the hardware, not assumed from the ISA docs).
- g_i4s=1 on VSX: unlike AVX2, the f32 fallback on PowerPC is plain scalar C, so int4 IDOT pays even at S=1 (measured 5.5x within the VSX build). Same reasoning as the ARM SDOT case from PR #9.
- The f32 dequant kernels (matmul_q / matmul_i4 / matmul_i2) intentionally stay scalar on PowerPC: with g_i4s=1 they are only reached with IDOT=0, which is the exact A/B debug path.

Testing:

- New tests/test_idot.c compiles the engine and checks dot_i8i8 / dot_i4i8 exactness against a plain-C reference: odd sizes for the scalar tail, sub-vector sizes, w=-128 everywhere, and 64 random reps per size. Integer kernels have no rounding, so equality is exact. Passes on avx512-vnni (x86) and vsx (POWER8); it is wired into make test-c so it also covers avx2 and neon wherever those build.
- Tiny oracle on the POWER8 VSX build stays token-exact: teacher forcing 32/32, greedy 20/20.
- All pre-existing C test suites still pass on both x86 and POWER8.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -66,7 +66,7 @@ CUDA_ARCH ?= native
 NVCCFLAGS ?= -O3 -std=c++17 -arch=$(CUDA_ARCH) -Xcompiler=-Wall,-Wextra
 PYTHON    ?= python3
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE)
+TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_idot$(EXE)
 ifeq ($(CUDA),1)
 ifeq ($(UNAME_S),Darwin)
 $(error CUDA=1 is supported only on Linux)
@@ -117,6 +117,9 @@ tests/test_tier$(EXE): tests/test_tier.c tier.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_grammar$(EXE): tests/test_grammar.c grammar.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_idot$(EXE): tests/test_idot.c glm.c st.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 test-c: $(TEST_BINS)
--- a/c/Makefile
+++ b/c/Makefile
@@ -33,6 +33,18 @@ CFLAGS  = -D_FILE_OFFSET_BITS=64 -O3 -march=$(ARCH) -fopenmp -Wall -Wextra -Wno-
 LDFLAGS = -lm -fopenmp -static
 EXE     = .exe
 else
+UNAME_M := $(shell uname -m)
+ifneq (,$(filter ppc64le ppc64,$(UNAME_M)))
+# --- Linux PowerPC (POWER8/POWER9/POWER10) ---
+# PowerPC GCC uses -mcpu, not -march. ARCH=native works on gcc >= 4.7.
+# The AVX2/NEON kernels fall back to the portable scalar C path
+# (validated token-exact vs the transformers oracle on a POWER8 S824).
+CC      = gcc
+ARCH   ?= native
+CFLAGS  = -O3 -mcpu=$(ARCH) -fopenmp -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function
+LDFLAGS = -lm -fopenmp
+EXE     =
+else
 # --- Linux x86-64 (percorso originale, invariato) ---
 CC      = gcc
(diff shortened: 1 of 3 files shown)


JustVugg (approved):
Validated on x86: glm.c is purely additive (no existing lines removed), build 0 warnings, oracle 32/32 TF + 20/20 greedy — the VSX kernels are behind `#ifdef __VSX__` guards so the x86/ARM/scalar paths are provably untouched. The POWER8 speedups (12.8x int8, 7.6x int4 over scalar) I take on your S824 hardware validation, same trust model as the NEON and AVX-512-VNNI community kernels. Nice companion to the #97 build support — POWER8 now has both the build and the fast kernels. Merging to dev.

Scottcjn:
Real-model follow-up: the full GLM-5.2 int4 container (383 GB, 144 shards) now runs on the same POWER8 S824 through these kernels.

Cold first run: 48 tokens at 0.20 tok/s, expert hit rate climbing 57 to 72 percent within the run, model load 6.4s, resident dense 9.9 GB. Second run: 64 tokens at 0.28 tok/s, hit rate 82 percent, MTP acceptance 43 percent at 2.29 tokens/forward with the int8 head, matching the 39-59 percent community numbers from #8. Banner confirms idot: vsx.

This box has 512 GB RAM, so the engine auto-raised the expert cache cap to 256 per layer, the entire routed population (projected peak 389 GB). Warm-run profile for 64 tokens: expert-disk 57s, expert-matmul 97s, attention 56s. The disk term keeps shrinking as page cache fills, which leaves the integer matmul path these kernels serve as the dominant cost. That is exactly the regime where the VSX speedup pays.

One unrelated observation while running this: the filesystem heuristic flags the ext4 NVMe mount as '9p/Windows filesystem; fadvise is ineffective' on ppc64le, so readahead hints were off for all of the above. I will look into that separately.

JustVugg:
Superb end-to-end validation — POWER8 running the full 383 GB int4 container through the VSX kernels, banner confirming `idot: vsx`, and **43% MTP acceptance at 2.29 tok/forward with the int8 head, matching the 39-59% community range from #8**. That's the third independent confirmation the int8 head works exactly as advertised — which, alongside the 0%-reports (all traceable to the int4 head from the wrong mirror, e.g. #102), pins that whole mystery down. And your warm profile (disk 57 → shrinking as page cache fills, matmul 97) lands right in the regime these kernels were built for. Great work landing POWER8 in the tent, kernels and real model both.

On the filesystem warning — you found a real (if cosmetic) bug, and let me clarify the mechanism so you don't chase a ghost: that `/mnt/` check is **only a printed warning; it does not disable fadvise or readahead**. Two separate things were going on:
1. **The false positive is real**: the heuristic checks `strncmp(snap, "/mnt/", 5)` — meant to catch WSL2's `/mnt/c` 9p mounts, but `/mnt/` is a perfectly normal mount point on native Linux, so it wrongly flags your ext4 NVMe. I'm fixing it properly — a `statfs` f_type check (9p is `0x01021997`) instead of a path-prefix guess — so it only warns on an actual 9p mount. Queued behind a build that's currently occupied.
2. **But your readahead wasn't off because of that warning.** The next-block `WILLNEED` (async readahead) runs regardless. The *cross-layer* prefetch is off by default (`PREFETCH=0`) by design — `PREFETCH=1` enables it. So if you want the cross-layer hints, that flag turns them on; the `/mnt/` warning never gated them.

So nothing was actually crippled — just a misleading banner on native Linux, which I'll silence at the source. Thank you for the precision, as always.

PR #110, opened 2026-07-13
attention: heap per-thread score buffers — fix segfault past 8192 context without DSA

balisujohn:
## human note
pending human self-review, the bug is definitely real

## Bug

Both attention score buffers are fixed stack arrays, `float sc[8192]` (absorb
path and dense path). The score count `nt` is only bounded by `index_topk`
(2048) when DSA selection covers the layer. When it does not — snapshot
converted **without** `--indexer` (`has_dsa=0`), `DSA=0`, or the MTP layer —
`nt = pos+1-kv_start`, i.e. the full context width.

So on a non-indexer snapshot, any request whose context passes ~8192 tokens
writes past the end of `sc[]` on every OpenMP worker's stack:

- positions ~8192–9400: **silent stack corruption** (no fault, wrong results
  possible);
- beyond that: segfault when the writes reach the thread-stack guard page.

Observed in production on a Framework Desktop (Ryzen AI Max+ 395), int4
snapshot without indexer weights, 14.7k-token prompt at `--ctx 32768`: engine
segfaulted seconds after `[prefill] layer 1/78`, kernel log showing three OMP
workers faulting simultaneously in `attention._omp_fn.2` on the
`sc[jj]=a*c->attn_scale` store (fault address = score index ~9400).

## Fix

Allocate the score scratch once per attention call on the heap, sized
`omp_get_max_threads() × (Tk - kv_start)` (the true `nt` upper bound for both
the dense range and the DSA top-k list), and give each worker its slice via
`omp_get_thread_num()`. Non-OpenMP builds get inline fallbacks (1 thread /
id 0), keeping the dependency-free CPU path intact.

Cost: one `falloc` per attention call — e.g. 32 threads × 14.7k tokens ≈ 1.9 MB,
freed on exit; no change to the inner loops.

## Validation

- `make check`: portable build (0 warnings), C unit tests, 38 Python tests pass.
- Build without `-fopenmp` compiles (fallback path).
- Real snapshot (GLM-5.2 int4, no indexer, `KVSAVE=0 CTX=10240 DRAFT=0`),
  same binary flags, two-turn session per binary:
  - short greedy turn: **old and fixed binaries emit identical tokens**
    (`' they came into the harbor. He counted 18 ships at'` — no numeric
    change);
  - 10,232-token prompt: **old binary segfaults (rc=-11) during
    `[prefill] layer 1/78`** (reproduces the production crash); **fixed
    binary prefills layers 1–9+ cleanly** (~78 s/layer, expert-streaming
    disk bound; killed after layer 9 as the attention code under test had
    long since passed the old crash point).

Hardware: Framework Desktop (Ryzen AI Max+ 395, 128 GB), avx512-vnni build.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -35,8 +35,13 @@
 #include "tok.h"
 #include "tier.h"
 #include "grammar.h"                              /* metodo F: draft grammaticali (#48) */
+#ifdef _OPENMP
+#include <omp.h>                                  /* scratch per-thread nell'attention */
+#else
+static inline int omp_get_max_threads(void){ return 1; }
+static inline int omp_get_thread_num(void){ return 0; }
+#endif
 #ifdef COLI_CUDA
-#include <omp.h>
 #include "backend_cuda.h"
 #endif
 #ifdef __AVX2__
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Real bug, correct fix, validated. Both attention paths used `float sc[8192]` on each OMP worker's stack, but `nt` is bounded by index_topk (2048) only when DSA selection covers the layer — on a non-indexer snapshot (`has_dsa=0`), `DSA=0`, or the MTP layer, `nt` is the full context width, so any context past ~8192 tokens writes off the end of the stack on every worker. Segfault at best, silent stack corruption just under the line. Genuine and nasty.

The fix is clean: heap `sc_all = falloc(max_threads * sc_cap)` sized to the actual max nt (`Tk - kv_start`, with the MTP `kv_start=-1` +1 handled), each thread indexing its own slice — same math, just off the stack. The `omp_get_max_threads/thread_num` shims for non-OpenMP builds are correct. Validated here: build 0 warnings, oracle 32/32 TF + 20/20 greedy, and **ABSORB=1 also 20/20** (the absorb path — the other sc[8192] site — is byte-exact too). The per-call heap alloc is freed at both exits. Merging to dev — thank you, this is exactly the kind of latent overflow that only bites someone with a long chat on a non-DSA model.

Issue #108, opened 2026-07-13
Quality benchmark (the one you asked for): int4 GLM-5.2 scores 62.5% mean acc_norm — but the scoring protocol is a confound; here's the experiment that would settle it


dnnspaul:
You flagged this as "the single most valuable thing a faster machine can contribute," so here it is: a full `./coli bench` run on the int4 container.

**Headline: the int4 model scores ~62.5% mean acc_norm, well below the 85–95% the README anticipates for full-precision GLM-5.2. The gap is larger than the sampling error — but I cannot attribute it to quantization yet, and I don't think anyone should until the control below is run.**

## Result

```
task                  n     acc  acc_norm
hellaswag            40   60.0%     67.5%
arc_challenge        40   62.5%     60.0%
mmlu                 40   60.0%     60.0%

MEAN acc_norm: 62.5% across 3 tasks
```

With Wilson 95% intervals (correct for n=40; the normal approximation is not):

| task | acc_norm | 95% CI |
|---|---|---|
| hellaswag | 67.5% | [52.0%, 79.9%] |
| arc_challenge | 60.0% | [44.6%, 73.7%] |
| mmlu | 60.0% | [44.6%, 73.7%] |
| **pooled (n=120)** | **62.5%** | **[53.6%, 70.6%]** |

Random baseline is 25% (4 options), so the model is clearly working — it is not broken, just well short of published figures. The pooled upper bound (70.6%) still sits ~15 points under the low end of your 85–95% expectation, and the deficit is *consistent across three independent tasks*, which is harder to explain as noise.

## Why I am NOT concluding "int4 costs 25 points"

Two confounds have to die first, and neither can be killed with this run alone:

**1. Scoring method vs published numbers.** This harness does 0-shot log-likelihood scoring (EleutherAI style). Published frontier-model scores of 85–95% are typically few-shot and/or CoT-prompted. GLM-5.2 is a *reasoning* model, and reasoning models are known to underperform their headline scores badly under raw log-likelihood MC scoring, because you never let them think. Some — possibly most — of this gap may be the scoring protocol, not the quantization. Notably `REFERENCE` in `tools/eval_glm.py` is still unfilled, so the harness itself has no calibrated comparison point.

**2. n=40 is small.** ±14pp per task. I'd want `--limit 200`+ before quoting any single task's number.

[...]

Issue #81, opened 2026-07-12
Rotation-preconditioned quant for streamed experts (int2/int3 quality) worth considering?


locallypwned:
Hello.
First off, this is a great project and the honest numbers table is refreshing to see.
A question, not a claim, because I'm not really a coder and I got here by reasoning from hardware constraints rather than from the math.
Your engine is disk-bound: a cold token costs ~11GB of expert reads, and you list int2/int3 quality sweeps as a next step. So bytes-per-expert is basically the whole lever for tok/s. The usual reason naive int2 isn't viable is that it collapses on quality. The standard fix for that is preconditioning the weights with an orthogonal rotation (WHT plus random sign flips) before quantizing, so the outliers get spread evenly across the block and the rounding hurts less. QuaRot, QuIP#, and the TurboQuant weight-quant follow-ups all use this.
The reason I think it fits your engine specifically is because the one real cost of the rotation is the inverse transform at dequant-on-use, which is compute. But you're disk-bound, not compute-bound (until the big-RAM machines where it goes matmul-bound). So in the regime you actually ship in, that extra compute should hide in the shadow of the disk wait you're already paying. If int2 becomes viable via the rotation, that's roughly half the disk reads per token, without the quality hit that kills naive int2, and the cost is close to free where you are.
I might be missing something obvious about why this doesn't map to your kernels, or it may already be on your list. If so I'd honestly like to know why it doesn't work. If it isn't, it seemed worth raising given your disk-bound numbers and your stated int2/int3 goal. Happy to point at the specific papers and threads if useful.

ZacharyZcR:
https://github.com/kacper-daftcode/vLLM-Moet Maybe you could take a look at this? I think it might be helpful. If you have any other ideas, please feel free to share them.

locallypwned:
> https://github.com/kacper-daftcode/vLLM-Moet Maybe you could take a look at this? I think it might be helpful. If you have any other ideas, please feel free to share them.

Thanks, but I don't think this is quite it. vLLM-Moet's tiering is VRAM-to-RAM only (no SSD), and its codec is a fixed codebook, not a Hadamard rotation. Still curious if anyone's tried the rotation angle specifically on a disk-paging setup.

ZacharyZcR:
I took another close look at your proposal. Are there any additional resources or projects I can use to learn more? Thank you.

locallypwned:
> I took another close look at your proposal. Are there any additional resources or projects I can use to learn more? Thank you.

[...]

PR #115, opened 2026-07-13
tools: add quant_ablation.py — A/B any quantization scheme against fp16, engine-free

dnnspaul:
Requested in #108. An engine-free way to A/B a quantization scheme against fp16 **before** committing to a 370 GB conversion.

## What it does

Comparing colibrì's score to a published model-card number can't measure quantization: this harness scores 0-shot log-likelihood, published numbers are few-shot/CoT, and that protocol gap can swamp the effect entirely.

This removes the confound **by construction** — take an fp16 model, push its weights through colibrì's *own* quantizer (quantize → dequantize, in place), and score both with the **same harness, same questions, same machine**. The only variable is the quantizer, so the delta *is* the quantization cost.

Quantizer math is replicated from `tools/convert_fp8_to_int4.py` (symmetric absmax, per-row scales) and generalised with an optional group size, so grouped schemes can be compared head-to-head with what ships today. Scheme grammar: `fp16 | int{2,4,8}[-g<N>][-nohead]`.

```sh
python tools/fetch_benchmarks.py --out ./bench --tasks hellaswag,arc_challenge,mmlu --limit 200
python tools/quant_ablation.py --model allenai/OLMoE-1B-7B-0924 --data ./bench \
    --tasks hellaswag,arc_challenge,mmlu --limit 200 \
    --schemes fp16,int4,int4-g64,int4-g128
```

## Results it produced (OLMoE-1B-7B, n=200/task — #108)

| scheme | hellaswag | arc_c | mmlu | mean | Δ vs fp16 |
|---|---|---|---|---|---|
| fp16 | 77.0% | 47.0% | 47.0% | 57.0% | — |
| **int4 (as shipped)** | 74.0% | 41.0% | **31.5%** | 48.8% | **−8.2pp** |
| int4-nohead | 73.5% | 40.5% | 37.5% | 50.5% | −6.5pp |
| **int4-g128** | 78.5% | 45.5% | 38.0% | 54.0% | **−3.0pp** |
| int4-g128-nohead | 78.5% | 46.5% | 38.0% | 54.3% | −2.7pp |

The per-row int4 container costs ~8pp, **concentrated on the hard task** — MMLU falls to 31.5% against a 25% random baseline while easy HellaSwag barely moves. Per-row scales eat the small logit margins hard questions depend on: the same margin erosion that flips near-tie tokens in #100. `group=128` recovers ~63% of it. Keeping `lm_head`/`embed` in fp16 is **not** the fix (+1.7pp alone, +0.3pp on top of grouping).

## Coverage assert (please keep this)

`transformers` fuses MoE experts into 3D tensors (`mlp.experts.gate_up_proj = [E, in, out]`) even when the checkpoint stores one 2D matrix per expert. A `p.ndim == 2` filter therefore skips **every expert**, leaving ~85% of the model in fp16 while appearing to work — my first run did exactly that and produced a plausible, entirely fictional number. The tool prints quantized-param coverage and **exits non-zero below `--min-coverage` (default 95%)**, so that failure is loud instead of believable.

## Notes

- Dev-only tool (`torch` + `transformers`); the engine's dependency-free path is untouched.
- Scoring mirrors `tools/eval_glm.py` exactly, including `acc_norm` normalising by **character** length of the continuation (not tokens) and the same seed/shuffle.
- `make check` passes (38 tests, clean build).
- Targeted at `dev` per CONTRIBUTING.md.

Should slot straight into ranking grouped-scale / error-feedback / rotation (#81) schemes in minutes rather than days.

### DIFF
--- a/c/tools/quant_ablation.py
+++ b/c/tools/quant_ablation.py
@@ -1,0 +1,235 @@
+"""
+A/B any quantization scheme against fp16 — WITHOUT converting a 370 GB model first.
+
+Why this exists
+---------------
+Measuring "what does int4 cost us?" by comparing colibri's score to a published model-card
+number does not work: the harness scores 0-shot log-likelihood, published numbers are
+few-shot/CoT, and that protocol gap can swamp the quantization effect (see #108).
+
+This tool removes the confound by construction. It takes an fp16 model, pushes its weights
+through colibri's OWN quantizer (quantize -> dequantize, in place), and scores both with the
+SAME harness, on the SAME questions, on the SAME machine. The only variable is the quantizer,
+so the delta IS the quantization cost.
+
+It runs on a small model (OLMoE) in minutes, so a scheme can be ranked BEFORE committing to
+a multi-hour GLM conversion. The quantizer math is replicated from tools/convert_fp8_to_int4.py
+(symmetric absmax, per-row scales) and generalised with an optional group size.
+
+Measured with this tool on OLMoE-1B-7B, n=200/task (issue #108):
+
+    scheme            hellaswag  arc_c   mmlu   mean   delta
+    fp16                  77.0%  47.0%  47.0%  57.0%      --
+    int4       (shipped)  74.0%  41.0%  31.5%  48.8%   -8.2pp
+    int4-nohead           73.5%  40.5%  37.5%  50.5%   -6.5pp
+    int4-g128             78.5%  45.5%  38.0%  54.0%   -3.0pp
+    int4-g128-nohead      78.5%  46.5%  38.0%  54.3%   -2.7pp
+
+  -> the per-row int4 container costs ~8pp, concentrated on the HARD task (MMLU falls to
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Exactly the tool the quantization work needs — thank you for turning the offer into code. Security read: clean (the only `eval` is `model.eval()`, PyTorch mode-setting; no os.system/subprocess/shell/pickle/raw-network), standard ML deps (torch/transformers), and it's a self-contained research script in `tools/` — zero engine impact, can't affect default users. It does the confound-free thing you proved in #108: fake-quant (quantize→dequantize→fp16 matmul) to isolate weight error, apply-scheme parametrized (grouped, per-row, head/embed exclusions), and it should include the ~100% coverage assert to avoid the 3D-fused-experts trap you flagged. This lets us rank grouped/error-feedback/rotation (#81) schemes in minutes before a 370 GB conversion — invaluable. Merging to dev.

PR #118, opened 2026-07-13
openai: type tool arguments from the schema, implement tool_choice, add the first tool-calling tests

dnnspaul:
Two OpenAI-compatibility bugs in the new tool calling, both found by driving it against the real GLM-5.2 model. There were **no tool-calling tests** — which is how both shipped. This adds nine.

(Depends on #117 to exercise at all: `serve` currently dies on the first request.)

## 1. String-typed arguments are coerced to numbers

`parse_tool_calls` runs `json.loads()` on every `<arg_value>` **without consulting the schema**, so a string parameter whose value looks numeric comes back as an `int`. Real request, real model:

```jsonc
// tool declares:
"parameters": {"properties": {"order_id": {"type": "string"}}}

// server delivers:
"tool_calls":[{"function":{"name":"lookup_order","arguments":"{\"order_id\": 12345}"}}]
//                                                                          ^^^^^ int, not "12345"
```

Any tool with a string **id, SKU, phone number, or zip** receives the wrong type. (A zip like `"01234"` survives only by accident — the leading zero makes it invalid JSON.)

**Fix:** the declared type decides. `string` is kept verbatim and never parsed; `integer`/`number`/`boolean`/`object`/`array` decode as before; a parameter *absent* from the schema keeps the old permissive behaviour, so nothing regresses. `true` is no longer accepted as a number.

## 2. `tool_choice` is ignored entirely

`"none"`, `"required"` and `{"type":"function","function":{"name":…}}` all rendered the **identical** prompt — a client that explicitly *forbade* tools could still get `tool_calls` back. This also contradicts the server's own stated principle: unsupported parameters *"return an explicit error rather than being silently ignored."*

**Fix:**

| `tool_choice` | behaviour |
|---|---|
| `none` | tools are not declared in the prompt, and `tool_calls` are never surfaced even if the model emits markers anyway |
| `auto` | unchanged (default) |
| `required` | tools declared + instructed to call one |
| `{"type":"function",…}` | **only that function** is declared, and it is named in the instruction |
| invalid | **400** — unknown string, unknown function name, or `tool_choice` without `tools` |

## Verified on the real 744B model

Ryzen 7 9800X3D · RTX 5090 · 384 GB int4 container:

| case | before | after |
|---|---|---|
| `order_id` declared `string` | `12345` (int) | **`'12345'` (str)** |
| `tool_choice: "none"` | model called the tool | **`tool_calls: None`, `finish_reason: stop`** |
| `tool_choice: "maybe"` | silently ignored | **400 `unsupported_value`** |

## Tests

`c/tests/test_openai_server.py` had no tool-calling coverage. Added `ToolArgumentTypeTest` and `ToolChoiceTest` — 9 tests covering string-vs-numeric arguments, booleans, schema-less parameters, all four `tool_choice` modes, and the validation errors. **36 pass.**

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -196,23 +196,62 @@ def _tool_param_order(tools):
     return out
 
 
+def _tool_param_types(tools):
+    """name -> {param: declared JSON-schema type}. The model emits every argument as text;
+    without the schema a string-typed value that happens to look numeric ("12345" for an
+    order id, an SKU, a phone number) would be json.loads()'d into an int and the tool would
+    receive the wrong type."""
+    out = {}
+    for tool in (tools or []):
+        fn = tool.get("function", tool) if isinstance(tool, dict) else {}
+        name = fn.get("name")
+        if not name:
+            continue
+        props = ((fn.get("parameters") or {}).get("properties") or {})
+        types = {}
+        for key, spec in props.items():
+            if isinstance(spec, dict):
+                t = spec.get("type")
+                if isinstance(t, list):          # {"type": ["string", "null"]}
+                    t = next((x for x in t if x != "null"), None)
+                types[key] = t
+        out[name] = types
+    return out
+
+
+def _coerce_arg(value, declared):
+    """Decode a raw <arg_value> according to the declared schema type.
+
+    A string-typed parameter is kept verbatim -- never parsed as JSON. Everything else keeps
+    the previous permissive behaviour (parse if it parses, otherwise leave as text)."""
+    if declared == "string":
+        return value
+    try:
+        parsed = json.loads(value)
+    except (json.JSONDecodeError, TypeError):
+        return value
+    if declared in ("integer", "number") and isinstance(parsed, bool):
+        return value                              # `true` is not a number
+    if declared and declared not in ("integer", "number", "boolean", "object", "array"):
+        return value
+    return parsed
+
+
 def parse_tool_calls(reply, tools=None):
     """Return (content, tool_calls). Strict GLM parse; optional de-mangler (COLI_TOOL_SALVAGE=1)
     rescues malformed int4 output by mapping a lone payload onto the tool's primary parameter."""
     param_order = _tool_param_order(tools)
+    param_types = _tool_param_types(tools)
     calls, salvaged = [], []
     for match in _BOX_RE.finditer(reply):
         inner = match.group(1)
         name_match = _NAME_RE.match(inner)
         name = name_match.group(1) if name_match else inner.strip()
         args = {}
+        types = param_types.get(name, {})
         for arg in _ARG_RE.finditer(inner):
             key, value = arg.group(1), arg.group(2)
-            try:
-                value = json.loads(value)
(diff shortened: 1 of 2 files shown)


JustVugg:
Merged to `dev` (`15bf8ca`). I re-ran the suite independently in a clean worktree — **36/36 green**, including all nine new tool-calling tests. Both fixes are exactly the right call: the declared type deciding coercion (with `string` kept verbatim and `true` rejected as a number) is the correct contract, and honoring `tool_choice` — especially `none` never surfacing calls even if the model emits markers — closes a real correctness hole. Adding the tests that were missing is the part I appreciate most.

Noted the dependency on #117 (`serve` heap-overflow on first request) — that's the ASan fix that makes any of this reachable end-to-end, so I'll review it next. Thanks for driving these against the real 744B model; that's where they surfaced and that's what makes them trustworthy.

dnnspaul:
Pushed a second commit: I went back and audited the implementation against the **full OpenAI tool-calling contract** rather than only the bugs I happened to trip over. Three more gaps, one of which is a silent correctness bug.

## 3. Tool results were bound by **position**; `tool_call_id` was never read

This is the serious one. GLM's template renders results as a positional run of `<tool_response>` blocks, so the model pairs the Nth result with the Nth call. The OpenAI contract binds them by **`tool_call_id`** and does **not** require the client to preserve order.

```jsonc
assistant: tool_calls [ call_W → get_weather, call_T → get_time ]
tool: {"tool_call_id": "call_T", "content": "14:05"}        // legal: out of order
tool: {"tool_call_id": "call_W", "content": "22C, sunny"}
```
rendered:
```
<|observation|><tool_response>14:05</tool_response><tool_response>22C, sunny</tool_response>
```
→ **the time answers the `get_weather` call.** No error — just a wrong answer, deterministically.

Now each consecutive run of tool messages is reordered to match the ids from the preceding assistant turn; an id matching no pending call is a 400; untagged results keep client order (lenient for callers that omit the field).

## 4. `parallel_tool_calls` was ignored entirely (0 occurrences in the file)

Contract: at most one call per turn when `false`. The parser returned as many as the model emitted. Now `false` both instructs the model and truncates to the first call; a non-boolean is a 400.

## 5. `strict` was silently stripped

`openai_server.py:310` dropped `strict` from the declaration. Structured-output guarantees aren't implemented, so a caller who explicitly asked for one got unvalidated arguments and no error. Now a 400, consistent with how `stop`/`logprobs`/`seed` are already handled.

**This one is a judgement call and I'd defer to you:** rejecting means clients that set `strict: true` by default start failing. If you'd rather log-and-ignore, it's a one-line change — say which and I'll adjust.

## Audited and already correct — no change needed

**Streaming.** The delta shape is spec-conformant: per-call `index`, `id`, `type: "function"`, `function.{name,arguments}`, and `finish_reason: "tool_calls"`. Emitting arguments in one chunk rather than incrementally is valid — clients accumulate. Credit where due; I went looking for a bug here and didn't find one.

Tests now **45 pass** (18 new). `make check` clean.

PR #122, opened 2026-07-13
fix: 8k-token stack overflow in MLA attention (fixed sc[8192] score buffer)

Scottcjn:
## The bug

The MLA attention path holds one score per attended key in a fixed-size `float sc[8192]` on the stack (both the absorb/decode loop and the dense/prefill loop). On layers that run DSA sparse selection, `nt` is capped at `index_topk` (~2048), so it fits. But on layers **without** selection — the `first_k_dense_replace` dense layers, or any run with the indexer off — `nt` is the full context length. Prefill a prompt longer than **~8192 tokens** and the score loop writes past `sc[8192]` and smashes the stack.

## Reproduction

On GLM-5.2, any prompt over ~8k tokens crashes at `prefill layer 1` (a dense layer). Captured on a POWER8 build: a SIGSEGV with the fault address in the stack region and a float-store faulting instruction — textbook stack overflow. A 2,827-token prompt worked; a 12,334-token prompt died ~40s in, every time.

Raising the stack limit only *hides* it: the out-of-bounds writes still clobber adjacent stack state, so a longer prompt would produce silently wrong output rather than crash. The buffer has to leave the stack.

## The fix

Heap-size the score buffer to `nt` in both attention loops. No behavioral change on prompts that already fit — the computation is identical, the buffer is just correctly sized.

## Validation

- Token-exact after the change: teacher-forcing 32/32 and greedy 20/20 vs the transformers oracle.
- The 12,334-token prompt that reliably segfaulted at ~40s now runs 18+ minutes past that point with no crash.

Arch-independent: reproduced on a POWER build, but the fixed stack array overflows the same way on x86 at the same context length.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -472,6 +472,22 @@ static void quant_scratch(size_t xn, size_t sn, int8_t **xq, float **sx){
     *xq=g_qscratch.xq; *sx=g_qscratch.sx;
 }
 
+/* Per-thread attention score scratch (mirrors g_qscratch). Holds one score per
+ * attended key; nt reaches the full context on layers without DSA selection, so
+ * a fixed sc[8192] stack array smashed the stack past ~8k tokens. A thread-local
+ * buffer that grows only when a longer context needs it avoids both that overflow
+ * and a malloc/free per (s,h) in the parallel attention loops. */
+static _Thread_local float *g_scscratch=NULL; static _Thread_local size_t g_scscratch_cap=0;
+static float *score_scratch(int nt){
+    size_t n = nt>0 ? (size_t)nt : 1;
+    if(n>g_scscratch_cap){
+        float *p=realloc(g_scscratch,n*sizeof(float));
+        if(!p){ fprintf(stderr,"OOM attention scores\n"); exit(1); }
+        g_scscratch=p; g_scscratch_cap=n;
+    }
+    return g_scscratch;
+}
+
 static void matmul_qt(float *y, const float *x, QT *w, int S){
 #ifdef COLI_CUDA
     /* The CUDA backend owns persistent copies only for model-resident tensors.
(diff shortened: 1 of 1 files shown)


dnnspaul:
Your diagnosis is spot-on, but I think this may already be fixed on `main` — worth checking before you spend more time on it.

**#110 (`342ceaf`, merged 08:52 today) removed exactly this buffer**: *"attention: heap per-thread score buffers — fix stack overflow past 8192 ctx on non-DSA snapshots (both absorb + dense paths, oracle-exact)"*. On current `main`:

```console
$ git show origin/main:c/glm.c | grep -c 'float sc\[8192\]'
0
```

The standalone `attention()` body your diff patches (around line 1203) was also refactored away — `attention()` at :1630 is now a thin wrapper over `attention_rows()`, and both score buffers there are `falloc`'d. This PR was opened at 15:17, so it looks like the branch predates #110 landing rather than anything being wrong with the analysis.

One thing I'd flag if you do rebase: #110 allocates **one per-thread buffer per layer**, whereas this patch does `malloc`/`free` **inside the OpenMP loop, per (sequence, head)** — that's an allocation per head per row in the hottest loop in the engine. Even where it's correct it would likely cost measurable throughput.

**Your repro is valuable independently, though**, and it corroborates something I'm hitting right now: I'm running an 11,102-token prompt (an agentic client's system prompt + tool schemas) on the real 744B model, and it grinds at prefill layer 1 with zero disk I/O — same fingerprint as yours, same context range. On post-#110 `main` it isn't a stack smash, it's just **O(T²) attention with no DSA**: the pre-converted HF int4 repos ship **no `out-idx-*` weights**, so the lightning indexer never engages and `nt` is the full context on *every* layer, not just the dense ones. I'm extracting the indexer weights now (`--indexer`, 20 shards) to confirm that's what makes long contexts tractable.

So if you can still reproduce a crash on **current** `main` (not a pre-#110 base), that's a different and still-live bug and I'd very much like to see it — there is one more sizing bug in that same function that #110 didn't catch (batched serve path: #117).

Scottcjn:
Revised the fix after an adversarial review pass. The first version heap-allocated the score buffer per (s,h) iteration, which (1) put a malloc/free in the hot parallel attention loop and (2) called exit() on OOM from inside an OpenMP worker. The current version uses a per-thread growing scratch buffer (score_scratch) modeled on the existing g_qscratch in the same file: allocated once per thread, grown only when a longer context needs it, reused across iterations, with the OOM check only on grow. Same correctness (token-exact vs the oracle, TF 32/32 / greedy 20/20), no per-iteration allocation, no fixed cap.

JustVugg:
Thank you for this, @Scottcjn — the analysis is exactly right, and you found it independently. The catch: the same bug was already fixed in **#110** (merged to `main`), which moved that score buffer off the stack entirely — `float sc[8192]` no longer exists anywhere in `main` or `dev` (it's now a heap allocation sized by context length, so >8k prompts grow the buffer instead of smashing the stack). That's why this PR shows as CONFLICTING: it edits code that's already gone.

I verified before writing this: `grep 'float sc\[' c/glm.c` is empty on current `main`. So there's nothing left to fix here — but you spotted a genuine bug on your own, and a *related* sizing bug in the batched serve path (also in this buffer) just landed as #117. If you can still reproduce a >8k-token crash on current `main`, please reopen with the trace — I'd take that very seriously. Closing as already-fixed by #110, with thanks.

Issue #113, opened 2026-07-13
[Performance]: native Windows 11 (MinGW) — i5-12600K / 32 GB / WD SN770 → 0.08–0.14 tok/s (first native-Windows datapoint)


bopof:
### Commit

5254470f95b8340a50c3517fe4f420ce3fd914c4

### Hardware and storage

- **CPU:** Intel Core i5-12600K (10C / 16T, AVX2, **no AVX-512**)
- **RAM:** 32 GB
- **GPU / VRAM:** none used — CPU-only build (`CUDA=0`); colibrì streams experts from disk, GPU tier not compiled
- **Disk:** WD_BLACK SN770 1 TB (PCIe4 NVMe) — model stored here
- **Filesystem:** NTFS

### Software environment

- **OS:** Windows 11 (build 26200), **native — no WSL**
- **Compiler:** WinLibs MinGW-w64 **GCC 16.1.0** (`x86_64-ucrt-posix-seh`, POSIX threads / OpenMP)
- **CUDA / driver:** n/a (CPU-only build)
- **Relevant env vars:** `SNAP`, `PROMPT`, `NGEN=20`; `TOPP=0.7` and `PIN` / `PIN_GB=4` per run. (`HF_HUB_DISABLE_XET=1` was needed only for the *model download* — the Xet transfer path stalled on the plain-LFS safetensors — not at inference.)

### Build and benchmark commands

```shell
# --- build (static, this machine) ---
# NOTE: compiled glm.c directly. The Makefile's uname-based Windows detection
# doesn't fire from plain PowerShell/CMD, so `make` fell through to the Linux
# branch (no .exe / -static). Building from Git-Bash / MSYS2 uses the MINGW branch.
gcc -D_FILE_OFFSET_BITS=64 -O3 -march=native -fopenmp -static glm.c    -o glm.exe    -lm -fopenmp
gcc -D_FILE_OFFSET_BITS=64 -O3 -march=native -fopenmp -static iobench.c -o iobench.exe -lm -fopenmp

# --- disk (see caveat in Results: O_DIRECT unavailable on this MinGW build) ---
# measured against a 64 GB file (= 2x RAM) to defeat the page cache
iobench.exe <64GB-file> 19 128 8 0

# --- inference: 3 configs, same prompt, NGEN=20, cold ---
set SNAP=<model-dir>
set PROMPT=Q: In one short paragraph, explain what a Mixture-of-Experts language model is. A:
set NGEN=20

[...]

PR #129, opened 2026-07-13
Makefile: detect Windows from $(OS) so make works from native PowerShell/CMD

bopof:
Follow-up to #113 (first native-Windows benchmark datapoint). Fixes the Makefile Windows-detection papercut noted there.

## Problem
`c/Makefile` gates the Windows branch on `uname -s` (matching `MINGW*`/`MSYS*`). But `uname` is not on `PATH` in a plain **PowerShell/CMD** shell, so `$(shell uname -s)` returns empty. The `Darwin` and `IS_WIN` checks then both fail and the build falls through to the **Linux branch**, producing an extensionless, non-static binary instead of a static `glm.exe`. Today the only native-shell workaround is compiling `glm.c` directly.

## Fix
Detect Windows from the **`OS` environment variable** instead. GNU Make imports it, and Windows sets it to `Windows_NT` in **every** shell — cmd, PowerShell, MSYS2, and Git-Bash. `uname` is now only invoked when *not* on Windows.

- Native PowerShell/CMD: `make` now takes the Windows branch → `-march=x86-64-v3 -static`, emits `glm.exe`.
- MSYS2 / Git-Bash: **unchanged** — they also see `OS=Windows_NT`, so they take the same Windows branch and emit the same binary as before.
- macOS / Linux / PowerPC branches: untouched (`uname -s`/`uname -m` still drive them).

## Verification
`mingw32-make glm` from a **native PowerShell** shell (no MSYS2, no Git-Bash) with WinLibs MinGW-w64 GCC 16.1.0:

```
gcc -D_FILE_OFFSET_BITS=64 -O3 -march=x86-64-v3 -fopenmp ... glm.c -o glm.exe -lm -fopenmp -static
```

Exit 0, `glm.exe` produced. Top-level `mingw32-make` delegation into `c/` also works from the native shell.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1,7 +1,18 @@
-UNAME_S := $(shell uname -s)
-MINGW    := $(findstring MINGW,$(UNAME_S))
-MSYS     := $(findstring MSYS,$(UNAME_S))
-IS_WIN   := $(MINGW)$(MSYS)
+# --- OS detection ---
+# On Windows the OS env var is 'Windows_NT' in EVERY shell (cmd, PowerShell,
+# MSYS2, Git-Bash), and GNU Make imports it. Detect Windows from it FIRST so
+# `make` works from a native PowerShell/CMD shell — there `uname` is not on
+# PATH, so the old uname-only check returned empty and fell through to the
+# Linux branch (no .exe, no -static). We only call `uname` when NOT on Windows.
+ifeq ($(OS),Windows_NT)
+IS_WIN   := 1
+UNAME_S  :=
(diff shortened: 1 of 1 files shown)


rofl0r:
uname is actually quite broken for detecting the **target** platform. mingw is essentially a crosscompiler, but it (and any other gcc or clang based compiler) can be queried with `$(CC) -dumpmachine`, which displays the target triple (e.g. x86_64-unknown-linux-musl). using that result you can then e.g. detect cygwin or mingw in that string (or linux, or apple, etc).

Issue #123, opened 2026-07-13
[Bug]:  Windows Build Failure – Required Fixes for `glm.c` and `compat.h`


DevLenn:
### Commit or version

latest

## Description

While building the project on Windows (MinGW/MSYS2), I encountered several compilation and compatibility issues related to POSIX-specific functionality that is not available or behaves differently on Windows.

The build was failing because `glm.c` unconditionally included `sys/select.h`, and additional runtime/build issues appeared around `fstat`, `mmap`, `madvise`, and `select()` usage.

### 1. Guard `sys/select.h` in `glm.c`

I replaced:

```c
#include <sys/select.h>
```

with:

```c
#ifndef _WIN32
#include <sys/select.h>
#endif
```

Using PowerShell:

```powershell
(Get-Content glm.c -Raw) -replace '#include <sys/select\.h>', "#ifndef _WIN32`r`n#include <sys/select.h>`r`n#endif" | Set-Content glm.c
```

### 2. Add Windows compatibility layer to `compat.h`

I added the following block directly before:

```c
#endif /* _WIN32 */
```

This provides:

* Proper `struct stat` visibility.
* A 64-bit-safe `fstat()` implementation using `_fstat64`.
* Minimal read-only `mmap()`/`munmap()` support using Windows file mappings.
* A no-op `madvise()` implementation.
* A replacement for `select()` on stdin, since Windows `select()` only supports sockets.

```c
/* --- struct stat / off_t: ensure sys/stat.h is visible on this path --- */
#include <sys/stat.h>
#include <string.h>

/* --- fstat -> _fstat64: force a real 64-bit file size regardless of
 * whichever off_t width the ambient MinGW headers picked. Model shards can
 * exceed 4 GB, so a 32-bit size would silently truncate/wrap. --- */
static inline int compat_fstat(int fd, struct stat *buf){
    struct __stat64 sb;
    if(_fstat64(fd, &sb) != 0) return -1;
    memset(buf, 0, sizeof(*buf));
    buf->st_size = (off_t)sb.st_size;
    buf->st_mode = sb.st_mode;
    return 0;
}
#define fstat(fd,buf) compat_fstat(fd,buf)

/* --- mmap/munmap/madvise: minimal read-only support for model-shard mapping ---
 * Only the subset glm.c actually uses (PROT_READ, MAP_SHARED, whole/partial
 * read-only views) is implemented; this is not a general POSIX mmap. */
#ifndef PROT_READ
[...]

Issue #128, opened 2026-07-13
Windows native benchmark: Core Ultra 9 185H + RTX 5070 Ti — 0.03→0.5 tok/s after warmup, + CUDA/AVX-VNNI/RAM/pipe fixes


woolcoxm:
## Hardware

| component | spec |
|---|---|
| CPU | Intel Core Ultra 9 185H (Meteor Lake, 16C/22T, AVX2 + AVX-VNNI) |
| RAM | 32 GB |
| GPU | NVIDIA GeForce RTX 5070 Ti, 16 GB VRAM (Blackwell, sm_120) |
| Disk | Crucial P3 1TB NVMe (QLC), NTFS — model on this drive |
| OS | Windows 11 Pro (native, no WSL) |
| Toolchain | MinGW-w64 GCC 15.2 + MSVC 2022 Build Tools 14.44 + CUDA 12.8 (nvcc) |

## Model

Pre-converted int4 from `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp` — the **int8 MTP head** version (sizes `3527131672 / 5366238584 / 1065950496`, confirmed). 144 shards, 383.7 GB.

## What I had to fix to get it running natively on Windows

The `Windows` branch compiled but had four runtime blockers. All four are fixed in commit woolcoxm/colibri@72ddeed (pushed to my fork's `Windows` branch).

### 1. RAM detection returned 0 → expert cache capped at 0/layer
`resource_plan.py:memory_available()` only read `/proc/meminfo`. On native Windows Python that path doesn't exist, so it returned `0`, the budget defaulted to 8 GB, and `coli doctor` reported `cap 0/layer`. The C engine itself was unaffected (its `compat.h` uses `GlobalMemoryStatusEx`), but the Python planning layer (doctor/plan/chat auto-budget) was broken.

**Fix:** Windows branch using `GlobalMemoryStatusEx` (`ullAvailPhys`), matching the C engine's `compat_meminfo`, with a `GetPhysicallyInstalledSystemMemory` fallback.

### 2. `coli chat` hung at ~10 GB resident (serve-mode pipe deadlock)
[...]

Issue #101, opened 2026-07-12
Benchmark: RTX 5090 + Ryzen 9800X3D + PCIe 5.0 (WSL2) — 10.51 GB/s VHDX; CUDA expert tier ≈ 0% once OMP tuning is on [corrected]


dnnspaul:
The README asks for datapoints from faster machines, and notes the CUDA tier "intentionally makes no end-to-end speedup claim before the full model is benchmarked." Here is that measurement, plus a `PILOT=1` result.

**Headline: this machine is matmul-bound, not disk-bound — in every single config.** The CUDA expert tier is worth **+14%** end-to-end, and it buys that entirely by cutting matmul time, not disk time.

## Machine

| | |
|---|---|
| CPU | AMD Ryzen 7 9800X3D (8C/16T), AVX-512 VNNI |
| RAM | 94 GB physical · **70 GB** allocated to WSL2 · `RAM_GB=65` budget |
| GPU | RTX 5090, 32 GB (`sm_120`), CUDA 13.0 |
| Disk | Samsung 9100 PRO PCIe 5.0 (1 TB), ext4 **inside a WSL2 VHDX** |
| OS | Windows 11 + WSL2 Debian, kernel 6.6.87.2 · gcc 14.2.0 |
| Model | `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp` (int8 MTP head) |
| colibrì | `f38e74e`, `make glm CUDA=1 ARCH=native` |

### iobench

```
O_DIRECT x8 threads: 64 x 19MB = 1.3 GB in 0.12s -> 10.51 GB/s
buffered x8 threads: 64 x 19MB = 1.3 GB in 0.15s ->  8.62 GB/s
```

**10.51 GB/s O_DIRECT through a WSL2 VHDX** — worth flagging, since the README treats WSL2 VHDX as a ~1 GB/s ceiling. That was the old dev box's drive, not the virtualization layer. VHDX is not inherently slow.

## Decode (greedy, `--ngen 32`, `RAM_GB=65`, same prompt)

| config | tok/s | expert hit | expert-disk | expert-matmul |
|---|---|---|---|---|
| CPU only, MTP on | 0.29 | 47.4% | 36.9 s | 55.9 s |
| **+ CUDA expert tier** (24.4 GB VRAM, 1281 experts) | **0.33** | 44.8% | 35.1 s | **46.1 s** |
[...]

PR #131, opened 2026-07-13
Windows native: CUDA via runtime DLL, AVX-VNNI kernel, RAM detection, serve-mode pipe fix

woolcoxm:
## What this adds

Four Windows-port fixes that make colibrì run natively on Windows 11 with full CPU+GPU acceleration — no WSL required. All validated on real hardware (benchmark datapoint in #128).

### 1. RAM detection (`resource_plan.py` + test)
`memory_available()` only read `/proc/meminfo`. On native Windows Python that path does not exist, so it returned `0`, the RAM budget defaulted to 8 GB, and the expert cache was sized to **0 slots/layer** — making the engine effectively unusable through `coli chat`/`doctor`/`plan`. The C engine itself was unaffected (its `compat.h` uses `GlobalMemoryStatusEx`), but the Python planning layer was never given a Windows path.

**Fix:** Windows branch using `GlobalMemoryStatusEx` (`ullAvailPhys`), matching the C engine's `compat_meminfo`, with a `GetPhysicallyInstalledSystemMemory` fallback for sandboxed callers. Regression test included (`test_memory_available_is_positive`).

### 2. Serve-mode pipe fix (`glm.c` + `coli`)
The engine speaks a byte protocol with the `coli` wrapper (`\x01\x01READY\x01\x01\n`). On Windows the MinGW CRT opens stdout in **TEXT mode**, translating `\n`→`\r\n`, so the sentinel never matched and `coli chat` hung at ~10 GB resident (engine loaded fine in ~4s, then sat blocked on stdin while the wrapper waited for a READY that arrived as `...\r\n`).

**Fix:** `_setmode(_fileno(stdin/stdout), _O_BINARY)` in `run_serve()` so the protocol bytes are exact. Also moved child stderr from a temp-file redirect (which stalled the Windows pipe layer) to a PIPE with a bounded non-blocking drain into the errlog.

### 3. AVX-VNNI 128-bit kernel (`glm.c`)
Intel Alder Lake+ and Meteor Lake+ CPUs have `VPDPBUSD` (AVX-VNNI, a 128-bit int8 dot-product instruction), but the engine only checked for the 512-bit AVX512-VNNI variant, so these CPUs ran the slower AVX2 `maddubs` path.

**Fix:** Added a `__AVXVNNI__ && __AVX2__` branch to `dot_i8i8` and `dot_i4i8` using `_mm_dpbusd_epi32` (~30 lines total). **Validated bit-identical to the existing AVX2 kernel** across all tested lengths, for both int8 and int4. Build with `make glm.exe ARCH=native` to enable; `x86-64-v3` default compiles it out safely via the `#ifdef` guards. Also fixed a pre-existing intrinsic typo (`_mm256_extracti128_si128` → `_mm256_extracti128_si256`) that blocked `-march=native` builds entirely.

### 4. CUDA via runtime DLL (`backend_loader.c` + `backend_cuda.h` + `Makefile`)
MinGW gcc cannot compile `.cu` (nvcc needs MSVC `cl.exe`), and cross-linking MSVC objects into a gcc binary is fragile. The Makefile hard-errored: *"CUDA=1 on Windows requires G1"*.

**Fix:** Build `backend_cuda.cu` into a standalone `coli_cuda.dll` with nvcc+MSVC (`-shared -DCOLI_CUDA_BUILDING_DLL`), then the host `glm.exe` loads it at runtime via `LoadLibrary`/`GetProcAddress`. `backend_loader.c` resolves the 11 `extern "C"` API symbols on first use; if the DLL is absent the engine falls back to CPU without error. `ColiCudaTensor*` is opaque to the host (stored, never dereferenced), so the MSVC-allocated struct is safe across the ABI boundary.

New Makefile targets: `make cuda-dll` (builds the DLL, requires `vcvars64.bat`) and `CUDA_DLL=1` (builds the host with the loader instead of cudart). **Verified on RTX 5070 Ti (sm_120)** — see #128 for the startup log.

## Also included
- `warmup.ps1` — overnight expert-cache priming script (runs `coli run` in a loop with diverse prompts to build the `.coli_usage` histogram)
- README Windows section updated with CUDA + AVX-VNNI build instructions
- `.gitignore` updated for Windows build artifacts

## Benchmark (from #128)

| metric | value |
|---|---|
| Cold decode (hit 1%) | 0.03 tok/s |
| Warm decode (after warmup) | ~0.5 tok/s (**~17× from cache warming**) |
| Load time | ~4–8s |
| Profile (cold) | expert-disk 5.2s · expert-matmul 0.78s (disk-bound) |
| GPU tier | 103 hot experts → 1.95 GB VRAM ✅ |

## Files changed
- `c/resource_plan.py` + `c/tests/test_resource_plan.py` — RAM detection + test
- `c/glm.c` — serve-mode binary stdin/stdout, AVX-VNNI kernel, intrinsic typo fix
- `c/coli` — stderr pipe drain
- `c/backend_loader.c` (new) — LoadLibrary shim
- `c/backend_cuda.h` — `__declspec(dllexport)` for DLL exports
- `c/Makefile` — `cuda-dll` + `CUDA_DLL=1` targets
- `c/warmup.ps1` (new) — cache priming script
- `README.md` + `.gitignore`

Happy to split into smaller PRs if that's preferred. The fixes are independent — each solves a distinct blocker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,15 +11,26 @@ desktop/src-tauri/gen/
 
 # binari compilati (si rigenerano con make / coli build)
 c/glm
+c/glm.exe
 c/olmoe
+c/olmoe.exe
 c/iobench
+c/iobench.exe
 c/tok_test
 c/backend_cuda.o
 c/backend_cuda_test
+c/backend_loader.o
+c/coli_cuda.dll
+c/coli_cuda.lib
+c/coli_cuda.exp
 c/tests/test_json
+c/tests/test_json.exe
 c/tests/test_st
+c/tests/test_st.exe
 c/tests/test_tier
+c/tests/test_tier.exe
 c/tests/test_grammar
+c/tests/test_grammar.exe
 
 # oracoli tiny generati (make_glm_oracle.py) e dati benchmark scaricati
 c/glm_tiny/
--- a/README.md
+++ b/README.md
@@ -160,6 +160,12 @@ make iobench.exe        # disk I/O benchmark
 make test-c             # run C tests
 make test-python        # run Python tests (requires python)
 
+# AVX-VNNI: Intel Alder Lake+ (and Meteor Lake+) CPUs have a 128-bit int8
+# dot-product instruction (VPDPBUSD) the engine can use for ~1.3x faster
+# quantized matmul. The x86-64-v3 default (portable AVX2) compiles it out;
+# build for THIS machine to enable it:
+make glm.exe ARCH=native                       # banner prints "idot: avx-vnni"
+
 # Verify (tiny model, 2.4 MB):
 pip install torch transformers safetensors huggingface_hub
 python tools/make_glm_oracle.py                # generate tiny oracle
@@ -171,9 +177,49 @@ python coli chat --model D:\glm52_i4            # interactive chat
 python coli serve --model D:\glm52_i4            # OpenAI-compatible API
 ```
 
-**Status:** Phase 1 complete (compiles, correct, static-linked). O_DIRECT (Phase 2),
-GPU via `LoadLibrary` on `coli_cuda.dll` (Phases G0–G2), and full-model validation
-are separate workstreams. See `PORT_WINDOWS_PLAN.md` for the full plan.
+**Warmup (overnight cache priming):** the engine's expert cache learns from
+your workload. The included `warmup.ps1` script runs `coli run` in a loop with
+diverse prompts to build the `.coli_usage` histogram unattended, so the next
+real session starts with a large, accurate hot-expert pin. Each run saves usage
+atomically on clean completion.
+
+```powershell
+.\warmup.ps1 -Rounds 1 -Ngen 32               # ~60-90 min, durable progress
+```
+
+**NVIDIA GPU (optional, via runtime DLL):** on Windows the engine is built with
+MinGW gcc but CUDA kernels require MSVC + nvcc. The split is clean: build the
+CUDA backend into a standalone `coli_cuda.dll` (nvcc + MSVC), then the host
+`glm.exe` loads it at runtime via `LoadLibrary` (`c/backend_loader.c`). The host
+never links cudart directly; if the DLL is absent the engine falls back to CPU
+without error.
+
+```powershell
+# Prerequisites: CUDA Toolkit + MSVC Build Tools (cl.exe) + nvcc on PATH.
+# Build the DLL from a shell with the MSVC environment set (vcvars64.bat or
+# "x64 Native Tools Command Prompt for VS"):
+make cuda-dll CUDA_HOME="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.8" CUDA_ARCH=sm_120
+
+# Build the host with the runtime loader (CUDA_DLL=1 adds -DCOLI_CUDA and
+# links backend_loader.o instead of cudart):
+make glm.exe CUDA_DLL=1 ARCH=native
+
+# Run with the GPU expert tier (8 GB VRAM budget here; scale to your free VRAM):
+$env:COLI_CUDA="1"; $env:COLI_GPU="0"; $env:CUDA_EXPERT_GB="8"
+python coli chat --model D:\glm52_i4 --topp 0.7
+```
+
+The DLL exports 11 `extern "C"` symbols (`coli_cuda_init`, `coli_cuda_matmul`,
+etc.); `backend_loader.c` resolves them via `GetProcAddress` on first use.
+`ColiCudaTensor*` is opaque to the host (stored, never dereferenced), so the
+MSVC-allocated struct is safe across the ABI boundary. `CUDA_ARCH` must match
+your GPU's compute capability (e.g. `sm_120` for Blackwell / RTX 50-series,
+`sm_89` for Ada / RTX 40-series).
+
+**Status:** Phase 1 complete (compiles, correct, static-linked). The Windows
+GPU tier (runtime `coli_cuda.dll` via `LoadLibrary`) is implemented and
+verified on RTX 50-series (sm_120). O_DIRECT (Phase 2) and full-model
+validation against the transformers oracle remain separate workstreams.
 
 ### OpenAI-compatible API
 
--- a/c/Makefile
+++ b/c/Makefile
@@ -24,9 +24,12 @@ EXE     =
 else ifneq ($(IS_WIN),)
 # --- Windows 11 x86-64 (MinGW-w64 / MSYS2) ---
 # GCC + libgomp + winpthreads: pthread, OpenMP, clock_gettime, opendir/readdir,
-# AVX2 intrinsics — tutto gratis, nessun porting.
-# ARCH default = x86-64-v3 (binario portabile con AVX2). Per max velocita'
-# su QUESTA macchina: make ARCH=native
+# AVX2 intrinsics - tutto gratis, nessun porting.
+# ARCH default = x86-64-v3 (portable binary with AVX2). For max speed on THIS
+# machine use ARCH=native: on AVX-VNNI CPUs (Intel Alder Lake+, Meteor Lake+)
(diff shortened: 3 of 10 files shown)


JustVugg:
Good news from a real Windows validation: I installed a MinGW-w64 cross-compiler and built this branch's CPU-only `glm.c` to a Windows `glm.exe`, then ran it via WSL interop — it compiles clean (`-D_FILE_OFFSET_BITS=64`) and loads the model + runs inference on Windows. The `sys/select.h` / mmap / madvise / select portability path works. So the build-failure half of #123 is solid.

One blocker before it can land: this branch is based on an older `dev` (its `make check` runs 35 tests vs 58 on current `dev`), so it would revert the recent serve/tool-calling/gpu-flag fixes (#117/#118/#125) that just went to `main`. Could you **rebase onto current `dev`**? Once rebased I can integrate the CPU-portability path directly. The CUDA-on-Windows (DLL) parts I can't test here — no NVIDIA on this box — so we'd review/gate those separately with a CUDA-Windows validator. Thanks for the port work; the hard part is done and verified.

woolcoxm:
doing it now will post back when its done )

woolcoxm:
Rebased onto `dev` (`4418e16`) as requested. All recent dev work (#117 serve heap-overflow, #118 tool-calling, #125 gpu-flag handling, #95 AVX-512 i4 accumulator, #68 continuous batching) is preserved — our commit sits cleanly on top.

**Conflict resolutions during rebase:**
- `c/Makefile` — merged dev's expanded TEST_BINS (test_decode_batch, test_idot, test_i4_acc512) with our CUDA_DLL path and cuda-dll target
- `c/backend_cuda.h` — kept dev's new fused CUDA API (expert_mlp, expert_group, attention_absorb, group_stats) and added `COLI_CUDA_DLLEXPORT` to them so they export from the DLL too
- `c/backend_loader.c` — added LoadLibrary resolution + wrappers for the 4 new CUDA functions dev added
- `c/glm.c` — guarded 3 POSIX-only paths from dev's recent commits so they compile on native MinGW:
  - `#include <sys/select.h>` → `__APPLE__ || __linux__` guard (continuous-batching serve uses Unix `select()`)
  - `map_of_fd()` mmap/fstat → POSIX guard (returns NULL on Windows; COLI_MMAP defaults off)
  - `madvise(MADV_WILLNEED)` → POSIX guard
  - `run_serve_mux()` → POSIX guard with a Windows stub that reports SERVE_BATCH is unavailable (falls back to the single-sequence serve path)

The CPU portability path (what you validated) is untouched. The CUDA-on-Windows DLL parts are isolated in `backend_loader.c` + the `cuda-dll` Makefile target and only activate under `CUDA_DLL=1`.

Builds clean on native MinGW (`make glm.exe CUDA_DLL=1 ARCH=native`), all C tests pass.

woolcoxm:
Thanks — happy to clarify for the README row.

**The 0.5 tok/s requires #131. Stock main cannot reach it.** Here's why:

Stock main on native Windows hangs at the READY sentinel — `coli chat` loads the engine (~10 GB resident in ~4s), then the wrapper spins forever because the MinGW CRT translates `\n`→`\r\n` on stdout and the `\x01\x01READY\x01\x01\n` sentinel never matches. So stock main literally cannot run interactive chat on Windows; it deadlocks on the first prompt. The 0.5 tok/s session required the serve-mode pipe fix (`_setmode _O_BINARY`) just to reach a prompt.

What actually drove the 0.03→0.5 jump is **cache warming**, not the code fixes — but cache warming only works if (a) the engine runs at all (pipe fix) and (b) the expert cache is sized correctly (RAM detection fix, which took it from `cap 0/layer` to `cap 9/layer`). So:

| component | needed for 0.5 tok/s? |
|---|---|
| serve-mode pipe fix | **yes** — stock main hangs without it |
| RAM detection fix | **yes** — stock sizes cache to 0/layer on Windows |
| AVX-VNNI kernel | modest — faster matmul, but disk-bound so small effect |
| warmup data (`.coli_usage`) | **yes** — this is what turned cold reads into hits |
| GPU (`COLI_CUDA=1`) | **no** — that session didn't use it; measured neutral/slightly-negative separately |

**For the README row, label it:** *"with #131 (pipe + RAM fixes), warm cache, no GPU"*. The honest framing is: the number demonstrates cache warming on Windows once the portability blockers are fixed — not a GPU or kernel win. The GPU datapoint (tested separately, ~0% at low hit rate) is in #128 and matches the #101 finding.

**On the commit split:** good call on isolating the AVX-VNNI kernel for A/B. I'll split the PR into:
1. CPU portability (pipe fix + RAM detection + POSIX guards) — what you validated
2. AVX-VNNI kernel — standalone, A/B-able on Meteor Lake
3. CUDA DLL path — gated separately

That also lets (1) land independently since you've already confirmed it compiles and runs. I'll rebase onto dev and push the split commits today.

woolcoxm:
Split into 3 commits as you suggested, rebased onto `dev` (`4418e16`):

```
b20bf92 Windows CUDA: runtime coli_cuda.dll via LoadLibrary (nvcc+MSVC split)
d1a62ca AVX-VNNI 128-bit int8/int4 dot kernel (Intel Alder Lake+ / Meteor Lake+)
ba242dd Windows: serve-mode pipe fix, RAM detection, POSIX guards, warmup script
```

**Commit 1 (CPU portability)** — what you cross-compiled and validated: serve-mode pipe fix, RAM detection, POSIX guards for dev's select/mmap/madvise code, warmup script. This can land independently.

**Commit 2 (AVX-VNNI kernel)** — standalone, 1 file (`glm.c`), 37 lines. A/B-able on Meteor Lake: build with `ARCH=native` (enables `__AVXVNNI__`, banner prints `idot: avx-vnni`) vs `ARCH=x86-64-v3` (compiles out, falls back to AVX2 `maddubs`). Validated bit-identical output to the AVX2 kernel. Also includes the `_mm256_extracti128_si256` typo fix that was blocking `-march=native` builds.

**Commit 3 (CUDA DLL)** — gated under `CUDA_DLL=1`, isolated in `backend_loader.c` + the `cuda-dll` Makefile target. The host never links cudart; if the DLL is absent the engine falls back to CPU silently. Verified on RTX 5070 Ti (sm_120).

Each commit builds clean on its own (tested by checking out each and running `make glm.exe`). The conflict resolutions from the rebase are in commit 1 (Makefile TEST_BINS, POSIX guards) and commit 3 (backend_cuda.h dllexport, backend_loader.c wrappers for dev's new fused CUDA API).

JustVugg:
Merged to `dev` (`2319b94`) — this is exactly the shape I hoped for, thank you. The 3-way split made it reviewable and the honest #128 breakdown (cache warming, not GPU/kernel; stock main *hangs* on the `\r\n` sentinel) is the kind of framing that keeps the benchmark table trustworthy.

Validation I ran on the rebased head before merging:
- `make check`: **59/59** (no regression on the default path — the gated CUDA-DLL and AVX-VNNI both compile out cleanly on a non-VNNI Linux box)
- oracle `TF=1`: **32/32** teacher-forcing token-exact
- Windows cross-compile (MinGW-w64) clean, and `glm.exe` loads the model + runs inference via WSL interop

Notes for anyone reading later:
- Commit 1 (pipe fix + RAM detection + POSIX guards) is what fixes #123 — stock main genuinely could not run interactive chat on native Windows before this.
- Commit 2 (AVX-VNNI) is bit-identical to AVX2 and compiles out where `__AVXVNNI__` isn't defined; the `_mm256_extracti128_si256` typo fix also unblocks `-march=native` on Alder Lake+.
- Commit 3 (CUDA DLL) is gated behind `CUDA_DLL=1`, host never links cudart, silent CPU fallback — I can't test the Windows-CUDA path here (no NVIDIA + MSVC toolchain), so that part rests on your RTX 5070 Ti verification, which the gating makes safe to carry.

I'll add the #128 row to the README with your exact label — *'with #131 (pipe + RAM fixes), warm cache, no GPU'*. First-rate contribution.

Issue #133, opened 2026-07-13
[Bug]: shipped ref.json does not match OLMoE-1B-7B-0125-Instruct — olmoe validator reports 5/12 out of the box (engine itself is token-exact vs a fresh oracle)


lEWFkRAD:
### Commit or version

97c756a06472b2f632453f35a541b2f0553bec00

### Environment

Windows 11 host / WSL2 Ubuntu, Intel Core Ultra 9 285K (24c, AVX2), 31 GB VM RAM, model on ext4 (NVMe), gcc 15.2.0. Oracle: transformers 5.13.1 + torch 2.13.0 (CPU, bf16).

### Reproduction steps

```sh
make -C c olmoe
# model: allenai/OLMoE-1B-7B-0125-Instruct (bf16 safetensors, 3 shards)
SNAP=/path/olmoe OMP_NUM_THREADS=16 ./c/olmoe 64 8 ref.json
```

### Expected behavior

`Matching tokens: 12/12` against the shipped `ref.json`, per the engine's token-exact acceptance criterion.

### Actual behavior and logs

```
Reference: 7785 15 187 187 510 5347 273 253 1986 2077 310 5041
C engine : 7785 15 187 187 510 3565 3448 273 6181 310 5112 15
Matching tokens: 5/12
```

**The engine is not at fault — the shipped `ref.json` appears to be generated from a different checkpoint.** I regenerated the oracle independently (transformers 5.13.1, bf16 CPU, greedy) on the exact checkpoint above:

- oracle greedy continuation: ` Paris.\n\nThe official language of France is French.` = tokens `[7785, 15, 187, 187, 510, 3565, 3448, 273, 6181, 310, 5112, 15]`
- **oracle vs C engine (int8, cap=64): 12/12 — token-exact.** Also 12/12 at cap=8 (disk-streamed, 18% cache hit), i.e. the streaming path is exact too.
- oracle vs shipped `ref.json`: 5/12.
- Teacher-forced argmax over the shipped `full_ids` on this checkpoint: 9/12, first divergence at position 10 (`' capital'` expected by ref vs `' official'` argmax) — this checkpoint's greedy path cannot produce the shipped continuation.

So the engine passes its own validation, but a fresh clone + the documented checkpoint reports 5/12 out of the box, which reads as an engine failure.

[...]

PR #142, opened 2026-07-13
serve: fix Windows mux-loop busy-spin — winsock select() cannot poll stdin (#139)

### LINKED ISSUE #139: [Bug]: SERVE mux loop busy-spins on native Windows — winsock select() can't poll stdin (runtime gap left after #131; serve path also untestable in-tree)
### Commit or version

6d3ed7e (current main/dev)

### Environment

Windows 11 Pro 26200, MSYS2 UCRT64, gcc 16.1.0, Intel Core Ultra 9 285K.

### Reproduction steps

Static + mechanism-level report; runtime repro is blocked by a secondary gap (below).

```sh
# builds clean since #131, but:
grep -n 'select(STDIN_FILENO' c/glm.c   # -> unguarded at glm.c:3147
```

### Expected behavior

`SERVE=1` mux loop accepts requests on stdin on native Windows, as on POSIX.

### Actual behavior and logs

#131 guarded the `sys/select.h` *include*, but the serve loop still calls POSIX-style `select(STDIN_FILENO+1, ...)` (glm.c:3147). On MinGW this resolves to **winsock** `select()`, which (a) requires `WSAStartup`, and (b) only accepts sockets — fd 0 is not a socket, so it returns `SOCKET_ERROR` (-1) on every iteration. `ready > 0` is never true, `mux_submit` is never called, and the `for(;;)` loop busy-spins forever: serve mode on Windows compiles and prints `READY` but can never receive a request.

**Secondary finding — serve mode is untestable with in-tree assets:** the tiny oracle (`glm_tiny`) has no `tokenizer.json`, and serve mode calls `tok_load` before entering the loop. So the smallest runtime repro of any serve-loop issue currently requires a full multi-hundred-GB model. A tiny tokenizer fixture (or a `TOK=none` byte-fallback for the self-test) would make the serve path smoke-testable in `make check`.

**Proposed fix** (validated at your merge bar on my machine when #137 was open — clean UCRT64 build, `test-c` 6/6, tiny oracle TF 32/32 + greedy 20/20): keep POSIX behavior unchanged; on `_WIN32`, fall back to a blocking `mux_submit` whenever no slot is decoding (sequential serve semantics):

```c
#ifndef _WIN32
        fd_set rfds; FD_ZERO(&rfds); FD_SET(STDIN_FILENO,&rfds);
        struct timeval tv={0,0}, *ptv=active?&tv:NULL;
        int ready=eof?0:select(STDIN_FILENO+1,&rfds,NULL,NULL,ptv);
        if(ready>0 && FD_ISSET(STDIN_FILENO,&rfds)) if(mux_submit(m,&T,ctx,req,nctx,maxctx,eos)<0) eof=1;
#else
        /* winsock select() cannot poll stdin (not a socket): block for input when idle */
        if(!eof && !active) if(mux_submit(m,&T,ctx,req,nctx,maxctx,eos)<0) eof=1;
#endif
```

A proper overlapped-I/O poll (`WaitForSingleObject` on `GetStdHandle(STD_INPUT_HANDLE)`) could restore concurrent submit-while-decoding later; the blocking fallback at least makes serve *functional* sequentially. PR against `dev` incoming.

lEWFkRAD:
## Summary

Fixes #139. On native Windows the SERVE mux loop's `select(STDIN_FILENO+1, ...)` resolves to **winsock** `select()` — which needs `WSAStartup` and only accepts sockets — so polling fd 0 returns `SOCKET_ERROR` every iteration, `mux_submit` is never reached, and serve mode busy-spins forever after printing `READY`. #131 fixed the *compile* (guarded include); this fixes the *runtime*.

Smallest change: keep the POSIX poll byte-identical; on `_WIN32`, block in `mux_submit` whenever no slot is decoding (sequential serve). A `WaitForSingleObject(GetStdHandle(STD_INPUT_HANDLE))` poll can restore submit-while-decoding on Windows later without touching this structure.

## Validation

- [x] `make -C c check` — C suites 6/6 on MSYS2 UCRT64 (gcc 16.1, Core Ultra 9 285K); tiny oracle **TF 32/32 + greedy 20/20** on the patched binary
- [ ] CUDA changes — n/a
- [x] No performance claims

Note: 3 warnings exist on this toolchain **before** this diff (compat.h:187 `#pragma comment` under gcc, glm.c:1219 `-Wformat-truncation`) — pre-existing on `dev`, untouched here, happy to fix separately if wanted. Full runtime serve round-trip on Windows is still blocked by the in-tree fixture gap from #139 (glm_tiny ships no tokenizer.json) — verified instead: build, unit suites, oracle, and that the POSIX branch is preprocessor-identical to current dev.

## Compatibility

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -3142,10 +3142,20 @@ static void run_serve_mux(Model *m, const char *snap){
     int eof=0;
     for(;;){
         int active=0; for(int i=0;i<nctx;i++) active+=req[i].active;
+#ifndef _WIN32
         fd_set rfds; FD_ZERO(&rfds); FD_SET(STDIN_FILENO,&rfds);
         struct timeval tv={0,0}, *ptv=active?&tv:NULL;
         int ready=eof?0:select(STDIN_FILENO+1,&rfds,NULL,NULL,ptv);
         if(ready>0 && FD_ISSET(STDIN_FILENO,&rfds)) if(mux_submit(m,&T,ctx,req,nctx,maxctx,eos)<0) eof=1;
+#else
+        /* Native Windows: MinGW resolves select() to winsock, which needs WSAStartup and
+         * only accepts sockets — polling STDIN_FILENO returns SOCKET_ERROR every time, so
+         * the loop spins and no request is ever read (#139). Fall back to a BLOCKING
+         * submit whenever no slot is decoding: sequential serve, POSIX path untouched.
+         * (A WaitForSingleObject(GetStdHandle(STD_INPUT_HANDLE)) poll can restore
+         * submit-while-decoding later.) */
+        if(!eof && !active) if(mux_submit(m,&T,ctx,req,nctx,maxctx,eos)<0) eof=1;
+#endif
         active=0; for(int i=0;i<nctx;i++) active+=req[i].active;
         if(!active){ if(eof) break; continue; }
         DecodeRow rows[16]; int slots[16], S=0;


JustVugg:
Thank you for the fix attempt, but this is unfortunately a no-op — and it's the same preprocessor blind-spot as #139, so no fault on the intent. Your `#else` branch is added **inside** the outer `#if defined(__APPLE__) || defined(__linux__)` (glm.c:3129) that already wraps the entire mux loop:

- On **Windows**: the outer `#if` is false, so your `#else` (and the whole loop) isn't compiled at all — Windows compiles the warning stub at glm.c:3171 instead. Your Windows blocking-submit never exists.
- On **Apple/Linux**: your inner `#ifndef _WIN32` takes the POSIX `select()` branch, so your `#else` is skipped there too.

So the branch is unreachable on every platform. `make check` 6/6 + oracle 32/32 pass because (a) on Windows the new code doesn't compile, and (b) neither `make check` nor the oracle exercises the serve loop — which is exactly the untestable-serve gap you flagged in #139. There's no busy-spin in current `main` to fix (the mux is already guarded), so I'm going to close this without merging.

If you want to *actually* bring SERVE_BATCH to native Windows (a real, welcome feature), the move is to replace the stub at glm.c:3172-3178 with a real `WaitForSingleObject(GetStdHandle(STD_INPUT_HANDLE))` poll — and land the tiny-tokenizer serve fixture from #139 first, so it can be validated end-to-end. Happy to help scope that. Thanks for the sharp Windows auditing — #133/#134/#137 were real finds.

Issue #140, opened 2026-07-13
[Feature]: CI — run make check on ubuntu/windows/macos for PRs (today's #137 would have been caught pre-merge; repo currently has no workflows)


lEWFkRAD:
Today's cycle (#137: PR #68 unknowingly broke the whole Windows build; caught same-day only because someone happened to build Windows at HEAD) suggests the project has outgrown manual pre-merge verification. Right now `.github/` has no workflows at all — every "clean build, 0 warnings, oracle 32/32 + greedy 20/20" gate in CONTRIBUTING runs on the maintainer's machines, by hand, for every PR on every platform.

`make check` is already designed for CI: dependency-free, no model download, no CUDA. Proposal — a single workflow running it on the three claimed platforms:

- **ubuntu-latest** — `make -C c check` (gcc, the reference path)
- **windows-latest** — MSYS2/UCRT64 via `msys2/setup-msys2` (gcc + make + python), `make -C c check` — this is the job that would have flagged #68/#137 pre-merge
- **macos-latest** — clang; builds single-threaded without libomp per the Makefile's existing fallback (or `brew install libomp` in the job for the threaded path)

Triggers: `pull_request` + `push` to `main`/`dev`. Nothing gates on it until you're comfortable — informational at first, required later if it earns its keep.

Deliberately out of scope for v1: CUDA (needs self-hosted GPU), Metal (runtime shader compile is covered by the plain macOS build), oracle runs vs real models (model downloads).

I have this written and validated against the exact MSYS2/UCRT64 toolchain from #137. PR against `dev` incoming — happy to adjust triggers/matrix to taste.

JustVugg:
**Update — accepted, this is going on the roadmap.** Cost is confirmed a non-issue: on a public repo, GitHub-hosted standard runners are unlimited-free and don't even count against metered/included usage, so there's no billing reason not to.

Planned shape: a single `.github/workflows/check.yml` running `make -C c check` (the dependency-free gate — build + C unit tests + Python tests; no model, no CUDA) across **ubuntu-latest, macos-latest, and windows-latest (MSYS2/UCRT64)**, on every `pull_request` and push to `main`/`dev`. That's the exact matrix that would have caught #123/#137 pre-merge.

[...]

Issue #137, opened 2026-07-13
[Bug]: Windows (MinGW-w64) build broken at HEAD by PR #68 — unguarded sys/select.h + mmap/madvise in COLI_MMAP path (patch included, TF 32/32 + greedy 20/20 on Windows)


lEWFkRAD:
### Commit or version

97c756a06472b2f632453f35a541b2f0553bec00 (HEAD)

### Environment

Windows 11 Pro 26200, Intel Core Ultra 9 285K (24c, AVX2), 63.5 GB RAM, NVMe; MSYS2 UCRT64, gcc 16.1.0, `make glm ARCH=native`.

### Reproduction steps

```sh
# MSYS2 UCRT64 shell, HEAD
make -C c glm ARCH=native
```

### Expected behavior

Windows build succeeds (native MinGW-w64 port, PR #40 / README "Windows 11" support).

### Actual behavior and logs

```
glm.c:30:10: fatal error: sys/select.h: No such file or directory
```

and after clearing that, from the same commit:

```
glm.c:1084:23: error: implicit declaration of function 'fstat'
glm.c:1086:17: error: implicit declaration of function 'mmap'
glm.c:1086:31: error: 'PROT_READ' undeclared
glm.c:1153:17: error: implicit declaration of function 'madvise'
glm.c:1153:68: error: 'MADV_WILLNEED' undeclared
```

**Cause:** cbd5990 (PR #68, merged today) introduced unguarded POSIX in `glm.c`:
1. `#include <sys/select.h>` + a `select()` stdin poll in the new mux serve loop;
2. the `COLI_MMAP` path (`map_of_fd`, `mmap`/`fstat`/`madvise` in `expert_load`) outside the existing `__APPLE__||__linux__` guards that the includes at the top of the file already use.

The Windows port (PR #40) predates it; every commit since cbd5990 fails to build on MinGW-w64.

**Patch (validated):** minimal `#ifndef _WIN32` guards — mmap path compiled out on Windows with `COLI_MMAP=1` demoted to a warning + pread slab fallback; serve loop falls back to blocking stdin submit when no slot is decoding (sequential serve; POSIX behavior unchanged). With this patch on Windows: clean build, `make test-c` all green, tiny oracle **TF 32/32 + greedy 20/20**.

```diff
diff --git a/c/glm.c b/c/glm.c
index 1f1e603..5caf331 100644
--- a/c/glm.c
+++ b/c/glm.c
@@ -27,7 +27,9 @@
[...]

Issue #141, opened 2026-07-13
[Bug]: FAIL: test_non_executable_engine_and_excessive_ram_budget_fail


galmok:
### Commit or version

6d3ed7e62b1b4c05d8e656a5263e91b983aa26ba

### Environment

Windows 11, build following the Windows guide.
9850X3D, 9070XT (I know it isn't supported yet), used "scoop install mingw-winlibs" to install (just now).
I added this to the path as the makefile needed uname and I found that git had one stashed away somewhere:
$env:Path += ';C:\Program Files\Git\usr\bin'

I am using TheRock ROCm 7.13 and not HIP 7.11. This could be the cause, but I don't know how to test it without changing my setup significantly. I don't know if colibri even uses this.

### Reproduction steps

I ran these commands (had to install scoop first):
scoop install mingw-winlibs
make glm.exe
make olmoe.exe
make iobench.exe
make test-c
make test-python

The last command failed with:

AIL: test_non_executable_engine_and_excessive_ram_budget_fail (test_doctor.DoctorTest.test_non_executable_engine_and_excessive_ram_budget_fail)

python --version
Python 3.14.3

### Expected behavior

I expected the test to succeed.

### Actual behavior and logs

```shell
Full make python-test:
make test-python
python3 -m unittest discover -s tests -p 'test_*.py'
..........F........[api] 127.0.0.1 - "OPTIONS /v1/chat/completions HTTP/1.1" 204 -
.[api] 127.0.0.1 - "POST /v1/chat/completions HTTP/1.1" 200 -
.[api] 127.0.0.1 - "GET /health HTTP/1.1" 200 -
.[api] 127.0.0.1 - "POST /v1/completions HTTP/1.1" 200 -
.[api] 127.0.0.1 - "GET /v1/models HTTP/1.1" 200 -
[api] 127.0.0.1 - "GET /v1/models HTTP/1.1" 401 -
C:\Users\USERNAME\AppData\Local\Python\pythoncore-3.14-64\Lib\tempfile.py:484: ResourceWarning: Implicitly cleaning up <HTTPError 401: 'Unauthorized'>
  _warnings.warn(self.warn_message, ResourceWarning)
.[api] 127.0.0.1 - "POST /v1/chat/completions HTTP/1.1" 400 -
[...]

PR #143, opened 2026-07-13
ci: make check on ubuntu / windows (MSYS2 UCRT64) / macos — and it already caught a Windows test bug (#140)

lEWFkRAD:
## Summary

Implements #140: a `check` workflow running the repo's own dependency-free gate (`make -C c check`) on **ubuntu-latest**, **windows-latest** (MSYS2/UCRT64 — the toolchain from #137, the job that would have caught #68 pre-merge), and **macos-latest** (clang + libomp), on every PR and push to `main`/`dev`. No CUDA, no Metal, no model downloads — v1 is exactly the gate CONTRIBUTING already requires, automated.

It caught its first bug before it was even filed as a PR: on native Windows, `make check` was **58/59** — `test_doctor.test_non_executable_engine_and_excessive_ram_budget_fail` asserts that `chmod(0o644)` makes the engine non-executable, but Windows has no executable bit, so `os.access(X_OK)` legitimately passes. Second commit guards that single POSIX-semantics assertion (the RAM-overbudget half still runs everywhere). With it, `make check` is fully green on all three platforms I can test (ubuntu via WSL2 gcc 15.2, native Windows MSYS2/UCRT64 gcc 16.1; macos job is standard clang + libomp per the Makefile's existing Darwin branch).

## Validation

- [x] `make -C c check` — green on native Windows (MSYS2 UCRT64) and Linux with these changes; 59 python tests OK, C suites 6/6
- [ ] CUDA changes — n/a
- [x] No performance claims

## Compatibility

- [x] The default CPU build remains dependency-free (workflow-only + one test guard)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,49 @@
+# CI: run the repo's own dependency-free gate (`make check` = clean + portable
+# CPU build + C unit suites + Python stdlib tests) on the three claimed
+# platforms. No model downloads, no CUDA, no external deps — by design (#140).
+name: check
+
+on:
+  push:
+    branches: [main, dev]
+  pull_request:
+    branches: [main, dev]
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: make check
+        run: make -C c check
+
+  windows:
+    # The job that would have caught #68/#137 pre-merge: native MinGW-w64
+    # (MSYS2/UCRT64), the exact toolchain the README's Windows port targets.
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: UCRT64
+          update: false
+          install: >-
+            make
+            mingw-w64-ucrt-x86_64-gcc
+            mingw-w64-ucrt-x86_64-python
+      - name: make check
+        run: make -C c check
+
+  macos:
+    # clang; libomp for the threaded path (Makefile falls back to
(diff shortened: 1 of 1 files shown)


lEWFkRAD:
Re the acceptance plan in #140: this PR is exactly that shape, and its second commit is the prerequisite you named — it fixes the `os.access(X_OK)` Windows test gap now tracked as **#141** (galmok hit the same 58/59 independently; the guard keeps the RAM-overbudget half of the test running everywhere). So merging this PR delivers the workflow *and* makes the windows job start green in one go. Happy to split the test fix into its own PR against #141 if you'd rather land them separately.

rajpratham1 (approved):
This looks good to me.

Adding a cross-platform `make check` workflow is a valuable improvement for the project, and using the same command across Linux, Windows (MSYS2/UCRT64), and macOS keeps the CI simple and consistent.

The accompanying test update also correctly handles the platform difference in executable permissions on Windows, making the test portable instead of assuming POSIX semantics.

Overall, this should improve CI reliability and help catch platform-specific regressions earlier. Nice work!

lEWFkRAD:
Rebase-free update now that the tree's moved: **#178 already landed the doctor test fix (#141)** I'd flagged as this PR's prerequisite, so my offer to split that out is moot — thanks to whoever grabbed it. That makes #143 simpler: it's now purely `.github/workflows/check.yml`, no code overlap with #178, and the Windows `make check` job should come up green off #178 rather than needing anything from here.

`dev` still has no `.github/workflows`, so this is still the whole CI gap. Clean/mergeable as-is. Happy to add `concurrency: cancel-in-progress` or pin runner image tags before merge if you'd prefer — otherwise it's ready.

(Standing offer on the RTX PRO 6000 as a Windows-CUDA self-hosted runner still holds.)

PR #144, opened 2026-07-13
ci: GitHub Actions — engine, web, Python, CUDA syntax (free for public repos)

ZacharyZcR:
Four lightweight jobs, triggered on push/PR to dev and main:

| Job | What it catches | Time |
|---|---|---|
| **engine** | `make glm` build + `make test-c` (7 tests) | ~30s |
| **cuda-syntax** | `nvcc -c backend_cuda.cu` (no GPU needed) | ~40s |
| **web** | `npm run build` + 17 vitest | ~25s |
| **python** | `unittest discover test_*.py` | ~5s |

All free on public repos (GitHub-hosted runners). The whole matrix finishes in under 2 minutes.

Born from recent pain: a textually clean 3-way merge silently regressed prefill 4× (OMP tuning vs CUDA), an executable bit was lost in transit, and a CORS mismatch hid behind a build that looked fine. This pipeline would have caught all three at the PR stage.

Intentionally minimal — no GPU runner (those aren't free), no model download, no integration tests. The bar is 'does it compile, do the unit tests pass, does the web UI build.' Everything else stays on the lab machine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,66 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [dev, main]
+  push:
+    branches: [dev, main]
+
+jobs:
+  engine:
+    name: Engine (Linux, CPU)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build glm
+        run: cd c && make glm
+      - name: C test suite
+        run: cd c && make test-c
+
+  engine-cuda-syntax:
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This looks good to me.

The workflow provides a solid baseline CI pipeline by validating the C engine, web application, Python test suite, and CUDA source compilation. The separation into independent jobs also makes failures easier to diagnose and should improve contributor confidence before merging changes.

I don't see any issues with this implementation. Nice addition to the project's automation.

Issue #146, opened 2026-07-13
Write-up: how the grammar-forced draft source (#48 → PR #70) lifts constrained-output decode — mechanism, A/B numbers, honest prior art, and why streaming-MoE is the regime where it pays most


fabio-rovai:
PR #70 merged the `GRAMMAR=` draft source and the README now documents the knob, but the mechanism and the *reason it pays disproportionately in this engine* are spread across #48, the PR thread, and commit messages. Since other threads are starting to reference the feature (#127 lists it among the engine's signature capabilities), here is the consolidated write-up: what it does, what we measured, where it does and doesn't help, and the prior art it builds on.

## The trick in one paragraph

On constrained-output workloads (JSON/NDJSON, function calling, structured extraction) the grammar itself is a draft source. A byte-level GBNF subset is compiled into a set-of-stacks PDA; wherever the grammar admits **exactly one** legal next byte (braces, quotes, key names, enum bodies, fixed separators), the engine walks that forced span, tokenizes it, and injects it as pre-accepted draft tokens into the **same batch-union verify forward** used by MTP and n-gram drafts. It never constrains sampling: forced spans are *verified by the target model*, so a wrong, stale, or desynced grammar cannot change the output — worst case is rejected drafts, and an adaptive guard disables the source below 50% acceptance. It composes with `DRAFT`/MTP, which fill the free-text gaps between forced spans.

## Measured A/B (M3 Max, 128 GB, macOS)

Container `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp`, greedy (`TEMP=0`), `MTP=0 DRAFT=0` (isolating the grammar source), NDJSON scoring workload (our production tender-scoring prompt, fixed schema):

| config | tok/s | tok/forward | forced-span acceptance | output |
|---|---|---|---|---|
| baseline | 0.37 | 1.00 | — | reference |
| `GRAMMAR=fit.gbnf` | **0.50** (+35%) | **1.60** | **48/48 (100%)** | **byte-identical** |

The tokenization-boundary concern (forced byte spans not landing on token boundaries) did not bite on this workload. A/B evidence (outputs, grammar, prompt) is preserved and can be attached on request.

## Why a disk-streaming MoE engine is the best-case regime for this

In this engine the marginal cost of a decode step is **unique expert bytes off disk**, not FLOPs. The batch-union forward loads each unique expert once for *all* positions in the verify batch. Two consequences:

[...]

PR #148, opened 2026-07-13
SCHEMA=<file.json>: JSON-Schema → GBNF compiler for grammar-forced drafts (follow-up to #70, seam discussed in #146)

fabio-rovai:
Follow-up to #70, and the first half of the `response_format` seam discussed in #146: a **JSON-Schema → GBNF compiler** (`c/schema_gbnf.h`, header-only, reuses the in-tree `json.h`), wired in as `SCHEMA=<file.json>` next to `GRAMMAR=<file.gbnf>`.

## What

```bash
SCHEMA=score_schema.json PROMPT="..." ./glm 64 4 8   # instead of hand-writing GBNF
```

Supported subset (fail-closed — anything else logs one line and runs **without** a grammar, output unchanged):
- `object` + `properties` (+ `required` — must list every property, OpenAI structured-output *strict* semantics; a proper subset falls back)
- `string`, `number`, `integer`, `boolean`, `null`, `enum` (strings/numbers/bools), `const`
- `array` + `items` (+ `minItems` 0/1), nesting to depth 32
- annotations (`title`, `description`, `$schema`, `default`, `additionalProperties`) ignored

The emitted grammar is compact (no inter-token whitespace, declared key order) because strictness is free here: **the grammar is a draft source, not a constraint** (#70's design) — a schema stricter than the model's actual output only loses draft acceptance, it cannot change the output. That's also why a partial-subset compiler is safe to ship where a constrained-decoding engine would need completeness.

## Why

- On the NDJSON scoring workload in #70/#146, forced spans (keys, separators, enum bodies) were 100%-accepted drafts; every structured workload has a schema lying around, almost none have GBNF.
- Second half (proposed in #146, needs your call on the wire protocol): accept OpenAI `response_format: {"type":"json_schema"}` in `openai_server.py` (today it 400s at `generation_options`), compile server-side with this same subset, and thread the grammar per-request through `SUBMIT` — plus per-slot `GrState` so grammar drafting can engage in `run_serve_mux` (today the mux path runs with speculation disabled entirely). Happy to do that as PR 2 if you want it — protocol shape (new SUBMIT field vs side-frame) is your design call.

## Tests

`tests/test_schema_gbnf.c` (in `TEST_BINS`): 8 groups compile schemas and walk the result end-to-end through the `grammar.h` PDA — forced-span extraction (`{"score":` from a two-key strict object), enum disambiguation (one byte picks the branch, remainder is forced again), nested object/array instances byte-walked to `can_end`, escaped keys/values, JSON-number edge cases (leading-zero rejection, `-1e-3`), and fail-closed fallbacks (`oneOf`, `pattern`, subset-`required`).

`make test-c` green on Apple Silicon (M3 Max). One note from that run: `tests/test_decode_batch` and `tests/test_i4_acc512` are **committed x86-64 Linux ELF binaries** in the repo — on any non-x86 host `make test` tries to execute them and dies until they're deleted (make sees them as up-to-date). Not touched in this PR to keep it focused, but you probably want `git rm` + a `.gitignore` entry (related to the CI discussions in #140/#143).


fabio-rovai:
Pushed the whitespace-tolerance commit promised in #146: the compiler now emits `jws` at separators. Rationale + measurement in the commit message — short version: on current main's sloppy-JSON continuations a compact grammar forfeits most spans by desyncing at the first stray space; `jws` positions aren't forced themselves but keep the walker alive so every span after them still drafts. Verified drafts make this strictly acceptance-positive. Tests updated (sloppy-instance walk included).

JustVugg:

fabio-rovai:
Rebased onto current main (37ffb61) — only conflict was the `TEST_BINS` line against the new `test_compat_direct`, merged additively. `make glm` + full `make test-c` green on Apple Silicon post-rebase (nice touch, the new `tools/run_tests.py`). Ready to merge. One drive-by note, not from this PR: current main emits one Darwin-only warning — `glm.c:3540 unused variable 'c'` in `hwinfo_emit` (the non-/proc path doesn't use the Cfg) — trivially silenceable whenever someone's in there.

JustVugg:
Merged (with the same additive TEST_BINS resolution you describe — we raced, sorry for the overlap). Thanks for the Apple Silicon validation of `tools/run_tests.py` post-rebase. The Darwin `unused variable 'c'` in `hwinfo_emit` is confirmed on Linux too — it'll be silenced in the next cleanup commit on dev.

PR #152, opened 2026-07-13
prefill: batch the attention input projections (bit-identical, -4.5% prefill)

dnnspaul:
The three attention input projections (`q_a`, `q_b`, `kv_a`) are called **one row at a time** during prefill:

```c
for(int s=0;s<S;s++){
    matmul_qt(qresid, xs,     &l->q_a, 1);   // rows = 1
    matmul_qt(qfull,  qresid, &l->q_b, 1);   // rows = 1
    matmul_qt(comp,   xs,     &l->kv_a, 1);  // rows = 1
}
```

…while `o_proj` twenty lines below already runs batched (`matmul_qt(out, ctx, &l->o, S)`) and `moe()` already does its batch-union. This PR batches them over all S rows, so each weight row is read once for the whole prompt instead of once per token.

### Result (upstream/dev @ 6d3ed7e, Zen5 + RTX 5090, GLM-5.2 int4, 78 layers, DSA on, OMP tuning on)

Prefill run to completion; numbers are the engine's own `PROFILE` output.

| prompt | projection/RoPE | total prefill |
|--------|-----------------|---------------|
| 256    | 22.7s → **17.7s** (−22%) | 120.9s → **115.4s** (−4.5%) |
| 512    | 45.5s → **36.3s** (−20%) | 231.4s → **222.7s** (−3.8%) |

Every other profile term is unchanged (expert-disk, expert-matmul, score-softmax-value, o_proj all within noise) — only the targeted term moves.

### Output is bit-identical

Summed log-likelihood of a fixed 1023-token passage, `SCORE` mode, CPU (deterministic):

```
dev  : -4987.458316
+PR  : -4987.458316
```

Decode (S=1) is untouched: it makes the same three calls with the same kernel.

### Why `matmul_qt_ex(..., allow_idot=0)`

This is the part worth reviewing. `matmul_qt` gates the int4 IDOT kernel behind `S >= g_i4s` (2 on x86), so batching **alone** would silently pull these three projections onto IDOT's int8-quantized activations. I measured that, and it is not free:

| projections on… | 256-tok prefill | log-lik (README) | log-lik (prose) |
|---|---|---|---|
| exact kernel, per-token (dev today) | 120.9s | −4987.46 | −3452.74 |
| exact kernel, batched (**this PR**) | 115.4s | −4987.46 | — |
| IDOT kernel (`I4S=1`) | 113.8s | −5160.47 | −3538.43 |

IDOT costs **+0.169 nats/token** on markdown/code and **+0.084** on prose — perplexity **+18.4%** and **+8.7%** — to buy **1.4%** more speed over this PR. Since #95 gave the exact int4 path an AVX-512 accumulator (faster *and* more accurate), the IDOT gate is no longer a good trade for these tensors on x86. So the PR pins them to the exact kernel and takes the batching win only.

I opened an issue with the full profile and a note about `g_i4s=1` on ARM/POWER, where these projections already run on IDOT by default.

### Also in here

`matmul_qt`'s comment documents the threshold as *"configurable via `I4S`"*, but `getenv("I4S")` was never called — the knob did nothing. Wired up; it is what let me separate the kernel switch from the batching win.

### Testing

`make test-c` and `make test-python` pass (49 python tests). Built with `CUDA=1`, no new warnings.


dnnspaul:
Heads-up before you review: **the quality rationale in this PR's description is weaker than I wrote it.**

The perplexity numbers I cited (IDOT costing +18.4% / +8.7%) were measured **without GLM's `[gMASK]<sop>` prefix** — the out-of-distribution scoring bug @bokiko found in #108. Redone in-distribution, the penalty roughly halves and **flips sign on one of two corpora** (+9.0% on markdown/code, −0.7% on prose). Full retraction and numbers: #153 (comment).

**The PR itself is unaffected.** It is bit-identical to `dev` by construction — summed log-lik `-4987.458316` before and after, and that equality holds with the prefix too, since no scoring protocol can change a bit-identical result. The −4.5% prefill is wall-clock.

What changes is only the *argument* for `matmul_qt_ex(..., allow_idot=0)`. The version that survives:

> On current `dev`, letting the projections ride the IDOT gate buys **1.4%** (113.8s vs 115.4s at 256 tok) because #95 gave the exact int4 path an AVX-512 accumulator. Pinning them to the exact kernel therefore costs almost no speed, and avoids a numerics change whose quality effect I can no longer characterise cleanly.

That's a conservative-default argument, not the "IDOT is expensive" argument I originally made. If you'd rather take the 1.4% and let batching pull them onto IDOT, say so and I'll drop the `allow_idot` parameter — it's a one-line change and the batching win is independent of it.

fabio-rovai:
Here's the ARM datapoint for the open question in your last section ("a note about `g_i4s=1` on ARM/POWER, where these projections already run on IDOT by default").

**M3 Max (NEON/SDOT), current main (6d3ed7e), same method as yours** — summed log-likelihood, `SCORE` mode, greedy, deterministic CPU, GLM-5.2 int4 (`mateogrgic` container), one fixed 1023-token passage (this repo's README, so absolute values aren't comparable to your passage — the per-token delta is the number):

| config | summed log-lik | Δ nats/token |
|---|---|---|
| `IDOT=0` (exact f32) | −5089.81 | — |
| `IDOT=1` (ARM default) | −5263.14 | **+0.169** |

**+0.169 nats/token — numerically almost identical to your x86 markdown figure (+0.169), which is what you'd expect: the error source is the int8 *activation* quantization (`qrow_i8`), and that's architecture-independent.** Perplexity +18.5% on this passage.

One scope difference to be explicit about: your `I4S=1` row isolates the three attention projections; without your `matmul_qt_ex(allow_idot=...)` plumbing I can only toggle IDOT globally, and on ARM `g_i4s=1` means the *entire* engine (decode included) runs on IDOT by default. So my delta is the full-engine cost, which upper-bounds the projections' share — but it also means the ARM default ships this tax on every token, not just prefill. The ARM speed case for IDOT at S=1 was measured at ~+14% end-to-end (that's why `g_i4s=1` exists); +18% perplexity for +14% speed looks like a bad default trade in the same way your table shows for x86, and your per-tensor `allow_idot` mechanism is exactly the right knob to fix it — pinning the attention projections (and arguably `lm_head`) to the exact kernel on ARM too would be a natural follow-up to this PR, and I'm happy to run the isolated A/B on this box once your plumbing lands.

fabio-rovai:
Correction to my ARM datapoint above, prompted by your #153 retraction — **our measurement had the same `[gMASK]<sop>` confound** (raw README tokens, no prefix). Redone in-distribution (prefix prepended, `ctxlen` set so the same 1023 tokens are scored — directly comparable):

| ARM full-engine, M3 Max | summed log-lik | Δ nats/token |
|---|---|---|
| `IDOT=0`, in-distribution | −3253.52 | — |
| `IDOT=1`, in-distribution | −3294.23 | **+0.0398 (+4.1% ppl)** |
| (both, out-of-distribution — retracted) | −5089.81 / −5263.14 | ~~+0.169~~ |

So the confound was ~4× the true effect, same direction as your finding. **And the conclusion flips accordingly:** at +4.1% perplexity for ~+14% end-to-end speed, ARM's `g_i4s=1` default is a *defensible* trade, not the bad one I claimed — I retract that sentence. What survives on the ARM side matches your revised conservative-default argument: the cost is real but modest, it's full-engine here (upper-bounding the projections' share, which is likely well under 4%), and per-tensor `allow_idot` control remains the right mechanism for anyone who wants the exact path where it's nearly free. Offer stands to run the isolated projections A/B on this box once your plumbing lands.

Methodology footnote for anyone reproducing on this container: the in-distribution absolute log-lik is ~35% better than un-prefixed (−3253 vs −5090 for the identical scored tokens) — the prefix matters enormously on GLM-5.2, #108/#153 are right that no quality number should be quoted without it.

dnnspaul:
@fabio-rovai This is the datapoint the ARM question needed, and you caught the prefix confound in your own data before I could even flag it. Two independent boxes, same confound, same direction — that's about as good as this gets.

**Our numbers reconcile cleanly.** Mine (x86, isolated to `q_a`/`q_b`/`kv_a`, 5 corpora, in-distribution): **+0.019 nats/token, +1.9% ppl, 95% CI −0.012…+0.050 — contains zero.** Yours (ARM, full-engine): **+0.040 nats/token, +4.1% ppl.** Full-engine necessarily upper-bounds the projections' share, so 4.1% ⊇ 1.9% is exactly the relationship you'd expect if the projections carry only part of it. Nothing to resolve between us.

And I agree with your revised conclusion: **+4.1% ppl for +14% end-to-end is a defensible default**, not a bad one. I've told @JustVugg the same on #153 — no ARM/POWER default change is warranted on the evidence either of us has actually produced.

## One correction, and it opens a better experiment

> pinning the attention projections (and arguably `lm_head`) to the exact kernel on ARM too

`lm_head` is a different animal, and I only found this by building the probe:

```
[PROBE] lm_head fmt=1        <- int8, not int4
```

The gate is `(w->fmt==1 || (w->fmt==2 && S>=g_i4s))`, so **`fmt==1` short-circuits the threshold entirely: `lm_head` already runs on IDOT everywhere — on x86 too, at every batch size, today.** `g_i4s` never touches it. So it's *not* part of the ARM/x86 difference at all (my own lm_head hypothesis on #153 died on exactly this — I'd tried to force IDOT *on* and got a bit-identical no-op, which is what tipped me off).

Which means the interesting direction is the one neither of us has tried: **pinning `lm_head` OFF IDOT.** `matmul_qt_ex(..., allow_idot=0)` on `fmt==1` falls through to `matmul_q` — exact int8 weights, **f32 activations**. That's a knob that didn't exist before this PR, it would apply to **every platform**, and it lands on the logits directly, which is the one place a per-row int8 activation scale has nowhere to hide. It's also one matmul per token, so the speed cost should be negligible.

I'm running that now: `lm_head` exact vs IDOT, 5 corpora, in-distribution, isolated. If it's real, it's a free quality win for everyone rather than an ARM-only trade, and it's a one-line change on top of this PR.

Your offer to run the isolated projections A/B on the M3 Max once this lands — yes please, and if you're up for it, the lm_head arm too. Between your box and mine we'd have both architectures on both tensors, which is more than either of us can claim alone.

dnnspaul:
## lm_head result: pinning it off IDOT does nothing. Hypothesis dead.

@fabio-rovai @JustVugg — I said I'd measure rather than assert, so here it is, and it kills my own idea.

`LMHEAD_EXACT=1` routes all 7 `lm_head` call sites through `matmul_qt_ex(..., allow_idot=0)`, so `fmt==1` falls through to `matmul_q` (exact int8 weights, **f32 activations**) instead of IDOT. 5 corpora, in-distribution, CPU, deterministic. `LMHEAD_EXACT=0` reproduces pristine `dev` bit-for-bit on all five (self-check).

| corpus | IDOT (today) | exact | Δ/token | ΔPPL |
|---|---|---|---|---|
| prose | −2295.2 | −2295.8 | +0.00058 | +0.06% |
| factual | −1845.1 | −1845.3 | +0.00016 | +0.02% |
| academic | −1969.8 | −1970.2 | +0.00039 | +0.04% |
| code | −1415.7 | −1414.9 | −0.00078 | −0.08% |
| markdown | −3272.1 | −3270.3 | −0.00177 | −0.18% |

**mean −0.0003 nats/token (−0.03% ppl).** Two corpora marginally better, three marginally worse, every delta under 2 nats across 1023 tokens. That is nothing — three orders of magnitude below the effect I originally (wrongly) claimed. Decode speed: 0.31 vs 0.30 tok/s, within noise.

So the intuition that lm_head is "small and sensitive, sitting straight on the logits, so int8 activations must hurt there" is **wrong, measured.** It's the fourth plausible story I've had killed by an experiment in this thread, and I'd rather report the null than quietly drop it.

**Practical upshot: there is nothing to fix at lm_head, on any platform.** Don't add a carve-out for it.

### Where that leaves the ARM number

@fabio-rovai's in-distribution full-engine cost is **+0.040 nats/token (+4.1% ppl)**. We can now account for it:

| tensor | in-distribution cost | status |
|---|---|---|
| attention projections (`q_a`/`q_b`/`kv_a`) | +0.019 nats/tok (+1.9%), **CI contains zero** | measured, x86, isolated |
| `lm_head` | **−0.0003 nats/tok (≈0)** | measured, x86, isolated |
| everything else (experts, DSA indexer, o_proj at S=1) | **the remainder** | not isolated |

So the ARM cost is **not** concentrated in the two tensors we suspected. Whatever is left lives in the expert FFNs and/or the DSA indexer projections at S=1 — which is a much less surprising place for activation quantization to show up (far more of the model by parameter count), and it makes `g_i4s=1` look like an even more reasonable default than it did: the +14% speed is buying a cost that's spread thin across the bulk of the network rather than concentrated in a protectable, sensitive tensor.

@fabio-rovai — if you still want to run something on the M3 Max once this lands, the isolated **projections** arm (`PROJ_IDOT`) is now the only one worth the machine time; lm_head is settled. Though honestly, given the CI on the projections already contains zero on x86 and your full-engine number is modest, I'd say the ARM default question is **closed** and neither of us should spend more on it.

dnnspaul:
Pushed a second commit to this PR (`162030e`) — same fix, one block down.

**The DSA indexer has the identical per-token bug.** `ix_wk` is called one row at a time inside the prefill loop, so its weights get re-read for every token on every DSA layer:

```c
for(int s=0;s<S;s++){
    matmul_qt(kd, xs, &m->ix_wk[layer], 1);   // rows = 1
    layernorm(...); rope_interleave(...);
}
```

Batched over all S rows, pinned to the exact kernel with `matmul_qt_ex(..., 0)` for the same reason as the attention projections (batching alone would push S past the `S>=g_i4s` gate and silently swap the activation quantization).

**Bit-identical**, in-distribution, 1023 tokens: prose `−2295.245383` and markdown `−3272.146403` — unchanged to the last digit from the first commit.

**Speed** (256-token prompt, **3 alternating reps each**, baseline = this PR's first commit):

| | base | +indexer | Δ |
|---|---|---|---|
| projection/RoPE | 17.54s (sd 0.20) | **16.96s** (sd 0.15) | **−3.3%** |
| total prefill | 115.34s (sd 0.82) | **113.69s** (sd 0.35) | **−1.4%** |

Both separated beyond 2 sd. I ran the reps because the single-shot number (−1.8%) sat inside the run-to-run spread I'd seen earlier (~3% on a cold cache), and I wasn't willing to quote it until it was distinguishable from noise. Warm, the spread is well under 1%.

It's a small win — but it's free and provably bit-identical, so I'd rather it ride along than sit in a follow-up.

**`ix_wq`/`ix_wp` I deliberately left alone.** They sit inside an `omp parallel for` and are skipped entirely below `index_topk` context, so batching them is a different change with a different risk profile (nested OpenMP, conditional per-token execution). Happy to look at it separately if you think it's worth it — on an 11k-token prompt they *do* run, so there may be something there for exactly the long-prompt case #153 is about.

Squash or split as you prefer; the two commits are independent.

JustVugg:

dnnspaul:
## ⚠️ My mistake: the merged commit contains debug scaffolding I never described. Please hold before promoting to `main`.

You merged before I pushed the rebase, and the branch you merged carries **leftover probe code from my own measurements** — I only found it when the rebase conflicted on it. This is entirely my fault: `git checkout -B` carries uncommitted working-tree edits forward, and my `LMHEAD_EXACT` probe (from the lm_head study I posted above) rode into the indexer commit without me noticing.

**What's in `dev` now that shouldn't be** (9 references):

```c
static int g_lmhead_exact=0;   /* PROBE: LMHEAD_EXACT=1 -> pin lm_head to the exact int8 kernel */
...
g_lmhead_exact = getenv("LMHEAD_EXACT")?atoi(getenv("LMHEAD_EXACT")):0;
...
matmul_qt_ex(lo, row, &m->lm_head, 1, !g_lmhead_exact);   /* x7 call sites */
```

I see you annotated one of them `/* exact-mode knob (#152) */` — **it isn't a knob, it's scaffolding.** It was never in the PR description, and I'd argue it should not exist:

1. **It does nothing.** I measured it (comment above): pinning lm_head off IDOT is **−0.03% perplexity** across 5 corpora — three orders of magnitude below anything actionable. There's no reason for the knob to exist.
2. **It's a silent footgun.** `LMHEAD_EXACT=1` would quietly change lm_head numerics for anyone who set it, with no documentation and no test coverage.
3. It's dead weight in a file whose whole virtue is that it's readable.

**Good news: it is a functional no-op as merged.** The flag defaults to 0, so `!g_lmhead_exact == 1`, which is exactly `matmul_qt`'s own `allow_idot`. Nothing in `dev` behaves differently, which is presumably why your 60/60 passed — it would. And **`main` is clean** (the probe is only on `dev`), so nothing has escaped yet.

**Fix incoming.** I've stripped all 9 references, rebuilt, and I'm running the bit-identical check right now (dev-with-probe vs dev-minus-probe must produce the same log-likelihood to the last digit — it should, since it's a no-op, but I'm not asserting that either). PR up as soon as it's green.

The two real changes are untouched and remain exactly what the PR claims: the three attention projections and `ix_wk`, batched, pinned to the exact kernel via `matmul_qt_ex(..., 0)`, bit-identical output.

Sorry for the churn — you did the right thing merging a PR that said what it did; the PR just didn't say everything it did. I should have diffed the branch against `dev` before pushing instead of trusting my local state, and I'll do that from now on.

JustVugg:
Hold honored — **no promotion to `main` until your cleanup PR lands.** Thanks for catching it fast and for the precise inventory (9 refs, functional no-op since the flag defaults 0 — which indeed explains my 60/60 passing). For the record on my side: when I resolved the merge conflict I kept `matmul_qt_ex(lo, row, &m->lm_head, 1, !g_lmhead_exact)` believing it was part of the PR's design, so the probe riding along is on both our diligence, not just yours. Your split is exactly right: the batched projections pinned via `matmul_qt_ex(...,0)` stay (that's the PR), the `LMHEAD_EXACT` scaffolding goes. Send the strip PR and I'll merge it straight away — dev is frozen for promotion until then.

JustVugg:
Cleanup #197 merged and the whole batch promoted to `main` — validated post-strip: make check 60/60, oracle TF 32/32 + greedy 20/20, MinGW cross-compile clean, and zero `g_lmhead_exact` references left. The legit batched projections (`matmul_qt_ex(...,0)`) are live in main. Thanks for the fast catch and the clean strip.

Issue #138, opened 2026-07-13
[Feature]: Support 2 bit quantization


nonbasketless:
### Problem

Slow.

### Proposed solution

Incredible work first of all :).

Very humble suggestion: many of us have 128 gb unified memory. At two bits, more than half the model would fit in memory. With a larger fraction of experts residing in RAM, I'd expect generation speed to improve sharply. Perhaps more than a factor of two if this results in significantly fewer expert reloads. Right?

Evidence that smaller quants are very worthwhile:

https://www.reddit.com/r/LocalLLaMA/comments/1uimjdi/glm_52_q1_s_vs_qwen_27b_q8/

### Alternatives considered

Spending tens of thousands USD.

### Scope and compatibility

Most useful for machines with a lot of RAM, but "not enough"

JustVugg:
Your intuition is exactly right, and it's on the roadmap — on a disk-bound (or RAM-bandwidth-bound) box, fewer bits = fewer bytes/token = directly faster, and at int2 a 128 GB box holds a much larger resident fraction. The catch is quality: **naive** int2 is too lossy (~45% output error in our ablations — the absmax scale gets held hostage by one outlier coordinate per group). The lever that makes int2 *usable* is **incoherence preprocessing** — QuaRot/QuIP#-style rotation that spreads the outliers across the block before quantizing — which is being tracked in #81, and the ablation tool now measures it (`-rot` in #132). The staged plan: **int3-g64 first** (PR #168, ~−25% bytes vs int4, quality validated — I'm running that A/B right now), then **int2 + rotation** once int3 is proven. So: yes, coming, and #81 is the issue to watch. Thanks for the push and the kind words.

JustVugg:
Consolidating into #81 (rotation-preconditioned int2/int3), which is the tracking issue for exactly this — naive int2 is too lossy, rotation is the enabler, and int3-g64 (the first step) just validated at −25% bytes / +6.8% ppl. Follow #81 for progress; closing as duplicate-of-roadmap.

nonbasketless:
Interesting and thank you!

PR #111, opened 2026-07-13
GPU resident pipeline: batch CUDA attention, head-sharded kv_b, prefill expert groups, W4A16 mixed dispatch

ZacharyZcR:
Stacks on #80 (which stacks on #68). Two commits: the lab-qualified CUDA substrate (batch MLA attention, kv_b head-sharding, fused o_proj, per-device expert-group dispatch, W4A16 kernels) and the resident-pipeline layer on top (`COLI_CUDA_PIPE=1`).

## The problem it attacks

With full expert residency the six GPUs averaged **4.5% SM utilization**: the residual stream returned to the CPU between every operation, and three independent negative results (GPU scatter-add -43.8%, shared-expert fused kernel -38.8%, W4A16-with-transfers -17%) all died on the same PCIe round-trips. The fix is structural: keep the data plane on the layer's home device; keep control flow (routing, cache/pin management, speculation) on CPU.

## What's in here

- **Batch MLA attention on device** (`COLI_CUDA_ATTN=1`), optional **kv_b head-sharding** across cards with no weight duplication (`COLI_CUDA_ATTN_SHARD=1`), attention→o_proj fusion. 914-token request: 154.8s → 94.05s.
- **Device-resident prefill attention chain** (`COLI_CUDA_PIPE=1`): projections, norms, RoPE, batch attention and o_proj run as one chain on the layer device (q_a/q_b/kv_a colocated with kv_b); only the final [S,D] and the new KV rows return to host. Host KV stays canonical. Attention phase 41.2s → 30.8s on a 1571-token benchmark.
- **Prefill batch-union through the GPU expert groups**: previously gated to S≤64, so during prefill every VRAM-resident expert sat idle (measured: 21ms of GPU expert time inside a 148s prefill). Expert phase 78.9s → 69.0s.
- **W4A16 tensor-core expert groups** (`COLI_CUDA_TC_W4A16=1`) with row-count **mixed dispatch**: ≥16 rows/expert use tensor cores, smaller batches keep the naive kernel — tensor cores measured *negative* below ~16 rows (decode 4.24 → 3.37 tok/s before the fix, restored after). Expert phase → 64.3s.
- Batched router matmul (bit-identical), prefill-phase PROFILO breakdown, device-primitive unit tests (`tests/test_pipe_cuda.cu`, verified against the engine's CPU math on SM120).

[...]

PR #168, opened 2026-07-14
int3-g64 (fmt=5): per-group-scale 3-bit weights, the size/quality point your own ablation (#132) says the next container should sit at

fabio-rovai:
This adds a fourth weight format end-to-end — **int3 with per-64-group f32 scales (3.5 bits/weight)** — engine, converter, and tests, so producing an int3-g64 GLM container becomes a one-flag job (`--ebits 3 --xbits 3`) for anyone holding the FP8 source.

## Why this format, with data

Three measured inputs converge on it:

1. **Quality:** #132's own ablation (OLMoE, fp16 A/B) found `int3-g64` *beats the shipped per-row int4* (−7.5pp vs −9.3pp) at ~14% fewer bits — grouped scales matter more than the 4th bit. This PR ships that exact quantization math (mirrors `quant_ablation._quant_last_dim(bits=3, group=64)` bit-for-bit; our test suite reproduces the effect: on outlier rows int3-g64's reconstruction RMS is **3.3× lower** than per-row int4).
2. **Transport:** the engine is fetch-bound for most users. 0.86× bytes ≈ −14% on every cold expert read and ~50 GB off the container (357 → ~307 GB) — and in a policy simulation we ran over real routing telemetry for the #146 work, flat int3-g64 recompression beat both plain LRU and heat-pinned int4 at *every* RAM budget (8/25/60 GB), which is more than we can say for the fancier tiering schemes we also simulated (they only win by serving 47–90% of calls degraded).
3. **Demand:** #138 asks for 2-bit; per #132 (and our own int2 measurements) 2-bit steady-state is a quality cliff — int3-g64 is the honest answer to that request.

## Design

- **Layout (fmt=4):** per row, per 64-input group: 24 bytes = 16B low plane (2 bits/val, `matmul_i2`'s packing) + 8B high plane (1 bit/val); values in [-4,3] stored v+4; one f32 scale per group in `NAME.qs`.
- **Format detection:** fmt isn't stored on disk and row formats are inferred from weight byte counts — which int3 byte counts can alias. So the **`.qs` scale-tensor size is the tag**: `O` floats → row-scale formats, `O*ceil(I/64)` → fmt=4 (`qt_fmt_from_bytes` centralizes this for the dense loader and both `expert_load` paths, with a byte-count fallback for the degenerate I≤64 case).
- **Kernel:** `matmul_i3`, exact-f32 path with per-group scale accumulation. NEON: low plane reuses `matmul_i2`'s unpack, high plane expanded via `vtst` on bit masks. Scalar reference everywhere (x86 SIMD is a follow-up — the target regime is fetch-bound). **No IDOT path in v1**: per-group accumulation doesn't compose with the current int8-activation kernels without restructuring, and the quality economics of IDOT are under separate scrutiny anyway (#152 and the ARM datapoint we posted there: +0.169 nats/token).
- **GPU:** Metal/CUDA gates already exclude unknown formats → clean CPU fallback, no shader changes needed for v1.
- Runtime-quantize path too: `./glm 64 3 3` on an unquantized checkpoint now produces fmt=4 (previously bits=3 silently became int4).

## Tests (all green on Apple Silicon, `make test-c` + python suite)

- `tests/test_int3.c` — bit-exact pack/unpack vs a plain-C reference; `matmul_i3` (NEON + scalar tail) vs dequant-matmul across shapes including short tail groups and the real GLM `I=7168`; `qt_alloc/qt_fill/qt_bytes/matmul_qt/qt_addrow/qt_matvec_rows` plumbing; the format tag; the outlier-rows RMS assertion (int3-g64 < per-row int4, ratio 3.34).
- `tests/test_int3_load.c` — writes a real single-shard `.safetensors` fixture (fmt=4 tensor + int4 control), indexes it with `st_init`, loads through `qt_from_disk`, asserts both formats are detected and dequantize bit-identically to the packer's output.
- `tests/test_int3_convert.py` — NumPy converter round-trip against an independent decoder, plus the outlier-vs-int4 property.

## What this PR does NOT claim

GLM-native quality validation requires quantizing **from the FP8 source** (int4→int3 requantization stacks errors and is not the honest test). We don't have the 1.51 TB checkpoint; @mateogrgic (or anyone holding it) can produce the container with `--ebits 3 --xbits 3` unchanged otherwise, and #108's harness plus `TF=1` are the right acceptance gates. Happy to coordinate and to run the full decode/quality ladder on Apple Silicon once a container exists.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -171,7 +171,7 @@ else
 PYTHON    ?= python3
 endif
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_pipe_block$(EXE)
+TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_int3$(EXE) tests/test_int3_load$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_pipe_block$(EXE)
 ifneq (,$(LINUX))
 TEST_BINS += tests/test_uring$(EXE)
 endif
@@ -341,6 +341,12 @@ tests/test_sample_nan$(EXE): tests/test_sample_nan.c colibri.c st.h uring.h json
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_kv_alloc$(EXE): tests/test_kv_alloc.c colibri.c st.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_int3$(EXE): tests/test_int3.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_int3_load$(EXE): tests/test_int3_load.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_logit_nan$(EXE): tests/test_logit_nan.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -94,7 +94,13 @@ typedef struct {
  *   fmt=1 INT8  -> q8 (1 byte/param) + scala per riga
  *   fmt=2 INT4  -> q4 (2 valori per byte, impacchettati) + scala per riga
  * INT4 e' cio' che fa stare la densa residente nei 15 GB (0.5 byte/param). */
-/* fmt: 0 F32, 1 INT8, 2 INT4 (2/byte), 3 INT2 (4/byte). q4 ospita sia int4 che int2 packed. */
+/* fmt: 0 F32, 1 INT8, 2 INT4 (2/byte), 3 INT2 (4/byte), 4 INT4-GROUPED, 5 INT3-G64.
+ * q4 ospita int4/int2/int3 packed. fmt=4 (grouped int4, #242): per-row nibbles + one f32
+ * scale per group of `gs` inputs (s has O*ceil(I/gs) entries).
+ * fmt=5 (int3, per-GROUP scales, group=64, see quant.h I3_*): values in [-4,3] stored per
+ * 64-input group as 24 bytes = 16B low plane (2 bits/val, int2 layout) + 8B high plane
+ * (1 bit/val), plus ONE f32 scale PER GROUP (s has O*ceil(I/64) entries, not O). 3.5
+ * bits/weight effective — the quality/size sweet spot measured in the #132 ablation. */
 typedef struct {
     int fmt; float *qf; int8_t *q8; uint8_t *q4; float *s; int O, I, gs;  /* gs=group size (0=per-row, 128=grouped) */
 #ifdef COLI_CUDA
@@ -110,6 +116,10 @@ static int64_t qt_bytes(const QT *t){    /* byte residenti del tensore */
     if(t->fmt==4){ /* int4 grouped: packed nibbles + O*ceil(I/gs) scales */
         int ng=(t->I+t->gs-1)/t->gs;
         return (int64_t)t->O*((t->I+1)/2) + (int64_t)t->O*ng*4; }
+    if(t->fmt==5){ /* int3-g64: 24B/group weights + one f32 scale per group (I3_* in quant.h,
+                    * included below — keep the arithmetic literal here) */
+        int64_t ng=((int64_t)t->I+63)/64;
+        return (int64_t)t->O*ng*24 + (int64_t)t->O*ng*4; }
     return (int64_t)t->O*((t->I+1)/2) + (int64_t)t->O*4;  /* fmt=2 int4 per-row */
 }
 
@@ -267,6 +277,7 @@ static void qt_cuda_reset(QT *t){
     t->cuda_failed=0;
 }
 static int qt_cuda_upload(QT *t){
+    if(t->fmt==5) return 0;   /* int3-g64: no CUDA kernel yet — tensor stays CPU-side */
     const void *weights = t->fmt==0 ? (const void*)t->qf
                         : t->fmt==1 ? (const void*)t->q8 : (const void*)t->q4;
     if(t->fmt==4)   /* grouped int4 (#334): scales are [O, ceil(I/gs)] — the plain
@@ -445,7 +456,7 @@ static void matmul_qt_ex(float *y, const float *x, QT *w, int S, int allow_idot)
     }
 #endif
 #ifdef COLI_CUDA
-    if(g_cuda_enabled && w->cuda_eligible && !w->cuda_failed && !omp_in_parallel()){
+    if(g_cuda_enabled && w->cuda_eligible && !w->cuda_failed && w->fmt!=5 && !omp_in_parallel()){
         const void *weights = w->fmt==0 ? (const void*)w->qf
                             : w->fmt==1 ? (const void*)w->q8 : (const void*)w->q4;
         if(coli_cuda_matmul(&w->cuda,y,x,weights,w->s,w->fmt,S,w->I,w->O,w->cuda_device)) return;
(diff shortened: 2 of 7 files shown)


rajpratham1 (changes requested):
This is a substantial feature that introduces a new int3-g64 quantization format across the quantization pipeline, runtime kernels, loader logic, conversion tools, and test suite.

The implementation appears well thought out and includes comprehensive tests covering packing, loading, conversion, and numerical correctness, which is great to see.

Before merging, I'd appreciate review from maintainers on a few aspects:
- The long-term compatibility of introducing `fmt=4` into the existing serialization/loading logic.
- Whether the new format-detection heuristics in `qt_fmt_from_bytes()` remain robust for all supported tensor shapes and future formats.
- The maintenance implications of adding a separate int3 execution path (especially the scalar/x86 fallback versus NEON implementation).

Overall the change looks promising, but given its scope and impact on core quantization/inference paths, I think it would benefit from additional review before approval.

fabio-rovai:
Rebased onto current main (62419af, post today's merge wave — #111/#195/#199/#201/#202/#204/#205/#206/#212). One conflict, in `c/Makefile` `TEST_BINS` (union of main's new `test_schema_gbnf`/`test_compat_direct` with this branch's `test_int3`/`test_int3_load`); `glm.c` and the converter merged clean. Full `make test-c` + python suite (64 tests) green on Apple Silicon, including the int3 outlier-rows assertion (RMS ratio 3.34 vs per-row int4).

On @rajpratham1's three review questions — all three are design decisions the PR already commits to, so stating them here explicitly:

1. **fmt=4 serialization compatibility** — nothing is added to the on-disk header; fmt was never stored and still isn't. The tag is the `.qs` scale-tensor size, which is a property existing containers already have: `O` floats → the three existing row-scale formats, `O*ceil(I/64)` → fmt=4. Old containers are untouched and undetectable-as-int3 by construction (the two sizes only coincide when `I≤64`, which is exactly the degenerate case `qt_fmt_from_bytes` resolves by weight byte count instead — covered in `test_int3_load`).

2. **Detection robustness for future formats** — that's the reason detection was centralized into `qt_fmt_from_bytes()` rather than left inline in the dense loader and the two `expert_load` paths: any future format adds one arm in one function, and the existing arms are pinned by the loader test, which round-trips a real `.safetensors` fixture containing an fmt=4 tensor and an int4 control side by side.

3. **Maintenance cost of the int3 path** — the contract is scalar-reference-everywhere: every platform runs the exact-f32 scalar kernel; NEON is an optimization layered on top and tested bit-identical against it (including short tail groups and the real GLM `I=7168`). x86 SIMD is deliberately not in v1 because the target regime is fetch-bound — the kernel is not the bottleneck this format exists to move. And there is intentionally no IDOT path (per-group accumulation doesn't compose with the current int8-activation kernels, and #152 puts IDOT's quality economics under separate scrutiny), so the surface being maintained is one packing, one scalar kernel, one NEON overlay.

JustVugg:
Heads-up: #242 (group-scaled int4) just merged to `dev` and it took the `fmt=4` identifier. Your int3-g64 here also uses `fmt=4`. The on-disk *detection* doesn't collide (int3's weight-byte count differs from int4's, so it won't be mis-read as #242's format), but the `t->fmt` integer can only mean one thing in the dispatch/kernel switch. Could you renumber this to **fmt=5** on the next rebase? That keeps int4-grouped (4) and int3-g64 (5) cleanly separable. Everything else about the ablation still stands.

fabio-rovai:
New independent evidence for this PR's size/quality claim, from a per-projection ablation run last night.

I extended `tools/quant_ablation.py` with per-expert-projection overrides (scheme grammar gains `-gate<b>/-up<b>/-down<b>`, e.g. `int4-g64-up2` = experts at int4-g64 but `up_proj` at 2 bits) and ran OLMoE-1B-7B-0924, n=200/task, quantized honestly from the bf16 source, 100% parameter coverage on every scheme:

| scheme | hellaswag | arc_c | mmlu | mean acc_norm | Δ vs fp16 |
|---|---|---|---|---|---|
| fp16 | 80.0 | 54.5 | 49.5 | 61.3 | — |
| int4 (per-row, shipped) | 76.5 | 48.5 | 37.0 | 54.0 | −7.3pp |
| int4-g64 | 78.5 | 52.0 | 37.0 | 55.8 | −5.5pp |
| int4-g64-**up2** | 68.0 | 39.0 | 29.0 | 45.3 | **−16.0pp** |
| int4-g64-**up3** | 78.5 | 51.5 | 38.5 | **56.2** | −5.1pp |

Two takeaways relevant here:

1. **3-bit `up` is free.** `int4-g64-up3` matches (nominally beats) `int4-g64` — so the most conservative deployment tranche of this PR's fmt=4 is *up-only*: keep gate/down at int4, put `up_proj` at int3-g64, and a container drops ~8% of expert bytes (disk *and* per-token RAM traffic) at no measured quality cost. For FP8-source container builders that's a one-flag change with the kernels this PR already provides. (FloE, arXiv 2505.05950, motivated testing `up` specifically.)

2. **2-bit is a cliff even confined to `up`** (−16pp, the worst single move on the table) — consistent with #132's ablation and, I'd argue, settles the #138-class "2-bit tier" question for absmax-style quantization: 3 bits with group scales is the floor, which is exactly where this PR sits.

Caveats: OLMoE as proxy model, n=200/task, 0-shot acc_norm; MMLU didn't respond to grouping in this run (recovery showed on hellaswag/arc). Happy to fold the per-projection scheme support into this PR or send it separately — it's ~15 lines in the scheme parser.

JustVugg:

**It's 121 commits behind `dev`.** On `glm.c`, which changes several times a day, that isn't a rebase — it's a rewrite, and doing it *for* you would mean rewriting your design decisions without understanding why you made them. So I'd rather ask.

Your timing turned out to be better than anyone realised. #225 and #307 are converging on per-row int4 scales being the root cause of incoherent output — three users report a repetition loop, and @woolcoxm says gs=64 makes it disappear. Grouped scales (fmt=4) are now in the tree for CPU, and #298 is adding the CUDA half. int3-g64 sits exactly on that thread, so a rebased version would be reviewed against a live question rather than a hypothetical one.

fabio-rovai:
Rebased onto current `dev` and renumbered to `fmt=5` as you asked. It now sits next to #242's grouped int4 (`fmt=4`) rather than on top of it.

One detection subtlety worth flagging: int3-g64 and grouped-int4 at gs=64 carry the same scale count (`O*ceil(I/64)`), so scale size alone cannot separate them. int3-g64 is tagged instead by its distinct weight-byte count (`O*ceil(I/64)*24`), while grouped-int4 keeps its `detect_group_size` path. `qt_fmt_from_bytes` returns fmt 1/2/3/5 from the weight bytes, then the loader upgrades fmt=2 to fmt=4 via the scale array. They coexist in `qt_from_disk` and both `expert_load` paths, and `test_i4_grouped` still passes alongside the new int3 tests.

The converter now carries two orthogonal axes: your per-tensor-type `bits_map` (`--shared-bits` etc.) and this PR's per-projection `PROJ_BITS` (`--up-bits` etc.). `bits==3` routes to int3-g64 ahead of the grouped-int4 branch; everything else is unchanged.

Kept to the two original commits, 6 files, +503/-18. All 8 CI jobs pass. The first run earned its keep: it caught two real bugs in my own tests (an unconditional `numpy` import and a POSIX two-arg `mkdir`), now fixed to skip-if-absent and guard on `_WIN32` respectively.

On the live question: if #225/#307 converge on per-row int4 scales being the repetition-loop cause, int3-g64 is the 3.5-bit point on that same grouped-scale thread. `test_int3` measures 3.3x lower reconstruction error than per-row int4 on outlier rows. The Metal/CUDA gates fall back to CPU for fmt=5 today; happy to add the CUDA fmt=5 kernel to track #298, here or as a follow-up, whichever you prefer.

JustVugg:

JustVugg:
@fabio-rovai rebase heads-up: I tried to rebase this onto current `dev`, but it has **semantic** conflicts in `colibri.c` (8 hunks) that I won't resolve on your behalf — your int3-g64 (fmt=5) quant path overlaps code that moved a lot in `dev`: the #391 refactor (`glm.c` → `colibri.c`), the grouped-int4 fmt=4 work, and #413's `qt_resolve_fmt` which now validates weight/scale byte counts against `[O,I]` before use (so a new fmt=5 needs to register its expected layout there). Merging fmt=5 correctly against all that is a quant-layout judgment call I can't validate without measuring quality. Could you rebase and fold fmt=5 into the new `qt_resolve_fmt` gate? The converter side (`convert_fp8_to_int4.py`) also conflicts with #404's resume work. Happy to help with the mechanical bits once the layout call is made.

fabio-rovai:
Rebased onto current `dev` (post-#426 head), layout call made. How the 8 hunks resolved:

**`qt_resolve_fmt` registration (the judgment call you asked for):** fmt=5 registers its distinct weight-byte layout `O*ceil(I/64)*24` and its scale cardinality `O*ceil(I/64)` in #413's gate, validated against `[O,I]` like every other format. The ordering matters and is now documented in the code: int3-g64 and grouped-int4-at-gs=64 carry the **same** scale count, so the weight bytes are the int3 tag; row formats keep precedence for degenerate small-I shapes where byte counts can coincide (only possible at I<=256, where per-row is also legitimately a 1-group layout). `test_int3.c` now exercises the resolver directly, including the same-scale-count fmt=4/fmt=5 disambiguation pair at the real GLM I=7168.

**#391 placement:** `matmul_i3` + `pack_int3_g64` + the I3_* layout helpers went into `quant.h` next to their kernel family; the QT/loader/dispatch branches into `colibri.c`. One thing the rebase surfaced: the io_uring expert path (`uring_finalize_load`) still used the raw `?1:?2:3` byte inference — it never detects fmt=4 grouping and never sets `gs`, and would have mis-tagged int3 as int2. Converted it to `qt_resolve_fmt` like the other two expert paths; that closes the latent fmt-4 gap on Linux/uring as a side effect.

**Backends:** `qt_cuda_upload` returns 0 for fmt=5 (tensor stays CPU-side — same graceful pattern as your fmt=4 comment describes), the dense CUDA matmul gate excludes fmt=5, and Metal's existing gates (`gemm fmt<=3`, `moe fmt 1/2`) already reject it. Happy to add the CUDA fmt=5 kernel tracking #298's structure as a follow-up.

**Converter / #404:** reconciled — and one finding you'll want regardless of this PR: current `dev`'s resume manifests record `dict(PROJ_BITS)` in four places (`check_or_record_params` + the `--indir` progress file) but the global is never defined, so those paths **NameError at runtime today**. Commit 2 supplies the definition those sites depend on, plus the `--up-bits/--gate-bits/--down-bits` flags.

Kept to the two original commits: 7 files, +528/-12. `make check` green locally on macOS (105 python tests + all C bins including the two new int3 tests).

fabio-rovai:
@rajpratham1 your 14 Jul review asked for three things, and the version now on the branch answers each structurally rather than argumentatively:

1. **Serialization/loading compatibility** — the format no longer goes through ad-hoc detection. fmt=5 is registered in `qt_resolve_fmt` (#413), the same validated gate every other format passes: weight bytes AND scale bytes are checked against `[O,I]` before use, and an unknown combination refuses instead of misreading.
2. **Robustness of `qt_fmt_from_bytes` heuristics** — that function no longer exists; it was superseded by `qt_resolve_fmt` upstream and this PR folds into it. The one genuine ambiguity (int3-g64 vs grouped-int4-at-gs=64 share scale cardinality; row vs grouped coincide at I<=256) is documented at the resolution site and covered by dedicated assertions in `test_int3.c` at the real GLM shape.
3. **Maintenance of the separate int3 path** — post-#391 it lives in `quant.h` directly beside the int2/int4/int4-grouped kernel family it mirrors, scalar reference everywhere with NEON as an overlay, same structure as its siblings.

All 8 CI jobs are green on the rebased branch. Would you re-review when you get a chance?

Issue #163, opened 2026-07-14
MTP acceptance regression: 33% -> 4% (same machine/config/int8 head) between pre-#68 main and current dev


juicyroots:
### What happened

MTP draft acceptance dropped from **33% to 4%** on the identical machine, model, prompt, and configuration between pre-#68 `main` (commit 1bdaeee) and current `dev` HEAD (6d3ed7e). At 4% the adaptive guard disables drafts after 24 proposals, so speculation effectively never engages anymore.

### A/B data

GLM-5.2 744B int4 + **int8 MTP head** (mateogrgic checkpoint), Ryzen 9 9950X3D, 126 GB, Windows 11 native (MinGW GCC 16.1). Same prompt ("Summarize the plot of Romeo and Juliet in three sentences."), same warm `.coli_usage`, `--ram 75 --ngen 32 --topp 0.7`, 40 GB pin, MLOCK=1:

| engine | speculation | acceptance |
|---|---|---|
| main @ 1bdaeee (pre-#68) | 2.00 tok/forward (16 fw / 32 tok) | **33% (16/48)** |
| dev @ 6d3ed7e | 1.07 tok/forward (30 fw / 32 tok) | **4% (1/24)** → auto-disabled |

Reproduced at temp 0.7 and at `--temp 0` (greedy), and identically with `DIRECT=0` vs `DIRECT=1` (byte-identical greedy outputs across the I/O paths, so it's not a data/read issue). On the older engine the same head reached 56–83% acceptance on other prompts; the 33% row is the strict same-config comparison.

Possibly related: #156 reports MTP as a net loss on Windows (DRAFT=0 faster) — if that machine ran a post-#68 build, this regression rather than cold-cache draft cost may be the real cause.

### Suspects

The #68 decode-batching rework touched the draft/verify path (`spec_decode`/`run_serve_mux` area). Acceptance collapsing by ~8× with the same int8 head suggests the verification distribution or the draft context changed, not the head quality. One more observation from the same runs, possibly the same root cause: greedy decode at `--topp 0.7` now opens the R&J summary with off-distribution text ("Aegeanin (assuming you meant …") — deterministic and identical across I/O paths.

Happy to run bisects or instrumented builds on this rig (it's fast to iterate: warm cache, ~60 s per 32-token run).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JustVugg:
[...]

PR #174, opened 2026-07-14
Skip OMP hot-thread tuning on GPU builds (fixes #159 ~3x regression, fixes #116 Metal -39%)

### LINKED ISSUE #159: [Bug]: ~3x throughput regression on x86+CUDA since 92011de — OMP hot-spin tuning vs new threads (I4_ACC512 exonerated; COLI_NO_OMP_TUNE=1 recovers and then some)
### Commit or version

Regression present at current main (6d3ed7e); last known good is 92011de. (Also reproduces on `dev`, where we first hit it.)

### Environment

Linux (kernel 7.0.0-27, native, no WSL2), gcc 15, CUDA 13.2 / driver 595.71.05, `make glm CUDA=1 ARCH=native`. AMD Ryzen 9 9950X3D2 (16C/32T, AVX-512 VNNI), 121 GB DDR5, 2× RTX 5090 (sm_120), Gen5 NVMe (11.5 GB/s O_DIRECT per iobench). GLM-5.2 int4 weights, warmed `.coli_usage` (3.3M+ selections).

### Reproduction steps

```shell
# identical config on both commits (the README's Gen5-NVMe + 5090 row, #120):
MTP=0 DIRECT=1 PIPE_WORKERS=16 PREFETCH=1 \
./coli run --model <model> --auto-tier --ram 90 --gpu 1 --vram 28 --repin 16 --ngen 256 "<essay prompt>"
```

### Expected behavior

Comparable or better throughput than 92011de (I4_ACC512 #95 should be ~+7% on the CPU-served matmul share).

### Actual behavior and logs

**~3× throughput regression on x86 + CUDA:** 92011de → **1.17 tok/s**; 6d3ed7e → **0.41 tok/s**, same config, same warmed pins, back-to-back on the same box. Every CPU profile bucket inflates 4–5×:

```
92011de: expert-disk 89.9s | expert-matmul 49.9s | attention 44.8s | other 33.4s   (256 tok)
6d3ed7e: expert-disk 96.4s | expert-matmul 220.5s | attention 240.1s | other 169.8s
```

**Isolation matrix** (96-token heats, current main, same config):

| variant | tok/s |
|---|---|
| stock | 0.39 |
| `COLI_NO_OMP_TUNE=1` | **1.29** |
| `I4_ACC512=0` | 0.32 |
| `IDOT=0` | 0.37 |
| `OMP_NUM_THREADS=12`, spin on | 0.93 |
| `OMP_NUM_THREADS=14`, spin on | 0.92 |

So: the new kernels (#95) are innocent — in fact with the spin disabled, current main is **faster** than 92011de (1.29 vs 1.17, consistent with the promised I4_ACC512 gain). The hot-thread tuning block itself is byte-identical between the two commits, so the regression is the `OMP_WAIT_POLICY=active` + `GOMP_SPINCOUNT=200000` policy interacting with something that spawns new threads on current main (async pipe / CUDA-sync paths?). Shrinking the team does **not** fix it — the policy has to go off entirely.

This looks like the exact x86 twin of the M5 Metal finding in #116 (OMP hot-team spin throttling the GPU, NO_OMP recovers): the same kill-switch takes this box from a 3× loss to a new best (details in #120 — the full re-run numbers are being posted there).

Suggested short-term mitigation: default the spin tuning off when the CUDA backend is active (or a docs warning for x86+CUDA), until the thread interaction is understood. Happy to A/B candidate patches on this hardware and to share full logs/profiles for every run above.

ZacharyZcR:
One-line gate: the `OMP_WAIT_POLICY=active` + `GOMP_SPINCOUNT` re-exec tuning (from #77) is skipped when `COLI_CUDA` or `COLI_METAL` is set.

## The problem

The hot-thread spin helps CPU-only decode (+3× on Zen5, measured in #77), but **hurts whenever a GPU backend does the work**:
- **x86 + CUDA** (#159): spinning OMP workers contend with CUDA dispatch threads → **~3× throughput regression** (0.41 vs 1.17 tok/s). `COLI_NO_OMP_TUNE=1` recovers instantly (1.29 tok/s, a new best with #95).
- **Apple Metal** (#116): active spin steals SoC power budget from the GPU → **−39%**. Same kill-switch recovers.

Two independent GPU platforms, same mechanism — that's a strong signal.

## The fix

```c
if(!getenv("COLI_OMP_TUNED") && !getenv("COLI_NO_OMP_TUNE") &&
   !getenv("COLI_CUDA") && !getenv("COLI_METAL")){
```

CPU-only builds are completely unchanged. GPU users who set `COLI_NO_OMP_TUNE=1` as a workaround can drop it.

## Verification

The identical gate was tested on our 6×RTX 5090 lab machine during the #111 pipeline work: without it, prefill regressed 4× (CPU stuck at 1.8 cores); with it, full throughput recovered (122.7s prefill, 4.17 tok/s decode). The #159 reporter independently confirmed `COLI_NO_OMP_TUNE=1` takes their box from 0.39 to 1.29 tok/s.

Targets `main` directly — this is a regression fix for code already on main.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -3667,7 +3667,8 @@ int main(int argc, char **argv){
      * re-exec + tuning path (distinct from the internal COLI_OMP_TUNED sentinel).
      *
      * Must remain the FIRST statement in main(): argv is passed verbatim to execv(). */
-    if(!getenv("COLI_OMP_TUNED") && !getenv("COLI_NO_OMP_TUNE")){
+    if(!getenv("COLI_OMP_TUNED") && !getenv("COLI_NO_OMP_TUNE") &&
+       !getenv("COLI_CUDA") && !getenv("COLI_METAL")){
         setenv("OMP_WAIT_POLICY","active",0);  /* keep the team hot across the tiny per-expert matmul regions */
         setenv("GOMP_SPINCOUNT","200000",0);   /* spin briefly, then yield so long disk waits don't burn a core */
         setenv("OMP_PROC_BIND","close",0);     /* pack the team onto adjacent cores for cache locality */
--- a/c/glm.c
+++ b/c/glm.c
@@ -392,8 +392,12 @@ static void matmul_i4_pair(float *yg, float *yu, const float *x,
 }
 
 static void matmul_qt(float *y,const float *x,QT *w,int S);
+static int g_no_fused_pair=0;  /* COLI_NO_FUSED_PAIR=1: disable the gate+up kernel fusion
+ * that changes OMP scheduling vs separate matmul_qt calls — this
+ * shifts floating-point accumulation order and can collapse MTP
+ * draft acceptance by flipping near-ties (#163). */
 static void expert_gate_up(float *g,float *u,const float *x,QT *wg,QT *wu,int S){
-    if(S==1&&wg->fmt==2&&wu->fmt==2&&wg->I==wu->I&&wg->O==wu->O)
+    if(!g_no_fused_pair&&S==1&&wg->fmt==2&&wu->fmt==2&&wg->I==wu->I&&wg->O==wu->O)
         matmul_i4_pair(g,u,x,wg->q4,wg->s,wu->q4,wu->s,wg->I,wg->O);
     else { matmul_qt(g,x,wg,S); matmul_qt(u,x,wu,S); }
 }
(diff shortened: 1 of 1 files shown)


ZacharyZcR:
Added a second commit for **#163** (MTP acceptance regression) on top of the OMP fix:

The fused `expert_gate_up` kernel (`matmul_i4_pair`) changes the OMP static-schedule thread assignment vs the pre-#68 separate `matmul_qt` calls — same math, different accumulation order, near-tie argmax shifts through 78 layers, acceptance collapses from 33% to 4%.

`COLI_NO_FUSED_PAIR=1` disables the fusion as an immediate diagnostic/workaround. If the #163 reporter confirms acceptance recovers with this flag, the mechanism is proven and the proper fix is making draft (S=1) and verify (S>1) use identical kernels.

Both fixes are on this branch — one gate for OMP spin (#159), one gate for the pair kernel (#163).

PR #190, opened 2026-07-14
bench: use XDG cache home for --data default

bennypowers:
## Summary

- Move bench `--data` default from `c/bench/` (source tree) to the platform cache directory
  - Linux/macOS: `$XDG_CACHE_HOME/colibri/bench` (defaults to `~/.cache/colibri/bench`)
  - Windows: `%LOCALAPPDATA%\colibri\bench` (defaults to `C:\Users\<user>\AppData\Local\colibri\bench`)
- Benchmark datasets are downloaded artifacts, not source files, so they belong in the user's cache directory
- `fetch_benchmarks.py` already calls `os.makedirs(out, exist_ok=True)`, so the new path is created on first use

## Test plan

- [ ] Run `coli bench --help` and verify `--data` default shows the cache path
- [ ] Run `coli bench` and verify datasets download to `~/.cache/colibri/bench/`
- [ ] Verify `--data /custom/path` still overrides the default

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -621,7 +621,13 @@ def main():
     ps.add_argument("--queue-timeout",type=float,default=float(os.environ.get("COLI_QUEUE_TIMEOUT","300")))
     ps.add_argument("--kv-slots",type=int,default=int(os.environ.get("COLI_KV_SLOTS","1")))
     pb=sub.add_parser("bench", parents=[common]); pb.add_argument("tasks", nargs="*")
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
The change is straightforward and improves the default benchmark data location by following platform conventions.

Using `XDG_CACHE_HOME` on Unix-like systems and `LOCALAPPDATA` on Windows makes the default cache location more appropriate while still allowing users to override it with `--data`. The implementation is simple, readable, and preserves cross-platform compatibility.

LGTM!

PR #193, opened 2026-07-14
stats: opt-in disk-load split by speculation context and layer kind (DISK_SPLIT=1)

### LINKED ISSUE #182: Cache-aware drafting: a measured null, plus draft-vs-verify disk-load instrumentation
Following up on my fleet benchmarks (#107): I prototyped "cache-aware speculative drafting" — the idea that draft tokens, whose quality only affects acceptance rate (verify is exact), could route around disk misses. It's a null, and the instrumentation that proves it seems worth having upstream. All on M4 Pro / 48 GB / Metal, dev branch, 353 GB int4 model, int8 MTP heads.

1. Instrumentation. I added counters splitting expert disk loads by context (mtp_draft / mtp_absorb / verify+main) and by layer kind (MTP layer's int8 experts vs main int4). One 64-token run, --ram 38:

disk-load split: draft 628 + absorb 333 + verify/main 40869 misses
MTP-layer 961 loads 36.32 GB | main-layers 41436 loads 783.78 GB (MTP 4.4% of bytes)

The draft path is 1.5% of disk loads and 4.4% of bytes. The verify pass — which exactness makes untouchable — is where speculation's disk cost lives. Any draft-side cache trick is chasing a rounding error.

2. The experiment anyway (two modes, env-gated, default off, verify untouched):
- DRAFT_CACHED=1: draft routing restricted to resident experts, weights renormalized.
- DRAFT_CACHED=2: miss-budget gate — predict the verify's disk cost from the previous token's routing (eroute) vs residency; drop drafts when predicted misses exceed DRAFT_MISS_MAX.

A/B, same prompt/flags, baseline→experiment→baseline:

```
┌──────────────────────────┬───────┬────────┬────────────┬───────────────┐
│           run            │ tok/s │ tok/fw │ MTP accept │ verify misses │
├──────────────────────────┼───────┼────────┼────────────┼───────────────┤
│ baseline-1               │ 0.28  │ 2.56   │ 51%        │ 36,651        │
├──────────────────────────┼───────┼────────┼────────────┼───────────────┤
│ DRAFT_CACHED=2 (thr 192) │ 0.26  │ 2.06   │ 38%        │ 39,557        │
├──────────────────────────┼───────┼────────┼────────────┼───────────────┤
│ baseline-2               │ 0.30  │ 2.67   │ 54%        │ 35,116        │
└──────────────────────────┴───────┴────────┴────────────┴───────────────┘
```

3. Why it can't work (mechanism, not just numbers):
- The gate fired on 3/31 steps and the two baselines differ from each other more than from the experiment — at temp 0.7, generation noise (acceptance swung 42–54% across unmodified runs) dwarfs the effect. Detecting a 5% change here needs replay/greedy A/B pairs.
- The cost proxy measures the wrong thing: the previous token's eroute set is by construction the warmest possible expert set; the true incremental draft cost is routing divergence between draft positions and token 0, which that signal cannot see.
- Most fundamentally: at ~50% acceptance and 2.5+ tok/fw, speculation is already net-positive on this class of machine, so any gate that converges toward "speculate less" starts from a losing position.

4. Two things that might be worth keeping:
- The instrumentation (+~40 lines of the patch): the draft/verify disk split doesn't exist in the stats line today and would have killed this idea in one run. Happy to PR it separately, without the DRAFT_CACHED modes.
- An observed inefficiency: the MTP layer's LRU held only 7 slots in my config, missing ~90% of draft routings — 36 GB of int8 reads per 64 tokens from the model's single hottest layer. A small dedicated hot store (~32 MTP experts ≈ 1.2 GB) would erase that at zero acceptance cost. Expected gain is only ~3–4% so I didn't chase it against my noise floor, but it's free of the acceptance trade-off that sinks draft degradation.

Branch with instrumentation + both modes exists locally (+103/−1, one file); say the word and I'll PR whichever part is useful.

bokiko:
This adds the disk-load instrumentation discussed in #182. With DISK_SPLIT=1, the final stats line of coli run splits expert disk loads (LRU misses reaching expert_load) two ways: by context — loads issued while drafting (mtp_draft), while absorbing verified tokens into the MTP KV (mtp_absorb), and during verify/main forwards — and by layer kind — the MTP layer (int8 experts, ~2× the bytes) vs the main layers (int4), with exact bytes read (weights + scales). It answers "where do my disk bytes actually go?" when tuning speculation: on an M4 Pro 48 GB with GLM-5.2 int4, the draft path turns out to be ~1.5% of misses and the MTP layer ~4.5% of bytes.

Zero overhead when off, per the spec in #182: DISK_SPLIT is read once at startup into a static flag (same pattern as the existing g_* flags), every counter increment sits behind that flag check, and the extra stats line prints only when enabled — with the flag unset no atomic is ever touched and output is byte-identical to stock. The counters are __atomic relaxed adds because expert_load also runs on OMP/PIPE/pilot threads. Measurement only: routing, verification and generated output are untouched.

Example (16 tokens, METAL, --ram 38):
disk-load split: draft 153 + absorb 110 + verify/main 11481 misses | MTP-layer 263 loads 9.94 GB | main-layers 12048 loads 227.89 GB (MTP 4.2% of bytes)
Closes #182.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -174,6 +174,12 @@ typedef struct {
     double t_edisk, t_ewait, t_emm, t_attn, t_kvb, t_head;/* profiling: dove va il tempo */
     double t_aproj,t_acore,t_aout;                     /* attention breakdown */
     int64_t resident_bytes;
+    /* DISK_SPLIT=1: split dei DISK LOAD (miss LRU -> expert_load) per contesto e per tipo
+     * di layer. ld_ctx: 0=main/verify/prefill, 1=dentro mtp_draft, 2=dentro mtp_absorb. */
+    int ld_ctx;
+    uint64_t miss_draft, miss_absorb;            /* miss in moe() per contesto */
+    uint64_t ld_mtp, ld_main;                    /* expert_load per tipo layer (MTP int8 vs main int4) */
+    uint64_t bytes_mtp, bytes_main;              /* byte letti da disco per tipo layer */
 } Model;
 
 static void usage_save(Model *m);        /* cache che impara: definita accanto a stats_dump */
@@ -835,6 +841,11 @@ static int64_t la_hit[3], la_tot[3];
 static int la_pred[2][130][16]; static signed char la_val[2][130];
 static int g_pilot=0;    /* PILOT=1: prefetch pilotato dal router (vedi pilot_prefetch) */
 static int g_pilot_k=8;  /* PILOT_K=k: prefetcha solo le prime k predizioni per posizione */
+static int g_disk_split=0; /* DISK_SPLIT=1: contatori che spezzano i DISK LOAD (miss LRU) in
+                          * draft MTP / absorb / verify-main e in layer MTP (int8) vs main
+                          * (int4), con i byte letti. Default OFF: a flag spento gli atomic
+                          * non vengono MAI toccati (zero overhead), le righe extra di stats
+                          * non vengono stampate. Solo misura: nessun effetto sull'output. */
 /* Aligned allocator for dense QT weights/scales: under METAL, page-align + register so the
  * GPU reads them zero-copy (no upload duplicate). Plain malloc otherwise. */
 static void *qalloc(size_t n){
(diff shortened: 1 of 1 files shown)


JustVugg:
Yes — this is exactly the instrumentation I invited in #182, and an issue to spec it first is even better. Design ACK: `DISK_SPLIT=1`, default off, zero cost when unset, counters split by context (draft / absorb / verify+main) and layer kind (MTP vs main), dumped in the end-of-run stats line. It would have turned three recent debates (MTP net-loss, disk-vs-matmul regimes, draft-side caching) into one command each. Send the PR against `dev` whenever ready.

bokiko:
Just to close the loop on sequencing: this PR *is* that delivery — branch cut from `dev`, spec exactly as ACKed (default off, zero cost unset, context + layer-kind split in the stats line). Ready for review.

Issue #153, opened 2026-07-13
Prefill is attention-bound (not disk-bound), and int4 IDOT costs ~9-18% perplexity on the attention projections — ARM/POWER enable it by default


dnnspaul:
Follow-up to #101 (where I measured *decode* on this box and confirmed it is disk-bound). **Prefill behaves completely differently, and I think the project's mental model is off for it.** Everything below is on `upstream/dev` @ 6d3ed7e, Zen5 (16c) + RTX 5090, GLM-5.2 744B int4, 78 layers, DSA on, MTP=0, OMP hot-thread tuning **on**, prefill always run to completion (the engine's own `--- N tokens in X.XXs` line is the only thing I count).

## 1. Prefill is attention-bound. Disk is flat.

Engine `PROFILE`, unpatched dev:

| prompt | total | expert-disk | expert-matmul | attention | ├ proj/RoPE | └ score-softmax |
|--------|-------|-------------|---------------|-----------|-------------|-----------------|
| 256 | 120.9s | 19.4s | 34.6s | **49.2s** | 22.7s | 9.3s |
| 512 | 231.4s | 21.2s | 58.3s | **118.0s** | 45.5s | 38.6s |

**expert-disk is essentially constant** (19.4s → 21.2s) while attention roughly doubles. The whole model streams off disk once per prefill and that is *not* the bottleneck — a faster NVMe buys you nothing here. Attention is 41% → 51% of prefill and climbing.

`score-softmax-value` is quadratic (9.3 → 38.6 for 2× tokens). DSA's top-2048 caps it beyond 2048 ctx, but below that it is uncapped.

**Consequence:** prompts of the size any agentic client sends (opencode's default system prompt is ~11k tokens) are *hours* of prefill on this box, and no amount of disk or expert-cache work fixes that. I could not get opencode usable against colibrì for this reason.

## 2. The attention projections sit on the wrong side of the IDOT gate

`q_a`/`q_b`/`kv_a` are called one row at a time in the prefill loop, so they never reach `S >= g_i4s` and were the only prefill matmuls still on the exact int4 kernel. I batched them in #152 — bit-identical output, −4.5% prefill.

But while doing it I found the interesting part. **Letting those projections use the int4 IDOT kernel costs real quality.** Two 1023-token passages, `SCORE` mode, CPU (deterministic — I verified by running the reference twice and getting bit-identical output; note the CUDA path is *not* run-to-run deterministic, see §4):

[...]

PR #194, opened 2026-07-14
eval: default the [gMASK]<sop> prefix ON for GLM snapshots (#108)

bokiko:
Half of the fix agreed with @dnnspaul in #108 (he's covering tools/quant_ablation.py). eval_glm.py now detects a GLM snapshot from config.json's model_type and prepends [gMASK]<sop> to every scoring context by default, with a stderr notice. EVAL_PREFIX still overrides for research use — including EVAL_PREFIX= (empty) to deliberately reproduce the old out-of-distribution behavior.

Rationale is in #108: GLM sees this prefix at the start of every training sequence, so scoring raw completions without it silently depresses scores (hellaswag 60→90 acc_norm at n=40 with the fix) and — worse, per @dnnspaul's #153 retraction — distorts A/B sensitivity. Defaulting ON makes the bug impossible to hit by omission, which is where it's been biting people.

Validated on GLM-5.2 int4: auto-detect fires on model_type=glm_moe_dsa (contexts gain the two prefix tokens, notice printed), EVAL_PREFIX= output is byte-identical to stock, --selftest passes.

### DIFF
--- a/c/tools/eval_glm.py
+++ b/c/tools/eval_glm.py
@@ -48,11 +48,24 @@ def load_docs(task, data_dir, limit, seed):
     random.Random(seed).shuffle(docs)
     return docs[:limit] if limit else docs
 
-def build_requests(tk, docs_by_task):
+def detect_prefix(snap):
+    """GLM sees [gMASK]<sop> at the start of every training sequence; scoring raw text
+    without it is out-of-distribution and silently depresses/distorts scores (#108).
+    Default the prefix ON for GLM snapshots; EVAL_PREFIX (even empty) overrides."""
+    if "EVAL_PREFIX" in os.environ: return os.environ["EVAL_PREFIX"]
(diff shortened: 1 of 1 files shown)


JustVugg:
Agreed — defaulting `[gMASK]<sop>` ON for GLM snapshots in eval/scoring is the right call: every result we've had without it turned out misleading (#108's confound, the retracted 9-18% in #153), and #155 already conditions the ablation tool. The engine's `SCORE` path should do the same detection (GLM config → prefix on, override via env). Happy to take a PR, or I'll fold it into the next eval-correctness pass on `dev`.

bokiko:
Thanks! I will take the `SCORE`-path follow-up as a separate PR: same detection (config `model_type` contains "glm"), env override, plus one guard the eval side makes necessary — token streams that already begin with `[gMASK]<sop>` (e.g. from this PR's patched `eval_glm.py`) must not get double-prefixed. Will validate token-exact against GLM-5.2 int4 on known scores before sending.

Issue #136, opened 2026-07-13
[Performance]: Dell Pro Max GB10 (DGX Spark) — 0.50 tok/s warm CPU-only, unified-memory CUDA tier tested, disk 5.58 GB/s


blakemartz:
**Hardware:** Dell Pro Max GB10 (DGX Spark platform — same silicon as the NVIDIA unit in #76 and the ASUS GX10): Grace 10× Cortex-X925 + 10× A725 (aarch64, `i8mm`/`sve2`), GB10 sm_121, 121 GB unified LPDDR5x, Dell OEM NVMe (ESL04TBTLCZ 4 TB). Linux 6.17, gcc 13.3, CUDA 13.0, colibri @ `a78a06f`, container `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp`, `./glm 64 4 8`, PROMPT mode, TEMP=0, ~109 GB MemAvailable during runs.

**Correctness first:** everything from the #76 resolution verified on this unit — `make_glm_oracle.py` regen byte-identical to the committed `ref_glm.json`, tiny self-test TF **32/32** / greedy **20/20** on both CPU NEON and CUDA sm_121, and the 744B container produces coherent text in PROMPT mode.

**Disk (measured):** `iobench` 19 MB random × 8 threads O_DIRECT → **5.58 GB/s**. Note for other GB10 owners: the NVIDIA-branded unit in #76 reports 10.74 GB/s — same platform, different OEM SSD, ~2× disk spread. Measure your own before comparing tok/s.

**Decode:**

| run | tok/s | expert hit | MTP acceptance | tok/forward |
|---|---|---|---|---|
| cold, first session (NGEN=48) | 0.21 | 64% | 20% | 1.57 |
| warm, second session (NGEN=128) | **0.50** | **83.2%** | **73%** | 3.20 |

Warm profile per 128 tokens: expert-disk 57.9 s / **expert-matmul 129.8 s** / attention 49.7 s — once the learning cache warms, this box is **matmul-bound, not disk-bound**.

**The unified-memory CUDA question** (from the #76 thread — the "no-PCIe-cost CUDA tier"): tested, and it does *not* invert the discrete-GPU finding, for a new reason:

| config | tok/s | what happened |
|---|---|---|
| CPU only (warm) | **0.50** | baseline |
[...]

PR #217, opened 2026-07-14
aarch64: add runtime i8mm expert kernel

RonitBStudent:
## Summary

Add a runtime-dispatched AArch64 i8mm expert kernel for the existing int8-activation IDOT path.

- compile `SMMLA` only inside per-function i8mm targets, preserving the baseline AArch64 binary
- detect i8mm at runtime with `hw.optional.arm.FEAT_I8MM` on macOS and `AT_HWCAP2/HWCAP2_I8MM` on Linux
- compute a 2-output × 2-activation tile so all four `SMMLA` lanes do useful work
- use four independent accumulators to hide matrix-dot latency
- support both int8 and packed-int4 weights, arbitrary tails, odd `S`, and odd output counts
- keep the measured-slower int4 `S=1, I<6144` case on the existing NEON path
- add `I8MM=0` for same-binary A/B measurements

This keeps the dependency-free default CPU path and file formats unchanged.

## Why

The warm profile in #136 is matmul-bound on AArch64, while the current NEON IDOT path uses SDOT but not the available i8mm/SMMLA instructions. A direct instruction substitution was slower; the useful unit is a 2×2 tile with activation-pair packing and multiple independent accumulators.

## Validation

Apple M2 Pro (12 cores, 32 GB), macOS arm64, Apple Clang 21.0.0. The benchmark was single-threaded because local libomp headers were unavailable. It uses the real GLM-5.2 expert shapes, one untimed call per path, 21 timed A/B pairs with alternating order, and reports medians.

```text
$ make -C c OMPDIR= i8mm-bench
format shape              S   NEON-ms SMMLA-ms speedup  dispatch
i8   gate/up            S=1    0.867    0.526   1.648x  SMMLA
i8   down               S=1    0.652    0.529   1.233x  SMMLA
i8   gate/up            S=2    1.737    0.542   3.205x  SMMLA
i8   down               S=2    1.423    0.556   2.559x  SMMLA
i8   gate/up            S=4    3.445    1.015   3.394x  SMMLA
i8   down               S=4    2.638    0.986   2.675x  SMMLA
i4   gate/up            S=1    0.863    0.853   1.012x  SMMLA
i4   down               S=1    0.656    0.653   1.005x  NEON guard
i4   gate/up            S=2    1.725    0.850   2.029x  SMMLA
i4   down               S=2    1.305    0.853   1.530x  SMMLA
i4   gate/up            S=4    3.493    1.730   2.019x  SMMLA
i4   down               S=4    2.658    1.703   1.561x  SMMLA
```

Checks run:

- `make check OMPDIR=` — clean CPU build, C suite, and 62 Python tests pass with zero compiler warnings
- `./c/tests/test_idot` — exact against scalar references for int8/int4 tiles, random full-range weights, `-128` edges, tails, and `S=1..3` / odd-output dispatch
- x86_64 compile-only checks for `glm.c`, `test_idot.c`, and `bench_i8mm.c` — clean compile-out of the AArch64 path
- `git diff --check`

## Still needed before ready

I do not have the full model snapshot locally. Please run the token-exact oracle (32/32 TF + 20/20 greedy) and `OMP_NUM_THREADS=1 make -C c i8mm-bench` on an i8mm Linux machine/GB10 or Grace. `I8MM=0` provides the same-binary NEON control.

Tracks the aarch64 i8mm frontier item in discussion #209.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -229,6 +229,12 @@ cuda-bench: backend_cuda.cu backend_cuda.h tests/bench_tensor_core.cu
 	$(NVCC) $(NVCCFLAGS) backend_cuda.cu tests/bench_tensor_core.cu -o backend_cuda_bench$(EXE)
 	./backend_cuda_bench$(EXE)
 
+i8mm_bench$(EXE): tests/bench_i8mm.c glm.c st.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+i8mm-bench: i8mm_bench$(EXE)
+	./i8mm_bench$(EXE)
+
 olmoe$(EXE): olmoe.c st.h json.h compat.h
 	$(CC) $(CFLAGS) olmoe.c -o olmoe$(EXE) $(LDFLAGS)
 
@@ -299,4 +305,4 @@ clean:
 
 bench: iobench$(EXE)
 	@if [ -n "$(ARGS)" ]; then ./iobench$(EXE) $(ARGS); else echo "built iobench$(EXE) — run: ./iobench$(EXE) <file> <MB> <iters> <threads> <direct 0|1>"; fi
-.PHONY: all glm cuda-test cuda-bench cuda-dll portable test-c test-python test check clean install uninstall bench
+.PHONY: all glm cuda-test cuda-bench cuda-dll i8mm-bench portable test-c test-python test check clean install uninstall bench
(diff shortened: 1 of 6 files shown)


JustVugg:
Closing this draft — it's stalled (171 behind, no update since the 14th) and, honestly, we have no way to verify it: no aarch64 hardware here and no ARM job in CI, so the i8mm kernel could only ever be "it compiles," which isn't enough for a correctness-critical dot kernel.

Not a no to ARM — the opposite. If you reopen against current `dev` with (a) the i8mm path guarded so a non-i8mm ARM build still works, and (b) an exactness test like `tests/test_idot.c` that checks the kernel against a scalar reference (that test runs in CI on x86 and would run on ARM once there's a runner), it becomes reviewable and mergeable. The test is what turns "trust the SIMD" into "CI proves it." Ping me and I'll help wire the test harness.

PR #223, opened 2026-07-14
Nearly double the hit-rate

withinboredom:
## Summary

I saw a video on YouTube about this project and realized the problems you're trying to solve are very similar to ours over on [Swytch](https://github.com/swytchdb/swytch) (A leaderless distributed cache with serializable guarantees). The caching algorithm that makes it possible is called "cloxcache" and we're currently writing a paper on it. The caching family would be CLOCK-LRU-K -- and it adapts LFU/LRU on-the-fly. I've adapted it from Go to C, and your cache structures. To our knowledge, it's the most performant cache-eviction algorithm currently in existence on large caches, outperforming other eviction algorithms and ~99.998% as good as a perfect cache can ever be.

Anyway, enough with the marketing speak, here's the data:

image

Reproduction:

```sh
export SNAP=$MODEL PROMPT="Explain how MoE offloading works in detail." NGEN=256
export TEMP=0 DRAFT=8 AUTOPIN=0 DIRECT=1
export RAM_GB=32
./glm 8
```

Finals:
| Run | hit rate | decode time | disk io | rss |
| --- | ----     |  ----------- | ------ | --- |
| baseline | 9.4% | 3,865s | 1,176s | 32.9gb |
| cloxcache | 15.1% | 3,628s | 1,103s | 31.67gb |
| delta | +60% | -237s | -73s | -- |

Hardware:
```
Linux beta 6.8.0-134-generic #134-Ubuntu SMP PREEMPT_DYNAMIC Fri Jun 26 18:43:11 UTC 2026 x86_64 x86_64 x86_64 GNU/Linu
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             43 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      12
  On-line CPU(s) list:       0-11
Vendor ID:                   AuthenticAMD
  BIOS Vendor ID:            Advanced Micro Devices, Inc.
  Model name:                AMD Ryzen 5 3600 6-Core Processor
    BIOS Model name:         AMD Ryzen 5 3600 6-Core Processor               Unknown CPU @ 3.6GHz
    BIOS CPU family:         107
    CPU family:              23
    Model:                   113
    Thread(s) per core:      2
    Core(s) per socket:      6
    Socket(s):               1
    Stepping:                0
    Frequency boost:         enabled
    CPU(s) scaling MHz:      59%
    CPU max MHz:             4208.2031
    CPU min MHz:             2200.0000
    BogoMIPS:                7199.68
    Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt
                             pdpe1gb rdtscp lm constant_tsc rep_good nopl nonstop_tsc cpuid extd_apicid aperfmperf rapl pni pclmulqdq monitor ssse3 fma cx16
                             sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefet
                             ch osvw ibs skinit wdt tce topoext perfctr_core perfctr_nb bpext perfctr_llc mwaitx cpb cat_l3 cdp_l3 hw_pstate ssbd mba ibpb st
                             ibp vmmcall fsgsbase bmi1 avx2 smep bmi2 cqm rdt_a rdseed adx smap clflushopt clwb sha_ni xsaveopt xsavec xgetbv1 cqm_llc cqm_oc
                             cup_llc cqm_mbm_total cqm_mbm_local clzero irperf xsaveerptr rdpru wbnoinvd arat npt lbrv svm_lock nrip_save tsc_scale vmcb_clea
                             n flushbyasid decodeassists pausefilter pfthreshold avic v_vmsave_vmload vgif v_spec_ctrl umip rdpid overflow_recov succor smca
                             sev sev_es ibpb_exit_to_user
Virtualization features:
  Virtualization:            AMD-V
Caches (sum of all):
  L1d:                       192 KiB (6 instances)
  L1i:                       192 KiB (6 instances)
  L2:                        3 MiB (6 instances)
  L3:                        32 MiB (2 instances)
NUMA:
  NUMA node(s):              1
  NUMA node0 CPU(s):         0-11
Vulnerabilities:
  Gather data sampling:      Not affected
  Indirect target selection: Not affected
  Itlb multihit:             Not affected
  L1tf:                      Not affected
  Mds:                       Not affected
  Meltdown:                  Not affected
  Mmio stale data:           Not affected
  Reg file data sampling:    Not affected
  Retbleed:                  Mitigation; untrained return thunk; SMT enabled with STIBP protection
  Spec rstack overflow:      Mitigation; Safe RET
  Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
  Spectre v1:                Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:                Mitigation; Retpolines; IBPB conditional; STIBP always-on; RSB filling; PBRSB-eIBRS Not affected; BHI Not affected
  Srbds:                     Not affected
  Tsa:                       Not affected
  Tsx async abort:           Not affected
  Vmscape:                   Mitigation; IBPB before exit to userspace
free -h
               total        used        free      shared  buff/cache   available
Mem:            62Gi        12Gi        31Gi       198Mi        19Gi        49Gi
Swap:             0B          0B          0B
```

## Validation

- [x] `make -C c check`

## Compatibility


JustVugg:
Thank you for this — I ran it on our reference box and I owe you the numbers, because they're interesting in a way that isn't "yes" or "no".

## What I measured

ABA (dev → dev+223 → dev), same prompt, same env, `DIRECT=1 URING=1 PIPE=1 AUTOPIN=0`, `RAM_GB=22`, `NGEN=16`, machine idle and verified idle between runs:

| run | hit rate | **experts loaded/token** | decode |
|---|---|---|---|
| A1 — dev | 3.8% | **876.5** (11.69/layer) | 251s |
| B — dev + #223 | 3.7% | **876.5** (11.69/layer) | 393s |
| A2 — dev | 3.7% | **876.5** (11.69/layer) | 785s |

**Identical to the decimal.** The SLRU changes nothing here — not "a little", exactly nothing.

## Why, and why it isn't your bug

```
[RAM_GB=22.0] resident 10.6 GB + reserve 6.2 GB -> cap lowered 8->3
```

**Three slots per layer, eight experts wanted per token.** With three seats and eight arrivals, every eviction policy loses identically — there's no room left to be clever. Your algorithm isn't failing; it has nothing to work with.

You ran `RAM_GB=32` and reported baseline hit 9.4% → 15.1%. That's a cap around 8, where choosing the victim well actually changes who's resident next step. **I think your result is real and our box simply cannot reach the regime where it applies**: to get cap=8 here we'd need ~28 GB of budget on a 25.4 GB machine. So this is not a refutation — it's a boundary condition your PR doesn't state.

**The ask**: what's the lowest cap at which cloxcache starts to win? If it's ~6+, this is a good PR for large-memory hosts and should say so — the README line would be "on hosts with cache headroom" rather than a general claim. That's a much stronger PR than one that quietly does nothing for the low-RAM users this engine targets.

## The other half, which is the useful part for both of us

Look at the decode column: **251s, 393s, 785s**. A1 and A2 are the **same binary**, same prompt, same machine, twenty minutes apart — a **3.1× spread**.

Which means: had I run only A1→B, I'd have posted *"#223 is 56% slower, regression."* Had I run only A2→B, I'd have posted *"#223 is 2× faster, merging."* Both would have been garbage, and B's 393s sits neatly between two runs of the same code.

On a disk-bound box wall-clock is noise. The only stable signals are `experts loaded/token` and hit rate — and those are what I'd suggest you quote, because your own table leads with decode time (3,865s → 3,628s, -6%), which on our hardware is well inside the noise floor. Your hit-rate number is the honest one and it's the one doing the work.

One small thing while I'm here: the title says "nearly double the hit-rate", but 9.4% → 15.1% is **+60%**, not double. I'd retitle — the real result doesn't need the help, and this repo has had a rough week with numbers that promised more than they delivered (see #303).

## Where this leaves it

I'm not merging it yet, and I want to be straight about why: it adds real surface (`tier.h`, `TierAdapt`, per-layer `efreq` allocations) for a benefit measured at **zero** on the machine that represents this project's thesis. That's not a verdict on the work — it merges cleanly onto `dev` (only a README conflict, on an expert count: 75 layers × 256 + the MTP head = 19,456, so I kept ours and took your SLRU wording), builds with zero warnings, and passes the full suite including your `test_tier`.

It's a judgment call about scope, not correctness, and it's an easy yes the moment we know the cap threshold. If you can give me that number — or a run at a cap this class of hardware can actually reach — I'll take it.

Genuinely appreciate the rigour, and the fact that this is the safe kind of change: it touches eviction only, never which expert gets selected, so it cannot alter output. That distinction matters a lot around here lately.

KingIcyCreamProjects:
Following the ABA results — two things from a code-level read of the eviction path on dev that bear directly on the cap-threshold question:

1. **The engine currently can't attribute hits to the tier this PR changes.** `m->hits` counts a pin-tier hit and an ecache hit identically (both lookup branches bump the same counter; the PROF harness reports only the lumped value). At any given config the ecache could be serving ~0% of hits or a large share — which is exactly the ambiguity behind "what's the lowest cap at which cloxcache starts to win." I've split the counters on a branch (`hit_pin` / `hit_ecache`, surfaced in the serve stats line, REPLAY summary, and `[PROF]` report): https://github.com/KingIcyCreamProjects/colibri/tree/telemetry/pin-vs-ecache-hits — happy to send it as a PR. It would let the cap sweep show *why* a policy is flat (pin tier absorbing everything vs a genuine floor effect) before anyone spends another A/B on it.

2. **There's a third eviction site, Linux-only.** This PR converts the `moe()` promotion and `pilot_realload` scans, but `pilot_uring_batch` (`c/glm.c:3472` on dev, `#ifdef __linux__`) still runs the old min-`used` scan — so on a `URING=1` Linux host the uring pilot loader would evict under a different policy than the two converted sites. No-op on non-uring builds, but worth converting for consistency since the claim is about the array's global eviction order.

withinboredom:
> The ask: what's the lowest cap at which cloxcache starts to win?

The honest answer is a ratio, not a cap: it starts winning when cap exceeds roughly a third of the per-token unique expert demand per layer, peaks around 0.4–0.8×, and thins out again above 1× (where plain recency already holds the working set). Your box makes the arithmetic concrete: 876.5/75 ≈ 11.7 loads/layer/token at 3.8% hit ⇒ ~12 unique experts demanded per layer per token, against 3 slots. That's a 4× full flush every token — no resident survives to the next token under any policy, so "identical to the decimal" is the expected outcome, not a near-miss. You're right that it isn't my bug, and right that it isn't your refutation: cap 3 against demand 12 is ratio 0.25, below the floor. My RAM_GB=32 number sits at cap 8 against the same demand — ratio 0.66, mid-sweet-spot. Both measurements are the same curve.

And you don't have to take the curve from me, because:

> On a disk-bound box wall-clock is noise.

— is correct, and I won't argue with it. Since the PR I built an offline replayer that removes the hardware from the measurement entirely, and it's on the branch: capture a routing trace during any normal run (one env var, one line per routing decision, negligible overhead), then replay the identical request stream through plain LRU, this PR's SLRU, and Belady's clairvoyant OPT:

ROUTE_TRACE=run.trace ./glm <your usual args>
make tools/route_sim
for c in 3 4 5 6 8 12; do tools/route_sim -c $c -p 0 run.trace; done

Run any cap, including the ones your RAM can't reach. It also scores each eviction decision against hindsight (was the victim re-requested before another resident would have been?), which is the metric hit rate can't see. For calibration, my traced workload (256 experts/layer, demand ≈ 8/layer/token, -p 0 matching your AUTOPIN=0):

```
┌─────┬────────────┬──────────┬───────────┬─────────────┐
│ cap │ cap/demand │ LRU hit% │ SLRU hit% │ OPT ceiling │
├─────┼────────────┼──────────┼───────────┼─────────────┤
│ 3   │ 0.38       │ 8.0      │ 9.4       │ 25.6        │
├─────┼────────────┼──────────┼───────────┼─────────────┤
│ 4   │ 0.50       │ 11.8     │ 13.6      │ 31.8        │
├─────┼────────────┼──────────┼───────────┼─────────────┤
│ 5   │ 0.63       │ 16.3     │ 18.5      │ 36.8        │
├─────┼────────────┼──────────┼───────────┼─────────────┤
│ 6   │ 0.75       │ 20.8     │ 22.8      │ 41.0        │
├─────┼────────────┼──────────┼───────────┼─────────────┤
│ 8   │ 1.00       │ 31.5     │ 31.7      │ 47.1        │
└─────┴────────────┴──────────┴───────────┴─────────────┘
```

By your scaling (~2 GB/cap between RAM_GB=22→cap 3 and 32→cap 8), cap 4–5 is RAM_GB ≈ 24–26, so cap 4 may even be physically reachable on the 25.4 GB box if you want confirmation — but the simulator will tell you the answer for your routing distribution either way. The OPT column is worth a look independently of this PR: it prices exactly what each additional GB of cache budget buys on your workload before anyone tunes another heuristic.

> the README line would be "on hosts with cache headroom" rather than a general claim

Agreed, and I'd go further — proposed wording: "SLRU eviction improves expert-cache hit rate when per-layer cache capacity exceeds ~⅓ of per-token expert demand (measured +60% hit rate at cap/demand ≈ 0.65); below that threshold the cache fully flushes each token and the policy is provably inert — identical residency to LRU, zero cost, zero risk." The inertness your A/B demonstrated is worth stating as the safety property it is.

> I'd retitle — the real result doesn't need the help

Also agreed, no defense offered. "+60% hit rate" it is.

One relevant note: the branch has moved since you tested. Building the simulator caught two real defects in my own implementation (ghost-eviction ordering that silently destroyed ghost memory under exactly the LRU-victim pattern; adaptation windows mis-paced so the hit-rate feedback learned from noise) and added eviction-driven frequency decay, which the replayer shows matters most at small caps — the regime you care about. All verified against OPT on trace replay; test_tier covers each. Worth retesting from head rather than the revision you benchmarked — though at cap 3, I expect and now predict on the record that you'll still measure exactly nothing.

Happy to have the merge gate on the documented threshold; it's the right bar. Run the replay on your trace and the number in the README can be yours rather than mine.

JustVugg:

JustVugg:
@withinboredom heads-up on the rebase: I tried to rebase this onto current `dev` for you (we're taking the #391 refactor's rebase cost off contributors where we can), but this one has **semantic** conflicts I don't want to resolve on your behalf — it overlaps the DISK-CLASS recency/frequency tracking that just landed in `dev` (#438). Both your hit-rate cache state (`efreq`, `TierAdapt`) and DISK-CLASS's private recency state touch the same Model struct and eviction path, so merging them correctly is a judgment call about how the two cache-tracking approaches should coexist — and I can't measure the hit-rate impact without the full model. Could you rebase and reconcile with the DISK-CLASS state? Note `glm.c` is now `colibri.c` + header modules (your `tier.h` slots in cleanly). Happy to help place the mechanical hunks once you've made the semantic call. Sorry for the churn.

JustVugg:
Still keen on this — the hit-rate win is real. It's been in conflict for a while though, and `dev` has moved a lot (the #391 refactor split the engine into colibri.c + headers, and cache-adjacent code has churned). Are you still able to rebase? If not, tell me and I'll take the rebase over myself — I'd rather carry it than lose the improvement.

JustVugg:
Still want this — the hit-rate win is real. To merge:

**1. Rebase** on current `dev` (CI green; cache-adjacent code churned after #391 and #474's LFRU eviction guard).

**2. Prove it doesn't change outputs.** This changes cache behaviour, so I need a token-exact run (tiny model, `TF=1` → 32/32) alongside the measured hit-rate delta on a real streaming run. If it composes with #474's eviction guard, even better. Rebase + those two numbers and it lands. Happy to help if the conflict is mechanical. Thanks @withinboredom.

JustVugg:
Status: this conflicts with `dev`, and the reason is structural rather than a normal rebase — the PR modifies `c/glm.c`, which **no longer exists**. #391 renamed the engine to `c/colibri.c` and split parts into header modules; `make glm` survives only as an alias. So this is a forward port onto a moved file, not a `git rebase` that resolves cleanly.

I don't want to lose the work: "nearly double the hit-rate" is exactly the axis that matters most here, since expert bytes per token are the wall on every disk-bound host. `c/tier.h` — where most of the substance lives — is still `c/tier.h` on dev, so the tier logic should port with far less friction than the `glm.c` hunks suggest.

Two things that would help me help you: (1) are the measurements in the PR still reproducible on current `dev`? The cache and eviction paths have changed since (a measured-RSS guard landed in #403, and the pilot eviction guard was fixed in #497), so the baseline you compared against has moved. (2) If you'd rather not carry the forward port yourself, say so and I'll look at porting the `tier.h` core with credit to you, rather than leaving it open indefinitely.

Either way, tell me if you're still on it — I'd rather ask than silently let a good PR rot.

JustVugg:

JustVugg:

bherald:
@withinboredom @JustVugg — I forward-ported the CLOCK-LRU-K policy primitives and route simulator to current `dev` (`7fb1159`) while preserving Robert’s authorship credit. I also added exact global pin replay from a frozen `.coli_usage` history, call-window filters for separating prefill from decode, and exact miss-reduction reporting.

Before changing the runtime path, I tested the simulator on a disk-bound Windows host: Ryzen 9 7845HX, GLM-5.2, CPU-only, `DIRECT=1`, `PIPE=1`, `PILOT_REAL=1`, `RAM_GB=32`, 37 globally ranked pins, and cap 4 per layer. The frozen 24-token replay produced 20,228 unpinned probes.

Results:

- Full trace: LRU 18,794 misses versus CLOCK-LRU-K 18,506, a 1.53% reduction.
- Decode-only: LRU 12,979 misses versus CLOCK-LRU-K 12,690, a 2.23% reduction.
- Decay cadences 4, 8, 16, 32, and 64 did not reach the predeclared 3% promotion gate.
- The corresponding runtime baseline was 0.51 tok/s, 21.9% total expert hit rate, and 254.241 GB fetched over 24 tokens.

Because the result was positive but below the gate, I stopped before modifying the current demand, PILOT, io_uring, or RSS-guard cache paths. The simulator/tooling branch passes the complete current check suite: all C tests and 289 Python tests passed, with 22 expected skips.

Proposed branch: https://github.com/bherald/colibri/tree/perf/cache-clox-rebase

Would you prefer these simulator and evidence commits folded into #223, or kept as reproducible evidence while the runtime PR remains unchanged?

withinboredom:
@bherald -- I'm just returning back from some vacation. Can you try pinning k to `2` instead of letting it adapt? For this workload, adaption may not be ideal -- adaption is meant for varying workloads (skew, etc) while this is a mostly static workload.

bherald:
Yes, I'll take a look. Thank you.

bherald:
@withinboredom — I ran the requested fixed `k=2` control against the same frozen Windows trace and exact global-pin history, then repeated the replay after forward-porting the simulator to current `dev` (`a54b4da`).

Inputs:

- Trace SHA-256: `d0a68d0bf0e5fe78c6c282e013adf1ed07c0818fc77522e91a66bdec7a2ce1c2`
- Pin-history SHA-256: `2a4273e54a011f737ae04b2ed2a32f270aab6fbc13bb6fd4b1c7c61680e17f4d`
- Cap 4 per layer, 37 exact global pins, decay every 8
- Full window: 20,228 probes
- Decode window (`--min-call 78`): 14,330 probes

| Window | LRU misses | Adaptive misses | Adaptive reduction | Fixed k=2 misses | Fixed reduction |
| Full | 18,794 | 18,506 | 1.53% | 18,669 | 0.67% |
| Decode | 12,979 | 12,690 | 2.23% | 12,879 | 0.77% |

Adaptive converged to an average final `k` of 1.01 on the full trace and 1.00 during decode. Freezing it at 2 therefore underperformed adaptive on both windows, despite this being a mostly static workload. Both variants remain below the predeclared 3% promotion gate, so I am not proposing a runtime-path change from this trace.

The refreshed simulator/evidence branch passes all native gates plus 340 Python tests with 30 expected skips:

https://github.com/bherald/colibri/tree/perf/cache-clox-fixed-k2

The content-free replay receipt SHA-256 is `538d7c27d7cf511821c12e5347953e546f79ada1dfcc27452181a3fea9f7564c`.

JustVugg:
@withinboredom honest supersession question on the oldest PR in the queue: since this was opened, usage-decay (#780) and the pin/cap budget rework (#869, #914) landed, both in the same hit-rate territory, and the branch now conflicts. Is the mechanism here still additive on top of those, or did they eat the win? If you believe it's still additive, a rebase + one fresh measurement against current dev would restart the review; otherwise I'd close with credit — the idea clearly pushed in the right direction.

JustVugg:
Closing the oldest PR in the queue — but not silently, because there is real work in here.

Since it opened, usage-decay (#780) and the pin/cap budget rework (#869, #914) landed in the same hit-rate territory, and this branch now conflicts across exactly the regions they touched. What I don't want buried is @bherald's replay work in this thread: the fixed-`k=2` control against the frozen Windows trace and the exact global-pin history, re-run after forward-porting the simulator, is a better-controlled evaluation than most *merged* changes in this repo get.

If the adaptive-`k` mechanism is still additive on top of decay plus the new budgets, reopen or re-cut against current `dev` and I'll take it straight into review — that's a genuine offer, not a formality.

PR #232, opened 2026-07-15
Profiling: PROF=1 opt-in performance profile + live per-turn Profiling page in the web dashboard

nbeerbower:
Adds an opt-in profiling stack, additive only: with `PROF` unset every mode's output is byte-identical. Four linear commits:

1. **`PROF=1` opt-in performance profile** — answers "where does it slow down on THIS machine with THIS config": startup header (machine + effective knobs, so saved logs are self-describing), per-forward decode latency percentiles (p50/p90/p99/max) with a tail diagnosis, expert I/O accounting at the pread/mmap-touch sites (GB fetched, MB/token, hit rate, tier fill), phase shares of wall time, and a plain-language verdict naming the knob most likely to move tok/s.

2. **PROF in mux serve mode** — `run_serve_mux` (SERVE_BATCH, the `coli web` path) completes requests in `mux_done`, so the per-turn hook never fired there; snapshot the window where `hits0` is taken and report on stderr at DONE (stdout stays the framed protocol).

3. **Profiling page in the web dashboard** — `glm` emits a per-turn `PROF` protocol line next to TIERS/HITS, `openai_server.py` keeps a 120-turn rolling window served read-only at `/profile`, and a new Profiling tab renders stat tiles, wall-time composition bars, per-turn stacked phase columns with hover readouts, and a recent-turns table. Disk service is drawn apart from the stack: it overlaps compute, so only the I/O wait the compute thread felt counts inside wall time. Phase colours are CVD-validated with legend + table so identity never rides on colour alone.

4. **Phase-accounting fix** — 'Other' no longer swallows the disk stall.

Validated on a tiny-random unquantized fixture (REPLAY with PROF on/off; a RAM_GB squeeze flips hit rate 96.9%->26.6% and the verdict follows). `make check` 64 tests green, vitest 18 green, tsc clean. Field note: the page's verdict on a 187 GB RAM + RTX A6000 box correctly identified an 84% expert-I/O-wait bottleneck per turn.

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,9 +540,10 @@ What you get:
 
 - **Chat** with live metrics: a flashing token counter while generating, then tok/s, time-to-first-token, prompt→completion counts and queue wait;
 - **Runtime panel**: your hardware (CPU, GPUs + VRAM, RAM, cores), the scheduler, and the live expert-tier bar — how many of the 19,456 experts sit in VRAM / RAM / disk right now;
-- **Brain**: the whole model as a 76×256 cortex, one cell per expert. Colour = tier, brightness = routing heat, and the experts routed in each turn flash white and decay — you watch the model think. Hover any cell for its tier, heat and [measured topic affinity](https://github.com/JustVugg/colibri/issues/175) (specialists for code, Chinese, math, law… live in layers 11–22).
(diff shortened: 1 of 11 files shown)


jazir555:
You are a G, ballin'.

PR #235, opened 2026-07-15
feat: implement I/O-aware RAM fallback router and native string-indexed memory

dannythevibe:
This PR introduces two major performance features:

1. **I/O-Aware RAM Router:** Detects if the model is running on a slow HDD and actively tracks disk latency during the generation loop. If the disk starts bottlenecking, the router falls back to picking experts exclusively from the RAM-resident pool to maintain steady generation speeds.

2. **Native Local Memory:** Implemented a lightweight MemoryManager in the API server that logs conversation turns to a local text file and uses fast literal string matching to pull relevant historical context into the system prompt, entirely bypassing the need for heavy RAG databases.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -159,8 +159,27 @@ def resource_request(a, env):
         [int(value) for value in gpu.split(",")])
     return ram,ctx,devices,vram
 
+def detect_slow_drive(model_path):
+    if os.environ.get("COLI_FORCE_NVME") == "1": return False
+    if os.environ.get("COLI_FORCE_HDD") == "1": return True
+    try:
+        if sys.platform == "win32":
+            out = subprocess.check_output(["powershell", "-NoProfile", "-Command", "Get-PhysicalDisk | Select-Object -ExpandProperty MediaType"], text=True, timeout=2).strip().lower()
+            return "hdd" in out
+        elif sys.platform.startswith("linux"):
+            out = subprocess.check_output(["lsblk", "-ndo", "ROTA", model_path], text=True, timeout=2).strip()
+            return out == "1"
+    except Exception:
+        pass
+    return False
+
 def env_for(a):
     e = dict(os.environ, SNAP=a.model)
+    if detect_slow_drive(a.model):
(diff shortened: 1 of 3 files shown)


JustVugg:

**It's 121 commits behind `dev`.** On `glm.c`, which changes several times a day, that isn't a rebase — it's a rewrite, and doing it *for* you would mean rewriting your design decisions without understanding why you made them. So I'd rather ask.

The RAM/budget path is exactly where this week's OOM work landed (cap_for_ram now refuses to start when the projected peak exceeds available memory). A rebased version will collide with that, so it's worth checking your router still says something the new budget code doesn't.

JustVugg:
Closing this — 171 commits behind, and it lands right on top of code that was substantially rewritten in the meantime.

The RAM/fallback area specifically: `cap_for_ram` now auto-sizes from measured available memory and refuses to start when the projected peak would OOM (the #305/#341 work), and the expert cache sizing changed with the budget/ws_b reconciliation. An I/O-aware router touching `glm.c` + `coli` + `openai_server.py` from 171 commits back can't be reconciled with that without a full rewrite.

The idea — routing that accounts for I/O cost — is a live and interesting one (it's adjacent to the PILOT/prefetch and cache-policy discussions in #333/#362). If you want to bring it back against current `dev`, I'd suggest the smallest version that does one measurable thing, with the `experts loaded/token` counter showing the win — happy to review that. Thanks for the contribution.

dannythevibe:
Thanks for the detailed breakdown and for taking the time to review it. I completely understand why it can't be merged as-is, especially with the recent memory management overhaul. No hard feelings at all

I'm glad to hear the core idea of I/O-aware routing is still of interest. I'll take a look at the current dev branch and see if I can carve out a minimal, focused PR that demonstrates the win with the new budget code in place. The new CI setup sounds fantastic and should definitely make testing this much easier.

I'll open a new PR when I have a smaller, measurable version ready for review. Thanks again for the transparent communication.

Issue #239, opened 2026-07-15
FP8 download tool: dual-source (ModelScope + HF), parallel, resume — for the 379 GB GLM-5.2-FP8 checkpoint


woolcoxm:
## TL;DR

New `download_fp8.py` tool for downloading the GLM-5.2-FP8 checkpoint (379 GB, 141 shards) with **dual-source fallback** (ModelScope primary, HuggingFace fallback), parallel shard download, stall recovery, and clean resume. Designed for the bandwidth-constrained, throttle-prone reality of downloading a 379 GB model.

Branch: `windows-dev` (commit `91b1035`)

---

## Why

The production int4 conversion requires the FP8 source checkpoint. At 379 GB across 141 shards, the download is the **multi-day bottleneck** of the entire pipeline. Two problems with the default HuggingFace download path:

1. **HuggingFace per-stream throttling**: HF rate-limits individual streams after ~50 GB, dropping throughput to a crawl. There's no documented workaround for unauthenticated bulk downloads.
2. **No clean resume**: `hf_hub_download` resumes by file, but a mid-shard interruption leaves a partial file that may or may not be revalidated correctly, and there's no visibility into which shards are complete.

ModelScope mirrors the same checkpoint (`ZhipuAI/GLM-5.2-FP8`) on a CDN that doesn't apply the same per-stream throttling, giving 3-5x higher sustained throughput in practice.

## What it does

- **Dual-source**: tries ModelScope first (`ZhipuAI/GLM-5.2-FP8`); if the API is unreachable, falls back to HuggingFace (`zai-org/GLM-5.2-FP8`). Force a source with `--source ms` or `--source hf`.
- **Parallel download**: N worker threads (default 3) pull shards concurrently from a shared work queue.
- **Resume by size**: skips any shard whose local file size matches the expected size. Partial shards are re-downloaded from scratch (shards are 5 GB each, so re-downloading one is cheap).
- **Stall recovery**: 3 retry attempts per shard with exponential backoff; curl fallback (`-C -` for byte-range resume) if the Python download path fails.
- **Progress display**: per-shard completion with running percentage, elapsed time, and ETA.

## Usage

```bash
# Auto (ModelScope first, HF fallback), 3 parallel
python download_fp8.py

# 4 parallel, force ModelScope
python download_fp8.py --parallel 4 --source ms

# Force HuggingFace
python download_fp8.py --source hf
```

Re-running resumes — completed shards are skipped by size, only missing/partial ones are re-downloaded.

[...]

PR #242, opened 2026-07-15
Group-scaled int4 (fmt=4): one scale per 128 elements — fixes incoherent output

### LINKED ISSUE #225: Root cause of incoherent output: per-row int4 scales — fix with group-scaled quantization (fmt=4)
## TL;DR

The int4 model produces **fluent-but-incoherent output** — grammatical English that is completely off-topic (SEO spam, homework-help boilerplate) instead of reasoned responses. The root cause is **per-row quantization scales**: one F32 scale per output row (e.g. 2048 scales for a 2048×6144 matrix), which is 48x coarser than the FP8 source's 128×128 block scales. This destroys fine-grained weight information that handles reasoning and instruction-following while preserving the large-magnitude weights that handle grammar and vocabulary fluency.

Branch: `experiment/grouped-quant` (based on latest `dev` at `62419af`)

---

## The problem — demonstrated

Prompt: *"Explain the concept of recursion in programming. Provide a simple example."*

**Baseline (no budget, no changes):**
> "Hire Some To Take Object-Oriented Programming Assignment"

**Budget=4:**
> "The world is a dangerous place, not so much because of the small percentage of people who are doing to do the little of people who are doing to"

**Budget=6:**
> "Elite Custom Essays"

**Budget=12:**
> "Sololearn: Learn to code for FREE! +1 # Explain A concept of recursion..."

This is not random gibberish — it's fluent English that is completely off-topic. This is the signature of **activations corrupted by coarse quantization**: the model retains language fluency (large weights survive) but loses instruction-following and reasoning (fine-grained weights are crushed to zero).

### How the current converter works

`convert_fp8_to_int4.py` (line 39-52) quantizes with **one scale per output row**:

```python
amax = np.abs(w).max(axis=1, keepdims=True)    # one max per ROW
s = np.maximum(amax / qmax, 1e-8)               # one scale per ROW
q = np.clip(np.rint(w / s), -8, qmax)            # quantize entire row with that scale
```

For a 2048×6144 expert weight matrix, that's **2048 scales** — one per row, each covering 6144 elements.

### Why this destroys quality

Consider a row of 6144 values where most are small (magnitude ~0.01) but a few are large (~0.5). The per-row scale is `0.5/7 = 0.071`. The small values become `0.01/0.071 = 0.14`, which rounds to **zero**. All fine-grained information in those small weights is lost.

This matters enormously for MoE models: each token activates only 8 of 256 experts. A poorly-quantized expert pollutes every token that routes to it, and there's no averaging from the other 248 experts (they're simply off). The error is **concentrated**, not diluted.

### What the FP8 source does right

The FP8 checkpoint uses **128×128 block scales** — the weight matrix is divided into 128-element chunks, each with its own scale. Small values in one chunk don't get crushed by large values in another. For a 2048×6144 matrix: `2048 × 48 = 98,304` scales — 48x more granularity.

The converter already dequants FP8 to f32 correctly (lines 196-201), preserving the block-scale information. But then it **throws it all away** by collapsing to a single per-row scale during int4 requantization.

### GLM-5.2 was QAT-trained for int4

The GLM-5 paper (arXiv 2602.15763, §2.4.3) states: *"To provide better accuracy at low-precision, we apply INT4 QAT in the SFT stage."* This means the model is **designed** to work at int4 — but only if the quantization is fine-grained enough to match what the model saw during training. The FP8 checkpoint's 128×128 block scales are the granularity the model expects. Per-row scaling is 48x coarser.

### Community confirmation

Every other project getting coherent GLM-5.2 int4 output uses calibrated or fine-grained quantization:
- **ubergarm/GLM-5.1-GGUF** — imatrix-calibrated with expert-specific patches
- **Unsloth/GLM-5-GGUF** — dynamic UD-Q4 quants
- **QuantTrio/GLM-5.2-Int4-Int8Mix** — mixed int4/int8 with channel-wise scales
- **llama.cpp** — Q4_K with block-level group scales (typically 32 or 64 elements)

None use naive per-row RTN int4 for production inference.

## The fix: group-scaled int4 (fmt=4)

Add a new quantization format with **one scale per 128 elements** along the input dimension, matching the FP8 source's natural granularity.

### What changed

**`c/glm.c`** — 122 lines added:

1. **QT struct** (line 98): Added `int gs` field (group size, 0=per-row for backward compat, 128=grouped).

2. **Format detection** (`qt_from_disk`, line 1064): Auto-detects fmt=4 by checking the `.qs` scale array size. If it has `O * ceil(I/128)` elements instead of `O`, it's grouped. **Old per-row models (fmt=2) work unchanged.**

3. **New kernel** (`matmul_i4_grouped`, line 379): Same AVX2 nibble unpacking as `matmul_i4`, but the accumulator resets at each 128-element group boundary: `dot(x[grp], w[grp]) * scale[grp]`. The scale changes every 8 vector iterations (128/16=8).

4. **Dispatch** (`matmul_qt_ex`, line 813): Routes to `matmul_i4_grouped` when `fmt==4`. Always uses exact kernels (no IDOT approximation — the whole point is quality).

5. **Expert loading** (`expert_load`, lines 1418 and 1538): Both the mmap and slab+pread paths detect fmt=4 from scale array size and set `gs=128`.

6. **`qt_bytes`** (line 104): Reports correct memory for fmt=4.

**`c/tools/convert_fp8_to_int4.py`** — `quant_int4_grouped()` + `--group-size` arg:

```python
def quant_int4_grouped(w, bits, gs=128):
    O, I = w.shape
    ngroups = (I + gs - 1) // gs
    wpad = np.zeros((O, ngroups * gs), np.float32)
    wpad[:, :I] = w
    wr = wpad.reshape(O, ngroups, gs)              # [O, ngroups, gs]
    amax = np.abs(wr).max(axis=2, keepdims=True)   # one max per GROUP
    s = np.maximum(amax / qmax, 1e-8)
    q = np.clip(np.rint(wr / s), -8, qmax)
    # ... same nibble packing as quant_int4 ...
    return packed_nibbles, s.reshape(-1)            # [O * ngroups] scales
```

Same packed-nibble format as existing int4 — only the scale array is larger.

### Verification

**Converter round-trip test** (weights with varying group magnitudes):

| Method | Mean relative error | Max abs error |
|--------|-------------------|---------------|
| Per-row int4 (current) | 0.2056 | 0.0127 |
| Grouped int4 (gs=128) | **0.1278** | 0.0127 |
| **Improvement** | **1.6x lower** | — |

**Engine kernel test** (AVX2 `matmul_i4_grouped` vs f32 reference):

| Output | Reference | Kernel | Error |
|--------|-----------|--------|-------|
| o=0 | -0.126368 | -0.126368 | 2.98e-08 |
| o=1 | 0.075913 | 0.075913 | 1.49e-08 |
| o=2 | 0.077136 | 0.077136 | 7.45e-09 |
| ... | ... | ... | ... |

Max error: **2.98e-08** — matches f32 reference to within float32 epsilon. **PASS.**

### Cost

| Metric | Per-row (current) | Grouped (new) |
|--------|-------------------|---------------|
| Expert weight size | ~18.9 MB | ~20.1 MB (+6%) |
| Scale array per matrix | O × 4 bytes | O × ceil(I/128) × 4 bytes |
| Total model size | ~370 GB | ~390 GB (+5%) |
| Disk I/O per miss | ~19 MB | ~20 MB (+6%) |
| Kernel speed | One scale multiply per row | One scale multiply per 128 elements |

The 6% I/O increase is negligible compared to the quality gain. The grouped kernel has slightly more scale-lookup overhead but remains within AVX2 throughput — the bottleneck is disk I/O, not matmul.

### Backward compatibility

- Old per-row models (fmt=2) **work unchanged** — format auto-detected from scale array size
- All existing features (PILOT, EXPERT_BUDGET, CUDA, MTP, PIPE) work identically — they don't touch the dequant path
- The fused gate+up pair path (`matmul_i4_pair`) falls back to separate `matmul_qt` calls for fmt=4 — minor perf cost, correctness preserved
- `--group-size 0` produces per-row output (backward compat for the converter)

### Convert with group scales

```bash
python tools/convert_fp8_to_int4.py \
  --indir /path/to/GLM-5.2-FP8 \
  --outdir /path/to/glm52_i4_grouped \
  --ebits 4 --io-bits 8 --group-size 128
```

### Test quality

```bash
# Old model (per-row):
SNAP=/path/to/glm52_i4 PROMPT="Explain recursion in programming." NGEN=32 ./glm 64

# New model (grouped):
SNAP=/path/to/glm52_i4_grouped PROMPT="Explain recursion in programming." NGEN=32 ./glm 64
```

Compare output text coherence.

## What this won't fix

- **Cold cache speed** — still 0.18-0.36 tok/s on 24 GB RAM. Quality and speed are independent axes.
- **All quantization error** — int4 is still 4-bit. Some degradation vs FP8 will remain. But it should be coherent degradation (slightly wrong word choices) rather than total reasoning failure (unrelated topics).
- **The IDOT +12% perplexity** — the approximate activation kernel (`IDOT=1`, default ON) still applies to non-grouped tensors (attention projections are already protected). Grouped int4 uses exact kernels by design.

## Prior art

- **GLM-5 paper** (arXiv 2602.15763, §2.4.3): "We apply INT4 QAT in the SFT stage" — the model is trained for int4, but assumes fine-grained quantization.
- **MxMoE** (ICML 2025): MoE experts exhibit divergent quantization sensitivity; uniform quant across all experts is suboptimal.
- **MoEQuant** (OpenReview): Naive int4 on MoE loses meaningful accuracy; calibrated framework needed.
- **Automated Fine-Grained MoE Quantization** (ACL 2025): Layer/expert-wise sensitivity variation; group-size scaling is the baseline improvement.
- **ubergarm/GLM-5.1-GGUF**: imatrix-calibrated quants with expert-specific patches — explicitly patches `quantize_row_q4_0_ref()` for routed experts.
- **llama.cpp Q4_K**: Uses block-level group scales (32 or 64 elements) as the standard int4 format.

## Conversion status

Re-converting from the FP8 source (`zai-org/GLM-5.2-FP8`) with `--group-size 128`. Downloading via ModelScope to avoid HuggingFace per-stream throttling. Will report quality A/B results once the conversion is complete.

woolcoxm:
## Summary

Adds a new int4 quantization format (**fmt=4**) with **one scale per 128 elements** along the input dimension, instead of one scale per output row. This matches the FP8 source checkpoint's 128×128 block-scale granularity and fixes the root cause of incoherent int4 output.

Closes #225.

## Problem

The int4 model produces fluent-but-incoherent output — grammatical English that is completely off-topic (SEO spam, homework-help boilerplate). The root cause is **per-row quantization scales**: one F32 scale per output row (e.g. 2048 scales for a 2048×6144 matrix), which is 48× coarser than the FP8 source's 128×128 block scales. This destroys fine-grained weight information that handles reasoning and instruction-following while preserving the large-magnitude weights that handle grammar and vocabulary fluency.

GLM-5.2 was QAT-trained for int4 (arXiv 2602.15763, §2.4.3), but only with fine-grained quantization matching what the model saw during training. Per-row scaling is 48× too coarse.

## Changes

**`c/glm.c`** (+122 lines):
- `QT` struct: added `int gs` field (group size, 0=per-row for backward compat, 128=grouped)
- `matmul_i4_grouped` kernel: AVX2 nibble unpacking, accumulator resets at each 128-element group boundary — verified to 2.98e-08 vs f32 reference
- Format detection (`qt_from_disk`): auto-detects fmt=4 from `.qs` scale array size. Old per-row models (fmt=2) work unchanged.
- Expert loading: both mmap and slab+pread paths handle fmt=4
- `qt_bytes`: fmt=4 case added
- Dispatch (`matmul_qt_ex`): routes to `matmul_i4_grouped` when `fmt==4`, always exact (no IDOT)

**`c/tools/convert_fp8_to_int4.py`** (+52 lines):
- New `quant_int4_grouped()` function + `--group-size` arg (0=per-row default, 128=grouped)
- Same packed-nibble format as existing int4 — only the scale array is larger

## Validation

| Metric | Per-row (current) | Grouped (new) |
|--------|-------------------|---------------|
| Mean relative error | 0.155 | **0.128** |
| Max abs error | 0.0127 | 0.0127 |
| Improvement | — | **1.14-1.22× lower** |
| Kernel vs f32 | — | 2.98e-08 (PASS) |
| Cost (model size) | ~370 GB | ~390 GB (+5%) |

## Backward compatibility

- Old per-row models (fmt=2) work unchanged — format auto-detected from scale array size
- All existing features (PILOT, EXPERT_BUDGET, CUDA, MTP, PIPE) work identically
- `--group-size 0` produces per-row output (converter backward compat)

## Test plan

- [x] Converter round-trip: grouped vs per-row dequant error measured
- [x] Engine kernel: AVX2 `matmul_i4_grouped` matches f32 reference to 2.98e-08
- [x] Engine load: both fmt=2 and fmt=4 models load clean, format auto-detected
- [ ] Production quality A/B: grouped vs per-row on the real 372 GB model (requires FP8 download to complete, #239)

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -96,7 +96,7 @@ typedef struct {
  * INT4 e' cio' che fa stare la densa residente nei 15 GB (0.5 byte/param). */
 /* fmt: 0 F32, 1 INT8, 2 INT4 (2/byte), 3 INT2 (4/byte). q4 ospita sia int4 che int2 packed. */
 typedef struct {
-    int fmt; float *qf; int8_t *q8; uint8_t *q4; float *s; int O, I;
+    int fmt; float *qf; int8_t *q8; uint8_t *q4; float *s; int O, I, gs;  /* gs=group size (0=per-row, 128=grouped) */
 #ifdef COLI_CUDA
     ColiCudaTensor *cuda;
 #endif
@@ -107,7 +107,10 @@ static int64_t qt_bytes(const QT *t){    /* byte residenti del tensore */
     if(t->fmt==0) return n*4;
     if(t->fmt==1) return n + (int64_t)t->O*4;
     if(t->fmt==3) return (int64_t)t->O*((t->I+3)/4) + (int64_t)t->O*4;
-    return (int64_t)t->O*((t->I+1)/2) + (int64_t)t->O*4;
+    if(t->fmt==4){ /* int4 grouped: packed nibbles + O*ceil(I/gs) scales */
+        int ng=(t->I+t->gs-1)/t->gs;
+        return (int64_t)t->O*((t->I+1)/2) + (int64_t)t->O*ng*4; }
+    return (int64_t)t->O*((t->I+1)/2) + (int64_t)t->O*4;  /* fmt=2 int4 per-row */
 }
 
 typedef struct {
@@ -373,6 +376,46 @@ static void matmul_i4(float *y, const float *x, const uint8_t *q4, const float *
                 a += xs[i]*(float)lo + xs[i+1]*(float)hi; }
             if(i<I){ uint8_t byte=w[i>>1]; int lo=(int)(byte&0xF)-8; a += xs[i]*(float)lo; }
             y[(int64_t)s*O+o]=a*sc; } }
+}
+/* y[S,O] = x[S,I] @ W^T with W int4 packed (2/byte) + per-GROUP scales (fmt=4).
+ * Same nibble math as matmul_i4, but the scale changes every `gs` elements along I.
+ * The accumulator resets at each group boundary: dot(x[grp], w[grp]) * scale[grp].
+ * gs MUST be a multiple of 16 (the AVX2 vector width). */
+static void matmul_i4_grouped(float *y, const float *x, const uint8_t *q4, const float *scale,
+                              int S, int I, int O, int gs){
+    int rb=(I+1)/2; int ng=(I+gs-1)/gs;
+    #pragma omp parallel for schedule(static)
+    for(int o=0;o<O;o++){
+        const uint8_t *w=q4+(int64_t)o*rb;
+        const float *scl=scale+(int64_t)o*ng;
+        for(int s=0;s<S;s++){
+            const float *xs=x+(int64_t)s*I; float a=0;
+            for(int g=0; g*gs<I; g++){
+                int base=g*gs; int glen=gs; if(base+glen>I) glen=I-base;
+                float sc=scl[g];
+                int i=base;
+#ifdef __AVX2__
+                const __m128i m4=_mm_set1_epi8(0x0F); const __m256i b8=_mm256_set1_epi32(8);
+                __m256 acc=_mm256_setzero_ps();
+                for(; i+16<=base+glen; i+=16){ __m128i by=_mm_loadl_epi64((const __m128i*)(w+(i>>1)));
+                    __m128i lo=_mm_and_si128(by,m4),hi=_mm_and_si128(_mm_srli_epi16(by,4),m4);
+                    __m128i nib=_mm_unpacklo_epi8(lo,hi);
+                    __m256 w0=_mm256_cvtepi32_ps(_mm256_sub_epi32(_mm256_cvtepu8_epi32(nib),b8));
+                    __m256 w1=_mm256_cvtepi32_ps(_mm256_sub_epi32(_mm256_cvtepu8_epi32(_mm_srli_si128(nib,8)),b8));
+                    acc=_mm256_fmadd_ps(_mm256_loadu_ps(xs+i),   w0, acc);
+                    acc=_mm256_fmadd_ps(_mm256_loadu_ps(xs+i+8), w1, acc); }
+                a+=hsum256(acc)*sc;
+#endif
+                /* scalar tail for the group remainder */
+                for(; i<base+glen; i+=2){
+                    if(i+1<base+glen){ uint8_t byte=w[i>>1];
+                        a+=(xs[i]*(float)((int)(byte&0xF)-8)+xs[i+1]*(float)((int)(byte>>4)-8))*sc; }
+                    else { uint8_t byte=w[i>>1]; a+=xs[i]*(float)((int)(byte&0xF)-8)*sc; }
+                }
+            }
+            y[(int64_t)s*O+o]=a;
+        }
+    }
 }
 /* Decode hot path for gate+up: same exact q4 dot products as matmul_i4, but one
  * OpenMP dispatch covers both matrices. KTransformers uses persistent pools;
@@ -766,6 +809,9 @@ static void matmul_qt_ex(float *y, const float *x, QT *w, int S, int allow_idot)
     }
 #endif
     if(w->fmt==0){ matmul(y,x,w->qf,S,w->I,w->O); return; }
+    /* fmt=4: grouped int4 — always use the exact grouped kernel (no IDOT approximation,
+     * since the whole point of grouped scales is better quality). */
+    if(w->fmt==4){ matmul_i4_grouped(y,x,w->q4,w->s,S,w->I,w->O,w->gs); return; }
     /* int8 IDOT vince sempre (1.4-2.5x). int4 IDOT: l'autore su AVX2 trovo' che a S=1
      * non ripaga (soglia S>=2); ma su ARM/SDOT il singolo token CONVIENE (vedi g_i4s /
      * PR #9 per il gemello VNNI). Soglia configurabile con I4S.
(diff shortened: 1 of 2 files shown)


JustVugg:
Merged to `dev`. Reviewed and verified before landing:

**Correctness** — `matmul_i4_grouped` is sound: accumulator zeroed per group, nibble unpacking identical to `matmul_i4`, groups are byte-aligned (gs=128 is even so no cross-byte split at boundaries), and the AVX2 loads stay in-bounds. Your isolated 2.98e-08-vs-f32 number holds up.

**Backward compatibility — proven, not assumed.** I built the tiny oracle and ran the same snapshot through `dev` HEAD (without this PR) and this branch: **byte-identical** results (14/20 oracle match, same 84/0 expert-hit trace on both). The fmt=2 path is untouched — the detection only promotes to fmt=4 when `ns > O*4` (a per-row model always emits exactly `O` scales, so it stays fmt=2), and `make check`'s int4 kernel tests (`test_i4_acc512`, `test_idot`) pass. So: **no existing model gets worse from this** — the grouped format is only active if you rebuild the container with `--group-size 128`.

**Two follow-ups (not blockers):**
1. **fmt number collision with #168.** That PR also claims `fmt=4`, for int3-g64. The *detection* doesn't actually clash (int3 has a different weight-byte count, so it falls through to fmt=3, not fmt=4), but the integer value `4` in `t->fmt` can't mean two things. Since this landed first, #242 keeps fmt=4 and I've asked #168 to take fmt=5.
2. **Real-model quality A/B is still open** (your task 4, blocked on the FP8 download #239). The dequant-error drop (0.155→0.128) is encouraging but it's an *isolated* metric — the end-to-end 'fixes incoherent output' claim in the title isn't measured yet on the 372 GB model. Worth keeping #225 open until that A/B lands. Great work regardless.

woolcoxm:
Thanks for the thorough review — and for independently verifying the backward compat (the byte-identical oracle match is exactly the right test, glad it held).

On the two follow-ups:

1. **fmt=4 collision** — agreed, that's the right call. We're `fmt=4`, #168 takes `fmt=5`. No action needed on our side.

2. **Real-model quality A/B** — agreed, #225 should stay open until the end-to-end A/B lands. The dequant-error metric (0.155→0.128) is necessary but not sufficient. The FP8 download is at 82/141 shards and in progress; once it completes I'll produce both containers (per-row int4 + grouped int4) from the same FP8 source and run the recursion prompt + a small reasoning set side-by-side. That's the test that actually validates the "fixes incoherent output" claim.

PR #268, opened 2026-07-15
ARM: 4 independent SDOT accumulators in dot_i8i8/dot_i4i8

michael-denyer:
## Mechanism

The NEON DOTPROD paths in `dot_i8i8` and `dot_i4i8` accumulate into a single `int32x4_t`, so each `vdotq_s32` depends on the previous one — the loop is bound by the instruction's latency, not its throughput. This change uses four independent accumulators over a 64-deep unroll and sums them once at the end. Integer addition is associative, so results are **bit-exact**; the existing `test_idot` exactness suite covers both kernels. The non-DOTPROD `vmull/vpadal` fallback is untouched, and nothing changes on x86.

These per-row kernels carry S=1 decode (routed experts, lm_head, MTP head), so this benefits any aarch64 build with DOTPROD — all Apple Silicon included — with no build-flag changes.

## Measurements (M5 Pro, 18 threads, matmul_qt_ex at GLM-5.2 expert shapes, best of 3 runs)

| shape | before | after |
|---|---|---|
| gate/up int4 S=1 | 297.4 GF/s | 325.2 GF/s (+9.3%) |
| down int4 S=1 | 334.4 GF/s | 370.8 GF/s (+10.9%) |
| gate/up int8 S=1 | 257.1 GF/s | 262.8 GF/s (noise) |

int8 moves less because it has no nibble-unpack work and sits closer to memory-bound. Numbers are from one machine; happy to rerun under any harness you prefer.

## Testing

`make test-c` green on macOS arm64 (including `idot kernel exactness (neon): ok`), `make glm METAL=1` builds clean. Note this composes with (but does not depend on) a separate i8mm/SMMLA PR for the S>=2 batched case.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -611,18 +611,28 @@ static inline int32_t dot_i8i8(const int8_t *w, const int8_t *x, int I){
 #elif defined(__ARM_NEON)
     /* ARM: SDOT nativo se disponibile (Apple Silicon: sempre); altrimenti vmull/vpadal.
      * Stesso bound anti-overflow del trucco AVX2: coppie <= 128*127*2 = 32512 < 32767. */
+#if defined(__ARM_FEATURE_DOTPROD)
+    /* 4 accumulatori indipendenti: una sola catena vdotq e' latency-bound (misurato).
+     * EN: 4 independent accumulators; a single chained vdotq is latency-bound. */
+    int32x4_t a0=vdupq_n_s32(0),a1=vdupq_n_s32(0),a2=vdupq_n_s32(0),a3=vdupq_n_s32(0);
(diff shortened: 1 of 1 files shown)


JustVugg:
Thanks — but this is already in `dev`: #264 (skeldoor/perf/neon-multiacc-sdot, merged) added the exact same 4-independent-accumulator NEON SDOT to both `dot_i8i8` (glm.c:625) and `dot_i4i8` (glm.c:727). Same mechanism (break the single-`vdotq` latency chain into a0–a3), same 64-elem stride. Closing as superseded to avoid a duplicate — but your **#269 (i8mm SMMLA)** is genuinely new and I'm reviewing it now. Appreciate both.

PR #275, opened 2026-07-15
resource_plan: count physical cores on Windows (GetLogicalProcessorInformationEx)

KingIcyCreamProjects:
On Windows `physical_cpu_count()` falls straight through to `os.cpu_count()`, which returns *logical* processors. On SMT machines the plan then sets `OMP_NUM_THREADS` to 2 threads per core, which oversubscribes the vector units during expert matmul (on a 9950X3D: 32 logical vs 16 physical).

This adds a win32 branch that counts `RelationProcessorCore` records via `GetLogicalProcessorInformationEx` (ctypes, no new dependency), SMT-agnostic, with the existing `lscpu`/`os.cpu_count()` fallbacks intact and the whole thing wrapped in try/except.

Verified on a 9950X3D (16C/32T): returns 16. Existing tests pass; the plan JSON's `physical_cores` now reads 16 instead of 32 on that box.

### DIFF
--- a/c/resource_plan.py
+++ b/c/resource_plan.py
@@ -167,6 +167,30 @@ def discover_gpus():
 
 
 def physical_cpu_count():
+    if sys.platform == "win32":
+        # os.cpu_count() conta i processori logici (SMT): 2 thread/core saturano
+        # le unita' AVX-512 e peggiorano il matmul. Contiamo i core fisici veri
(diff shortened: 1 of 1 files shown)


JustVugg:
Merged to `main`. Counting physical cores via `GetLogicalProcessorInformationEx(RelationProcessorCore)` with a graceful `lscpu` fallback is the correct fix — logical-core counts do hurt the matmul under SMT. Thank you.

PR #276, opened 2026-07-15
benchmark_cuda_fixture: accept the current PROFILE service/wait format

KingIcyCreamProjects:
`profile_print` now emits `expert-disk N.NNNs service / N.NNNs wait`, but the harness regex still expects the old single `expert-disk N.NNNs` field — so every run of `tools/benchmark_cuda_fixture.py` dies with `RuntimeError: benchmark output missing` even though the engine output is fine.

This makes the regex accept both formats and reports `disk = service + wait`, keeping the JSON schema unchanged.

Found and verified on Windows 11 running the 313M fixture (all 5 modes produce medians again).


JustVugg:
Merged to `main` — thanks!

PR #278, opened 2026-07-15
glm: hwinfo_emit Windows support — CPU brand, cores, RAM were hardcoded-zero

KingIcyCreamProjects:
On native Windows the HWINFO dashboard snapshot reads `/proc/cpuinfo`, `/proc/meminfo` and `sysconf(_SC_NPROCESSORS_ONLN)` — none of which exist under MinGW — so the web dashboard's runtime panel shows "0 GB RAM / 0 cores" with an empty CPU name, while the CUDA branch populates VRAM correctly.

Fix, no new dependencies: CPUID brand string (leaves 0x80000002–4, gcc `<cpuid.h>`), `GetSystemInfo` for logical cores, and the existing `compat_meminfo` (GlobalMemoryStatusEx) for total/available RAM. Linux/macOS paths untouched.

Verified live against `coli serve` on Windows 11: panel now reads "AMD Ryzen 9 9950X3D 16-Core Processor / 66 GB RAM 24 GB free / 32 cores". Builds clean on current main with MinGW UCRT64 gcc 16.1.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -36,6 +36,9 @@
 #include <sys/mman.h>                             /* mlock: inchioda le pagine in RAM / wire pages into RAM */
 #include <sys/stat.h>                             /* fstat per mmap degli shard (COLI_MMAP) */
 #endif
+#if defined(_WIN32) && (defined(__x86_64__) || defined(__i386__))
+#include <cpuid.h>                                /* hwinfo_emit: CPU brand string senza /proc */
+#endif
 #include "st.h"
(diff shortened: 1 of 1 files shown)


JustVugg:
Merged to `main`. The dashboard showing "0 GB RAM / 0 cores" on Windows was a real papercut — CPUID brand string + `GlobalMemoryStatusEx` via the existing `compat_meminfo` is the clean way to do it with no new deps. Thanks!

PR #279, opened 2026-07-15
coli: measured Windows launcher defaults — OMP tuning parity + DIRECT/PIPE/PILOT_REAL (opt-out)

KingIcyCreamProjects:
On Windows a stock run leaves large measured wins on the table: the glm.c self-exec OMP tuning is Linux/FreeBSD-only, and the `PREFETCH`/readahead path is a `posix_fadvise` no-op under `_WIN32`. This teaches the launcher to `setdefault`, on win32 only:

- the OMP block (`OMP_WAIT_POLICY=active`, `GOMP_SPINCOUNT`, `OMP_DYNAMIC=FALSE`, `OMP_NUM_THREADS=<physical>`), same scope + presence-based kill-switch (`COLI_NO_OMP_TUNE`) as the engine's own tuner; `OMP_PROC_BIND`/`OMP_PLACES` are deliberately omitted (and no longer emitted by `--auto-tier` on win32) because MinGW libgomp rejects affinity outright;
- `DIRECT=1`, `PIPE=1`, `PILOT_REAL=1` — all lossless (byte-identical output), all overridable by simply setting the variable (`DIRECT=0` wins).

Measured on a 9950X3D + Samsung 9100 PRO (Gen5) + RTX 5090, Windows 11 native, 96-token greedy REPLAY, one lever per step, medians of 3–4 runs:

| step | tok/s | note |
|---|---|---|
| stock | 0.45 | disk-bound, buffered reads ran ~3 GB/s |
| +OMP block | 0.48 | expert-matmul −15% |
| +DIRECT=1 | 1.02 | +113% (iobench: 10.68 GB/s O_DIRECT vs 9.03 buffered warm; consistent with #162's 1.47×) |
| +PIPE=1 | 1.10 | workers 4/8/16 flat → default 8 untouched |
| +PILOT_REAL=1 | 1.22 | hit rate 58→78% |

Full ladder (including the negatives: PIN_GB=40 regression, CUDA-tier-alone ≈0% replicating #101, MTP DRAFT=3 losing 0.65 vs 1.05 disk-bound) is written up in the fork's tuning doc: https://github.com/KingIcyCreamProjects/colibri/blob/polish/ui-docs/docs/tuning-9950x3d-5090.md

`tests/test_env_defaults.py` covers the defaults, explicit-override-wins, kill-switch scope, and the non-win32 no-op. If you'd rather ship these as opt-in (e.g. behind `COLI_WIN_TUNE=1`) than opt-out, happy to flip the gate — the measurement stands either way.

Note: includes PR 1's commit (physical-core count) as its base — `env_for` uses `physical_cpu_count()`.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -161,6 +161,34 @@ def resource_request(a, env):
 
 def env_for(a):
     e = dict(os.environ, SNAP=a.model)
+    if sys.platform == "win32":
+        # COLI_NO_OMP_TUNE spegne SOLO il blocco OMP (stesso perimetro del
+        # self-exec di glm.c; presence-based come nel motore: impostarla a
+        # qualsiasi valore, anche 0, disattiva). I default I/O piu' sotto
+        # restano attivi: kill-switch dedicati = le var stesse (DIRECT=0 ecc.)
+        if not e.get("COLI_NO_OMP_TUNE"):
+            # parita' col tuning OMP self-exec di glm.c (solo Linux/FreeBSD, e
+            # comunque saltato sotto CUDA/Metal): libgomp legge queste variabili
+            # prima di main, quindi su Windows vanno nell'ambiente del figlio.
+            # niente OMP_PROC_BIND/OMP_PLACES: la libgomp di MinGW non supporta
+            # l'affinity su Windows ("Affinity not supported on this configuration")
+            from resource_plan import physical_cpu_count
+            for k, v in (("OMP_WAIT_POLICY", "active"),
+                         ("GOMP_SPINCOUNT", "200000"),
+                         ("OMP_DYNAMIC", "FALSE"),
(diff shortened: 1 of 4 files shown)


JustVugg:
Merged to `main`. Verified: clean merge + all 14 tests pass (`test_env_defaults` + `test_resource_plan`). The `physical_cpu_count` win32 block is identical to #275 (merged just before this) so git folded it cleanly — no duplication. Good call making the I/O defaults `setdefault` (explicit override always wins) and scoping `COLI_NO_OMP_TUNE` to the OMP block only; DIRECT/PIPE/PILOT_REAL are all lossless so opt-out defaults are safe. Skipping OMP_PROC_BIND/OMP_PLACES on Windows (MinGW libgomp has no affinity) is the right fix too. Thank you.

PR #283, opened 2026-07-15
ci: GitHub Actions — engine/web/python/cuda-syntax on ubuntu/macos/windows

### LINKED ISSUE #140: [Feature]: CI — run make check on ubuntu/windows/macos for PRs (today's #137 would have been caught pre-merge; repo currently has no workflows)

AloysJehwin:
## What

Adds the CI workflow requested in #140 / #143 / #144.

## Jobs

| Job | Platforms | What it catches | Time |
|-----|-----------|-----------------|------|
| **Engine** | ubuntu-latest, windows-latest | `make portable` + `make test-c` + `make test-python` | ~30s |
| **Engine (macOS)** | macos-latest (ARM) | `make glm` + full test suite (separate job: `portable` target is x86-only) | ~60s |
| **Web UI** | ubuntu-latest | `npm ci` + `npm run build` + `npm test` | ~17s |
| **CUDA syntax** | ubuntu-latest (nvidia/cuda:12.4 container) | `nvcc -c backend_cuda.cu` — no GPU needed | ~67s |

Triggers on push/PR to `main` and `dev`.

## Notes

- macOS is a separate job because `make portable` hard-codes `ARCH=x86-64-v3` which clang rejects on ARM — macOS uses `make glm` instead (no `-march` flag, NEON is baseline per the Makefile comment)
- CUDA syntax uses the `nvidia/cuda:12.4.0-devel-ubuntu22.04` container — simpler than fighting the apt repo on ubuntu-24.04 runners

Closes #140

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,16 +62,11 @@ jobs:
   cuda-syntax:
     name: CUDA syntax
     runs-on: ubuntu-latest
+    container: nvidia/cuda:12.4.0-devel-ubuntu22.04
     defaults:
       run:
         working-directory: c
     steps:
       - uses: actions/checkout@v4
-      - name: Install nvcc
-        run: |
-          wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64/cuda-keyring_1.1-1_all.deb
-          sudo dpkg -i cuda-keyring_1.1-1_all.deb
-          sudo apt-get update -qq
-          sudo apt-get install -y --no-install-recommends cuda-nvcc-12-4
       - name: Syntax check (no GPU needed)
         run: nvcc -c backend_cuda.cu -std=c++17 -O0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install nvcc
         run: |
+          wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64/cuda-keyring_1.1-1_all.deb
+          sudo dpkg -i cuda-keyring_1.1-1_all.deb
           sudo apt-get update -qq
           sudo apt-get install -y --no-install-recommends cuda-nvcc-12-4
       - name: Syntax check (no GPU needed)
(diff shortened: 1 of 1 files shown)


JustVugg:
Thank you for this — closing it as a duplicate, and I want to be clear that's about timing, not quality.

We took @ZacharyZcR's #144 (opened first, same file `.github/workflows/ci.yml`, same jobs: engine + C suite, web, python, CUDA syntax). It's merged and green on both `dev` and `main` now.

Two things your version and his shared, which his first run exposed and we fixed on the way in — worth knowing if you write another workflow here:

- the pinned `Jimver/cuda-toolkit` tag's version table stops at 12.6.2, so `12.6.3` fails the install step before nvcc runs;
- `nvcc ... | head -40` takes its exit status from `head`, so the CUDA job could never fail. A check that can't fail is worse than no check — `backend_cuda.cu` is compiled by nothing else in this repo.

That the same idea arrived from three people in three days is a fair signal we should have had CI sooner. Sorry it cost you the work. If you want to keep contributing to the checks, the gap that's still open is cross-platform coverage — #144 is Linux-only.

PR #287, opened 2026-07-15
metal: guard fused decode fast-paths on !kvs (fixes #285, #286 — serve+Metal 1-token replies)

### LINKED ISSUE #285: [Bug] SERVE_BATCH + COLI_METAL=1: responses are 1 token or garbage — Metal fused decode paths ignore ragged kvs[]/positions[]
## Symptom

macOS, Apple Silicon (Mac Studio M4 Max, 64 GB), GLM-5.2 int4, engine built with `METAL=1`, running `COLI_METAL=1 coli serve`. Every `/v1/chat/completions` request returns either exactly **one token** followed by `finish_reason: "stop"` (e.g. `"To"` for "Compare RAM and VRAM caching"), or, depending on sampling, a longer stream of garbage tokens (`"To`20oor a72----- ... mine mine mine ..."`).

The same build answers perfectly via `coli run` / `coli chat`. The same serve setup answers perfectly **without** `COLI_METAL=1`. So: engine OK, server OK, Metal OK — only the *combination* serve+Metal is broken.

## Cause

`run_serve_mux` decodes through `step_decode_batch`, which forwards ragged rows: per-row `kvs[]` / `positions[]`, `pos_base=0`. The comment on `attention_rows` documents the contract:

> kvs/pos describe a ragged decode batch: each row may belong to a different sequence. NULL keeps the original contiguous, currently-bound KV path.

Both Metal fast paths check `S<=4`, dims, `m->kv_start[layer]==0` — but **not `kvs == NULL`**:

1. `attention_rows` (fused decode attention, `coli_metal_attn_decode`) — passes `m->Lc[layer]`, `m->Rc[layer]`, `pos_base`, `m->kv_start[layer]`, i.e. the *bound* KV state and *contiguous* positions, ignoring the per-row arrays.
2. `layer_forward_rows` (full-layer command buffer, `coli_metal_layer_decode`) — same pattern, same problem, and it runs *before* attention_rows, so guarding only (1) still yields garbage.

With `pos_base=0` the GPU kernel attends at position 0 instead of the end of the context and writes KV rows to the wrong place → near-random logits → the sampler frequently draws a stop token immediately (hence the 1-token replies; the first token is always fine because prefill still goes through the contiguous `step()` path).

The first decode step after prefill happens inside `mux_submit` via `pick_tok` on prefill logits (correct); every subsequent token comes from the broken batched forward.

## Fix (tested)

Add the missing ragged-batch check to both gates, exactly as the contract comment prescribes:

```c
/* attention_rows */
if(g_metal_enabled && !kvs && S<=4 && ... )

/* layer_forward_rows, full-layer CB */
if(g_metal_enabled && !kvs && S<=4 && li<c->n_layers && l->sparse && ... )
```

Ragged batches then take the CPU absorb path (which handles kvs[]/positions[] per row); routed experts stay on the GPU (`[METAL] mode: batched routed experts on GPU` still active).

Verified end-to-end on the machine above: serve+Metal now streams coherent, correctly formatted answers. `coli run`/`chat` performance is unaffected (still hits the fused paths).

A follow-up could teach the Metal kernels to take per-row KV/positions so serve mode regains fused attention on GPU — the guard is the minimal correctness fix.

## Important follow-up: the bug poisons `.coli_kv` persistently

While the unguarded Metal paths are active, the misplaced KV writes end up in the KV rows that `mux_done`/`kv_disk_append` persist to `<model>/.coli_kv` (KVSAVE=1 default). Every later serve start then *resumes* the corrupted state ("[KV] resumed conversation from disk"), so the corruption **survives restarts and even engine rebuilds** — outputs show doubled tokens ("the the", "optimism**ism**"), degrading into repetition loops, and greedy decoding becomes non-reproducible across otherwise identical requests. Anyone who ever ran serve+Metal on the affected code should delete `<model>/.coli_kv*` after upgrading, otherwise the fix appears not to work. A `.coli_kv` integrity/version check (or disabling KV persistence in the multiplexed server) would make this failure mode much less sticky.

## Repro

```
COLI_METAL=1 ./coli serve --model <glm52_i4> --ram 48 --port 8000
curl -s -X POST localhost:8000/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"glm-5.2","messages":[{"role":"user","content":"Compare RAM and VRAM caching"}],"max_tokens":48}'
# -> {"content":"To","finish_reason":"stop","completion_tokens":1}
```

Happy to open a PR with the two-line patch.

### LINKED ISSUE #286: [Bug]: coli serve (SERVE_BATCH / run_serve_mux) truncates every response to 1 token on macOS Apple Silicon
### Commit or version

62419af

### Environment

macOS 26.5.2, Apple MacBook Pro M5 Max 128GB

### Reproduction steps

Build: make glm METAL=1 (Metal backend + Homebrew libomp), 0 warnings
Model: mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp (int4, int8 MTP head)

Reproduction:

[BUG_REPORT.md](https://github.com/user-attachments/files/30060544/BUG_REPORT.md)

cd c && make glm METAL=1
M=/path/to/GLM-5.2-colibri-int4-with-int8-mtp

# (A) plain serve path — coli chat — WORKS (full answer)
printf 'Write a Python factorial function. Keep it short.\n:q\n' \
  | COLI_MODEL=$M ./coli chat --ram 90
#   -> "Here is a concise Python factorial function ... def factorial(n): ..."

# (B) OpenAI serve path — SERVE_BATCH mux — FAILS (1 token)
COLI_MODEL=$M ./coli serve --host 127.0.0.1 --port 8000 --ram 90 &
curl -s localhost:8000/v1/chat/completions -H 'content-type: application/json' -d '{
  "model":"glm-5.2-colibri","max_tokens":120,"temperature":0,
  "messages":[{"role":"user","content":"Write a Python factorial function. Keep it short."}]}'
#   -> {... "message":{"content":"Here"}, "finish_reason":"stop",
#            "usage":{"completion_tokens":1, ...}}

### Expected behavior

OpenAI serve path expected to provide same answer content as chat path.

[FIX_NOTES.md](https://github.com/user-attachments/files/30060588/FIX_NOTES.md)

### Actual behavior and logs

```shell
OpenAI serve path stops at one token.
```

woolcoxm:
## What

Fixes the **serve + Metal** bug where every `/v1/chat/completions` reply is truncated to a single token (or streams garbage). Reported independently by two reporters — #285 and #286 are the same root cause.

Closes #285. Closes #286.

## Root cause

`run_serve_mux` decodes through `step_decode_batch`, which forwards **ragged rows**: per-row `kvs[]` / `positions[]`, `pos_base=0` (the contract documented on `attention_rows`). The two Metal fused fast paths checked `S<=4`, the GLM-5.2 dims, and `m->kv_start[layer]==0` — but **not `kvs == NULL`**:

1. `attention_rows` (`coli_metal_attn_decode`) — binds the contiguous `m->Lc[layer]` / `m->Rc[layer]` and `pos_base`, ignoring the per-row arrays.
2. `layer_forward_rows` (full-layer command buffer, `coli_metal_layer_decode`) — same problem, and it runs *before* `attention_rows`, so guarding only (1) still yields garbage.

With `pos_base=0` the GPU kernel attends at position 0 (start of context) and writes KV rows to the wrong place → near-random logits → the sampler frequently draws a stop token immediately. The first token is fine (prefill goes through the contiguous `step()` path); every subsequent token comes from the broken batched forward.

## Fix

Add the missing ragged-batch guard to both gates, exactly as the contract comment prescribes:

```c
/* attention_rows */
if(g_metal_enabled && !kvs && S<=4 && ...)

/* layer_forward_rows, full-layer CB */
if(g_metal_enabled && !kvs && !positions && S<=4 && li<c->n_layers && l->sparse && ...)
```

Ragged batches now take the CPU absorb path, which handles `kvs[]`/`positions[]` per row. Routed experts still run on the GPU (the `[METAL] mode: batched routed experts on GPU` path is unaffected). `coli run` / `coli chat` are unaffected — `kvs==NULL` there, so the fused paths still engage.

## Important: users who hit this should delete `.coli_kv`

Per #285: while the unguarded paths were active, the misplaced KV writes were persisted to `<model>/.coli_kv` (KVSAVE=1 default) and **survive restarts** — outputs show doubled tokens / repetition loops. Anyone who ran serve+Metal on affected code should delete `<model>/.coli_kv*` after upgrading, or the fix will appear not to work.

## Validation

- [x] `make glm` — compiles clean (Windows, default CPU build)
- [x] `gcc -DCOLI_METAL -fsyntax-only glm.c` — guards parse clean inside the `#ifdef COLI_METAL` block
- [ ] Metal runtime — verified end-to-end by the #285 reporter (Mac Studio M4 Max): serve+Metal now streams coherent answers; `coli run`/`chat` unaffected
- [x] No performance claims — `coli run`/`chat` keeps the fused paths; only serve+Metal (the broken combo) changes

## Compatibility

- [x] Default CPU build unaffected (`!kvs` guards live inside `#ifdef COLI_METAL`)
- [x] No model files, binaries, or benchmark artifacts
- [x] No CUDA changes

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -1789,8 +1789,12 @@ static void attention_rows(Model *m, Layer *l, int layer, float *x, int S, int p
     double ta0=now_s();
 #ifdef COLI_METAL
     /* Fused decode attention on GPU: whole layer in one command buffer (keeps the GPU hot).
-     * S<=4 absorption path with st0==0, DSA selection inactive, and GLM-5.2 int4 dims. */
-    if(g_metal_enabled && S<=4 && (g_absorb==1||(g_absorb<0&&S<=4)) && m->kv_start[layer]==0
+     * S<=4 absorption path with st0==0, DSA selection inactive, and GLM-5.2 int4 dims.
+     * kvs==NULL: the contiguous, currently-bound KV path. When the multiplexed server
+     * (run_serve_mux) forwards ragged rows, kvs[]/positions[] are per-sequence and the
+     * fused kernel must NOT run — it binds m->Lc/m->Rc and pos_base, which would attend at
+     * the wrong position and corrupt KV (#285/#286: serve+Metal → 1-token replies). */
+    if(g_metal_enabled && !kvs && S<=4 && (g_absorb==1||(g_absorb<0&&S<=4)) && m->kv_start[layer]==0
        && D==6144 && H==64 && c->q_lora==2048 && c->kv_lora==512 && c->qk_nope==192
        && c->qk_rope==64 && vh==256 && l->kv_b.fmt==2){
         int sel_active = m->has_dsa && layer<c->n_layers && c->idx_type[layer] && (pos_base+S) > c->index_topk;
@@ -2925,8 +2929,12 @@ static void layer_forward_rows(Model *m, Layer *l, int li, float *x, int S, int 
 #ifdef COLI_METAL
     /* FULL-LAYER CB: in_ln + attention + residuo + post_ln + shared expert + router/top-K
      * in un solo submit GPU; la CPU legge il routing e fa solo resolve/disk/expert-CB.
-     * Fallback: qualsiasi condizione mancante -> percorso CPU intero qui sotto. */
-    if(g_metal_enabled && S<=4 && li<c->n_layers && l->sparse
+     * Fallback: qualsiasi condizione mancante -> percorso CPU intero qui sotto.
+     * !kvs: contiguous single-sequence decode only — ragged mux rows take the CPU path
+     * below, which honours per-row kvs[]/positions[] (#285/#286). Routed experts still
+     * run on the GPU via the batched-expert path; only the fused attention/layer CB is
+     * gated off here. */
+    if(g_metal_enabled && !kvs && !positions && S<=4 && li<c->n_layers && l->sparse
        && (g_absorb==1||(g_absorb<0&&S<=4)) && m->kv_start[li]==0
        && D==6144 && c->n_heads==64 && c->q_lora==2048 && c->kv_lora==512
        && c->qk_nope==192 && c->qk_rope==64 && c->v_head==256 && l->kv_b.fmt==2


JustVugg:
Closing in favor of #288, which lands the same core fix — the `!kvs` guard on the fused Metal decode fast-paths — from @monotophic, who reported #286. Same mechanism (ragged mux rows carry per-row kvs[]/positions[] that the single-Lc/Rc+pos_base kernel can't express, so it ropes every row at position 0 and truncates to 1 token). Your extra `!positions` guard is a fine belt-and-braces, but kvs[] and positions[] are always passed together on the mux path, so `!kvs` alone is sufficient. Thanks for the fast turnaround on this — #288 gets the reporter the fix credit.

PR #290, opened 2026-07-15
glm: make MLOCK=1 actually wire pinned experts under COLI_MMAP

NeuralNotwerk:
## Problem

`MLOCK=1` is silently a no-op under `COLI_MMAP=1`: `pin_wire()` only locks the legacy private-slab path (`s->slab` / `s->fslab`), and both stay NULL for mmap experts. The "pinned" set is then just warm page cache — under memory pressure the kernel evicts it and decode re-reads experts from disk every token. Measured on a 503 GB host running GLM-5.2 int4 with `PIN_GB=all`: hundreds of MB/s of sustained disk reads mid-generation while the log claimed everything was resident.

## Why the obvious fix isn't enough

Wiring the mmap ranges of every pin slot over-locks by the whole VRAM tier: `expert_host_release()` early-returns for mmap experts (no slab to free) **without nulling `q8`/`q4`**, so GPU-uploaded experts keep live-looking host pointers forever. On a 4x RTX 5090 + 1x RTX 4090 rig that wired 363 GB instead of the 226 GB RAM tier — ~137 GB of never-touched locked pages, enough to starve the kernel into thrashing the *unlocked* hot pages (dense, shared experts, MTP head) instead.

## Changes

- `qt_wire_mmap()`: mlock each pinned expert's weight + scale ranges inside the file mappings; **skips `cuda_eligible` QTs** (VRAM-tier experts compute from device memory).
- `qt_unwire_mmap()`: REPIN `gpu_swap` promotions drop the promoted expert's host lock instead of leaking it on every swap.
- `expert_load()` deliberately does not wire (it also runs for the transient VRAM-staging pass); wiring happens once in `pin_wire()` on the final placement, after the GPU upload pass has set the eligibility flags.

## Measured

GLM-5.2 int4, 4x RTX 5090 + 1x RTX 4090, 503 GB RAM, `COLI_MMAP=1 MLOCK=1 PIN=stats.txt PIN_GB=all CUDA_EXPERT_GB=auto CUDA_DENSE=1`:

| | before | after |
|---|---|---|
| wired | 0 GB (silent no-op) / 363 GB (naive range wiring) | **226 GB** (exactly the RAM tier) |
| decode disk reads (warm) | 110–450 MB/s | **0** |

CPU-only and non-mmap builds are unchanged (`qt_unwire_mmap` is a no-op when `g_mmap` is off or wiring is disabled).

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -1365,6 +1365,14 @@ static void embed_row(Model *m, int tok, float *x){
 static int g_mmap=0;
 static struct { int fd; void *base; size_t len; } g_maps[512]; static int g_nmaps;
 static pthread_mutex_t g_map_mtx = PTHREAD_MUTEX_INITIALIZER;   /* expert_load e' OMP-parallel */
+/* forward decls: mem_should_wire/mem_wire live near pin_wire() further down, but
+ * qt_wire_mmap() (also further down, used by pin_wire()'s COLI_MMAP path) needs
+ * them declared before its own definition. Real mlock-ing of mmap'd pinned
+ * experts happens there, not in expert_load() -- see qt_wire_mmap() for why. */
+static int mem_should_wire(void);
+static int mem_wire(void *addr, size_t len);
+static void qt_unwire_mmap(QT *t);   /* def. presso pin_wire / defined near pin_wire */
+static int64_t g_mmap_wired=0; static long g_mmap_wire_failed=0;
 static void *map_of_fd(int fd){
     pthread_mutex_lock(&g_map_mtx);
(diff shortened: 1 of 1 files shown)


JustVugg:
Merged to `main`. Verified locally: clean merge + compiles with zero warnings (`-Wall -Wextra`). Careful work for a first contribution — wiring only the **final** RAM-tier resident set (skipping `cuda_eligible` slots) and un-wiring on REPIN GPU promotion is exactly right; the naive `q8/q4`-NULL check would have leaked locked pages for every VRAM-tier expert, as your comment notes. Thank you.

PR #295, opened 2026-07-15
Security/codeql findings

srinivasj7:
## Summary

## Validation

- [ ] `make -C c check`

## Compatibility


srinivasj7:
sorry, wrong destination. Please Ignore.

PR #296, opened 2026-07-16
Windows/CUDA port: thermal gate, storage probe, correct dashboard stats

David2024patton:
## Summary

Windows + CUDA portability pass so the engine and web dashboard run correctly on Windows with an NVIDIA GPU, plus a couple of small correctness fixes. Every change is guarded by `#ifdef COLI_CUDA` and cross-platform `#ifdef`s, so **Linux/CPU behavior is unchanged**.

### Engine (`c/glm.c`)
- **NVML thermal heartbeat + `thermal_gate()`** — a background thread samples GPU temperature (dynamically loaded `nvml.dll` / `libnvidia-ml.so`) and applies a proportional per-forward pause to bleed heat under sustained load. `SAFE_MODE=0` disables it; if NVML is absent the engine proceeds normally.
- **Storage bus probe** — `probe_storage_drive()` / `storage_bus_is_slow()` check free space and disk bus type via Win32 `IOCTL_STORAGE_QUERY_PROPERTY` and POSIX `statvfs`. `FORCE_SLOW_DRIVE=1` bypasses.
- **Win32 `hwinfo_emit` branch** — cores/RAM/CPU brand via `GetSystemInfo` / `GlobalMemoryStatusEx` / `__get_cpuid`, so the dashboard works without `/proc`.

### Dashboard correctness
- Report **RAM and VRAM in binary GiB** so 128 GiB / 16 GiB read as `128` / `16` instead of decimal `137` / `17` — matches the spec-sheet convention users expect.
- Show the **real GPU model** via NVML `nvmlDeviceGetName` (e.g. `NVIDIA GeForce RTX 4070 Ti SUPER`) instead of a generic label. The web panel now displays the model name and labels the logical-processor count as **threads**.

### Server (`c/openai_server.py`)
- `COLI_EBITS` / `COLI_DBITS` env vars to select quant bit-width (needed for the 4-bit config on 16 GB GPUs). Absent → engine defaults (8/8), i.e. upstream behavior preserved.

## Testing
- Built on Windows with MinGW gcc + nvcc/MSVC (CUDA 12.9), `coli_cuda.dll` unchanged.
- Ran the full GLM-5.2 INT4 checkpoint end to end (`COLI_CUDA=1 CUDA_DENSE=1 CUDA_EXPERT_GB=6 RAM_GB=110`, 4-bit) with the web UI; `/health` reports `ram_total_gb 127.7`, `vram_total_gb 16.0`, `gpu "NVIDIA GeForce RTX 4070 Ti SUPER"`, coherent generation, thermal + storage probes firing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,72 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [dev, main]
+  push:
+    branches: [dev, main]
+
+jobs:
+  engine:
+    name: Engine (Linux, CPU)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build glm
+        run: cd c && make glm
+      - name: C test suite
+        run: cd c && make test-c
+
+  engine-cuda-syntax:
+    name: CUDA syntax check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install CUDA toolkit (compiler only)
+        uses: Jimver/cuda-toolkit@v0.2.19
+        with:
+          # v0.2.19's version table stops at 12.6.2 — 12.6.3 fails the install
+          # step with "Version not available" before nvcc is ever reached.
+          cuda: '12.6.2'
+          method: network
+          sub-packages: '["nvcc"]'
+      - name: Compile backend_cuda.cu (syntax only, no GPU)
+        run: |
+          cd c
+          # NOT `nvcc ... | head -40`: a pipeline exits with the status of its
+          # LAST command, so head's 0 masked every nvcc error and the job could
+          # only ever be green. A check that cannot fail is worse than no check —
+          # it buys false confidence in the one file nothing else compiles.
+          nvcc -O2 -std=c++17 -arch=sm_80 -c backend_cuda.cu -o /dev/null \
+            -Xcompiler=-Wall,-Wextra
+          echo "CUDA syntax check passed"
+
+  web:
+    name: Web UI
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: web
+    steps:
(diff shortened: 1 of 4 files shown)


JustVugg:
Thanks for this — the Windows dashboard stats overlap with #278 (hwinfo CPU/cores/RAM via CPUID + `compat_meminfo`), which just merged to `main`, so this now conflicts. The thermal gate + storage probe are still valuable and distinct. With "allow edits from maintainers" on I can rebase onto the current tree and drop whatever #278 already covers, so you don't have to untangle it — want me to? Otherwise a rebase on your side works too.

JustVugg:

**It's 66 commits behind `dev`.** On `glm.c`, which changes several times a day, that isn't a rebase — it's a rewrite, and doing it *for* you would mean rewriting your design decisions without understanding why you made them. So I'd rather ask.

Windows is now covered by CI on MSYS2/UCRT64, so a rebased version gets tested automatically — which is what this PR most lacked. Also check against #314 (build-config stamp, CUDA_HOME quoting), which may already cover part of your delta.

JustVugg:
Closing — most of this "port" bundle has since landed piecemeal from other PRs, and what's left can't rebase cleanly (74 behind, touching ci.yml + glm.c + server + web that all moved).

The Windows/CUDA ground is now covered: fmt=4 grouped-int4 + the diagnostic harness via #298, the CUDA DLL loader + build fixes via #131/#314, the CI (including the Windows MSVC-host CUDA build job) via #143/#144/#327. A broad multi-area port PR against that is a rewrite, not a merge.

If a *specific* piece here isn't covered — a particular thermal-gate behavior, a storage-probe the current `doctor`/`iobench` lacks, or a dashboard stat that's wrong — open it as its own small PR against `dev` and it'll go in fast. Bundling four areas is what made this un-mergeable; one area at a time is the way now. Thanks for the Windows push.

PR #299, opened 2026-07-16
KV8=1: fp8 e4m3 latent KV cache (CPU + CUDA) with .coli_kv v2

NeuralNotwerk:
Opt-in (`KV8=1`, default off) fp8 e4m3 storage for the MLA latent KV — the same latent-KV quantization regime DeepSeek-V3 uses. Per token per layer, the Lc/Rc rows become fp8 bytes plus one f32 amax/448 scale each:

- **~3.9× less KV RAM** — a 256k-token context drops from 47.7 GB to ~13 GB, and `kv_pool_bytes` reports the smaller pool so `cap_for_ram`/`expert_avail` stop demoting experts to disk under multi-slot KV.
- **`.coli_kv` v2** (~46 KB/token instead of ~182): flows through the existing persistent-handle + staging-buffer append path; v1 files quantize on resume and stay intact on disk until the first save rewrites them (magic self-heal in `kv_disk_open`), so a crash before the first save loses nothing.
- **CUDA covered**: `attention_absorb_kernel8/_batch_kernel8` (hardware fp8 cvt, sm_89+) behind three new entry points at ¼ the PCIe upload traffic, wired at the `COLI_CUDA_ATTN` call sites and the Windows DLL loader. Metal and `COLI_CUDA_PIPE` still read f32 rows, so KV8 auto-disables under Metal and forces the pipe off.
- **No dependencies**: encoder/decoder are ~60 lines of bit-math (`c/kv_fp8.h`), exhaustively unit-tested (256-code round-trip, RNE ties incl. the binade-boundary mantissa carry, subnormal-amax guard).

Validation: tiny-oracle KV8=0 is bit-identical (TF 32/32, gen 20/20, CPU and CUDA); KV8=1 flips only two positions whose logit margins are so thin that transformers' own f32 forward flips them across versions; CPU and CUDA fp8 paths agree token-for-token; ASan/UBSan clean. On the real 744B int4 model, log-likelihood A/B over three ~450-token requests: ppl deltas +2.4% / +0.6% / −3.7% — net ~0, inside the int4/routing noise floor. New tests: `tests/test_kv_fp8.c`, `tests/test_kv_disk.c`, KV8 cases in `test_kv_alloc.c` and `test_backend_cuda.cu`; `make test-c` and `make cuda-test` green (sm_120 + sm_89).

A follow-up PR (stacked) removes the CUDA T caps for KV8 via a device-resident fp8 shadow + split-T kernels.


NeuralNotwerk:
Withdrawing for now — we're running more soak testing on our own fleet first (already found a VRAM budgeting interaction between the fp8 shadow and CUDA_EXPERT_GB=auto worth fixing before this lands). Will reopen once it has more hours on it.

NeuralNotwerk:
Superseded by #399 — the VRAM-budget interaction that prompted this withdrawal (fp8 shadow vs CUDA_EXPERT_GB=auto) is fixed there, plus the stack now includes the review fixes, vectorized fp8 kernels, a 4-bit TurboQuant tier, and full Metal-backend support with parity tests.

PR #300, opened 2026-07-16
KV8 long-context CUDA: fp8 device shadow, split-T kernels, DSA-aware gather

NeuralNotwerk:
**Stacked on #299** (first commit is that PR; review the top commit).

The absorb kernels keep all T scores in shared memory, so CUDA attention rejects T>4096 (decode) / 8192 (batch) and long-context attention falls back to CPU — measured 22–500 ms per layer-token at 16k–262k. Under KV8 this PR removes the caps:

- **`kv_dev_sync8`** — the fp8 latent KV lives in a device-resident shadow on each layer's home GPU (~149 MB/layer at 256k, vs 604 MB f32 that never fit in VRAM); decode uploads only the new rows. Lazy allocation, fail-soft to the host-upload paths.
- **Split-T kernels** (flash-decoding style) — `dim3(H,P)` partial online-softmax blocks plus a log-sum-exp merge. A one-block-per-head tiled kernel measured 103 ms/layer at 262k (64 blocks can't fill the GPU); split-T brings it to **24.5 ms — 11–17× faster than the CPU fallback** at 16k–262k. Batch shapes keep a tiled streaming kernel (S×H blocks already saturate) with the cap lifted.
- **DSA gather** — `attention_absorb_sel_kernel8` attends the selected rows straight from the shadow (absolute indices, NS ≤ 4096). Under KV8 the batch branches are skipped when a selection is active, so GPU decode **respects the DSA selection** exactly like the CPU reference. The f32 GPU flow is untouched and keeps its historical dense-over-selection behavior — flagging that divergence separately in case you want it addressed for f32 too.

Validation: kernel tests vs double-precision references on sm_120 and sm_89, capped-vs-shadow parity at T=3000; engine teacher-forcing at T=10,000 matches the CPU KV8 engine on 9999/10000 positions (one fp-summation-order near-tie); forced-selection generation is token-identical to dense with the gather kernel confirmed launching via nsys; a 9936-step REPLAY through the split path runs clean.


NeuralNotwerk:
Withdrawing along with #299 (stacked) — same reason: more soak testing on our fleet first.

NeuralNotwerk:
Superseded by #399 (see comment on #299).

PR #302, opened 2026-07-16
Fix: set stdout to O_BINARY on Windows to fix READY sentinel

alekseysorokin68:
## Summary

## Validation

- [ ] `make -C c check`

## Compatibility

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -5749,6 +5749,9 @@ int main(int argc, char **argv){
         perror("[OMP] execv self-reexec failed, running untuned");
 #endif
     }
+#ifdef _WIN32
+    _setmode(fileno(stdout), O_BINARY);
+#endif
 #if defined(__AVX512F__) && defined(__AVX512BW__)
     if(getenv("I4_ACC512")) g_i4_acc512=atoi(getenv("I4_ACC512"))!=0;
     if(getenv("I4_ACC512_TEST")){


JustVugg:
Thank you for this — the `_setmode` fix is real and I want it. The rest of the diff I can't take, and I'd like to explain why rather than just decline.

**The part I want** (3 lines): `_setmode(fileno(stdout), O_BINARY)`. Both sentinels end with `\n` (`c/coli:69-70`), so under CRT text mode the engine emits `\x01\x01READY\x01\x01\r\n`, `endswith(sentinel)` never matches, and chat hangs forever waiting for a byte pattern that can't arrive. That's a genuine hazard and the fix is the right one — it matches the existing belt-and-braces reasoning in `compat.h:88`.

One question on the premise, though: if stdout translation were happening on the documented build path, **no Windows user could chat at all** — and they can (see #307, Windows 11, MinGW, chat working). So I don't think this reproduces on w64devkit/MinGW. **Which toolchain are you building with?** If it's MSVC, that's a build path we don't document and I'd like to know it works. It changes the fix from "bug" to "defense in depth", which I'd still merge — I just want the commit message to be true.

**The part I can't take** is the ~50-line POSIX emulation layer, and it's not a style objection:

1. **`mmap()` as `malloc` + `ReadFile` is the opposite of what this engine is.** It allocates `len` bytes of anonymous memory and reads the whole thing in. The entire premise of colibrì is streaming a 744B model through a small RAM budget; `COLI_MMAP=1` exists to make experts *zero-copy views* backed by the page cache. Emulating it with a full-size `malloc` turns the low-RAM path into the highest-RAM path — silently, since it still returns a valid pointer. Windows has `CreateFileMapping`/`MapViewOfFile` if we ever want this for real.
2. **`select()` → `_kbhit()` ignores the fd entirely** and only works on a console. `coli` gives the engine stdin over a **pipe**, so the serve loop this is meant to support would break in exactly the configuration that matters.
3. **It drops an architecture guard.** Moving `#include <cpuid.h>` inside `#ifdef _WIN32` loses the `(defined(__x86_64__) || defined(__i386__))` condition it had, which breaks Windows on ARM64.
4. **It duplicates `compat.h`**, which is this project's Windows strategy: native Win32 handles behind small explicit shims, not POSIX emulation. Two layers disagreeing about what `mmap` means is worse than either alone.

Could you split this into a PR with just the `_setmode` call? I'll merge that quickly. If you're hitting real build failures on Windows that motivated the stubs, please open an issue with the toolchain and the errors — those are worth fixing properly in `compat.h`, and I'd rather solve the actual build problem than take an emulation layer whose semantics quietly differ from the real thing.

`maintainerCanModify` is on, so I can do the split for you if you'd prefer — just say the word and it stays your commit.

alekseysorokin68:
Updated to 3 lines only — just the _setmode fix, no POSIX emulation layer.
Build environment:
- OS: Windows 11 (MSYS_NT-10.0-26200)
- Toolchain: MinGW-w64 GCC (MSYS2 mingw64, gcc -march=x86-64-v3)
- Make: MSYS2 /usr/bin/make
Why I can't compile-test the full codebase on Windows:
glm.c uses POSIX-only symbols (mmap, madvise, select, fd_set, struct stat/fstat) that are unavailable on native MinGW without platform guards. The codebase compiles successfully on Linux (Ubuntu 24.04, GCC 13) and macOS (Apple clang). Windows compilation requires wrapping those POSIX calls in #if defined(__APPLE__) || defined(__linux__) || defined(__FreeBSD__) guards — that's a separate concern noted in the commit message.
The _setmode fix itself is trivial and self-contained: a standard Windows CRT call that switches stdout to binary mode, matching the existing O_BINARY reasoning in compat.h:88.

Issue #307, opened 2026-07-16
[Bug]: python coli chat doesn't end


galmok:
### Commit or version

d4b4f33f22d0ddfb29d3386111a05ff1b2485322

### Environment

Windows 11, 9850X3D, 128GB, 9070XT 16GB.
Built according to Readme.

### Reproduction steps

I ran:
python coli chat --model T:\LLMs\colibri\glm52git\GLM-5.2-colibri-int4-with-int8-mtp
I entered some test prompts about jokes, and which it did answer, but the last response: has an infinite response problem:

>   Ah, yes! That is a legendary classic. It sounds like you're recalling a variation of the famous "3 envelopeed peas" (3 EPs) joke, which is the undisputed Mount Everest of anti-jokes / shaggy dog stories!
> 
>   It starts out so normally: A guy has a friend named either "MacGregor" or "Duncan." For his friend's birthdays, he always sends him exactly three envelopes packed with peas (3 EPs). He does this for years, or even decades. Finally, after thirty or forty years, the friend finally asks, "Why do you always send me three envelopes packed with peas?"
> 
[...]

Issue #303, opened 2026-07-16
EXPERT_BUDGET (merged in #254, now on main): measured empty operating window — needs a quality gate and an MTP-contract decision


bokiko:
This carries forward the three-host review I posted on #220 ([comment](https://github.com/JustVugg/colibri/issues/220#issuecomment-4983627320)) — it landed a few hours after the issue was closed, so filing the actionable part as its own issue now that the feature is on `main`.

## Summary of the data (full tables in the linked comment)

Three hosts (5950X/31 GB, M4 Pro/48 GB, 9955HX/92 GB), `8d29686` reviewed diff, same model files and prompt as my #107 rows:

- **The speed claims validate** — decode up to 3.4× and prefill 18× at `EXPERT_BUDGET=4` on the 92 GB host (hit rate 96%).
- **Quality collapses at the same settings.** hellaswag (n=40, prefix-fixed harness from #194): **30.0% acc_norm at budget=8 vs 90.0% budget-off** — barely above the 25% chance floor. At budget=4 the decode output is literal noise (baseline on the same host, same prompt, opens `"When you type a URL (like www.example.com)..."`; budget=4 produces `://.- .°}- **'s =[ }**0 @____`).
- **MTP acceptance collapses at every useful setting** (56%→0% at budget=4; 42%→0% on the 92 GB host).
- **The operating window may be empty**: by budget=16 — the loosest setting tested — decode speed is back to baseline while acceptance is still wrecked (8%). Every measured point is either "no faster" or "no longer coherent."

One scope note: @woolcoxm's observation on #220 (baseline already incoherent on his per-row int4 checkpoint, so budget can't visibly worsen it) is real on his box but doesn't transfer to these runs — the baselines above are coherent (90% hellaswag, clean decode text). The clean token-agreement A/B does wait on the grouped-int4 container (#225); the quality-collapse and acceptance data don't.

## Why this is worth an issue rather than a footnote

#294 just established the engine invariant that fixed #163: **draft and verify must compute the same function.** EXPERT_BUDGET violates the same invariant by design, through cache state instead of kernel choice — which experts survive the cap depends on residency at the moment of the forward, so the verify forward doesn't reproduce the drafter's computation. `SPEC_PIN` can't fix that (and shouldn't try — it's a feature semantics question, not a kernel-dispatch bug).

## Concrete asks

[...]

PR #165, opened 2026-07-14
Add DeepSeek V4 target-only CPU inference

DrewZt:
## Summary

This PR now contains the **DeepSeek V4 target-only CPU engine** requested in the latest review. DSpark speculative decoding has been removed from #165 and preserved on `DrewZt:pr165-full-dspark-backup` for a separate stacked follow-up.

The target engine:

- loads official sharded DeepSeek V4 safetensors checkpoints;
- implements target prefill and greedy decode, compressed attention, mHC, routed/shared experts, and RAM-tiered ExpertStore caching;
- keeps target dense residency independent of DSpark;
- uses shared `st.h` for safetensors indexing/range I/O;
- uses shared `quant.h` for canonical fmt7 MXFP4 matmul;
- is wired into `coli run`, `coli chat`, `coli serve`, and `coli web`;
- accepts `--no-dspark` as a compatibility no-op.

Production is consolidated in `c/deepseek_v4.c`. The standalone `c/v4` launcher, DSpark runtime/build units, DSpark fixtures/tests, and committed `.safetensors` fixtures have been removed.

## Shared infrastructure status

| Checkpoint path | Current implementation | Status |
|---|---|---|
| Safetensors indexing/range reads | shared `st.h` | migrated |
| fmt7 standard MXFP4 matmul | shared `quant.h` | migrated |
| fmt7 resident rows16 expert cache | minimal V4-private layout | `TODO(upstream-fmt7-rows16)`: migrate when shared quant exposes a resident rows16 API |
| fmt8 E4M3 + UE8M0 128x128 scales | minimal V4-private decoder | `TODO(upstream-fmt8-ue8m0)`: replace when shared fmt8 UE8M0 decode exists |

The last two paths remain only to keep the target engine usable and are explicitly marked for migration.

## Unified serving

`openai_server.py` detects `deepseek_v4`, renders native multi-turn V4 markers, and launches the persistent `SUBMIT`/`DATA`/`DONE` protocol. Serving is target-only, greedy, one KV slot, and rejects tools/grammar. Requests re-prefill context while the engine, dense tensors, head, and expert cache stay warm.

## Fixture and CI

The target-only fixture is generated from pinned PyTorch `2.13.0+cpu`, Transformers `5.14.1`, and safetensors `0.8.0`. Its reference comes from official `DeepseekV4ForCausalLM`; there is no C-engine oracle fallback. Generated safetensors are ignored and not committed.

[...]

PR #312, opened 2026-07-16
Inkling (Thinking Machines 975B MoE): new engine, o200k tokenizer, bf16->int4 converter

nbeerbower:
## What

Support for [Thinking Machines **Inkling**](https://thinkingmachines.ai/news/introducing-inkling/) (975B total / 41B active MoE, Apache 2.0) — a second full architecture for colibri, following the `olmoe.c` standalone-engine precedent:

- **`c/inkling.c`** — Stage-A engine: GQA with interleaved sliding-window (512, 16 KV heads) / global (8 KV heads) attention at 5:1, learned relative-position bias banks (no RoPE), log-length tau scaling past 128k, four depthwise causal short-convs per layer (fp32, residual inside, cached conv state for decode), dense + MoE layers, sigmoid router with loss-free bias and jointly-normalized routed+shared weights, mup logit scaling. Routed experts stream per-expert from the fused 3D tensors (LRU cache, colibri int4/int8 container or f32). `-p "prompt"` streaming greedy mode; ref-oracle harness mode.
- **`c/tok.h`** — o200k pre-tokenizer support (case-aware Split regex), auto-detected from `tokenizer.json` so the GLM/cl100k path is byte-identical to before. New `tok_unicode_o200k.h` range tables.
- **`c/tools/convert_inkling_int4.py`** — bf16 checkpoint → colibri container (int4 experts + `.qs` per-row scales, bf16 residents, f32 norms/convs/router). Threaded quantization, shard-resumable, `--watch` converts while the download is still running. Note for reviewers: TML's fused `w13` tensors store gate/up rows **interleaved**; the converter de-interleaves exactly per `transformers`' conversion mapping.
- **`c/tools/make_tiny_inkling.py`** — tiny random-init oracle generator (HF transformers ≥ 5.14).
- Zen-4-friendly kernels in `inkling.c`: bf16 residents dotted via AVX512-BF16 `vdpbf16ps` (halves resident RAM vs f32 — required: f32 residents would be ~172 GB), int8 expert dot via AVX-512 VNNI (IDOT family; `IDOT=0` scalar path preserved). Scalar fallbacks keep the dependency-free default CPU path intact.

## Validation

- **Token-exact oracle** (the colibri standard): tiny random-init Inkling via HF transformers — **36/36 teacher-forced argmax, 24/24 greedy**, in f32, int4-container+VNNI, IDOT=0 scalar, and bf16-resident modes.
- **Tokenizer**: 357/357 strings encode identically to HF `tokenizers` (case transitions, contractions, CJK, combining marks, emoji+modifiers, 300 mixed-charset fuzz cases).
- **Converter e2e** (`--selftest-e2e`): fabricates a TML-named checkpoint from the tiny oracle (inverse mapping, re-fused, re-interleaved), converts back, engine reproduces the oracle exactly in both passthrough and int4 modes.
- **Real checkpoint** (975B, int4 experts + bf16 residents, 469 GiB snapshot):

  ```
  $ SNAP=~/Models/inkling_i4 ./inkling -p "The capital of France is" -n 24
  The capital of France is Paris.
  The capital of the UK is London.
  The capital of Russia is Moscow.
  The capital of Italy
  ```

  Converted weights published at [nbeerbower/Inkling-colibri-int4](https://huggingface.co/nbeerbower/Inkling-colibri-int4).

## Perf status (honest numbers)

Ryzen 9 7900 (24t), 187 GB RAM, NVMe snapshot, commit b60f881, `make CUDA=0`, single run, cold cache: RSS 90.5 GB, prefill 150 s (5 tok), decode 0.06 tok/s at cache 16/layer — expert streaming bound, as expected for Stage A with a plain LRU. The architecture is unusually kind to long context (55/66 layers have a fixed 512-token window; 1M-token KV ≈ 46 GB bf16, RAM-resident), so the usual roadmap applies and should pay off more than it did for GLM: hot-store/pinning + prefetch (`tier.h`), then a GQA CUDA tier — the attention here needs no MLA-style custom kernels.

`make check` passes (62 tests), `-Wall -Wextra` clean, GLM oracle path untouched (tok.h dispatch is additive).

Happy to split this into smaller PRs if you prefer (tokenizer / engine / converter).

https://claude.ai/code/session_016KLDCYJSyHxd39ChTr4T27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Build colibri
-        run: cd c && make colibri
+      - name: Build colibri + inkling
+        run: cd c && make colibri inkling
       - name: C test suite
         run: cd c && make test-c
 
@@ -40,6 +40,38 @@ jobs:
           nvcc -O2 -std=c++17 -arch=sm_80 -c backend_cuda.cu -o /dev/null \
             -Xcompiler=-Wall,-Wextra
           echo "CUDA syntax check passed"
+      - name: Compile backend_cuda_ink.cu (syntax only, no GPU)
+        run: |
+          cd c
+          # inkling's own CUDA backend (bf16 residents + int4 expert GEMM); same
+          # no-pipeline rule as above so a compile error can actually fail the job.
+          nvcc -O2 -std=c++17 -arch=sm_80 -c backend_cuda_ink.cu -o /dev/null \
+            -Xcompiler=-Wall,-Wextra
+          echo "inkling CUDA syntax check passed"
+
+  inkling-oracle:
+    name: Inkling oracle (token-exact vs transformers)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: pip
+          cache-dependency-path: c/tools/oracle-requirements.txt
+      - name: Install torch (CPU) + transformers
+        run: pip install -r c/tools/oracle-requirements.txt
+      - name: Build inkling
+        run: cd c && make inkling
+      - name: Tiny-model fixture + token-exact oracle (f32, bit-exact experts)
+        run: |
+          cd c
+          # Random-init InklingForCausalLM via HF transformers, then the C engine
+          # reproduces it token-for-token. bits=0 keeps experts f32 (bit-exact
+          # forward); the greedy pass exits non-zero on any mismatch, so this
+          # gate fails the moment a shared-code change regresses the engine.
+          python3 tools/make_tiny_inkling.py tiny_inkling
+          SNAP=tiny_inkling ./inkling 8 0 tiny_inkling/ref_inkling.json
 
   engine-hip-syntax:
     name: HIP syntax check
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ olmoe_merged/
 olmoe_i4/
 c/olmoe_merged/
 c/olmoe_i4/
+c/tiny_inkling/
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ c/colibri.exe
 c/glm
 c/glm.exe
 c/olmoe
+c/inkling
 c/olmoe.exe
 c/iobench
 c/iobench.exe
--- a/c/Makefile
+++ b/c/Makefile
@@ -227,6 +227,7 @@ endif
 CFLAGS   += -DCOLI_CUDA
 LDFLAGS  += -L$(CUDA_HOME)/lib64 -Wl,-rpath,$(CUDA_HOME)/lib64 -lcudart -lstdc++
 CUDA_OBJ  = backend_cuda.o
+INK_CUDA_OBJ = backend_cuda_ink.o
 endif
 
 ifeq ($(HIP),1)
@@ -300,6 +301,10 @@ cuda-dll: backend_cuda.cu backend_cuda.h
 		-L"$(CUDA_HOME)/lib/x64" -lcudart \
 		backend_cuda.cu -o coli_cuda.dll
 
+backend_cuda_ink.o: backend_cuda_ink.cu backend_cuda_ink.h .build-config
+	@command -v "$(NVCC)" >/dev/null 2>&1 || { echo "nvcc not found: set CUDA_HOME or NVCC" >&2; exit 1; }
+	"$(NVCC)" $(NVCCFLAGS) -c backend_cuda_ink.cu -o $@
+
 backend_cuda.o: backend_cuda.cu backend_cuda.h backend_gpu_compat.h .build-config
 	@command -v "$(GPUCC)" >/dev/null 2>&1 || { echo "$(GPUCC_NAME) not found" >&2; exit 1; }
 	"$(GPUCC)" $(GPUFLAGS) -c backend_cuda.cu -o $@
@@ -339,8 +344,8 @@ cuda-bench: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/bench_tens
 olmoe$(EXE): olmoe.c st.h json.h compat.h
 	$(CC) $(CFLAGS) olmoe.c -o olmoe$(EXE) $(LDFLAGS)
 
-inkling$(EXE): inkling.c st.h json.h compat.h
-	$(CC) $(CFLAGS) inkling.c -o inkling$(EXE) $(LDFLAGS)
+inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ)
+	$(CC) $(CFLAGS) inkling.c $(INK_CUDA_OBJ) -o inkling$(EXE) $(LDFLAGS)
 
 # Use a baseline that matches the compiler target. macOS already targets a
 # portable baseline when ARCH is empty; forcing the x86 value there breaks
--- a/c/Makefile
+++ b/c/Makefile
@@ -339,6 +339,9 @@ cuda-bench: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/bench_tens
 olmoe$(EXE): olmoe.c st.h json.h compat.h
 	$(CC) $(CFLAGS) olmoe.c -o olmoe$(EXE) $(LDFLAGS)
 
+inkling$(EXE): inkling.c st.h json.h compat.h
+	$(CC) $(CFLAGS) inkling.c -o inkling$(EXE) $(LDFLAGS)
+
 # Use a baseline that matches the compiler target. macOS already targets a
 # portable baseline when ARCH is empty; forcing the x86 value there breaks
 # Apple Silicon. Unknown targets use native rather than an invalid x86 flag.
--- a/c/backend_cuda_ink.cu
+++ b/c/backend_cuda_ink.cu
@@ -16,6 +16,10 @@ static size_t g_xcap = 0, g_ycap = 0;
 static int g_ok = 0;
 
 extern "C" int ink_cuda_init(int dev) {
+    /* ~727 tiny matmul calls per decoded token: the default yield-based sync
+     * costs ~0.5-1 ms of wakeup latency per call. Spin-sync in the driver is
+     * one busy thread — affordable since the OMP team no longer spins. */
+    cudaSetDeviceFlags(cudaDeviceScheduleSpin);
     if (cudaSetDevice(dev) != cudaSuccess) return -1;
     if (cudaStreamCreate(&g_st) != cudaSuccess) return -1;
     /* fail fast on a broken runtime instead of at the first matmul */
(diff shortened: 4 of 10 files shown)


nbeerbower:
Pushed a Stage-B performance commit (3ef6775) — same architecture, the expert path grew the glm.c playbook:

- cache slots keep container rows **packed** (int4 stays 4-bit in RAM, ~28 MB/expert → double the cache per GB); new `matmul_q4` unpacks nibbles in-register, numerically identical to the int8 path
- `moe()` routes everything first, then fills **all** missing experts in one parallel burst (prefill batches the whole sequence's misses, so the NVMe finally sees queue depth), then computes
- usage-history pinning (`SNAP/.coli_usage`, top `PIN_N`/layer non-evictable) + RAM-sized auto cap

Measured on the 975B checkpoint (Ryzen 9 7900, NVMe, cap auto=64/layer, 1024 pinned):

| | Stage A | Stage B cold | Stage B pinned+warm |
|---|---|---|---|
| prefill (5 tok) | 150.2 s | 31.5 s | **21.7 s** |
| decode | 0.06 tok/s | 0.21 tok/s | **0.25 tok/s** |
| cache hit | ~0% | 72.0% | **81.5%** |

All four oracle modes (f32, int4-container VNNI, IDOT=0 scalar, bf16 residents) remain token-exact, `make check` 68/68, 0 warnings.

https://claude.ai/code/session_016KLDCYJSyHxd39ChTr4T27

nbeerbower:
Final push for this PR — it now includes the CUDA resident tier and cache warming, bringing inkling to feature parity with what glm already has (no concepts new to colibri):

- **CUDA tier** (`backend_cuda_ink.cu`, `CUDA=1`): bf16 residents (~36 GB) live in VRAM on a warp-per-row kernel with the same numeric contract as the CPU vdpbf16ps path; freed RAM goes to the expert cache. Pure-CPU default build unchanged and dependency-free.
- **Cache warming** (glm's `.coli_usage` idea): per-expert usage counts persist and accumulate; top `PIN_N`/layer pinned at startup. Fully toggleable (`PIN=off`, `PIN_N`, `USAGE_SAVE=0`). The `-f` promptfile mode exists to build the history for a binary with no serve mode yet — happy to drop it if it reads as scope.
- **`docs/inkling.md`**: quickstart, env reference, benchmark methodology.

Measured on the 975B (Ryzen 9 7900 / A6000 / NVMe, commit 87981c1, greedy):

| Configuration | Prefill | Decode | Hit |
|---|---|---|---|
| Stage A | 150.2 s | 0.06 tok/s | ~0% |
| CPU perf pass | 21.1 s | 0.25 tok/s | 81.5% |
| + CUDA residents | 18.4 s | 0.32 tok/s | 83.6% |
| trained-prompt pins (best case) | 1.9 s | 2.51 tok/s | 100% |
| **steady state (novel prompt, diverse history)** | 35.4 s | **0.25 tok/s** | 82.2% |

The honest number is the last row; it grows toward the best case as usage history accumulates (counts persist across runs). Phase timers show the remaining time is ~90% CPU expert matmul at high hit rates — GPU expert compute is the natural follow-up PR, not included here.

All modes remain token-exact vs the transformers oracle (f32, int4-container VNNI + IDOT=0, bf16 CPU, bf16 CUDA); `make check` 68/68, 0 warnings on both builds.

https://claude.ai/code/session_016KLDCYJSyHxd39ChTr4T27

JustVugg:
Thank you for this — real work, and I want to be straight with you about the bar rather than let it sit open for another week.

**The maintainer's rule, and it's the right one: a new engine lands only if he can run it on his own machine first.** Not "the CI compiles it" — *run it*. Right now neither of us can: there's no checkpoint here, so the only thing that could be verified is that gcc doesn't complain. That's not verification, that's a spell-check.

Why the bar is that high, concretely. This week, on the **one** engine we can run:

- `EXPERT_BUDGET` shipped on numbers nobody checked and had to be quarantined (#303) — measured 0.13 tok/s against a 0.30 baseline while claiming a speedup, and loading *more* experts per token than not using it at all;
- three separate users are getting incoherent output and the root cause is still unconfirmed (#307, #225);
- a CI check went in that could never fail;
- two tests I wrote myself "found" bugs that didn't exist.

That's the state of the engine that gets daily attention. A second engine that nobody here can execute doesn't get daily attention — it gets zero, until the month it breaks and the person holding it is the maintainer, who never had the weights.

**What would change the answer**, and I think it's reachable:

1. **A model he can actually run**, or a tiny one the CI can execute end-to-end. This repo already has the pattern: `make_glm_oracle.py` builds a `glm_tiny/` and the engine is scored token-exact against it. An engine with an oracle test is an engine that can't silently rot.
2. **A named maintainer** — you, on the record, for that file.
3. **Isolation from `glm.c`'s path.** This is the part I'd check hardest, and it's specific per PR (see below).

Until then the honest home is a **fork, linked from the README**. That's not a brush-off: it means people can find and use your work, and it doesn't hand a maintainer 10,000 lines he can't test. If it gets traction and someone can run it in CI, it comes back.

I'd rather tell you this now than leave it open for a month and close it then.

JustVugg:
Specific to this one — and it's a more favourable read than #165.

**You have a track record here**: #232 landed today after I merged and tested it (PROF=1 works, the instrumentation costs ~48µs/token against tokens that take 3-30 *seconds*, so it's free in practice). That matters. This isn't a first PR from someone we've never seen.

**`inkling.c` (+1,255) is isolated** and that's the right shape — a separate file, like `olmoe.c`. `backend_cuda_ink.{cu,h}` likewise. If the whole thing were that, my only objection would be "nobody can run it".

**The part I'd check hardest is `c/tok.h` (+115/-1)** — 118 changed lines in the tokenizer `glm.c` depends on. That's the one place this PR can break the flagship engine, and it's the reason I'd want a real review rather than a wave-through. If the o200k support is purely additive (new tables, new branches, glm's path untouched byte-for-byte), it's defensible. If it refactors anything shared, that needs to be visible.

Concretely, that's testable **without** an Inkling checkpoint: if the tokenizer change is additive, the GLM oracle still scores token-exact against `glm_tiny/`, and `tests/test_stops.c` (which loads a synthetic tokenizer and checks the `special` flags) still passes. If you can show those green with your `tok.h`, you've retired the only risk I can actually name.

The remaining one is the maintainer's rule: he needs to run Inkling here before it lands. 975B is not a small ask, so if there's a smaller Thinking Machines checkpoint — or a way to build a tiny one with `convert_inkling_int4.py`, the way `make_glm_oracle.py` builds `glm_tiny/` — that's the path, and it would also give the engine an oracle test so it can't rot silently.

nbeerbower:
Totally get wanting to try it yourself before merging. Just wanted to get pen on paper mostly.

Made a separate PR #330 for just the tokenizer bit.

JustVugg:
The model is interesting and there are many things we can try to bring it into the hummingbird catalog. Now with kimi 3 it would also be interesting. The open world is going wild!

JustVugg:
@nbeerbower this is a serious piece of work and a second full engine is exactly where the project wants to go — thank you. One thing before merge, and it's the same bar the GLM engine holds itself to: **a tiny test model + oracle check we can run in CI and locally**.

For GLM we keep `glm_tiny` (a few-MB randomly-initialized model in the real container format) and the engine's `TF=1` teacher-forcing mode prints `N/32 vs oracle` — token-exactness against a `transformers` reference. That's what lets us merge refactors all day without fear. Since none of the maintainers can run the real Inkling 975B, without a tiny model your engine would be effectively untestable by anyone but you — and the first refactor that touches shared code (`tok.h`, Makefile) could silently break it.

I see you already ship `tools/make_tiny_inkling.py` — could you: (1) generate + commit (or attach) a tiny Inkling snapshot, (2) wire a `TF=1`-style oracle check (32-token teacher-forcing vs HF `transformers` on the same tiny model), and (3) add the one-liner to CI? With that in place this merges as-is — the code is additive and CI is already green. Happy to help with the CI wiring.

JustVugg:
Status check, and sorry for the wait. This conflicts with `dev` now — the engine was renamed and split in #391 (`glm.c` → `colibri.c` plus header modules) and `c/Makefile` and `c/tok.h` have both moved since. Are you still working on it?

On the substance, the structural choice here is right and worth saying out loud: `inkling.c` as a **separate engine** alongside `colibri.c`, sharing the container and tokenizer infrastructure but not entangling the GLM path, is exactly how a second model family should arrive. It's the same shape as the DeepSeek V4 work in #165, and it's why both are reviewable at all at their size.

Two things before I can take it seriously as a merge candidate:

1. **A tiny-model oracle.** `make_tiny_inkling.py` is there, but what settles a new engine in this project is a token-exact check against a `transformers` reference on a tiny checkpoint that lives in the repo — the way `glm_tiny` does. If `tools/make_tiny_inkling.py` already produces that and there's a teacher-forcing comparison, point me at the command; if not, that's the piece I'd want most.
2. **CI must build it.** A second engine that isn't compiled by `make check` rots within weeks. #165 wires its build and tiny-oracle into the `check` target; the same approach would work here.

If you're still on it, rebase and tell me about the oracle. If you've moved on, say so and I'll close it as groundwork with credit rather than leaving it open — no hard feelings either way, and the design note above stands regardless.

nbeerbower:
Still on it — rebased onto `dev` and pushed, so it's mergeable again. The `glm.c`→`colibri.c` split was a clean rebase: `inkling.c` / `backend_cuda_ink.*` are isolated new files, and the tokenizer commit dropped out entirely since o200k is already upstream (that was #330), so `tok.h` is now byte-identical to `dev`. The only real re-apply was the `Makefile` inkling target onto the new `$(GPUCC)` build.

On the two asks — both are in the PR now and green in CI:

**1. Tiny-model oracle.** It's the `glm_tiny` shape exactly. `tools/make_tiny_inkling.py` builds a small random-init `InklingForCausalLM` via HF `transformers` (8 layers, hitting every branch: sliding + global attention, log-length tau, dense + sparse MLP, 2 shared experts, unpadded vocab) and writes `ref_inkling.json`. The engine reproduces it token-for-token:

```
python3 tools/make_tiny_inkling.py tiny_inkling
SNAP=tiny_inkling ./inkling 8 0 tiny_inkling/ref_inkling.json
#  teacher-forced argmax: 36/36 match
#  Matching tokens: 24/24
```

`bits=0` keeps experts f32 so the forward is bit-exact; the greedy pass exits non-zero on any mismatch, so it's a real gate, not a print. Green in f32, int4-container+VNNI, `IDOT=0` scalar, and bf16-resident modes.

**2. CI builds it.** The `Engine (Linux, CPU)` job now `make colibri inkling`, the CUDA-syntax job also `nvcc -c backend_cuda_ink.cu`, and there's a new `inkling-oracle` job that pip-installs CPU torch + `transformers>=5.14` and runs exactly the two commands above. All three pass on this push. I kept the oracle out of the dependency-free `make check` gate on purpose — that one runs on Windows/macOS where I haven't confirmed the engine builds yet, so it's a Linux CI job instead. Happy to fold it into `make check` cross-OS as a follow-up once I've verified the Windows/macOS builds.

One honest scope note: the engine has grown well past this PR in my fork (`Schneewolf-Labs/sabrewing`) — GPU expert tier, int8 residents, MTP draft/verify, LoRA serving, per-phase profiling. I'd suggest landing this Stage-A→CUDA-warming version + oracle + CI first (reviewable at size, the shape you described), and I'll bring the perf/CUDA/MTP work as follow-up PRs rather than dropping another 10k lines at once. Sound right?

PR #315, opened 2026-07-16
CI: GitHub Actions for make check + the zero-warnings review gate

nbeerbower:
Automates the two review checks from CONTRIBUTING.md so every PR gets them before a human looks:

- **`make check`** on `ubuntu-latest` **and** `macos-latest` — the documented no-model, no-CUDA suite (portable CPU build + C unit tests + Python stdlib tests). The macOS job installs `libomp` so CI exercises the real OpenMP build rather than the single-thread fallback.
- **Zero-warnings gate** — `glm` and `olmoe` built at `ARCH=x86-64-v3`; any compiler warning fails the job, matching the "clean build (0 warnings)" review bar.

Deliberately out of scope (stays manual, as today): CUDA builds (no GPU on hosted runners) and the token-exact model oracles (tiny snapshots aren't in the repo — happy to add an oracle job in a follow-up if you'd like the fixtures checked in or fetched).

Validated locally on `dev`: `make check` 68/68 on Linux, both builds warning-free at x86-64-v3. Triggers: `pull_request` + pushes to `main`/`dev`.

https://claude.ai/code/session_016KLDCYJSyHxd39ChTr4T27

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+# CI: the checks CONTRIBUTING.md asks reviewers to run by hand.
+#  - `make check`: one portable CPU build + C unit tests + Python stdlib tests
+#    (no model download, no CUDA — same contract as the local target)
+#  - zero-warnings gate: the review bar is "clean build (0 warnings)", so the
+#    glm build fails CI if gcc emits any warning
+# CUDA and model-oracle validation stay manual: runners have no GPU and the
+# tiny oracle snapshots are not in the repo.
+name: CI
+
+on:
+  push:
+    branches: [main, dev]
+  pull_request:
+
+jobs:
+  check:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


JustVugg:
Thank you for this — closing it as a duplicate, and I want to be clear that's about timing, not quality.

We took @ZacharyZcR's #144 (opened first, same file `.github/workflows/ci.yml`, same jobs: engine + C suite, web, python, CUDA syntax). It's merged and green on both `dev` and `main` now.

Two things your version and his shared, which his first run exposed and we fixed on the way in — worth knowing if you write another workflow here:

- the pinned `Jimver/cuda-toolkit` tag's version table stops at 12.6.2, so `12.6.3` fails the install step before nvcc runs;
- `nvcc ... | head -40` takes its exit status from `head`, so the CUDA job could never fail. A check that can't fail is worse than no check — `backend_cuda.cu` is compiled by nothing else in this repo.

That the same idea arrived from three people in three days is a fair signal we should have had CI sooner. Sorry it cost you the work. If you want to keep contributing to the checks, the gap that's still open is cross-platform coverage — #144 is Linux-only.

Issue #310, opened 2026-07-16
Make the runtime path Python-free by porting the CLI and API gateway to Go


Magnet-js:
## Idea

Right now the runtime path (`coli` CLI and `openai_server.py`) requires Python, even though the engine itself is pure C. I'd like to hear what the maintainer and the community think about porting just these two components to Go, while keeping Python for the offline tooling where it actually makes sense.

## What would change

- `coli` CLI rewritten in Go: a single static binary for process management, argument parsing and env setup
- `openai_server.py` rewritten in Go: HTTP, SSE streaming and JSON are covered by the standard library (`net/http`), no external dependencies needed

## What would stay in Python

- `convert_fp8_to_int4.py`: depends on torch and safetensors, reimplementing FP8 dequantization with block scales in Go would be a huge effort with high risk for a step that needs bit-exact results
- `coli bench`: needs the HuggingFace `datasets` and `tokenizers` packages, there is no Go equivalent
- Oracle generation and validation tools: their whole purpose is comparing against the transformers reference, so they have to stay in Python

## Expected benefits

1. **Zero-dependency runtime for end users.** Anyone downloading the pre-converted int4 model from Hugging Face (the recommended path) would need no Python at all. Just the C engine plus one Go binary.
2. **Consistent with the project philosophy.** The README already says "the engine at runtime is pure C". This would extend that promise to the whole runtime path.
3. **Simpler deployment.** A static binary makes packaging, containerizing and distributing much easier, no interpreter version issues, no pip.
4. **Better concurrency for the gateway.** Goroutines handle many parallel SSE connections more naturally than threading under the GIL. Not a bottleneck today since the engine serves one generation at a time, but it removes a future limitation.

## Honest downsides

- A third language and toolchain in the repo (C + Go + Python instead of C + Python)
- The current Python server is deliberately minimal (~1100 lines, stdlib only) and works well, so this is not fixing a bug
- RAM savings would be negligible (~20 MB in a project where the engine uses 10-20 GB)

## Questions

- Is a Python-free runtime path something you consider valuable for this project?
- Would Go be acceptable as a language, or would you prefer keeping it to C + Python?
- If there is interest, I'd be happy to help with a prototype of the gateway first, since it has a well-defined protocol boundary (the SUBMIT/DATA/DONE lines over stdin/stdout).

Curious to hear your thoughts before anyone writes code, as suggested in the discussions welcome post.
(I would like to do the work (without AI))

[...]

Issue #198, opened 2026-07-14
Reproducing Windows (native) installation


davidfcsilva:
Hi,

While following the steps, I have tried to install this tool, but got two issues that I can't work around,
1)  Running the test
PS F:\colibri\c> SNAP=./glm_tiny TF=1 ./glm.exe 64 16 16
SNAP=./glm_tiny : The term 'SNAP=./glm_tiny' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At line:1 char:1
+ SNAP=./glm_tiny TF=1 ./glm.exe 64 16 16
+ ~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (SNAP=./glm_tiny:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
2) Compiling NVIDIA CUDA
PS F:\colibri\c> make cuda-dll CUDA_HOME="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.3" CUDA_ARCH=sm_86
The system cannot find the path specified.
'{' is not recognized as an internal or external command,
operable program or batch file.
make: *** [makefile:149: cuda-dll] Error 1

Pretty sure it's just something missing, I have installed Microsoft 2019 via choco
Installed scoop requirements
I have python installed
All is in the PATH
If anyone knows how to fix it please let me know so I can move this forward.

In Linux run's fine:

dsilva@k8s:~/colibri/c$ COLI_MODEL=../../llm ./coli chat

     ▄▀▀▀▄  ▄        colibrì v1.0
  ▄▄▄▄▀▀▀▀▄▀▀        tiny engine, immense model
      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU
        ▀▀▀▀         chat · llm · ram -GB · topp off
          ▀
  ──────────────────────────────────────────────────────────
  type and press Enter · :more continues · :reset clears memory · :q exits

davidfcsilva:
Just adding additional information

PS F:\colibri\c> python3.14.exe coli chat --model F:\llm\

     ▄▀▀▀▄  ▄        colibrì v1.0
  ▄▄▄▄▀▀▀▀▄▀▀        tiny engine, immense model
      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU
[...]

PR #320, opened 2026-07-16
docs: add winget python one-liner to Windows install block (#310)

Magnet-js:
## Summary

Interim Windows fix discussed in #310: the `coli` CLI and `openai_server.py` need a real `python3` interpreter, but a fresh Windows resolves `python` to the Microsoft Store alias stub that opens the Store instead of running anything (the trap from #198).

This adds a documented `winget install` one-liner to the Windows toolchain install block in the README, as suggested by @ZacharyZcR:

```powershell
winget install -e --id Python.Python.3.12
```

Installing the real interpreter replaces the alias stub, so the existing `python coli chat/serve` commands in the same block work out of the box.

## Test plan

- [x] Docs-only change; verified the surrounding README block renders correctly
- [x] `winget install -e --id Python.Python.3.12` is the exact-match package ID from the winget community repo

Refs #310, #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,11 @@ GCC 16.1.0 (x86_64-ucrt-posix-seh).
 scoop install mingw-winlibs                    # portable, no shell needed
 # or: pacman -S mingw-w64-x86_64-gcc make     # via MSYS2
 
+# Python (needed by the `coli` CLI and API server). A fresh Windows resolves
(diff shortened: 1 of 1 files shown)


Magnet-js:
@JustVugg Im unsure if the mian branch is correct. 
Please let me know if I did something wrong. 🙏

Issue #306, opened 2026-07-16
Benchmark + Windows GPU-build fixes: GLM-5.2 on Core Ultra 9 285K + RTX 5080 (sm_120)


mohamedmastouri2000-boop:
<html>
<body>
<h1 class="text-text-100 mt-3 -mb-1 text-[1.375rem] font-bold" data-sourcepos="1:1-1:88;0-87">Benchmark + Windows GPU-build fixes: GLM-5.2 on Core Ultra 9 285K + RTX 5080 (sm_120)</h1>

<p class="font-claude-response-body break-words whitespace-normal" data-sourcepos="3:1-3:324;89-412">First — thank you for this project. Running a 744B frontier model correctly on a consumer desktop is genuinely remarkable, and the docs/tooling (iobench, resource_plan, warmup, doctor) made bring-up smooth. Sharing a new hardware data point you don't have in the table yet, plus two Windows build fixes and one observation.</p>

<h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold" data-sourcepos="5:1-5:10;414-423">System</h2>

<ul class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-disc flex flex-col gap-1 pl-8 mb-3" data-sourcepos="6:1-9:101;424-796">
<li class="font-claude-response-body whitespace-normal break-words pl-2" data-sourcepos="6:1-6:98;424-521"><strong>CPU:</strong> Intel Core Ultra 9 285K (Arrow Lake, 24C, AVX-VNNI) — banner confirms <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">idot: avx-vnni</code></li>
<li class="font-claude-response-body whitespace-normal break-words pl-2" data-sourcepos="7:1-7:83;522-604"><strong>GPU:</strong> RTX 5080 16GB (Blackwell, <strong>sm_120</strong>), driver 610.62, CUDA Toolkit 13.3</li>
[...]

PR #298, opened 2026-07-16
cuda+engine: full fmt=4 (grouped int4 gs=64) support + diagnostic harness

woolcoxm:
## Summary

Adds complete support for the new grouped-int4 (fmt=4, gs=64) model format across the CUDA backend, engine dispatch, and dequant helpers. Also adds a fused gate+up kernel for fmt=4 and a comprehensive diagnostic harness.

## Problem

The new g64 model quantizes experts with group-size 64 — one f32 scale per 64 elements instead of one per row. The existing code only supported fmt=2 (per-row int4). Three things broke:

1. **CUDA completely broken** — `row_bytes(fmt=4)` returned 0, so every tensor upload failed silently → every dense tensor fell back to CPU → **0.05 tok/s** (20x regression)
2. **No fused gate+up** — `matmul_i4_pair` gates on `fmt==2` only, so fmt=4 did 2 separate passes → expert-matmul 2x slower
3. **Missing fmt=4 branches** in `embed_row`, `qt_addrow`, `qt_matvec_rows`, kv_b shard computation — latent correctness bugs

### CUDA backend (`backend_cuda.cu`)
- `ColiCudaTensor`: added `gs`/`ng` fields
- `row_bytes`/`weight_at`: fmt=4 = same packed int4 layout as fmt=2
- `quant_matmul` kernel: per-group scale application for fmt=4
- `tensor_upload`/`tensor_update`: allocate `O*ng` scales for fmt=4
- `tensor_free`/`tensor_bytes`: VRAM accounting uses `O*ng`
- `expert_group`: returns 0 for fmt=4 (falls back to correct per-expert path)
- All 11 `quant_matmul<<<>>>` call sites pass `gs,ng`

### Engine (`glm.c`)
- **`matmul_i4_grouped_pair`** (new): fused gate+up for fmt=4 — reads x once instead of twice
- `expert_gate_up`: dispatches to fused grouped pair for fmt=4
- `run_text`: prepend `[gMASK]<sop>` for GLM models (#108 — PROMPT mode generated garbage without it)
- `embed_row`, `qt_addrow`, `qt_matvec_rows`: fmt=4 branches
- kv_b shard `rb`: fmt=4 case

### API (`backend_cuda.h`, `backend_loader.c`)
- `coli_cuda_tensor_upload` and `coli_cuda_matmul`: added `gs` param, threaded through DLL boundary

### Tests
[...]

PR #339, opened 2026-07-17
AMD GPU support: HIP/ROCm via single-source backend_gpu_compat.h (+ WMMA dispatch gate)

noobdev-ph:
## Summary

The AMD half of #112, rebuilt against `dev` as requested — one thing only: **the existing CUDA backend compiles unchanged for AMD GPUs via ROCm** (`make HIP=1`), through a 62-line mapping header, the same one-shim pattern `compat.h` uses for the Windows port. 5 files, +218/−15; `backend_cuda.cu` itself changes by 7 lines; `glm.c` untouched.

- `backend_gpu_compat.h`: under nvcc, a pass-through to `cuda_runtime.h`+`mma.h` — **the NVIDIA path's preprocessed source is byte-identical to today's**, which your CI can verify. Under hipcc, the CUDA runtime surface maps to HIP 1:1.
- **WMMA gating** (`COLI_GPU_HAS_WMMA`) — this part matters beyond AMD: the tensor-core kernel *bodies* are `__CUDA_ARCH__ >= 700`-guarded, but the host dispatch checks `compute_major >= 7` at runtime. gfx GPUs report `compute_major = 12`, so without the compile-time gate the dispatch selects **empty kernel bodies and silently returns garbage** on any non-CUDA target. HIP uses the portable kernels; rocWMMA is a natural follow-up.
- Makefile: `HIP=1`/`HIP_ARCH`, `hip-test`, `gpu-compile` (CI compile-only), a `GPUCC` indirection so all GPU rules stay single-path, and HIP flags joined to your new `.build-config` stamp so HIP↔CPU flag flips relink correctly (#306's fix extended, not bypassed).
- CI: an `engine-hip-syntax` job mirroring your `engine-cuda-syntax` (rocm/dev container — tag pinned and disk pre-freed; `latest` no longer fits hosted runners, details in the job comments).

## On "verified rather than trusted" — the honest position

Hosted CI proves compilation; runtime verification necessarily comes from community AMD hardware — the same epistemology as the README's benchmark table. What this PR offers concretely: (1) the CUDA path is *provably* unchanged (byte-identical preprocessing — verifiable by anyone); (2) `hip-test` runs the full upstream test suite (incl. `tensor_update`) on my RX 9070 XT — pass, 5/5 repeat runs; (3) end-to-end: GLM-5.2 at **0.33 tok/s / 60% expert hit with `CUDA_RELEASE_HOST=1`** on this `dev` rebase (CPU: 0.22–0.27 same machine) — full setup documented in `GPU_BACKENDS.md` so any AMD owner can reproduce; (4) I commit to running `hip-test` + an end-to-end check against `dev` on request and on releases, and to first-responding AMD issues. If you'd rather see a second independent AMD datapoint before merging, keeping this open until one arrives is a reasonable call — the branch is maintained either way.

## Validation

- [x] `make -C c check` — CPU untouched
- [x] `hip-test` on gfx1201 (ROCm 7.2.4): full suite pass
- [x] nvcc path: byte-identical preprocessed source (pass-through include)
- [x] `engine-hip-syntax` runs in your CI structure

## Compatibility

- [x] Default CPU build dependency-free; `HIP=1` opt-in, Linux-only, mutually exclusive with `CUDA=1`, errors early
- [x] No model files, binaries, or artifacts

Companion PR: #338 (vendor-neutral backend hardening — independently mergeable).


JustVugg:
Rebase needed, and this one is substantive: #298 just merged into `dev` and reworked `backend_cuda.cu`'s kernels — fmt=4 per-group scales (`absorb_scale`, `gs`/`ng` parameters on `quant_matmul` and the attention kernels). Since your HIP compat layer mirrors that file, the same per-group logic needs to flow through `backend_gpu_compat.h`, otherwise ROCm inherits the exact garbage-output bug #298 fixed on g64 containers. You're the right person for that port — shout if anything in #298's diff is unclear.

noobdev-ph:
Rebased onto `dev` (`fe6c164`) — and the good news is that #298 needed **no compat-layer work at all**.

**Why ROCm doesn't inherit the g64 bug.** The compat layer doesn't mirror `backend_cuda.cu` — it's a symbol-mapping header, and the kernels compile *unchanged* for both vendors. So #298's per-group logic (`absorb_scale`, `gs`/`ng`) flows through to ROCm automatically; there's no second copy that can drift. Concretely: all 33 CUDA runtime symbols `dev` now uses were already mapped, so `backend_gpu_compat.h` is untouched by this rebase. That single-source property is the main thing the design buys — a fix like #298 lands on AMD the moment it lands on NVIDIA, with no porting step and nothing to keep in sync.

Verified on an RX 9070 XT (gfx1201 / RDNA4, ROCm 7.2): the full kernel suite **including the fmt=4 per-group coverage** passes, 3/3 stable runs; engine builds; `make check` green.

Only one merge conflict, resolved as a union: the `COLI_GPU_HAS_WMMA` compile-gate now sits alongside #431's new row-threshold check on the W4A16 branch, so both guards apply.

---

**One thing you'll want independent of this PR: `tests/test_backend_cuda.cu` doesn't currently compile on `dev`.**

#298 added the trailing `gs` parameter to `coli_cuda_matmul` and split `tensor_upload` / `tensor_upload_g`, but **6 `coli_cuda_matmul` call sites and 7 `tensor_upload` call sites** were left on the old signatures (several in the tests that came in with #338). It fails identically under nvcc and hipcc — it's a plain C++ arity mismatch, not anything AMD-specific:

```
tests/test_backend_cuda.cu:52:10: error: no matching function for call to 'coli_cuda_matmul'
```

It's invisible to CI because `engine-cuda-syntax` compiles `backend_cuda.cu` alone and never builds the test binary — so the kernel-correctness test has been silently unbuildable since #298 merged. I've repaired all 13 sites in this branch (`gs=0` / the `_g` variant), but you may want the same fix on `dev` directly since it affects CUDA users too, and it means the fmt=4 kernels shipped without their test actually running.

If it's useful, the `gpu-compile` target in this PR is what surfaced it — it builds the backend **and** the test binary without needing a GPU, so it catches exactly this class of breakage in CI. Wiring that into `engine-cuda-syntax` (one line: `make -C c gpu-compile CUDA_ARCH=sm_80`) would close the gap for both vendors regardless of what happens to the rest of this PR.

---

Fair warning on my side: I'm out of development hours for a while, so I won't be able to keep chasing `dev` on this one. The branch is green and mergeable as of now, and `GPU_BACKENDS.md` documents the build/validation path if anyone wants to pick it up. Thanks for the review throughout — #338 landing was a genuine pleasure to see.

JustVugg:
To merge the ROCm backend I need two things. **(1) Rebase** on current `dev` (CI is green; it moved a lot — #298 reworked the CUDA kernels' fmt=4 per-group scales). **(2)** Since your HIP layer mirrors `backend_cuda.cu`, it must carry #298's per-group scale logic (`absorb_scale`, gs/ng), otherwise ROCm inherits the exact garbage-output bug #298 fixed on g64 containers — so a **fmt=4 (gs64) run on real AMD hardware showing coherent output**, token-exact vs CPU if possible, is the gate. Get CI green + that datapoint and it lands. Thanks @noobdev-ph.

noobdev-ph:
Both gates met. 🐦

## (1) Rebased on current `dev` — CI green

Merged the 38 new commits (`65be434`). No conflicts this round, and again **zero compat-layer work**: no new CUDA runtime symbols needed mapping, so `backend_gpu_compat.h` is untouched. All nine checks pass — `mergeStateStatus: CLEAN`.

## (2) fmt=4 (gs64) on real AMD hardware — token-exact vs CPU

I built an actual **gs64 container** rather than reasoning about it: the repo's own tiny oracle (`tools/make_glm_oracle.py`, real `glm_moe_dsa` architecture) converted with `tools/convert_fp8_to_int4.py --group-size 64` → `[PLAN] ... grouped gs=64 (fmt=4)`. Then CPU vs RX 9070 XT (gfx1201 / RDNA4, ROCm 7.2) on that container:

| test | result |
|---|---|
| Teacher-forcing, resident dense on GPU (`CUDA_DENSE=1`) | **every predicted token identical to CPU** |
| Teacher-forcing, **routed experts in VRAM** (`CUDA_RELEASE_HOST=1`, host copies freed) | **token-exact** |
| Generation mode (not just TF) | **identical output** |
| 3 repeat GPU runs vs the CPU baseline | **3/3 token-exact** |

Both paths that touch per-group scales are covered: the resident-dense fmt=4 tensors *and* `coli_cuda_expert_mlp` with the experts actually resident in VRAM and their host copies released — i.e. the g64 configuration your bug was about, with no CPU copy available to mask a wrong result.

**Control (the part that makes the above mean something):** re-running with `COLI_GPU_FAIL_AFTER=0` forces every GPU call to fail and reports **33 live GPU calls** being disabled. So the GPU genuinely executed those matmuls — the token-exactness isn't a silent CPU fallback.

**One honest limitation:** the tiny oracle has random weights, so I can't show *coherent prose* from it — it scores 7/32 against the FP32 oracle purely from int4 quantization noise, and CPU and GPU agree on that number and on every individual token. I no longer have the 744B container locally (reclaimed the disk), so a coherent-text run would need someone with the full model. That said, for the specific failure mode you were guarding against, token-exact CPU↔GPU on a real fmt=4 gs64 container is the stricter test: garbage output would show up as *any* divergence, and there is none.

## Also in this branch

The 13 stale call sites in `tests/test_backend_cuda.cu` are still repaired here (`gs=0` / `_g` variant) — `dev` still has them, so the kernel-correctness test remains unbuildable there. The `gpu-compile` target is what catches it; wiring it into `engine-cuda-syntax` (`make -C c gpu-compile CUDA_ARCH=sm_80`) would close that hole for CUDA too.

Reproduce the whole thing on any ROCm box:

```sh
python tools/make_glm_oracle.py
python tools/convert_fp8_to_int4.py --indir ./glm_tiny --outdir ./glm_tiny_gs64 --group-size 64
SNAP=./glm_tiny_gs64 TF=1 ./colibri 64 4 4                      # CPU baseline
HIP_VISIBLE_DEVICES=0 COLI_CUDA=1 COLI_GPU=0 CUDA_EXPERT_GB=1 CUDA_RELEASE_HOST=1 \
  PIN=stats.txt PIN_GB=0.05 SNAP=./glm_tiny_gs64 TF=1 ./colibri 64 4 4   # AMD GPU
```

Over to you — happy to answer anything, and thanks for the clear gate. It made this a short hop rather than a guessing game.

JustVugg:
Merged. The COLI_GPU_FAIL_AFTER=0 control is what made this land immediately — proving the GPU actually executed those 33 calls rules out the silent-CPU-fallback trap that would make token-exactness meaningless. Combined with token-exact on a real fmt=4 gs64 container in both the resident-dense and experts-in-VRAM configurations, that is a stricter check for the #298 failure mode than coherent prose would have been. Thanks for being explicit about the random-weight limitation too — that honesty is why I trust the rest. AMD is now a supported backend; if anyone with a full 744B container on ROCm wants to post a coherent-text run and tok/s, that would round it out nicely.

noobdev-ph:
Quick status so this doesn't look stalled: **I'm re-downloading the full 744B int4 container to close the one gap in the evidence above** — the coherent-output run on real weights.

When I posted the gs64 validation I'd already reclaimed the ~363 GB model from disk, so the token-exact CPU↔GPU comparison was done on the tiny `glm_moe_dsa` oracle. That's the stricter test for *numerical* correctness (any divergence would show), but it can't demonstrate coherent text, since the oracle has random weights. Rather than leave that caveat sitting on the PR, I'm pulling the model back down and will run the real thing on the RX 9070 XT.

Downloading now (~11 GB of ~363 GB in, roughly 2 hours out). Once it's in place I'll post:

- **coherent generation from the 744B model on AMD**, with the routed experts resident in VRAM (`CUDA_RELEASE_HOST=1`) — the configuration your g64 concern was actually about
- a **CPU vs GPU comparison on the same prompt** on real weights, so the numerical claim is backed on the full model too, not just the oracle
- **fresh throughput numbers** on the current tree — the last figures I published (0.33 tok/s, 61% expert hit) predate ~38 commits of dev GPU work, so they're stale and I'd rather give you current ones

No changes to the branch are needed for any of this — it's `CLEAN` and mergeable as it stands, and the rebase + CI + gs64 token-exactness already reported are unaffected. This is purely additional evidence.

I'll follow up here with the results.

JustVugg:
That's exactly the right instinct, and rare — re-downloading 363 GB to close a caveat you flagged yourself, *after* the PR already merged. The token-exact-on-the-oracle evidence was already the stricter numerical test (which is why it merged), so nobody would have chased you for the coherent-text run; doing it anyway is what makes the AMD support trustworthy rather than just green. Take your time. When it lands, the three things you listed — coherent generation with experts resident in VRAM, CPU-vs-GPU on real weights, and fresh throughput on the current tree — would be worth their own datapoint issue so ROCm users have a reference. Thanks 🐦

noobdev-ph:
Closing the loop with the full model. I rebuilt the 744B checkpoint locally as a **real fmt=4 gs64 container** (`convert_fp8_to_int4.py --group-size 64`, from `zai-org/GLM-5.2-FP8` → `grouped gs=64 (fmt=4)`, 141 shards + int8 MTP) and ran it on the RX 9070 XT (gfx1201 / RDNA4, ROCm 7.2.4). Three pieces of evidence, strongest last.

**1. Coherent *and executable* output from fmt=4 gs64.** Gave it a code-gen prompt (a token-bucket rate limiter with a self-test). It generated 603 tokens of clean Python — and the generated code **runs and passes its own four assertions**:

```
$ python3 ratelimiter_gs64.py
All tests passed
```

That's provably-correct output, not just plausible tokens — burst capacity, lazy refill after 1s, cap-never-exceeded, oversized-cost-denied all assert true.

**2. The GPU actually computed the fmt=4 grouped experts.** On a follow-up prompt with the expert tier resident in VRAM (`CUDA_EXPERT_GB=12 CUDA_RELEASE_HOST=1`), the engine reports:

```
CUDA expert tier: 565 resident experts (12.00 GB) | 607 calls served from VRAM
```

So this isn't the dense path with experts on CPU — 565 grouped-int4 (gs=64) experts were resident on the AMD card and 607 grouped-expert matmuls were served from VRAM.

**3. Token-identical to CPU on real weights.** Same prompt, greedy, `DRAFT=0`, CPU vs the GPU run above — the generated text is **byte-identical** (the only diff in the capture was the `CUDA expert tier:` diagnostic line itself):

> Sunlight is composed of all colors, but it's scattered in all directions by gases and particles in Earth's atmosphere. Blue light is scattered more than other colors because it travels as shorter, smaller waves

Identical on both paths — so on the full model, in fmt=4 gs64, the GPU's grouped-expert output matches the CPU exactly. Combined with the earlier tiny-oracle result (token-exact fmt=4 across the whole forward), the g64 garbage-output concern is covered from both ends: strict numerical equivalence on the oracle, and coherent + token-identical + GPU-computed on the real 744B model.

Throughput, for reference (fresh model, cold learned-cache — hit-rates climb with use): ~0.20–0.30 tok/s, 30–43% expert hit on this Gen3-capped 5700G + RX 9070 XT. Correctness was the point here, not speed.

CI is green and the branch is `CLEAN`. This is the datapoint I couldn't give you before (needed the full model); it's now on the record.

noobdev-ph:
@JustVugg — congrats on the merge, and thanks for shepherding it in. 🎉

Flagging the comment just above in case it's useful beyond this PR: I finally got the **full 744B model on real AMD hardware** (RX 9070 XT / gfx1201, fmt=4 gs64) — coherent, self-tests pass, and the GPU's grouped-expert output is **token-identical to CPU** with 565 experts resident in VRAM. That's the datapoint I couldn't produce earlier without the full checkpoint.

If you keep a hardware/benchmark table, happy for it to serve as the first RDNA4 row — and I'm glad the `COLI_TEMP` fix (#509) was the one that landed; @ab2m's root-cause approach was the right call.

PR #346, opened 2026-07-17
docs: add Traditional Chinese README

bm1016bm-svg:
## What changed

- add a complete Traditional Chinese README localized for readers in Taiwan
- add an English / Traditional Chinese language switcher to both READMEs
- preserve commands, URLs, benchmark values, product names, and code examples

## Why

Colibrì documents a technically unusual workflow spanning model conversion, memory tiers, runtime planning, and several launch modes. A native Traditional Chinese README makes that workflow accessible without changing any runtime behavior.

## Validation

- `git diff --cached --check`
- verified all relative links in both README files resolve
- verified the English and Traditional Chinese versions have matching heading and code-fence structure
- verified all commands and URLs remain intact

This is a documentation-only change. No engine, build, dependency, or runtime behavior is modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
   <img src="assets/colibri.svg" width="500" alt="colibrì — tiny engine, immense model">
 </p>
 
+<p align="center">
+  English · <a href="README.zh-TW.md">繁體中文</a>
+</p>
+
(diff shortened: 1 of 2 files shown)


wzxzhuxi:
Everyone can just communicate in English; AI can easily translate the README into whatever language they use

Issue #292, opened 2026-07-15
Diagnostic: token-exactness sweep + performance breakdown (pipe2 S=1 resident vs CPU vs CUDA — 4 prompts x 5 configs)


woolcoxm:
### TL;DR

Comprehensive sweep on the actual hardware (RTX 5070 Ti 16 GB, GLM-5.2 744B int4, Core Ultra 9 185H AVX-VNNI, 32 GB RAM, QLC NVMe). **The pipe2 S=1 resident pipeline introduces zero additional token divergence** vs the CUDA-off-pipe path — it's a pure execution-location optimization, not a math change. However, **CUDA offload itself (attention + dense on GPU) diverges from the pure-CPU path on 2 of 4 prompts** at greedy/TEMP=0. The engine is fully deterministic run-to-run. Full performance breakdown with optimization targets below.

---

### Setup

- **4 prompts** (factual / code / creative / narrative): "Summarize Romeo and Juliet", "Write a Python factorial function", "Explain ocean currents", "Opening paragraph of a mystery novel"
- **4 configs**, same binary (`windows-optimizations` @ `e56d483`), same model, `.coli_usage` restored before every run:
  - **BASE** — pure CPU, no CUDA, no pipe
  - **CUDAOFF** — CUDA attention + dense on GPU, pipe2 OFF (S_MIN=8)
  - **CUDARES** — CUDA attention + dense, pipe2 resident at decode (S_MIN=1) ← the PR #274 question
  - **DIRECT** — CUDA + DIRECT=1 PIPE=1 PIPE_WORKERS=8 REPIN=16
- `TEMP=0` (greedy) throughout — exactness = argmax reproduced at every position
- `TOKENS=1` dumps generated token ids to stderr (commit `e56d483`, opt-in)

### Results

| prompt | CUDAOFF vs BASE | CUDARES vs BASE | DIRECT vs BASE | CUDARES vs CUDAOFF |
|---|---|---|---|---|
| romeo | ✅ exact | ✅ exact | ✅ exact | ✅ exact |
| ocean | ✅ exact | ✅ exact | ✅ exact | ✅ exact |
| factorial | ❌ diverges @ tok 4 | ❌ diverges @ tok 4 | ❌ diverges @ tok 4 | ✅ exact |
| story | ❌ diverges @ tok 6 | ❌ diverges @ tok 6 | ❌ diverges @ tok 6 | ✅ exact |

[...]

PR #357, opened 2026-07-17
sampling: partial top-keep select in attention_rows DSA — O(nk) quickselect, not O(nk log nk) qsort (#356)

### LINKED ISSUE #356: sampling: DSA top-keep full-qsorts nk scores per layer per token — partial-select is ~10-24x cheaper (measured)
## What

`attention_rows()` finds the top-`index_topk` (2048 for GLM-5.2) context keys to attend to by **full-qsorting all `nk` attention scores** on every call, then reading the threshold off `tmp[keep-1]`. `nk` is the current context length, so this is `O(nk · log nk)` **per layer × per token** in decode, for every conversation that has grown past `index_topk`.

```c
// glm.c:2589-2596  (DSA lightning indexer, per-position selection)
float *tmp=falloc(nk); memcpy(tmp,isc,nk*sizeof(float));
qsort(tmp,nk,sizeof(float),cmp_fdesc);          // <-- full O(nk log nk) sort
float thr=tmp[keep-1];
int *dst=m->dsa_sel+(int64_t)s*dtopk, nd=0;
for(int t=0;t<nk && nd<keep;t++) if(isc[t]>thr) dst[nd++]=t;   // tie-break by position
for(int t=0;t<nk && nd<keep;t++) if(isc[t]==thr) dst[nd++]=t;
```

The sort exists only to find the `keep`-th largest score (the **threshold**); the two scans that follow pick the kept positions **in position order**, which is what fixes tie-ordering. So the qsort is doing far more work than the contract needs: it fully orders `nk` elements to extract one pivot.

This is the same class of inefficiency as **#335** (sampling `dist_build` full-vocab qsort → heap partial-select), but on the attention side, and it scales with **context length** rather than vocab size — so the cost *grows over the lifetime of a conversation*.

## Measured

Standalone microbenchmark (compiled `-O3 -march=native`, median of 2000 reps, `keep = index_topk = 2048`), old qsort vs a quickselect partial-select that partitions so the `keep` largest land in `[0,keep)`:

| nk (ctx len) | qsort (old) | partial-select (new) | speedup |
|---|---|---|---|
| 2049 | 111,000 ns | 5,200 ns | **21×** |
| 4096 | 260,700 ns | 18,200 ns | 14× |
| 8192 | 616,900 ns | 60,300 ns | 10× |
| 16384 | 1,340,000 ns | 68,500 ns | 20× |
| 32768 | 2,840,700 ns | 289,200 ns | 10× |
| 65536 | 6,374,300 ns | 260,700 ns | **24×** |

At a 32K-token conversation × 78 layers, the qsort path alone costs roughly **2.8 ms × 78 ≈ 220 ms per generated token** — time spent sorting an array whose only purpose is to yield a single threshold value. The partial-select drops that to single-digit ms total.

The gap widens with context because qsort is `O(nk · log nk)` while quickselect is `O(nk)` average. DSA only activates past `index_topk` (2048), so this bites precisely on long conversations — the regime where decode latency matters most to the user.

## Correctness contract

This one is actually **cleaner than #335**. The qsort is **unstable** and only the threshold value (`tmp[keep-1]`) is read from the sorted array; the actual kept-position *set* is determined entirely by the two position-order scans:

```c
for(int t=0;t<nk && nd<keep;t++) if(isc[t]>thr) dst[nd++]=t;   // strictly-above first
for(int t=0;t<nk && nd<keep;t++) if(isc[t]==thr) dst[nd++]=t;  // then ties, in position order
```

So **any** algorithm that produces the correct threshold value (the `keep`-th largest score, with ties resolved by the position scan) yields a **bit-identical** `dst[]` to the current code. The quickselect pivot *is* by definition the `keep`-th largest element, so the new threshold equals the old `tmp[keep-1]` exactly. No multiset relaxation needed (unlike #335, which could only promise multiset equality because the heap was unstable and changed accumulation order).

The only thing to handle is the `keep`-th value when there are ties at the boundary — but the position-order `==thr` scan already makes the kept set deterministic given `thr`, so as long as the threshold value is identical, the kept set is identical.

## Proposed fix

Replace the `qsort` with a `partial_select_desc()` quickselect (Hoare partition, desc) that moves the `keep` largest into `tmp[0..keep)` in `O(nk)` average, then take `thr = min(tmp[0..keep))`:

```c
static void partial_select_desc(float *a, int n, int k);
...
float *tmp=falloc(nk); memcpy(tmp,isc,nk*sizeof(float));
partial_select_desc(tmp,nk,keep);            /* O(nk) avg, not O(nk log nk) */
float thr=tmp[0]; for(int t=1;t<keep;t++) if(tmp[t]<thr) thr=tmp[t];
```

The two position-order scans are unchanged → identical tie handling → identical `dst[]`.

`cmp_fdesc` becomes unused after this; I'll leave the definition in place (the existing `-Wno-unused-function` keeps the build clean, and the test/bench reuse the same comparator convention).

## What I checked and did NOT file (so it isn't re-investigated)

During this review I measured three other suspected hotspots and found them **already optimal at GLM-5.2 scale** — recording here so nobody else spends the time:

- **MoE top-K selection** (`glm.c:2913-2917`) is `O(K·E)` repeated linear-max-with-skip. At E=256, K=8 that's 2048 comparisons, and a branch-predictable linear scan measures **~3 µs** vs **~10.5 µs** for a full argsort. **Changing it would be a regression.** (The repeated scan wins because K·E ≪ E·log E at these sizes.)
- **`softmax`** (`glm.c:1292`) is 3 passes (max, exp+sum, divide). Precomputing `1/s` (2-pass) measures **within 1%** — the compiler already hoists the loop-invariant reciprocal. Not worth touching.
- **`tok_encode`** (`tok.h:247`) looks `O(L²)` at first glance but is **`O(L·nsp·s)`** — the inner position loop breaks on the first special-token hit and the outer loop breaks when none is found. Linear in text length.

## Secondary (not in the primary PR)

The `CACHE_ROUTE=1` opt-in path (`glm.c:2834-2850`) does `O(Mwin²·E)` repeated linear-max-with-skip scans when building `rank_buf`. Non-default and experimental — worth a separate follow-up issue, not bundled with this hot-path fix.

## Deliverables

- [ ] `partial_select_desc()` in `glm.c`, replacing the qsort in the DSA path
- [ ] `c/tests/test_dsa_select.c` — element-wise identical kept-set across shapes/sizes (stronger than test_topp's multiset check)
- [ ] `c/tests/bench_dsa_select.c` — old qsort vs new partial-select, the table above reproduced in-repo
- [ ] Makefile wiring (test in `TEST_BINS`, bench on-demand like `bench_topp`)

/cc reference: #335 (the sampling-side precedent this mirrors)

woolcoxm:
## Fixes #356

The DSA lightning indexer in `attention_rows()` selects the top-`index_topk` (2048 for GLM-5.2) context keys to attend to. It previously did this by **full-qsorting all `nk` attention scores** (`O(nk log nk)`) every layer, every token — just to read a single pivot value, the `keep`-th largest score, used as the threshold. It then scanned the *original* (unsorted) array in position order to build the kept set.

This is the attention-side twin of **#335/#354** (sampling `dist_build` full-vocab qsort → heap partial-select), but it scales with **context length**, so the cost grows over the lifetime of a conversation and bites precisely in the long-context regime where decode latency matters most.

## The fix

Replace the qsort with `partial_select_desc` (Hoare quickselect, median-of-three, descending) — `O(nk)` average to partition the `keep` largest into `tmp[0..keep)`, then `thr = min` of that block:

```c
float *tmp=falloc(nk); memcpy(tmp,isc,nk*sizeof(float));
partial_select_desc(tmp,nk,keep);            /* O(nk) avg, not O(nk log nk) */
float thr=tmp[0]; for(int t=1;t<keep;t++) if(tmp[t]<thr) thr=tmp[t];
```

The two position-order scans (`>thr` then `==thr`) are **unchanged** → the kept-position set is bit-identical.

## Correctness contract — stronger than #335

The qsort is unstable and only the threshold `tmp[keep-1]` is read from the sorted array; the actual kept-position *set* is determined entirely by the position-order scans, which are byte-for-byte identical before and after this PR. The quickselect pivot **is**, by definition, the `keep`-th largest element, so the new threshold equals the old `tmp[keep-1]` exactly. Therefore the kept set is **element-wise identical** — not just multiset-equal (which is all #335's unstable sampling heap could promise, since it also changed accumulation order).

## Measured — `bench_dsa_select` (median of 2000 reps, `keep=2048`)

| shape | nk (ctx) | old qsort | new quickselect | speedup |
|---|---|---|---|---|
| realistic | 2049 | 118 µs | 6.4 µs | 18× |
| realistic | 8192 | 629 µs | 43 µs | 15× |
| realistic | 32768 | 3.4 ms | 0.29 ms | 12× |
| realistic | 65536 | 6.3 ms | 0.56 ms | 11× |
| uniform | 32768 | 3.0 ms | 0.32 ms | 9× |
| plateau (ties) | 32768 | 0.67 ms | 29 µs | 23× |
| plateau (ties) | 65536 | 1.4 ms | 58 µs | 25× |

At 32K context × 78 layers, the qsort path was costing ~2.8 ms × 78 ≈ **220 ms/token** spent sorting an array whose only purpose is to yield one threshold. The gap widens with context (linear vs n-log-n). DSA only activates past `index_topk`=2048, so this is invisible on short prompts but dominates long-conversation decode.

## Test — `test_dsa_select` (in `TEST_BINS`)

129 cases. Drives the **real** `partial_select_desc` (via the `include-glm.c` pattern) and compares the full kept-position set against an **independent** reference that re-implements the old algorithm (qsort + `tmp[keep-1]` threshold + position scans) on a private buffer:

- **Element-wise identical** `dst[]` (not multiset) across shapes: random, peaked, strictly-decreasing (quickselect worst case), strictly-increasing (other worst case), tie-plateau, all-equal.
- Sweeps `nk ∈ {1,2,8,64,2049,4097,8193}` × `keep ∈ {1,8,256,1024,2048}`.
- Direct partition-invariant check (`max(tail) ≤ min(top)`) — catches a subtly broken quickselect even if the threshold happened to come out right.
- Edges: `keep==nk`, `keep==1` (argmax), all-equal-scores (degenerate threshold, every kept slot is a tie → positions `0..keep-1`).

```
test_dsa_select: 129 cases run, 0 failure(s)
test_dsa_select: ok
```

## What I checked and did NOT change

Recorded in #356 so it isn't re-investigated — three other suspected attention-side hotspots, all measured to be already optimal at GLM-5.2 scale (E=256, K=8, index_topk=2048):

- **MoE top-K** (`glm.c:2913`, `O(K·E)` repeated linear-max-with-skip): the branch-predictable scan measures ~3 µs vs ~10.5 µs for a full argsort. Changing it would be a **regression**.
- **softmax 3-pass vs 2-pass**: within 1% (compiler hoists the loop-invariant reciprocal).
- **`tok_encode`**: looks `O(L²)` but is actually `O(L·nsp·s)` — the inner loop breaks on first hit.

## Verification

- [x] `make tests/test_dsa_select && ./tests/test_dsa_select` → **129 cases, 0 failures**
- [x] `make tests/bench_dsa_select && ./tests/bench_dsa_select` → reproduces the table above (6–25× across shapes)
- [x] `make glm.exe` → clean (only the two pre-existing unrelated warnings: `psapi` pragma, `g_numa_nodes`)
- [x] Full C test suite: **12/12 build+pass**. (`test_stops` fails to build on `dev` — pre-existing `mkdtemp`/#352 breakage, confirmed identical on clean `origin/dev`; unrelated to this PR.)

## Scope / base

- Branch `fix/dsa-partial-select-356` based off **`dev`** (per repo convention — `dev` is 42 commits ahead of `main`).
- Four files: `glm.c` (helper + one call-site), `Makefile` (test in `TEST_BINS` + bench on-demand rule), `test_dsa_select.c`, `bench_dsa_select.c`. No model files, no binaries, no benchmark artifacts.
- Cross-referenced on #335, #354, #153, #292.

🤖 Generated with [ZCode](https://z.ai)

---

> **Credit / co-author:** the partial-select algorithm this PR extends to the attention side is **@KingIcyCreamProjects**'s work from issue #335 (the original max-heap partial top-p select + the 9950X3D measurement). This PR is the attention-side twin of #354, which implements the same idea for sampling. Recorded in-thread per the note below.

If proper git-native `Co-Authored-By:` attribution is wanted for any of these, it needs a maintainer to file it as a tracked issue/decision on the repo — the commits here are already merged into `origin/dev`, and amending them for trailers would rewrite shared integration history (every contributor's `dev` diverges).

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -166,7 +166,7 @@ else
 PYTHON    ?= python3
 endif
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_kv_alloc$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE)
+TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_kv_alloc$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE)
 ifneq (,$(LINUX))
 TEST_BINS += tests/test_uring$(EXE)
 endif
@@ -324,6 +324,14 @@ tests/test_i4_acc512$(EXE): tests/test_i4_acc512.c
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_compat_direct$(EXE): tests/test_compat_direct.c compat.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_dsa_select$(EXE): tests/test_dsa_select.c glm.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+# bench_dsa_select is a microbenchmark (old qsort vs new quickselect partial-select, #356),
+# NOT a test gate -- intentionally absent from TEST_BINS. Build on demand: make tests/bench_dsa_select
+tests/bench_dsa_select$(EXE): tests/bench_dsa_select.c glm.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_uring$(EXE): tests/test_uring.c glm.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
--- a/c/glm.c
+++ b/c/glm.c
@@ -2304,6 +2304,43 @@ static int g_dsa_force=0; /* DSA_FORCE=1: selezione sempre attiva (test: top-min
 static int cmp_fdesc(const void *a,const void *b){
     float x=*(const float*)a, y=*(const float*)b; return x<y?1:x>y?-1:0; }
 
+/* PARTIAL SELECT (quickselect, Hoare partition, DESCending). After this call the k
+ * LARGEST elements of a[0..n) are in a[0..k) in unspecified order; the (k+1)-th and
+ * beyond are untouched-or-smaller. O(n) average, O(n^2) pathological (mitigated by
+ * median-of-three below) — and unlike a full qsort it never orders more than needed.
+ *
+ * Why this exists (#356): the DSA top-keep in attention_rows previously full-qsorted
+ * all nk context scores (O(nk log nk)) per layer per token just to read ONE value --
+ * the keep-th largest (the threshold). quickselect finds that pivot in O(nk) average,
+ * and the position-order scans that build dst[] are unchanged, so the kept set is
+ * bit-identical. Mirrors the sampling-side fix in #335 (heap partial-select there).
+ *
+ * NOT a stable partition: callers must derive the threshold and then re-scan the
+ * ORIGINAL array (the DSA code does exactly this) rather than reading a[0..k). */
+static void partial_select_desc(float *a, int n, int k){
+    if(k<=0) return;
+    if(k>=n) return;                 /* nothing to partition: all kept */
+    int lo=0, hi=n-1;
+    while(lo<hi){
+        /* median-of-three pivot to dodge the O(n^2) path on sorted/reverse input */
+        int mid=lo+((hi-lo)>>1);
+        if(a[mid]>a[lo]){ float t=a[lo]; a[lo]=a[mid]; a[mid]=t; }
+        if(a[hi]>a[lo]){  float t=a[lo]; a[lo]=a[hi];  a[hi]=t;  }
+        if(a[mid]>a[hi]){ float t=a[hi]; a[hi]=a[mid]; a[mid]=t; }
+        float piv=a[hi];
+        int i=lo, j=hi;
+        for(;;){
+            while(a[i]>piv) i++;     /* desc: large values go left */
+            while(j>lo && a[j]<piv) j--;
+            if(i>=j) break;
+            float t=a[i]; a[i]=a[j]; a[j]=t; i++; if(i>j) break; j--;
+        }
+        /* partition point: a[lo..i) are all >= piv, a[i..hi] are all <= piv */
+        if(k<=i-1) hi=i-1;          /* the k-th largest is in the left partition */
+        else       lo=i;            /* it's in the right partition */
+    }
+}
+
 /* attenzione MLA con KV-cache compressa, su token nuovi x[S,hidden], pos_base = pos del primo */
 /* kvs/pos describe a ragged decode batch: each row may belong to a different
  * sequence.  NULL keeps the original contiguous, currently-bound KV path. */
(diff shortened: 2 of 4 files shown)


woolcoxm:
@KingIcyCreamProjects — you were right on the 8192 spike, and I verified it against the code rather than just trusting the read. Two bugs compounded:

1. **One frozen input per cell** — exactly as you said. `partial_select_desc`'s median-of-three pivot is deterministic (no RNG), and the bench froze the input then ran 2000 reps on that *identical* array, so every rep hit the same pivot sequence. One lucky input → one spiked row, stable across re-runs because it's deterministic, not because it's real.
2. **`brng` was never reset between cells** — one you didn't mention. It's a static global initialized once and advanced by every prior cell's draws, so each `(shape, nk)` input depended on the whole history. Reordering `nks[]` silently shifted every later cell.

Fixed in **#378**: reseed per `(shape, nk, seed)`, report **median of 11 seeds × 2000 reps**. On my Meteor Lake box the 8192 row drops to ~10× (realistic) / ~9× (uniform); the whole table is now a clean 6–26× band with no spikes. Would value a 9950X3D re-run of that branch to confirm your 75× row lands in-band.

Your whip-up bench was honestly fine for 2 seconds of work — the only thing missing was the seed sweep, which is a non-obvious gotcha with deterministic pivots.

---

**On attribution** (your note from #354): done. Credited you by name on #354, #357, and the #335 thread — the partial-select algorithm and the original 9950X3D measurement are your work and the PRs now say so. I did **not** add a \`Co-Authored-By:\` commit trailer, and I want to be straight about why: those commits are already merged into \`origin/dev\`, and amending them for the trailer would rewrite shared integration history — every contributor's \`dev\` would diverge, and that creates a real mess in the repo. The proper git-native route needs a maintainer to make a call on it (file it as a tracked decision). In-thread credit was the honest alternative; if you want the trailer too, say so and I'll raise it with the maintainer rather than force-push quietly.

**On the coverage nit** (V=151936 in \`test_topp\`): I owe you an honest retraction here. I said it was a reference-precision artifact and spent a while rewriting the reference to match production — that was circular, and you'd have been right to call it out. The real situation: at float-accumulation boundaries (uniform@0.5, V=151936) the heap gives keep=75969 where exact arithmetic says 75968 — which is actually sitting in *your* #354 bench table. That's not a regression (any float impl does this), but it means the bit-equivalence claim can't be extended to full V without either a looser tolerance or an honest caveat. I reverted my test changes; the gated test stays at V≤1519. The invariant checks (id-indexed \`g_pbuf\`, tail zeroed, head sums to 1) *would* extend cleanly to full V if you want that — those are real and cheap.

**Serve-decode tok/s** is still the open item — happy to run it, or if you've got the 9950X3D warm, that box is the one that matters.

PR #361, opened 2026-07-17
win32: retry transient ReadFile errors in compat_pread (#307)

woolcoxm:
## Summary

On Windows, `compat_pread` mapped **every** non-EOF / non-`EBADF` `ReadFile` failure to `EIO` and returned `-1` on the **first** attempt. Under the engine's default Windows load — `DIRECT=1`, `PIPE=1`, `PIPE_WORKERS=8`, `PILOT_REAL=1` (8 pipe workers + pilot + main, all issuing `compat_pread` on the same shard handles) plus a synchronous `ReadFile` per `WILLNEED` hint — transient errors are normal and recoverable:

- `ERROR_LOCK_VIOLATION` (33) — momentary handle conflict
- `ERROR_SHARING_VIOLATION` (32) — momentary sharing conflict
- `ERROR_NOT_READY` (21) — device momentarily not ready (AV scanning the VHDX, virtual-disk detach/timeout)

A single flaky `ReadFile` killed the load:

```
[engine terminated: exit code 1]
pread qs: Input/output error (off 3728808, 0/8192 bytes)
```

…mid-generation, after the engine had run fine while the working set was warm. This is the new symptom reported on #307 after the stale-`.coli_kv` fix landed.

## Root cause — what it is *not*

Two investigations ruled out the obvious suspects:

- **Not the OMP fix (`d5327e2`).** That `KMP_BLOCKTIME` code is behind `#ifdef __linux__` (`glm.c:6009`) and never runs on Windows; the `coli` launcher's `env_for` (`c/coli:176-180`) does not even set `KMP_BLOCKTIME`. Idle-spin tuning has no path to a `ReadFile` returning failure.
- **Not a slab/ESlot concurrency race.** The three slot pools (`pin[]`, `ecache[]`, `ws[]`) are disjoint. The pilot writes only `ecache[L > g_cur_moe_layer]` while the main thread reads `ecache[current]`, gated by `g_pilot_inflight[layer]==0`. `expert_load_impl` sets `s->eid` last. No half-loaded slab can be matmul'd, and a slab-realloc race would corrupt weights (wrong tokens), not produce `EIO`-exit.

The defect is in the I/O wrapper's error classification — not tensor geometry or alignment.

## The fix

`compat_pread` now retries the three genuinely transient codes up to **3×** with **1/2/4 ms** backoff before giving up to `EIO`. `ERROR_HANDLE_EOF`, `EBADF`-class, and real errors remain immediate. A dying drive still fails (after the retries); a genuine transient survives.

`err` is gated on an explicit success flag (`ok`), so a stale error from an earlier retried attempt can never be misreported (e.g. transient-fail-then-successful-EOF reads as EOF, not `EIO`).

## What this does *not* change

- **Linux / macOS:** untouched. `compat_pread` is Windows-only (`#ifdef _WIN32`).
- **Throughput:** unchanged on a healthy system — retries only fire on actual transient errors.
- **Honest failure:** a persistent fault still surfaces as `EIO` after the retries are exhausted.

## Verification

Compiles clean against real MinGW-w64 `windows.h` with `-Wall -Wextra` (zero warnings in the changed code). Logic traced through all cases: success-first, transient-then-success, transient-exhausted, hard-error, EOF (success with `rd=0`), EOF (`ERROR_HANDLE_EOF`), partial read.

## Open question for @galmok

This fix makes the contention **survivable**. Whether it is **sufficient** for your case depends on whether your failures are transient-recoverable (this should let you run to completion) vs. genuine storage faults (the retries delay but won't prevent the eventual `EIO`). To tell them apart, please also grab:

- **Windows Event Viewer → Windows Logs → System** during a failing run — look for `disk`, `Ntfs`, `diskpart` / `storahci` errors at the crash timestamp.
- **`chkdsk`** on the model volume.
- Which physical drive + filesystem the model lives on (local NVMe? a VHDX? an external/USB drive?).

Refs #307.

### DIFF
--- a/c/compat.h
+++ b/c/compat.h
@@ -142,7 +142,34 @@ static inline int compat_fadvise(int fd, off_t off, off_t len, int advice){
 /* --- pread -> ReadFile + OVERLAPPED su raw OS handle ---
  * Thread-safe (no shared seek position). Gestisce offset >4 GB e chunking
  * per letture >2 GB (anche se i tensori individuali sono nell'ordine dei
- * MB-centinaia di MB, il wrapper e' robusto per ogni taglia). */
+ * MB-centinaia di MB, il wrapper e' robusto per ogni taglia).
+ *
+ * RETRY ON TRANSIENT ERRORS (#307): gli shard handle NON sono aperti con
+ * FILE_FLAG_OVERLAPPED (st.h:85 li apre buffered), quindi il kernel serializza
+ * le ReadFile concorrenti allo stesso handle. Sotto il carico del motore su
+ * Windows (default: DIRECT=1, PIPE=1, PIPE_WORKERS=8, PILOT_REAL=1 -> 8 worker
+ * + pilot + main, tutti compat_pread sugli stessi shard handle; piu' un'altra
+ * ReadFile sincrona per ogni hint WILLNEED in compat_fadvise) errori transitori
+ * come ERROR_LOCK_VIOLATION / ERROR_SHARING_VIOLATION / ERROR_NOT_READY sono
+ * normali e recuperabili: conflitto momentaneo sul handle, AV che scansiona il
+ * VHDX, un detach/timeout del disco virtuale. Mapparli a EIO e ritornare -1 al
+ * primo tentativo (il comportamento precedente) uccideva il load con un singolo
+ * ReadFile flaky -> "[engine terminated: exit code 1]" + "pread qs: Input/output
+ * error (off X, 0/N bytes)" nel bel mezzo di una generazione, dopo che il motore
+ * aveva girato bene finche' il working set era caldo. Ora questi tre errori si
+ * ritentano fino a 3 volte con backoff 1/2/4 ms, poi si arrendono a EIO: un
+ * disco morente fallira' comunque (dopo i tentativi), ma un transient vero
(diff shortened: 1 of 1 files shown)


woolcoxm:
Closing in favour of **#364**, which stacks this exact fix ( transient-retry) on top of #298 (the g64-CUDA prerequisite) alongside the  auto-enable fix, and verifies the two together end-to-end on Windows.

Why close this one:
- The identical change is commit `921644a` in #364 (verified: `git diff dd9d8d8 921644a -- c/compat.h` is empty) — no code is lost.
- This branch is now **CONFLICTING / DIRTY** against `dev` (dev advanced since it was opened); the content still merges cleanly but the PR UI shows a stale conflict, which is noise for reviewers.
- #364 is the verified integration path and absorbs this fix once #298 lands.

The fix itself is preserved and tracked. Reopen if you'd prefer to keep a standalone dev-based path.

woolcoxm:
Correction to the above (the shell ate some backticked words): the identical change is the `compat_pread` transient-retry commit `921644a` in #364, alongside the `coli chat` auto-enable fix. No code is lost — this PR is superseded by #364, which is the verified Windows + g64-CUDA integration path stacked on #298.

PR #365, opened 2026-07-17
cuda: batch ragged attention across independent streams

ZacharyZcR:
## Summary

- add a CUDA MLA attention path for batches whose rows belong to independent KV sequences
- raise the multiplex decode batch ceiling from 64 to 512 rows
- keep DSA and unsupported cases on the existing fallback path
- add a CUDA correctness test against per-sequence batch attention

## Validation

- `make -C c glm CUDA=0`
- CUDA build on RTX 5090 (sm_120)
- `tests/test_ragged_attention`: `ragged_relative_rms=0`

## Measured result

On the 6x RTX 5090 full-residency GLM-5.2 setup at B=256, the decode forward changed from 26.131 s to 14.344 s (1.82x), while the attention phase changed from 15.143 s to 3.303 s (4.58x). Disk reads remained zero.

### DIFF
--- a/c/backend_cuda.cu
+++ b/c/backend_cuda.cu
@@ -338,6 +338,40 @@ __global__ static void attention_absorb_batch_kernel(float *ctx,const float *q,
         ctx[((size_t)s*H+h)*V+v]=a*(fmt?wscale[row]:1.f);}
 }
 
+/* Independent KV sequence per row. latent/rope are packed as [S,T,*], while
+ * lengths selects the valid prefix for each row. */
+__global__ static void attention_absorb_ragged_kernel(float *ctx,const float *q,
+        const float *latent,const float *rope,const int *lengths,
+        const void *weights,const float *wscale,int fmt,int S,int H,int Q,int R,
+        int V,int K,int T,float scale){
+    int s=blockIdx.y,h=blockIdx.x,tid=threadIdx.x,nt=lengths[s],rbase=h*(Q+V);
+    if(s>=S||nt<1||nt>T)return;
+    extern __shared__ float sm[];float *qa=sm,*cl=qa+K,*scores=cl+K,*red=scores+T;
+    const float *qs=q+((size_t)s*H+h)*(Q+R);
+    const float *ls=latent+(size_t)s*T*K,*rs=rope+(size_t)s*T*R;
+    for(int k=tid;k<K;k+=blockDim.x){float a=0;for(int d=0;d<Q;d++)
+        a+=qs[d]*weight_at(weights,fmt,(size_t)(rbase+d)*row_bytes(fmt,K),k)*
+          (fmt?wscale[rbase+d]:1.f);qa[k]=a;}
+    __syncthreads();
+    for(int t=tid;t<nt;t+=blockDim.x){float a=0;const float *lt=ls+(size_t)t*K;
+        const float *rt=rs+(size_t)t*R;for(int k=0;k<K;k++)a+=qa[k]*lt[k];
+        for(int d=0;d<R;d++)a+=qs[Q+d]*rt[d];scores[t]=a*scale;}
+    __syncthreads();
+    float local=-3.402823466e+38F;for(int t=tid;t<nt;t+=blockDim.x)local=fmaxf(local,scores[t]);
+    red[tid]=local;__syncthreads();
+    for(int n=blockDim.x>>1;n;n>>=1){if(tid<n)red[tid]=fmaxf(red[tid],red[tid+n]);__syncthreads();}
+    float mx=red[0];local=0;for(int t=tid;t<nt;t+=blockDim.x){float e=expf(scores[t]-mx);scores[t]=e;local+=e;}
+    red[tid]=local;__syncthreads();
+    for(int n=blockDim.x>>1;n;n>>=1){if(tid<n)red[tid]+=red[tid+n];__syncthreads();}
+    float inv=1.f/red[0];for(int t=tid;t<nt;t+=blockDim.x)scores[t]*=inv;
+    __syncthreads();
+    for(int k=tid;k<K;k+=blockDim.x){float a=0;for(int t=0;t<nt;t++)a+=scores[t]*ls[(size_t)t*K+k];cl[k]=a;}
+    __syncthreads();
+    for(int v=tid;v<V;v+=blockDim.x){int row=rbase+Q+v;float a=0;size_t rb=row_bytes(fmt,K);
+        for(int k=0;k<K;k++)a+=cl[k]*weight_at(weights,fmt,(size_t)row*rb,k);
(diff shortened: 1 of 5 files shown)


JustVugg:
Reviewed the CPU-safety and it's clean — every glm.c change is behind `COLI_CUDA`, it merges cleanly on current `dev`, and the CPU build is warning-free. Track record noted (#132/#232/#255 all landed solid), so I'm inclined toward this.

The one thing holding the merge: it's an **attention batching kernel** — correctness-sensitive — and I have no GPU here, so I can't run `test_ragged_attention.cu`. That test is exactly the right artifact; I just can't execute it. Could someone on GPU hardware (@woolcoxm / @mohamedmastouri2000-boop are both active on CUDA right now) post the `test_ragged_attention` result on real silicon, ideally with a token-identity check vs the non-batched path on a short prompt? Ragged batching across streams is the kind of thing where a boundary/masking mistake produces subtly-wrong-but-not-crashing output, so the token-exact A/B is the gate.

With a green test from hardware, this merges. It stacks naturally with #298's CUDA work, so coordinating the two is worth it.

mohamedmastouri2000-boop:
Ran it on real silicon — **RTX 5080 / sm_120** (the #306 box), CUDA 12.8, nvcc built against the PR head (786f98d):

```
$ nvcc -O3 -std=c++17 -arch=sm_120 backend_cuda.cu tests/test_ragged_attention.cu -o tests/test_ragged_attention
$ ./tests/test_ragged_attention
[CUDA] device 0: NVIDIA GeForce RTX 5080, 17.1 GB VRAM, sm_120
ragged_relative_rms=0
$ echo $?
0
```

**`ragged_relative_rms=0`, exit 0** — the batched ragged path is bit-identical to the per-sequence `coli_cuda_attention_project_batch` reference across all three ragged sequences (n={1,2,3}), so this is the token-exact A/B you asked for, not just a low-rms pass. No boundary/masking drift on Blackwell.

That's a second sm_120 confirmation alongside the author's 5090. Stacks cleanly with #298's CUDA work (I've been running the g64 container on this same box). Happy to run the full `test-c` suite or a longer multiplex decode on real weights if that adds anything before merge.

woolcoxm:
running the tests now.

JustVugg:
Merged. @mohamedmastouri2000-boop's run is what unblocked it: `ragged_relative_rms=0`, exit 0 on real RTX 5080 / sm_120 — **bit-identical** to the per-sequence reference across all three ragged lengths, not just a low-rms pass. That's the token-exact gate I asked for, on the exact class of kernel (attention, ragged batching) where a boundary/masking mistake produces subtly-wrong-not-crashing output. Second sm_120 confirmation alongside your 5090, CPU build clean and CPU suite green here, all glm.c changes behind `COLI_CUDA`. Thanks to both of you — this is how a CUDA kernel should land: verified on silicon by someone other than the author.

woolcoxm:
Ran the gate you asked for on real silicon — **RTX 5070 Ti (sm_120), CUDA 12.8, Windows, `coli_cuda.dll` build** — and I think this should block rather than merge. The toy test passes but the kernel has an intermittent correctness bug that only shows up at production scale.

## 1. The shipped test passes (false green)

```
$ test_ragged_attention
ragged_relative_rms=0      (exit 0)
```

But it runs at toy dimensions — K=3, T=3, H=2, S=3. At that geometry the bug below doesn't trigger, so `ragged_relative_rms=0` here doesn't generalize.

## 2. Token-identity A/B on a short prompt fails

Drove `glm` in mux mode (`SERVE_BATCH=1`, 2 slots → genuinely ragged `step_decode_batch`, `COLI_CUDA_ATTN=1` vs unset), greedy (temp=0), GLM-5.2 nothinking template, real `glm52_i4_g64`:

- **Control** (CPU attention vs CPU attention, cold start): byte-identical → run-to-run is deterministic, so any A/B divergence is the attention path, not expert-routing noise.
- **Ragged CUDA vs CPU attention**: divergent — ragged emits garbage, CPU emits coherent tokens.

## 3. Root cause: non-deterministic corruption at production scale

Wrote a kernel-level A/B (ragged-CUDA vs batch-CUDA, **both on GPU**, same tensors — no model/routing/quant confound) at real MLA dims — K=512, Q=192, R=64, V=256, H=64:

A cold ragged S=4 call vs the per-sequence batch reference, **5 consecutive identical runs** (same inputs, same tensors, fresh process each time):

```
run 1: rms=0
run 2: rms=0
run 3: rms=0
run 4: rms=0
run 5: rms=0.0114        <-- corrupted
```

It's intermittent: ~1 in 5 identical runs diverges, by ~0.01–0.025 relative rms. That signature — non-deterministic, not reproducible every run, no crash — points to a **race / uninitialized-device-memory read** in the ragged path's use of the shared `DeviceContext` scratch (`dc->ac` / `dc->al` / `dc->y`), not the kernel math itself (the kernel is textually identical to the working batch kernel, and matches batch bit-for-bit when it isn't corrupted). `cudaMalloc` scratch is uninitialized and `reserve()` is grow-only, so a stale/larger buffer from a prior geometry can be read before it's overwritten.

This is exactly the "subtly-wrong-but-not-crashing" case you flagged — except worse, because it's not even reliably reproducible, so it'll flake in CI and silently degrade output in production.

## Suggestion

- Add a production-dimension + repeated-call case to `test_ragged_attention.cu` (K=512/H=64/T≈40, run the ragged path N≥20 times vs batch, gate on max rms across runs — a single run at K=3 will never catch this).
- The fix is in the host glue in `backend_cuda.cu` (`coli_cuda_attention_project_ragged`): make sure every scratch buffer the kernel reads is fully written this call — either zero the tail of the `[S,T,K]` latent/rope uploads (the `calloc` covers the host side, but verify the device copy covers the full `S*T*K`), or `cudaMemsetAsync` the device scratch to zero before the kernel when a buffer is reused and larger than the current write. The shared-memory reduction already bounds to `lengths[s]`, so the leak is most likely the device scratch, not the kernel.

Happy to post the repro test (`test_ragged_repro.cu`) if useful. Gate fails as-is from me — request changes.

woolcoxm:
Follow-up with the repro tests pushed to a branch on my fork, so anyone with GPU hardware can build/run them directly:

- **Branch:** [`pr365-repro`](https://github.com/woolcoxm/colibri/tree/pr365-repro) (on top of this PR's head, `786f98d`)
- **Commit:** [`b528c2f`](https://github.com/woolcoxm/colibri/commit/b528c2f) — adds `tests/test_ragged_repro.cu` (the intermittent-corruption repro), `tests/test_batch_stability.cu` (the control), `tests/test_ragged_attention_real.cu` (production-dim A/B), and `tests/RAGGED_REPRO.md` with build/run instructions + analysis.

The repro is intermittent (~1 in 5 runs), so run `test_ragged_repro` a handful of times — a single `rms=0` run is expected by chance. The batch control (`test_batch_stability`) should be 0/8 every time.

PR #367, opened 2026-07-17
telemetry: split expert hit counter into pin vs ecache (LRU) (#336)

### LINKED ISSUE #336: telemetry: hits conflates pin-tier and ecache hits — split them so eviction-policy changes are measurable
The expert lookup counts a hit identically whether it was served by the pinned tier or by the LRU ecache: both Phase C branches bump the same `m->hits` (`c/glm.c:3057` pin, `:3059` ecache, on dev), the struct has only the one lumped field (`:195`), and the PROF harness reports the lumped value only.

That makes every cache-policy question unanswerable from logs. The pin tier is checked first and — with a warm usage profile — absorbs most hot experts, so the ecache may be serving anywhere from ~0% to a large share of hits, and nobody can tell which. Concretely blocked by this:

- **#223's cap-threshold question** ("at what cap does cloxcache start to win?") — a flat A/B can mean "the pin tier absorbed everything" or "genuine floor effect", and the counter can't distinguish them.
- **Prefill-burst pollution**: prefill and decode share one `moe()` path and one ecache; between phases the counters reset (`:4633`) but the cache contents don't — whether decode actually pays for prefill pollution is currently unmeasurable.
- **PILOT_REAL eviction pressure** (the reason `PILOT_K` defaults 8→6 under real loads) — same problem.

Proposal (implemented on a branch): two counters `hit_pin` / `hit_ecache` bumped at the two lookup branches, surfaced in the periodic serve stats line (`hit N% (lru M%)` — pin share = N−M), the REPLAY summary (`expert hit N% (pin X% + lru Y%)`), and the `[PROF]` report (`hit N% (X pin + Y lru / Z load)`). The serve-mux `PROF` protocol line is untouched (openai_server.py parses it positionally). Zero cost — two increments on paths that already increment `m->hits`.

Branch: https://github.com/KingIcyCreamProjects/colibri/tree/telemetry/pin-vs-ecache-hits — happy to open the PR.

KingIcyCreamProjects:
## What

Splits the single `m->hits` expert-cache counter into **`hit_pin`** (HOT-STORE / pinned experts, never evicted) and **`hit_ecache`** (per-layer LRU). By construction `hit_pin + hit_ecache == hits` at every increment — the two `m->hits++` sites in `moe()` each gain a 1:1 tier-specific bump, so the total is preserved with no double-count. Resolves #336.

## Why

Today `hits` lumps together two very different events: a **pin-tier** hit (free, the expert is wired in RAM) and an **ecache/LRU** hit (subject to the eviction policy). That makes eviction-policy changes unmeasurable — you can't tell whether a policy "won" because it improved LRU reuse or because the pin tier simply absorbed everything.

This directly unblocks the evaluation of **#223** (cloxcache/SLRU): the maintainer's gating question there — *"the lowest cap at which cloxcache starts to win"* — is unanswerable while `hits` conflates the tiers. With the split, a cap sweep shows whether a flat A/B is "pin tier absorbed all the traffic" vs. a genuine LRU-policy floor.

## Surface

- `glm.c` only, +19/−14. New fields on `Model`; snapshot/reset updated in lockstep at both counter-reset sites; windowed deltas added.
- Split is surfaced in the three interactive/offline report paths: `emit_stream` (lru%), `[PROF]` (`X pin + Y lru / Z load`), and `REPLAY` (`pin X% + lru Y%`).
- **Zero cost** — two increments on paths that already increment `hits`. **Cannot change model output** (telemetry only).

## Notes for the reviewer

- The serve-mux `[PROF]` protocol line is **intentionally left untouched** — `openai_server.py` parses it positionally, so the pin/ecache breakdown is surfaced only in the human-facing report paths above.
- Two-line textual overlap with **#223** on the `moe()` lookup branches (both add to the pin/ecache lookup lines). No semantic conflict — this only reads/writes new counters, #223 only changes victim selection — so whichever lands second is a trivial keep-both merge.

### DIFF
--- a/c/glm.c
+++ b/c/glm.c
@@ -193,6 +193,7 @@ typedef struct {
     uint64_t mtp_prop, mtp_acc;                  /* statistica acceptance */
     int **eroute; int *enr;                      /* metodo C: routing dell'ULTIMO token per layer */
     uint64_t eclock, hits, miss, ereq;
+    uint64_t hit_pin, hit_ecache;                /* split di hits: tier pin vs LRU ecache */
     uint64_t gpu_expert_calls; int gpu_expert_count; int64_t gpu_expert_bytes;
     uint64_t n_fw, n_emit;                       /* metodo E: forward di decode / token emessi */
     uint64_t route_slots, route_swaps;            /* CACHE_ROUTE: slots chosen / substituted vs true top-K */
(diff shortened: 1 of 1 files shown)


KingIcyCreamProjects:
Superseded by @JustVugg's direct implementation on dev (679c074) — same design (hit_pin/hit_ecache at both lookup branches, all three summary surfaces). Verified working on the 9950X3D in #336. Closing.

PR #370, opened 2026-07-17
sampling: guard against non-finite logits — was silent token-0 spew (#369)

### LINKED ISSUE #369: sampling: a single NaN/+Inf logit makes dist_build emit token 0 forever, silently (default path)
## Summary

On the **default** serve/chat sampling path (`TEMP>0`, `0<NUCLEUS<1`), a single **NaN or +Inf** logit makes `dist_build()` produce an all-NaN distribution, after which the sampler silently emits **token 0 forever** — an unbroken run of one token, with **no error printed**. A diagnosable upstream numerical fault becomes silent corruption.

## Mechanism (traced against `glm.c`)

`dist_build()` softmax:
```c
float mx=lo[0]; for(int i=1;i<V;i++) if(lo[i]>mx) mx=lo[i];
double s=0; ...
for(int i=0;i<V;i++){ g_pbuf[i]=expf((lo[i]-mx)*invt); s+=g_pbuf[i]; }
for(int i=0;i<V;i++) g_pbuf[i]/=(float)s;
```
- A `NaN` logit → `g_pbuf[that]=expf(NaN)=NaN` → `s=NaN` → every `g_pbuf[i]/=(float)s` is `NaN`.
- (`+Inf` → `mx=+Inf` → `expf(Inf-Inf)=NaN` → same collapse.)
- Nucleus keep-loop `cum+=g_pbuf[...]; if(cum>=g_nuc)` never triggers (`NaN>=x` is false) → `keep=V`, `s2=NaN`, still all-NaN.
- `dist_sample`: `cum>=u` never true → fallback `for(i=V-1..0) if(g_pbuf[i]>0) return i` is false for all (`NaN>0` false) → **returns 0**.
- Greedy path is equally blind: `argmax_v` seeds `bv=lo[0]` and `lo[i]>NaN` is always false, so a `NaN` at index 0 pins the argmax to **0**.

There is no `isfinite`/`isnan` guard anywhere between the logits and the sampler.

## Reachability

The precondition is the *default*, not an edge case (`TEMP` auto≈0.7, `NUCLEUS` default 0.90). A non-finite logit is plausible in this NVMe-streaming engine: a garbage/short-loaded expert tile or an fp overflow in `matmul_qt` at a low-RAM eviction boundary. Both `dist_build` call sites are affected (the sampler and the speculative-verify path).

## Impact

Not a fault *originator* — it needs an upstream NaN/+Inf. It's a fault **amplifier**: it turns a catchable numeric fault into silent, unbroken token-0 output. Medium severity: it defeats loud failure on the default path.

## Fix (ready)

Minimal, O(1)/free on the happy path, on `fix/logit-nan-guard` (off `dev`) on my fork:
- `argmax_v`: skip NaN (`x==x`) and seed from `-inf` → returns the max finite/+Inf entry, not a NaN-pinned 0.
- `dist_build`: after the softmax sum, if `s` is non-finite or `<=0`, collapse `g_pbuf` to a one-hot over the finite argmax and warn once — **degrade + diagnose, never silently corrupt**.
- `test_logit_nan` (wired into `TEST_BINS`): fails on stock `dev`, passes with the guard.

Heads-up: since **#354** rewrites `dist_build`'s internals (heap partial-select), the `dist_build` half of this guard should be re-placed on top of that once it lands — happy to open the PR against whichever base you prefer, or fold it into the #354 work. @woolcoxm

KingIcyCreamProjects:
## What

Fixes #369: on the default sampling path (`TEMP>0`, `0<NUCLEUS<1`), a single **NaN/+Inf logit** poisoned softmax into an all-NaN `g_pbuf`, after which `dist_sample` silently emitted **token 0 forever** — no error, just an unbroken run of one token. The greedy path was equally blind (`argmax_v` NaN-pinned to index 0).

Three small changes, all O(1)/free on the happy path:

- **`argmax_v`**: skip NaN (`x==x`) and seed from `-INFINITY` → returns the max finite/+Inf entry instead of a NaN-pinned 0. Covers greedy decode and the speculative-verify argmax.
- **`dist_build` max scan** (per @woolcoxm's review on #369): seed `mx` from `-INFINITY` and NaN-skip the scan — the collapse actually started at the max-finding (`mx=lo[0]` with a NaN at 0 dooms every `lo[i]-mx` before the sum), not at the normalize. If nothing finite survives, fall back to `mx=0` and let the post-sum guard decide.
- **`dist_build` post-sum guard**: if `s` is non-finite or ≤0, collapse `g_pbuf` to a one-hot over the finite argmax and warn once — degrade + diagnose, never silently corrupt. Now the second line of defense, not the only one.

Clean logits take a byte-identical path.

## Placement vs #354

Rebased onto current `dev` (post-#354/#357). As predicted in review, the #354 heap rewrite only touched the nucleus tail — the guard sits right after the normalize, before the `if(g_nuc>0 && g_nuc<1.f)` block, so the heap never sees an all-NaN `g_pbuf`.

## Tests

`tests/test_logit_nan.c` (wired into `TEST_BINS`), 6 assertion blocks: argmax_v NaN-skip / +Inf-pick / all-NaN; dist_build with a NaN mid-vocab → finite one-hot on the max finite logit and `dist_sample` emits it (not 0); **NaN at index 0** (the mx-seed failure mode from review); all-NaN stays finite; and a clean-logits regression case. **Fails on stock `dev`** (first assert: argmax NaN-pinned to 0), passes with this change.

Also re-ran `test_topp` (123 cases) on this tree: green — no interaction with the #354 partial select.

@woolcoxm — rebased, mx-scan point folded in, ready for your review.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -166,7 +166,7 @@ else
 PYTHON    ?= python3
 endif
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_kv_alloc$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE)
+TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_kv_alloc$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE)
 ifneq (,$(LINUX))
 TEST_BINS += tests/test_uring$(EXE)
 endif
@@ -329,6 +329,9 @@ tests/bench_topp$(EXE): tests/bench_topp.c glm.c st.h uring.h json.h tok.h tok_u
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_kv_alloc$(EXE): tests/test_kv_alloc.c glm.c st.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_logit_nan$(EXE): tests/test_logit_nan.c glm.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_i4_acc512$(EXE): tests/test_i4_acc512.c
--- a/c/glm.c
+++ b/c/glm.c
@@ -4259,7 +4259,13 @@ static void topp_siftdown(int *h, int n, int i){
  * la coda troncata va AZZERATA in g_pbuf (dist_sample la legge direttamente per id). */
 static void dist_build(const float *lo, int V){
     if(!g_pbuf){ g_pbuf=falloc(V); g_pidx=malloc(V*sizeof(int)); }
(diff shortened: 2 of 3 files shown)


JustVugg:
Merged, reconciled with the direct #369 fix that landed on `dev` in the meantime — and your PR is what closed the real gap: my fix guarded `dist_build` (the sampling path) but left `argmax_v` NaN-pinned to index 0, which is exactly the greedy path (`TEMP=0`). Your `x==x` NaN-skip in `argmax_v` covers it, and your `test_logit_nan` proves it. Resolution kept the already-shipped `dist_build` guard and took your `argmax_v` + test on top; both NaN tests run green in the suite (71 tests). Sorry for the duplicate work — the maintainer-fixes-it-directly race is on us (same thing happened to your #367, which you closed graciously). Between the two PRs and the two direct fixes, non-finite logits are now handled on every path: sampling, greedy, and the all-NaN collapse, each with a named stderr warning instead of silence. Thanks for writing it anyway — it caught what the direct fix missed.

Issue #382, opened 2026-07-18
[Bug]: `coli serve` stops after 256 tokens


LordMZTE:
### Commit or version

8a1dccae3b1a4c82ea9173e3d1ea8c002f14e785 (current dev)

### Environment

NixOS, 2x EPYC 7252, 256GB RAM, No GPU, GCC, NVMe

### Reproduction steps

Run `coli serve --ctx 32768 --ngen 32768`, use a client of your choice to give the LLM a prompt that will motivate it to produce over 256 tokens.

### Expected behavior

We generate up to the given limit of tokens.

### Actual behavior and logs

Observe that the output stops abruptly after 256 tokens were generated; less than the given maximum of 32768.

This happens because, for some reason, the server limits the number of output tokens to 256 here:
https://github.com/JustVugg/colibri/blob/8a1dccae3b1a4c82ea9173e3d1ea8c002f14e785/c/openai_server.py#L409

JustVugg:
Fixed on `dev` (f8e4dc9) — please pull and retest. Your diagnosis was exact: a request without `max_tokens` fell into `maximum = min(256, limit)`, so the server's own `--ngen 32768` was ignored for any client that omits the field.

The default is now the operator's configured budget itself (`maximum = limit`): `coli serve --ctx 32768 --ngen 32768` means what it says for every client. Generation still ends at EOS, so it's a cap, not a target — short answers stay short. Clients that DO send `max_tokens` are unchanged (still clamped to the server cap per #260 instead of erroring).

```sh
git pull   # dev
./coli serve --ctx 32768 --ngen 32768
# same client, same prompt → should now run past 256 to EOS or your cap
```
If anything still cuts at a suspicious round number, paste the response's `usage` block and I'll dig. Thanks for the pinpoint report — line number and all.

LordMZTE:
I can confirm that this is fixed on `dev`.

PR #384, opened 2026-07-18
plan NUMA interleave on multi-socket Linux

ZacharyZcR:
## Summary

- detect physical CPU socket count in `resource_plan.py`
- enable selective `COLI_NUMA=1` in generated plans on multi-socket Linux
- preserve an explicit `COLI_NUMA=0` override
- document the measured 2-socket CUDA-host result

This uses colibri's existing per-region `mbind` path. It does **not** use blanket `numactl --interleave=all`, so CUDA DMA staging buffers retain local placement.

## On-silicon A/B

Same GLM-5.2 int4 model, 2x Xeon Silver 4510, 6x RTX 5090, fixed usage snapshot/prompt/seed, `TEMP=0 DRAFT=0`, 64 decode tokens:

| policy | CPU expert bandwidth | decode |
|---|---:|---:|
| default first-touch | 42.42 GB/s | 7.66 tok/s |
| blanket `numactl --interleave=all` | 54.56 GB/s | 8.66 tok/s |
| selective `COLI_NUMA=1` run 1 | 58.26 GB/s | 9.02 tok/s |
| selective `COLI_NUMA=1` run 2 | 65.89 GB/s | 9.17 tok/s |

All four generated outputs had the same SHA-256 (`abe3e910...aaa1307`). Selective interleave improved median decode by about 18.8% without changing output. `strace` also confirmed the existing `mbind(MPOL_INTERLEAVE, MPOL_MF_MOVE)` calls return success.

The host has only 2 populated DDR5-4400 channels per socket, which makes NUMA placement unusually visible; the planner should use all installed memory controllers automatically on this class of machine.

## Tests

- `python3 -m unittest tests.test_resource_plan` (12 passed)
- `python3 -m unittest discover -s tests -p "test_*.py"` (73 passed)
- `git diff --check`

### DIFF
--- a/c/resource_plan.py
+++ b/c/resource_plan.py
@@ -203,6 +203,22 @@ def physical_cpu_count():
     return os.cpu_count() or 1
 
 
+def cpu_socket_count():
+    """Return the number of physical CPU sockets visible to this process."""
+    if not sys.platform.startswith("linux"):
+        return 1
+    try:
+        result = subprocess.run(["lscpu", "-p=socket"], text=True,
+                                capture_output=True, check=True, timeout=5)
+        sockets = {int(line) for line in result.stdout.splitlines()
+                   if line and not line.startswith("#")}
+        if sockets:
+            return len(sockets)
+    except (OSError, ValueError, subprocess.SubprocessError):
+        pass
+    return 1
+
+
 POLICIES = {
     "quality": {"preserve_quantization": True, "preserve_router": True},
     "balanced": {"preserve_quantization": True, "preserve_router": True},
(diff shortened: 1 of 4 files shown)


JustVugg:
Merged. The evidence bar this repo asks for was fully met: on-silicon A/B with SHA-256-identical outputs across all four policies (the correctness gate), +18.8% median decode from selective interleave, the blanket-vs-selective distinction measured rather than assumed (blanket `--interleave=all` losing to selective `COLI_NUMA=1` is exactly the CUDA-DMA-staging effect you called out), and `strace`-confirmed `mbind` success. Verified here before merge: your 12 unit tests pass, the full suite is green (73 tests), and on a single-socket box the planner correctly emits nothing — the feature is invisible where it can't help. The explicit `COLI_NUMA=0` override is the right escape hatch. @LordMZTE runs 2× EPYC 7252 (#382) — a second multi-socket datapoint would be a nice confirmation if he's curious. Thanks — planner changes with token-identity proofs and a no-op path for everyone else are exactly how this should be done.

PR #409, opened 2026-07-19
nix: add darwin support, fix tests, minor changes

attilaolah:
## Summary

- ~Fix Nix build, since the tests now require Python~ already merged
- ~Set `COLI_ENGINE` so the built `coli` wrapper can find the `glm` engine~ already merged
- Add missing `version.py` to the environment
- Allow building the Nix flake on macOS (including nix-darwin)
- Set the `coli` wrapper as the main program entry point

With `coli` being the main entry point, it is possible to run e.g. directly a chat using Nix, without even checking out the source:

```sh
export COLI_MODEL=/path/to/glm52_i4
nix run github:attilaolah/colibri -- chat
```

(This is using my fork which has these changes on the main branch.)

## Validation

- [x] `make -C c check`

No code changes, but the flake was checked on both NixOS and macOS:

- [x] `nix build`
- [x] `nix run`
- [x] `nix run .#colibri`

## Compatibility

Note that the flake lockfile is now included, which probably should have been included in the original PR. This is what makes the flake reproducible, guaranteeing that it will build at the current commit even when the dependencies, compilers, etc. in `nixpkgs` are updated in the future.


attilaolah:
#416 adds CUDA support

flake.nix
```diff
@@ -100,23 +112,25 @@
...
           };
-          coli = {
+          glm = {
             type = "app";
-            program = "${colibri}/bin/coli";
+            program = "${colibri}/share/colibri/glm";
           };
         };
 
+        formatter = (import nixpkgs {inherit system;}).alejandra;
```
attilaolah, line 123: I just noticed I should have used `pkgs.alejandra` here, it is already bound.

PR #425, opened 2026-07-19
fix: gracefully handle mbind EPERM and remove unnecessary MPOL_MF_MOVE

snailwei:
- Remove MPOL_MF_MOVE flag from mbind (fresh mmap has no pages to migrate)
- Probe mbind capability at NUMA init before any allocation
- On EPERM (missing CAP_SYS_NICE), disable NUMA and warn instead of failing later
- Also handle late EPERM in numa_slab_bind by disabling NUMA on the spot

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -470,8 +470,11 @@ static void numa_slab_bind(void *p, size_t n){
     unsigned long mask=(1UL<<g_numa_nodes)-1;
     uintptr_t a=(uintptr_t)p & ~(uintptr_t)4095;
     size_t len=(((uintptr_t)p+n+4095) & ~(uintptr_t)4095) - a;
-    syscall(SYS_mbind,a,len,3/*MPOL_INTERLEAVE*/,&mask,
-            (unsigned long)(g_numa_nodes+1),(unsigned)2/*MPOL_MF_MOVE*/);
+    long rc = syscall(SYS_mbind,a,len,3/*MPOL_INTERLEAVE*/,&mask,
+            (unsigned long)(g_numa_nodes+1),(unsigned)0/*no MPOL_MF_MOVE — fresh mmap, no pages to migrate*/);
+    if(rc<0 && (int)-rc==1){ /* EPERM: no CAP_SYS_NICE — disable permanently */
+        g_numa_nodes=1;
+    }
 #else
     (void)p;(void)n;
 #endif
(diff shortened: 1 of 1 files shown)


ZacharyZcR (commented):
Ran the patch through a careful review — the intent is right (graceful degrade + no useless migration flag), but as written **the probe disables NUMA unconditionally, on every machine**, which is why chat works: this is functionally `COLI_NUMA=0`.

## The probe always fails — and not with EPERM

Two independent bugs compose:

**1. The probe buffer is not page-aligned.** `mbind` requires a page-aligned address; a stack `char buf[4096]` is 16-byte aligned at best, so the call returns **EINVAL**, not success. Reproduced with the probe verbatim:

```
buf addr 0x7fffe2e5dd90 page-aligned: NO
rc=-1 errno=22 (Invalid argument)
PR425 check (rc<0 && -rc==1): TRUE -> would DISABLE NUMA
```

**2. The error check tests the wrong value.** glibc's `syscall()` returns **-1 on any failure** and puts the code in `errno` — it does not return `-errno`. So `rc<0 && (int)-rc==1` reduces to `rc==-1`, i.e. *every* failure (EINVAL, ENOMEM, EFAULT…) is classified as "EPERM: no CAP_SYS_NICE".

Together: EINVAL from the unaligned probe → misread as EPERM → `g_numa_nodes=1` → every subsequent `numa_slab_bind` is a no-op → the interleave (and its measured **+13% 2S / +40% 4S**, #82) silently disappears. The crash goes away because the feature does.

(Also worth noting: plain `mbind` on your own memory doesn't require `CAP_SYS_NICE` — that requirement belongs to `MPOL_MF_MOVE_ALL`. So the EPERM premise itself is mostly moot.)

## Dropping MPOL_MF_MOVE is right — but it doesn't fix the OOM

Removing the flag is correct on its own merits: `numa_slab_bind` runs immediately after `posix_memalign`, before the pread fills the slab, so on the common path there is nothing to migrate. Keep that part.

But it does not address the crash. Re-ran the VMA measurement from https://github.com/JustVugg/colibri/issues/419#issuecomment-5016223536 with `flags=0` (this patch's exact call):

| scenario | VMA growth |
|---|---|
| no mbind | +1 total for 2000 slabs |
| mbind with MPOL_MF_MOVE | +2.00 per slab |
| **mbind with flags=0 (this patch)** | **+2.00 per slab** |

The VMA split comes from the per-region *policy*, not the migration flag. With NUMA actually enabled, `PIN_GB=all` still creates ~78k policy VMAs and still crosses the default `vm.max_map_count=65530` → the same `OOM slab`.

## Suggested shape

- Fix the probe: bind one page-aligned page (`posix_memalign(&p,4096,4096)`), check `errno==EPERM` specifically, free it after — keep the graceful-degrade idea as genuine hardening for constrained containers.
- Keep the `flags=0` change.
- For the actual crash, the fix has to remove the per-slab policy VMAs (scoped `set_mempolicy(MPOL_INTERLEAVE)` around the pin/dense load) or document the `vm.max_map_count` requirement. Happy to collaborate on that — but first, could you post `cat /proc/sys/vm/max_map_count` and a `wc -l /proc/<pid>/maps` at peak from your box? If your map count limit is raised, this patch will *appear* to work even with the probe fixed — while the default-config crash stays alive for everyone else.

PR #437, opened 2026-07-19
fix(serve): lift non-EOS stop tokens in C and fix redundant tools parsing in Python (#401)

### LINKED ISSUE #401: [Bug]: Tool Calling Fail in serve API
### Commit or version

main

### Environment

Debian 13
AMD 7950X
64GB DDR5
Samsung 990 Pro 2TB Gen 4
GT1030 2GB GDDR5
GT1030 2GB GDDR4

### Reproduction steps

When I use the Serve mode and connecting Visual Studio Code (Zoo Code) Plugin or OpenCoder

### Expected behavior

it wanna tool calls responses from the Model API

### Actual behavior and logs

```shell
but the API only answers with simple text
```

4ny3l:
This PR provides a complete, dual-layer fix for the tool-calling failures reported in #401 when interacting with real-world coding clients (such as Zoo Code / OpenCoder). 

### The Root Causes Addressed:
1. **C Engine (`c/glm.c`)**: In `serve` mode, the engine was inheriting non-EOS role-markers (`<|user|>` and `<|observation|>`) as hard stop tokens. Under heavy `int4` quantization, logit argmax noise frequently caused the model to hit these IDs immediately when attempting to open a tool block (e.g., emitting the character `<`). This prematurely cut generation short, returning a raw `<` string to the client.
2. **Python Gateway (`c/openai_server.py`)**: The `tools` array was being extracted from the request body twice. The second extraction inside `generation()` completely bypassed the tool filtering that `render_chat()` applies when `tool_choice` enforces a specific forced function call via a dictionary payload.

### Changes implemented:
* **`c/glm.c`**: Added a `SERVE` environment guard inside `stops_arm()` to filter out non-EOS stop tokens when running the API server. This lets the Python side control role boundaries while ensuring the C engine doesn't halt prematurely on noisy logit choices.
* **`c/openai_server.py`**: Refactored `generation()` to accept pre-processed `tools` and `tool_choice` directly from `chat_completion()`.
* **`c/openai_server.py`**: Added early structural schema validation for incoming `tools` arrays in `generation_options()` to return clear `HTTP 400` errors for malformed client inputs rather than failing deep inside the generation pipeline.

Fixes #401


JustVugg:
@4ny3l I rebased this onto current `dev` and pushed to your branch (maintainer edit) so you don't have to redo the work — the #391 refactor split `glm.c` into `colibri.c` + header modules, so your `stops_arm` change moved to `sample.h` (`stops_arm_tok`), and `openai_server.py` merged cleanly. Build-verified (colibri.c compiles, openai_server.py parses). Your fix is spot-on, by the way — I traced #401 independently and the reporter's own log shows the model emitting a bare `<` (it opens `<tool_call>`, then int4 argmax noise hits `<|user|>`/`<|observation|>` armed as stops → premature halt), which is exactly your root cause #1. No changes to your logic, just relocated the hunk. Now conflict-free and ready.

Issue #403, opened 2026-07-19
[Bug]: generation RSS overshoots the projected peak by ~40 GB on GB10 (Grace 20-core, 121 GB unified) — three kernel OOM kills; a low --cap contains it, --ram does not


anrasi:
### Commit or version

`72d3d37` (main), i8mm build (`idot: neon-i8mm`). The accounting looks unchanged on `dev` @ `caa49f7` (same `cap_for_ram` slack terms and auto-raise). The startup refusal added after #305 can't catch this class by construction: it only arms when even cap=1 doesn't fit the budget, and in all four runs below the cap was *raised* — the projection said everything fits.

### Environment

NVIDIA DGX Spark GB10 — Grace 10× Cortex-X925 + 10× A725 (aarch64, 20 cores), 121 GB unified LPDDR5x (`MemTotal 127535300 kB`), Ubuntu 24.04, kernel 6.17.0-1026-nvidia, gcc 13.3.0, NVMe. CPU-only decode (`[PIN] 0 VRAM`), OMP hot-thread re-exec (killed process shows as `exe`). Model: GLM-5.2 int4 + int8 MTP, converted from the official FP8 with the repo converter, token-exact vs the oracle (TF 32/32, greedy 20/20). Engine header: `experts@8-bit dense@8-bit`, resident dense 10.1 GB, load 12.8 s.

### Reproduction steps

```sh
# run C — explicit budget, killed by the kernel at t≈720:
THINK=1 COLI_MODEL=~/models/glm52_i4 \
  ./coli run --ctx 6144 --ngen 4096 --ram 75 "<550-token planning prompt>"

# run D — same budget, LRU capped low; the only survivor:
THINK=1 CAP_RAISE=0 COLI_MODEL=~/models/glm52_i4 \
  ./coli run --ctx 6144 --ngen 3072 --ram 75 --cap 16 "<same prompt>"
```

Runs A/B are run C without `--ram` (auto budget), `--ngen` 3072 / 4096 respectively. All four use the same 550-token prompt; long-form generation is the trigger and the per-16-token stats line is what makes the growth visible. (`THINK=1` is honored here via a local two-line patch to `coli run` — stock `run` hardcodes the nothink template, fix in PR #402 — so generations include the reasoning block. No engine changes: the binary is stock `72d3d37`.)

### Expected behavior

[...]

PR #445, opened 2026-07-19
pin: with CUDA_RELEASE_HOST the VRAM prefix must not consume the RAM pin budget — +72% on 6×5090

ZacharyZcR:
Found while chasing #431's walls down to their roots.

## The bug

`pin_load` computes `npin` from the RAM budget **first**, then carves the VRAM-ranked prefix out of those slots. With `CUDA_RELEASE_HOST` the prefix's host slabs are freed right after upload — so on a multi-GPU host the top-ranked experts consume the RAM budget **without occupying RAM**, and the CPU tier pins only the leftovers.

Measured on the 6× RTX 5090 / 251 GB box: `9,280 VRAM + 1,721 RAM (32.5 GB warm)` — on a machine whose RAM fits ~10k more. The cold tail paid disk on every token; hit rate ceilinged at 99.0% no matter how long the run.

## The fix

Hoist the VRAM budget estimate above the `npin` finalization and make the release-destined prefix **additive** to the RAM-derived count. Non-release configs are untouched (`prefix_est` stays 0; the arithmetic reduces to today's exactly).

## Measured (same box, same env + fix, `PIN_GB=all PIN_FILL=1 CUDA_RELEASE_HOST=1`, `TEMP=0 DRAFT=0`)

```
[PIN] placement: 9,280 VRAM + 10,176 RAM expert (192.5 GB warm)
expert hit rate 100.0% (pin 100.0% + lru 0.0%)   ← disk 0, from the FIRST token
```

| | before | after |
|---|---|---|
| 1,024-token greedy decode | 3.62 tok/s | **6.21 tok/s (+72%)** |
| warm late segment (t=768→1024) | 5.11 | **5.73 (+12%)** |
| prefill (25 tok) | 10.4 s | **8.8 s** |

The whole-run gain is the LRU warmup penalty disappearing — full residency from token one. The late-segment gain is the steady 0.9%-miss disk residue recovered. The #403 `rss_guard` still enforces the measured-RSS ceiling at runtime (peak RSS 195.9 GB on 251 GB here).

`make check` 77/77, zero warnings.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -4774,6 +4774,29 @@ static void pin_load(Model *m, const char *statspath, double gb){
         double avail=expert_avail(m,ram_env,m->ebits,est_ctx);
         npin=avail>0?(int)(avail/eb):0;
     } else npin=(int)(gb*1e9/eb);
+#ifdef COLI_CUDA
+    /* The VRAM budget must be known BEFORE npin is finalized: with
+     * CUDA_RELEASE_HOST the VRAM-ranked prefix's host slabs are freed right
+     * after upload, so those slots must NOT consume the RAM pin budget.
+     * Before this, a 6-GPU host lost its top ~9k ranked experts from the RAM
+     * count and pinned only the leftovers (measured: 9,280 VRAM + 1,721 RAM
+     * on a box whose RAM fits ~11k — the cold tail then paid disk forever). */
+    double remaining[COLI_CUDA_MAX_DEVICES]={0}, placed_b[COLI_CUDA_MAX_DEVICES]={0};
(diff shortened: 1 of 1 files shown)


ZacharyZcR:
@JustVugg flagging this one for priority review when you get a chance: it's a one-block budget-accounting fix with a measured **+72% decode** on the 6-GPU host (3.62 → 6.21 tok/s over 1,024 tokens, hit rate 99.0 → 100.0, disk reads to zero) — every multi-GPU `CUDA_RELEASE_HOST` deployment is currently leaving this on the table. Full A/B protocol in the description; happy to walk through it.

Issue #441, opened 2026-07-19
[Performance]: PILOT prefetch runs at NVMe queue-depth 1, and DIRECT=1 re-reads WILLNEED-prefetched pages


cdhdt:
These are two coupled observations about the speculative expert-prefetch path,
from reading `c/glm.c`, `c/st.h`, `c/uring.h` on `main` @ 72874f3. They are
code-level hypotheses, not yet measured end-to-end — I'll follow up with a
measured A/B on a real streaming setup (see "Measurement plan"). Filing now so the
analysis is on record and can be discussed.

### 1. Non-io_uring PILOT_REAL prefetch runs at NVMe queue-depth 1

`pilot_worker` is a **single** pthread, and on the non-`URING` `PILOT_REAL` path it
calls `pilot_realload` → one **blocking** `expert_load` at a time. All speculative
bandwidth therefore flows through a single in-flight request, so a drive that wants
QD 8–32 to reach its rated random-read bandwidth sees QD 1. The demand-miss pool
(`PIPE`, `g_pp`, 8 workers) already runs at depth; the speculative path does not.

**Idea:** spawn `PIPE_WORKERS` pilot threads, or route non-uring pilot loads through
the existing `g_pp` pool, so speculation runs at the same queue depth as demand
misses.

### 2. `DIRECT=1` and hint-only PILOT fight each other

With `DIRECT=1`, demand loads read through the O_DIRECT fd (bypassing the page
cache), but hint-only PILOT (`g_pilot_real==0`) warms the page cache via
`POSIX_FADV_WILLNEED` (`st_prefetch` / `expert_prefetch`). The O_DIRECT demand read
then **bypasses those warmed pages and re-reads from disk** — the prefetch cost is
spent on the scarcest resource for zero benefit. The two features are effectively
mutually exclusive as wired.

**Idea:** when `g_direct` is set, force `PILOT_REAL` (real loads into the next
layer's cache slots) and skip the WILLNEED path entirely.

### Why this could be high-impact

In the disk-streaming regime, ~11 GB of routed experts are read per token and the
drive is the bottleneck (per benchmarks.md). How much of that 11 GB is off the
critical path is set almost entirely by (a) prefetch queue depth and (b) not
wasting prefetched bytes — exactly the two things above.

### Measurement plan (why this is an issue, not yet a PR)

The honest A/B needs a real slow-ish NVMe + the full GLM-5.2 checkpoint, streaming
[...]

Issue #428, opened 2026-07-19
[Feature]: Metal fused-decode program — parallel r_top8 first (+6.7% measured, identical output), following the #136 principle (by Fable 5 in partnership with @Monotophic)


monotophic:
### Problem

This is the Metal follow-through on the principle from #136 ("on unified memory the lever
is a compute backend, not a placement tier"), and the companion analysis promised at the
close of #387. Coordinates: M5 Max 128 GB, macOS 26.5, GLM-5.2 int4, engine lineage off
`dev` `caa49f7` (+ the #392 residency set), champion decode config, 256-token windows,
interleaved brackets 2.04–2.11 tok/s.

Metal decode on this reference setup has been pinned at ~2.1 tok/s. We instrumented the
fused decode path (gap-free timing taxonomy, per-CB GPU timestamps, self-checked against
the PROFILE aggregates; replicated within 0.5%) and decomposed where a token's ~476 ms
actually goes:

| component | ms/token |
|---|---|
| disk (felt) | 308 |
| layer CB kernel exec (attention+dense+shared+router) | 73.7 |
| MoE CB kernel exec | 49.9 |
| per-CB scheduling slack | 34.3 |
| CPU orchestration (readback wait, resolve, submit) | 30.5 |

A kernel-ceiling microbench at the model's exact shapes then localized the layer CB's
inefficiency (it runs ~3.7× above its memory-bandwidth floor, while the MoE GEMVs are
already within ~2× of theirs): **`r_top8`, the top-8 expert selection, executes on a
single serial GPU thread — 0.465 ms/layer ≈ 36 ms/token ≈ 55% of the layer CB.**

### Proposed solution

A measured, env-gated, identical-output fused-decode program for the Metal backend, built
on our hardware as a reference implementation (we understand no maintainer currently has
Metal hardware). First installment already built, reviewed, and measured:

1. **Parallel r_top8** (32-lane exact-match replacement, including tie semantics and the
   topp/normk tail): **2.10 → 2.24 tok/s (+6.7%), generated output byte-identical across
   all arms.** Bit-exactness is proven two independent ways — a property-by-property
   algebra review of the reduction, and a 45-case adversarial/fuzz suite (ties, denormals,
   cross-lane boundaries, S up to 4 engine / 64 standalone) running the engine's own
   compiled pipelines — plus a SIMD-width guard that degrades to the serial kernel on any
[...]

Issue #398, opened 2026-07-18
[Feature]: Proposal: Creating an OpenAI-compatible API wrapper for Colibri using Claude Fable 5


ronnystevens-cmd:
### Problem

Hi everyone,

Like many of you, I am absolutely blown away by the Colibri engine. Being able to run a frontier-class model like GLM-5.2 (744B) via SSD streaming on mid-range and portable hardware is an absolute tipping point for local AI.

Currently, Colibri runs beautifully as a minimalist CLI engine. However, to truly unlock its potential, we need a way to connect it to external agentic frameworks (like Hermes, LangChain, or Autogen) and popular frontends (like Open WebUI or SillyTavern). To do that, we need an OpenAI-compatible API wrapper (specifically supporting the /v1/chat/completions endpoint).

Since the core engine is cleanly written in a single, compact C file, it is the perfect candidate for a high-context, advanced coding model like Claude Fable 5 to architect an API shell. Fable 5's massive context window and self-correction loops could easily generate a lightweight C-based HTTP server or a highly efficient Python/Go wrapper around the binary.

The Idea:
I am looking for a capable developer in the community who has experience with the Colibri codebase (or C-based inference engines) to spearhead this. You wouldn't have to start from scratch—we can leverage Fable 5 to feed it the Colibri source code, prompt it to build the API endpoint structure, and handle the integration.

I currently have a high-end testing environment ready (including an upcoming dedicated 128GB RAM setup with fast PCIe NVMe storage) and would love to help test, benchmark, and validate the API implementation once we get a prototype rolling.

If anyone is capable and interested in collaborating on prompting/building this Fable 5 API bridge, let’s team up and bridge the gap between Colibri's raw streaming power and the local agentic ecosystem!

Looking forward to your thoughts,
[Ronny Stevens aka "Flitsken"]

### Proposed solution

Creating an OpenAI-compatible API wrapper for Colibri using Claude Fable 5

ronnystevens-cmd:
a preliminary prompt direction:Role: Senior AI Systems Architect (Fable 5) Goal: Create a Python-based API Wrapper for the 'Colibri' inference engine to enable remote usage of GLM-5.2 (744B).

Context: We have a Linux Mint system with an RTX 3080 and 64GB RAM running Colibri Engine (~/colibri-ai/colibri). The model is located at ~/colibri-ai/model/glm-5-2-744b.bin. The engine uses SSD-streaming mode.

[...]

PR #446, opened 2026-07-19
io: skip the wasted WILLNEED on expert weights under DIRECT=1 (#441 safe half)

cdhdt:
> **Now measured** — see "Measurement" below. Two independent A/Bs exist: @michael-denyer's
> on Apple Silicon (a decisive **−36 % throughput** from the wasted WILLNEED) and mine on
> x86/Linux NVMe (directionally right but **much weaker: flat tok/s, −3 % disk service**).
> Reporting both, including the one that doesn't flatter the change. This is the **safe
> half of #441**.

## Summary

Under `DIRECT=1`, `expert_load` reads the expert **weights** through the O_DIRECT
fd (bypassing the page cache). But `expert_prefetch` still issues
`POSIX_FADV_WILLNEED` on those same weight tensors — warming pages the O_DIRECT
demand read never consumes. That's **wasted readahead on the disk**, the scarcest
resource when streaming.

The `.qs` **scales are always read buffered** (`pread_full` on the normal fd,
`expert_load` ~L1329), so their WILLNEED still helps — this PR keeps it and skips
only the weight WILLNEED under `g_direct`:

```c
for(int k=0;k<3;k++){
    snprintf(nm,...,suf[k]);
    if(!g_direct) st_prefetch(&m->S,nm);   // weights go O_DIRECT under DIRECT -> WILLNEED wasted
    char qs[320]; snprintf(qs,...); st_prefetch(&m->S,qs);  // scales always buffered -> keep
}
```

This only affects the hint-only PILOT path (`PILOT=1`, `PILOT_REAL=0`) combined
with `DIRECT=1`. `fadvise` is advisory, so **output is bit-identical** — no
rounding, no routing change.

## Why draft / what's left

The benefit is *avoided wasted disk I/O*, which only shows up when actually
cold-streaming experts. The measurement to run (and attach here before marking
ready):

```sh
# baseline dev vs this branch, same prompt, cold cache, greedy:
PILOT=1 DIRECT=1 ./coli run --temp 0 --model /path/glm52 "<prompt>"
# compare tok/s and the disk-service line, with/without the patch
```

The **riskier half of #441** — multi-threading the single pilot prefetch thread
(NVMe queue-depth 1 → 8-32) and forcing `PILOT_REAL` under DIRECT — is a separate
change that engages the cross-layer real-load machinery; it stays out of this PR
and lands with its own runtime A/B.

## Validation (so far)

- [x] `make -C c check` — portable build 0 warnings, all C unit + 83 Python tests pass
- [x] Native (`ARCH=native`) build 0 warnings
- [x] Bit-identical by construction (advisory `fadvise` only; no compute/routing touched)
- [x] **Cold-stream A/B on real NVMe** — done (see Measurement; weak on x86/Linux, decisive on Apple Silicon per @michael-denyer)

## Compatibility

- [x] Default CPU build stays dependency-free

Closes part of #441 (the weights-WILLNEED-under-DIRECT half).

---
*Prepared with AI assistance (Claude), per the project's AI-contribution precedent
(#428, #398). Build + `make check` run locally; cold-stream A/B now measured (below).*

---

### A — @michael-denyer, Apple Silicon (decisive; posted in the thread below)

M5 Pro, 64 GB, macOS 26.5.2, GLM-5.2 REAP-504B int4 (168 experts/layer),
`DIRECT=1 PIPE=1 PIPE_WORKERS=8`, greedy 128-token decode, interleaved runs:

| arm | tok/s | expert hit | disk-wait |
|---|---|---|---|
| baseline | 2.63 | 64.9 % | 26.0 s |
| `PILOT=1` hint-only WILLNEED | 1.68 | 65.0 % | 42.2 s |

Advisory WILLNEED under `DIRECT=1` cost **36 % throughput and +62 % disk-wait at a flat hit
rate** — precisely the mechanism this PR removes. They also confirmed the `.qs` distinction
this PR makes ("correct and finer than ours was") and adopted it.

### B — mine, x86/Linux (much weaker; reported as measured)

i7-12700H, 62 GB RAM, NVMe, full GLM-5.2 744B int4 (358 GB), `DIRECT=1 PILOT=1` (hint-only),
greedy `--temp 0`, 17-token prompt / 20-token decode, `dev` vs this branch:

| arm | tok/s | expert hit | disk service (decode) | disk wait |
|---|---|---|---|---|
| `dev` | 0.22 | 41.0 % | 602.7 s | 49.9 s |
| this PR | 0.22 | 42.1 % | **583.9 s** | 49.0 s |

**Flat on throughput, ~−3 % disk service.** Direction is right, magnitude is nowhere near the
Apple-Silicon result. Plausible reasons: Linux `POSIX_FADV_WILLNEED` + O_DIRECT behaves
differently from macOS `F_NOCACHE`; my decode window is short (20 vs 128 tokens) and my hit
rate regime differs (41 % vs 65 %). Single run per arm, no medians.

```sh
cd c && COLI_MODEL=/path/glm52_i4 DIRECT=1 PILOT=1 ./coli run --gpu none --temp 0 --ngen 20 \
  "Give three prime numbers and explain why each is prime."
# compare the decode-block PROFILE 'expert-disk … service' between dev and this branch
```

**Honest summary:** the change removes I/O that is *provably* wasted under `DIRECT=1`
(O_DIRECT bypasses exactly the pages the WILLNEED warms) and is bit-identical in output, so it
cannot be a regression. Whether it converts to throughput appears strongly platform-dependent:
decisive on Apple Silicon, marginal on this x86/Linux NVMe.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -1701,12 +1701,21 @@ static void expert_host_ensure(Model *m, int layer, ESlot *s){
 #endif
 
 /* prefetch asincrono dei pesi di un expert (e delle sue scale .qs): avvia il readahead
- * cosi' le letture sincrone successive trovano la page-cache calda. */
+ * cosi' le letture sincrone successive trovano la page-cache calda.
+ * Sotto g_direct i PESI vengono letti con O_DIRECT (bypassa la page-cache, vedi
+ * expert_load): il WILLNEED su di essi scalda pagine che la lettura di domanda non
+ * consuma -> readahead sprecato sul disco, la risorsa piu' scarsa nello streaming.
+ * Le scale .qs restano SEMPRE bufferizzate (pread sul fd normale), quindi il loro
+ * WILLNEED resta utile anche con DIRECT=1. fadvise e' solo consultivo: saltarlo non
+ * cambia mai l'output (bit-identico), riduce solo I/O sprecato.
+ * EN: under O_DIRECT the weights bypass the page cache, so their WILLNEED is wasted;
+ * the .qs scales are always buffered, so keep theirs. Advisory hint -> output-preserving. */
 static void expert_prefetch(Model *m, int layer, int eid){
     char nm[300];
     const char *suf[3]={"gate_proj.weight","up_proj.weight","down_proj.weight"};
     for(int k=0;k<3;k++){
-        snprintf(nm,sizeof(nm),"model.layers.%d.mlp.experts.%d.%s",layer,eid,suf[k]); st_prefetch(&m->S,nm);
+        snprintf(nm,sizeof(nm),"model.layers.%d.mlp.experts.%d.%s",layer,eid,suf[k]);
+        if(!g_direct) st_prefetch(&m->S,nm);
         char qs[320]; snprintf(qs,sizeof(qs),"%s.qs",nm); st_prefetch(&m->S,qs);
     }
 }


michael-denyer:
This matches something we measured on Apple Silicon two days before your issue, so here is the missing A/B for the exact half this draft implements.

Setup: M5 Pro, 64 GB, macOS 26.5.2, GLM-5.2 REAP-504B int4 container, expert-pruned to 168 per layer, DIRECT=1 PIPE=1 PIPE_WORKERS=8, greedy 128-token decode, interleaved runs on one binary.

| arm | tok/s | expert hit | disk-wait |
|---|---|---|---|
| baseline | 2.63 | 64.9% | 26.0s |
| PILOT=1 hint-only WILLNEED | 1.68 | 65.0% | 42.2s |

Advisory WILLNEED under DIRECT=1 cost 36% throughput and added 62% disk-wait at flat hit rate, which is exactly the mechanism in your issue: the drive spends bandwidth warming pages the O_DIRECT demand reads never consume. One caveat on the harness: pin-state warming ran across runs, which if anything favors the later PILOT run, so the direction is understated rather than inflated.

We have been running the equivalent gate locally since, and your .qs distinction is correct and finer than ours was: the scale reads go through pread_full on the buffered fd in every mode, so their WILLNEED stays useful under DIRECT. We adopted that refinement today (michael-denyer/colibri@bbf5d6d).

On the queue-depth half of #441, one tempering datapoint from the same campaign: real speculative loads (PILOT_REAL) raised hit rate from 64.9% to 85.3% and converted none of it to throughput here, 2.40 vs 2.63 tok/s. On a drive the demand pool already saturates, deeper speculation moves bytes earlier instead of moving fewer bytes, so the QD-1 fix may matter mainly on setups where the demand path leaves bandwidth idle.

cdhdt:
Thank you — that A/B is exactly the number this draft was missing, and it's a much cleaner
demonstration of the mechanism than anything I could produce here. Glad the `.qs` split was
useful; that one came out of tracing `expert_load` and noticing the scale tensors go through
`pread_full` on the buffered fd in *every* mode, so their WILLNEED survives `DIRECT=1` while the
weights' does not.

I've now measured the same half on x86/Linux, and I should report that **it is far weaker on my
setup** — I've put both A/Bs in the PR body rather than just the flattering one.

i7-12700H, 62 GB RAM, NVMe, full GLM-5.2 744B int4 (358 GB), `DIRECT=1 PILOT=1` (hint-only),
greedy, 17-token prompt / 20-token decode, `dev` vs the patch:

| arm | tok/s | expert hit | disk service (decode) | disk wait |
|---|---|---|---|---|
| `dev` | 0.22 | 41.0 % | 602.7 s | 49.9 s |
| patch | 0.22 | 42.1 % | 583.9 s | 49.0 s |

Flat on throughput, ~−3 % disk service. Single run per arm. Candidate reasons for the gap with
your −36 %: Linux `POSIX_FADV_WILLNEED` may simply be cheaper/lazier than macOS's readahead under
`F_NOCACHE`; my decode window is short (20 vs 128 tokens); and my hit-rate regime is different
(41 % vs 65 %). So on my box the change is best described as "removes provably-wasted I/O, can't
regress" rather than a throughput win — while on yours it plainly *is* one.

**On your queue-depth caveat: your instinct was right, and my data is a stronger version of it.**
I implemented the Part-1 half of #441 (multi-thread the blocking pilot so speculative loads run
at QD N instead of 1 — SPMC ring claim, plus the URING path's visible `-(eid+2)` reservation
ported to the blocking loader) and measured it on the same box:

| `PILOT_WORKERS` | decode tok/s | expert hit | disk service | disk wait |
|---|---|---|---|---|
| 1 | **0.26 / 0.25** | 58.8 / 59.6 % | 436 / 442 s | 40.7 / 42.6 s |
| 8 | 0.20 / 0.20 | 61.8 / 59.9 % | 927 / 975 s | 66.6 / 68.4 s |

(two runs each, second pass with the arm order reversed and a native build, to kill ordering and
build confounds — the numbers reproduce closely.)

So not merely "converts none of it to throughput" — here it actively **costs ~20 %**, doubling
disk service, and on the replication it didn't even buy extra residency (hit rate flat). Same
mechanism you named: the demand pool already saturates this drive, so speculation competes with
demand reads instead of saving bytes. I've posted the full negative result on #441 and I'm **not**
opening a perf PR for it. The implementation is output-preserving (byte-identical sha256 across
PILOT off / 1 / 8 on the full model) — it's simply not a win.

One side finding from that work, unrelated to throughput: `rss_guard` frees an evicted expert
slab *outside* `g_pilot_mx` after marking the slot `eid = -1`, so a pilot worker can reuse the
slot and `pread` into the slab as it's freed (UAF/double-free when the RAM guard is active). It
reproduces with a single pilot worker on `dev`. Filed as #448 with a one-hunk fix in #449.

Last thing, for the record: I had also posted a "−17 % `expert-matmul`" end-to-end figure on my
int8/int4 kernel PR (#444) from a single run per arm. Re-running it with two passes per arm showed
that comparison was warm-up drift, not the kernel — successive runs got faster regardless of arm,
and `dev` beat the patched build in a later pass. I've retracted it there; that PR now claims only
the isolated microbench (2.13–2.81×, bit-exact) and explicitly claims no end-to-end win. Mentioning
it here because it's the same lesson your interleaved-runs note points at.

---
*Measured and written with AI assistance (Claude), per the project's AI-contribution precedent
(#428, #398).*

Issue #448, opened 2026-07-19
[Bug]: rss_guard frees an evicted expert slab outside g_pilot_mx — use-after-free / double-free with the pilot prefetcher


cdhdt:
### Summary

`rss_guard` marks an evicted expert slot `eid = -1` **under `g_pilot_mx`, then unlocks, and only
afterwards frees `s->slab`**. In the window between the unlock and the free, the slot is visible
to every other thread as `{eid = -1, slab = still valid}` — which is exactly the state the pilot's
slot picker treats as "free, reuse me first". A pilot worker can therefore claim that slot and
start `pread`ing into `slab` while `rss_guard` frees it: **use-after-free**, and a **double-free**
if the loader takes its own realloc path.

Found while reviewing eviction paths for #441; it is **not** introduced by that work — it
reproduces on `dev` as-is.

### The code (`c/colibri.c`, `rss_guard`)

```c
ESlot *s = &m->ecache[l][lru];
s->eid = -1;                                  // slot published as "free" …
pthread_mutex_unlock(&g_pilot_mx);            // … and the lock is dropped here
int64_t sb = s->slab_cap + s->fslab_cap*4;
compat_aligned_free(s->slab); free(s->fslab); // … but the slab is freed only now
s->slab = NULL; s->fslab = NULL; ...
```

The victim scan in `pilot_realload` reuses free slots first:

```c
if(Sl[z].eid == -1){ slot = z; break; }       // takes rss_guard's slot, slab still non-NULL
```

and `expert_load` will happily reuse a non-NULL `slab` whose capacity fits.

### Interleaving

1. main thread, `rss_guard`: picks LRU slot `S`, sets `S.eid = -1`, **unlocks**.
2. pilot worker: locks, victim scan hits `S` (`eid == -1`), reserves it, unlocks, enters
   `expert_load` → begins a long `pread` **into `S.slab`**.
3. main thread resumes: `compat_aligned_free(S.slab)` → frees the buffer the worker is reading
   into; `S.slab = NULL` then clobbers the pointer. The worker may publish a "resident" expert
   with freed/NULL weights → later `moe()` reads it → crash or garbage.
   If the worker's `expert_load` took the realloc-and-free path first, step 3 is a **double-free**.

### Conditions

- Requires the RAM guard to be active: `RSS_GUARD_GB` set, or `g_ram_budget_gb > 0` (which
  `cap_for_ram` sets whenever a RAM budget is resolved — common on RAM-constrained hosts).
[...]

PR #449, opened 2026-07-19
rss_guard: free the evicted expert slab under g_pilot_mx (fixes a use-after-free with the pilot prefetcher)

cdhdt:
## Summary

`rss_guard` published an evicted expert slot as free (`eid = -1`) **under `g_pilot_mx`, then
unlocked, and only then freed `s->slab`**. In that window the slot reads as
`{eid = -1, slab still valid}` — precisely the state `pilot_realload`'s victim scan reuses first
(`if(Sl[z].eid == -1){ slot = z; break; }`). A pilot worker can claim it and `pread` into the
slab while `rss_guard` frees it: **use-after-free**, or a **double-free** if the loader takes its
own realloc path.

This PR keeps the free (and the NULLing) **inside the critical section**, so "slab valid" and
"slot reusable" are never simultaneously observable.

Full analysis in #448.

## Why it matters

- Triggers whenever the RAM guard is active (`RSS_GUARD_GB`, or any resolved `g_ram_budget_gb`)
  **and** the pilot is on — i.e. exactly on RAM-constrained hosts under memory pressure.
- **Reproduces with a single pilot worker**: `rss_guard` runs on the main thread while the pilot
  worker runs in the background. It is not tied to any multi-worker experiment.
- `rss_guard` also decrements `m->ecap`, pushing workers into the reuse branch — so the collision
  gets *more* likely exactly when the guard fires.

## The change

One hunk: move `compat_aligned_free(s->slab); free(s->fslab);` and the pointer/capacity NULLing
above the `pthread_mutex_unlock(&g_pilot_mx)`. The lock is now held across a `free()` (µs); pilot
workers only take it briefly for scan+reserve, so contention is unchanged in practice.

## Validation

- [x] `make -C c check` — portable build 0 warnings, all C unit tests + Python tests pass
- [x] Native (`ARCH=native`) build 0 warnings
- [x] No behaviour change on the non-guard path: the guard's logic, accounting (`freed`,
      `dropped`, `ecap--`) and log line are untouched; only the unlock moves.
- [ ] CUDA (`make -C c cuda-test`) — N/A, CPU-side eviction path only

## Compatibility

- [x] Default CPU build remains dependency-free

---
*Found and fixed with AI assistance (Claude) while reviewing eviction paths for #441, per the
project's AI-contribution precedent (#428, #398).*

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -4331,16 +4331,21 @@ static void rss_guard(Model *m){
             if(lru<0){ pthread_mutex_unlock(&g_pilot_mx); continue; }
             ESlot *s=&m->ecache[l][lru];
             s->eid=-1;                                    /* nascosto: nessun hit/evict altrui */
-            pthread_mutex_unlock(&g_pilot_mx);
             int64_t sb=s->slab_cap + s->fslab_cap*4;
 #ifdef COLI_METAL
             if(s->slab && g_metal_enabled) coli_metal_unregister(s->slab);
 #endif
+            /* La free resta SOTTO g_pilot_mx. Sbloccando prima, lo slot e' visibile come
+             * {eid=-1, slab ancora valido}: il pilota (pilot_realload) riusa per primo gli
+             * slot eid==-1, quindi puo' prenderlo e fare pread dentro lo slab MENTRE lo
+             * liberiamo -> use-after-free / double-free. Slab valido e slot riusabile
+             * devono restare mutuamente esclusivi finche' il puntatore non e' NULL. */
             compat_aligned_free(s->slab); free(s->fslab);
             s->slab=NULL; s->fslab=NULL; s->slab_cap=s->fslab_cap=0;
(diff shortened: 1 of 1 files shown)


JustVugg:
Reviewed and confirmed correct. Moving `compat_aligned_free(s->slab)` + `s->slab = NULL` (and the QT-view reset) **under `g_pilot_mx`**, with the unlock after, makes the eviction atomic from the pilot's point of view: the slot transitions `{eid valid, slab valid}` → `{eid = -1, slab = NULL}` with no intermediate `{eid = -1, slab valid}` window, so `pilot_realload` can never reserve the slot and `pread` into a slab that's being freed. Minimal and exactly the right shape — the slightly longer lock hold (free inside the critical section) is a non-issue since `free` is fast and correctness wins over a few microseconds of pilot contention. The Metal `coli_metal_unregister` moving under the lock is fine too. LGTM — I'll merge once CI is green. Thanks @cdhdt.

Issue #334, opened 2026-07-17
fmt=4 CUDA attention crash — FIXED (#298); benchmark tracker + follow-ups (PROFILE accounting, cache warmth)


woolcoxm:
## fmt=4 CUDA attention crash — fixed; benchmark tracker + follow-ups

Tracking the resolution of the g64-model hard-reboot crash and the benchmark status. **The crash is fixed** (PR #298, commit `22a224e`). This issue captures the benchmark data, the CUDA fix summary, and the open follow-ups so people following along have the full picture.

Supersedes/closes the crash hypothesis in #326 (the "possible VRAM-crash correlation" there is now confirmed and fixed — see below).

---

### The crash, and the fix

**Symptom:** the g64 (fmt=4 grouped int4, gs=64) model hard-crashed the machine (black screen → audio loop → reboot) when run with the full CUDA stack (`COLI_CUDA_ATTN=1 COLI_CUDA_PIPE=2 COLI_CUDA_MTP=1` + `REPIN=16`). The old per-row int4 model never crashed.

**Root cause:** the two MLA attention absorb kernels in `backend_cuda.cu` (`attention_absorb_kernel`, `attention_absorb_batch_kernel`) applied quantization scales **per output row** (`wscale[row]`) — the fmt=2 semantic. PR #298 added fmt=4 support to `quant_matmul` (per-group scales: `scales[o*ng + g]`) but missed these two kernels. The g64 `kv_b` tensor is fmt=4 (ng=8 groups/row), so its scale buffer is `O*ng` long; indexing it with a bare row index read with the wrong stride → GPU memory fault → Windows bugcheck `0x116 VIDEO_TDR_FAILURE` (`nvlddmkm.sys`, `STATUS_INSUFFICIENT_RESOURCES`).

**Fix (PR #298 `22a224e`):** added an `absorb_scale()` `__device__` helper mirroring `quant_matmul`'s fmt=4 branch (`wscale[row*ng + k/gs]` for fmt=4, `wscale[row]` otherwise), applied it inside the Q- and V-projection accumulation loops of both kernels, and threaded `w->gs, w->ng` through all six launch sites. For fmt≠4, `ng==1` → bit-identical to prior behavior. No `extern "C"` signature or header changes.

[...]

PR #429, opened 2026-07-19
numa: per-layer pin arenas — fix the PIN_GB=all VMA explosion (#419)

ZacharyZcR:
Fixes #419. Supersedes the mechanism-level parts of #425 (and keeps its two good ideas, done so they work).

## Root cause (measured, not guessed)

`OOM slab` on the 4 TB box is **VMA count exhaustion, not memory**: every per-slab `mbind` carries its own memory policy, so bound regions cannot merge with neighbours — measured **~2 VMAs per slab, with or without `MPOL_MF_MOVE`** ([experiment](https://github.com/JustVugg/colibri/issues/419#issuecomment-5016223536)). A `PIN_GB=all` load is 19,456 experts × (slab + fslab):

```
19,456 × 2 allocations × ~2 VMAs ≈ 78k  >  vm.max_map_count = 65,530 (default)
```

`posix_memalign` → `mmap` → `ENOMEM` at the limit, with terabytes free — and the failure lands mid-load, exactly as reported. It also explains #425's own `numastat` (362 GB on node 0): that patch's probe disables NUMA outright, so it loads by not interleaving at all.

## The fix

**One arena per layer instead of one policy per slab.** Experts of a layer share a tensor shape, so a layer's pins pack at a fixed stride into two arenas (weights + scales): **2 mbinds and a handful of VMAs per layer instead of ~500**. Whole-model: ~150 policy VMAs instead of ~78,000.

- Slices are pre-attached to the pin slots before the load loop; `slab_cap` covers `expert_load`'s realloc check, so its alloc branch never fires — **`expert_load` itself is untouched**.
- `aslab` marks arena ownership: `expert_host_release` detaches instead of freeing (a REPIN gpu-swap promotion must not `free()` an interior pointer), `expert_host_ensure` re-attaches the slice before reloading. GPU-prefix slots stay on individual allocs — their host backing really is released after upload, exactly as today.
- Kept from #425, fixed so they work: **`flags=0`** (every bind lands before the pread that first-touches the pages — nothing to migrate), and the **EPERM probe** — now on a page-aligned page (`mbind` rejects unaligned addresses with EINVAL) checking `errno==EPERM` specifically, so a constrained container degrades with a message and an EINVAL can never masquerade as a missing capability (the #425 probe disabled NUMA unconditionally on every machine — see the [review](https://github.com/JustVugg/colibri/pull/425#pullrequestreview-4731062620)).
[...]

PR #451, opened 2026-07-20
cuda: grouped-int4 (fmt=4) support in the expert-group kernels — opens the GPU tier to g64 and E8 containers (#334)

ZacharyZcR:
Makes #334's prevention note real, and unblocks the quantization roadmap's GPU path.

## Why

The grouped MoE kernels were **per-row-only**: `GroupDesc` had no group-size fields, `row_bytes()` returned 0 for fmt=4, the scale buffer was hardcoded to `O` floats — and a fmt=4 group that slipped to the generic path would have been silently decoded **as int2**. Net effect: the GPU expert tier is closed to every grouped container — the g64 quality line (#225, −63% of the int4 quality loss) and the E8 lattice route (#347, int3-e8-rot measuring *better* than shipped per-row int4 at 25% fewer bytes) could only run their experts on the CPU. On the bandwidth-walled decode path, bytes are speed — the grouped containers are pointless without this.

## What

- `ColiCudaTensor` gains `gs`/`scale_count`; upload allocates `O·ceil(I/gs)` scales for fmt=4 and applies the same offset→signed nibble conversion as fmt=2 (identical packing). New ABI entry `coli_cuda_tensor_upload_g` carries `gs` **without touching the existing symbol** — an old Windows DLL missing it returns 0 and the tensor stays CPU-side, no crash.
- `GroupDesc` gains per-tensor group sizes; new `grouped_hidden_g4_dual` / `grouped_down_g4` apply the per-group scale inside the accumulation. `gs` must be even (a packed byte never straddles a group); `gs=0` degrades to per-row, so fmt=2 members ride the same launch in mixed groups.
- `coli_cuda_expert_group` routes any group containing fmt=4 through the g4 kernels; **pure-fmt=2 groups keep today's paths byte-identical**. The generic fallback now explicitly rejects fmt=4 instead of decoding garbage.

## Verification

- `tests/test_grouped_g4_cuda.cu` — kernel-vs-CPU oracle (semantics cloned from `matmul_i4_grouped`): 50 trials × 3 experts covering gs=64, a non-divisible tail group (200 % 64 = 8), and a per-row member in the same launch — **zero mismatches** on a 5090.
- `make check` 77/77; CPU, CUDA and MinGW cross builds clean.

Next on this line: convert the g64/E8 containers and A/B the VRAM-resident expert tier end-to-end — this PR is the gate that had to open first.

### DIFF
--- a/c/backend_cuda.cu
+++ b/c/backend_cuda.cu
@@ -20,6 +20,8 @@ struct ColiCudaTensor {
     float *scales;
     size_t weight_bytes;
     int fmt, I, O, device;
+    int gs;                    /* quant group size; 0 = per-row scales (#334) */
+    size_t scale_count;        /* floats in `scales`: O per-row, O*ng grouped */
     int tracked;
     RaggedKVEntry ragged[512];
     int ragged_count;
@@ -43,6 +45,7 @@ typedef struct {
 typedef struct {
     const void *g,*u,*d; const float *gs,*us,*ds;
     int gf,uf,df,rows,offset;
+    int ggs,ugs,dgs;      /* per-tensor quant group size; 0 = per-row scales (#334 fmt=4) */
 } GroupDesc;
 
 static DeviceContext g_ctx[COLI_CUDA_MAX_DEVICES];
@@ -81,6 +84,7 @@ __host__ __device__ static size_t row_bytes(int fmt, int I) {
     if (fmt == 1) return (size_t)I;
     if (fmt == 2) return (size_t)(I + 1) / 2;
     if (fmt == 3) return (size_t)(I + 3) / 4;
+    if (fmt == 4) return (size_t)(I + 1) / 2;   /* grouped int4: nibbles like fmt 2 */
     return 0;
 }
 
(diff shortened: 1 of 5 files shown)


ZacharyZcR:
Merge-order note, since five of my open PRs touch `backend_cuda.cu` and all currently read CLEAN against `dev` — they will conflict with **each other** as they land, so here is the dependency map to save you the analysis:

**Independent, land anytime in any order:** #429 (colibri.c only), #453 (tools only), #424 (site/docs only).

**Ordered, all in `coli_cuda_expert_group`:**
1. **#433** — narrows the `TC_W4A16` branch gate (9 lines)
2. **#451** (this one) — adds the fmt=4 branch to the same if/else chain, so it wants #433's gate already in place
3. **#432** — device router (mostly new functions; small overlap)
4. **#434** — resident accumulation; **stacked on #432's branch**, so #432 must land first or the diff carries it
5. **#342** — group-dispatch overlap, touches the same scheduling region

#433 ↔ #451 is the sharp one (same if/else chain); the rest are near-additive.

**You do not need to follow this order** — merge whatever you have reviewed, and I will rebase the rest the same day, same as I did for #429. Just flagging so a merge doesn't look unexpectedly conflicted.

monotophic:

While preparing the Metal twin of this work (#457) we traced #451 against `dev` (@ e9b3614) for a
CUDA↔Metal parity comparison, and one finding seems worth putting on this thread's radar before the
merge window — especially given the merge-order plan upthread puts #451 in the ordered tier.

**Claim, hedged (traced across three files, not run):** if #451 merges *without* #298, there is a
plausible silent-wrong-answer path on the dense/attention-projection CUDA path:

- `qt_load` (`colibri.c:834`) sets `t.cuda_eligible=1` unconditionally by fmt under
  `CUDA_DENSE=1`; only `kv_b` gets an explicit opt-out (`:874`). `q_a`/`q_b`/`kv_a`/`o`/shared-expert
  do not.
- `matmul_qt_ex`'s CUDA dispatch (`colibri.c:419-450`) carries **no fmt gate** (the Metal block at
  the same call site gates on `w->fmt==1||w->fmt==2`).
- **Today this is accidentally safe:** `coli_cuda_tensor_upload`'s `row_bytes(4)==0` fails the
  upload, `cuda_failed=1` sticks, and the gs-aware CPU fallback (`matmul_i4_grouped`) serves the
  tensor — slow but correct.
- **After #451 alone:** `row_bytes(4)` is fixed, so the upload at this call site now *succeeds* —
  but through the plain (non-`_g`) upload wrapper, which sets `t->gs=0` and copies only `O` scale
  floats (not `O*ng`), and `quant_matmul`/`weight_at` then decodes fmt=4 through the untouched
  `fmt==3` (int2) fallback branch. Silent wrong answer on GPU.

**Bounding the trigger:** it fires only if a real checkpoint carries fmt=4 on a *dense*
(non-expert) tensor AND the user runs `CUDA_DENSE=1`. Your own commit message hedges the same way
("fmt=4 is primarily an expert format in practice"), and our Metal PR discloses the identical open
question on our side (#457 body, §9) — two independent traces landing on the same named uncertainty
is why we're flagging it rather than assuming baseline-safe-by-omission.

We hit exactly this "two changes must land together or the intermediate state is armed" class on
the Metal branch during review and disarmed it by commit reordering, so it felt worth passing on.
Any one-line guard disarms it here, e.g. a fmt gate at the `matmul_qt_ex` CUDA dispatch mirroring
the Metal site (smallest), routing this call site through the new `_g` upload path, or having the
plain upload refuse fmt=4. No opinion on which — and none needed if #298 is expected to land in the
same window, since its combined design covers the dense path.

Great to see the CUDA grouped tier moving — the formats contract lined up cleanly in our parity
check (packing, group math, scale layout), which should make the two platform tiers easy to review
as companions.

mohamedmastouri2000-boop:
@ZacharyZcR — nice work getting grouped-int4 (fmt=4) into the CUDA expert-group kernels here. Testing it on Blackwell (RTX 5080, sm_120) with the public g64 container, I hit one correctness issue you'll want to know about, and it's cleanly isolated.

The **expert-group path is correct**, but the **resident-dense fmt=4 path garbles output** when dense weights are pinned to VRAM. Same binary (current `dev`, this PR merged), same g64 container, one env var apart:

- `COLI_CUDA=1` + experts on GPU, **dense on CPU** (no `CUDA_DENSE`): coherent — "CUDA (Compute Unified Device Architecture…"
- `COLI_CUDA=1 CUDA_DENSE=1` (dense **resident on GPU**): **garbage** — "odesk odesk odesk…"

`CUDA_DENSE=1` is the usual g64-on-GPU config (it's what puts the ~12 GB dense in VRAM), so this hits the main path. My read is that the group size arriving via the global `g_upload_gs` isn't being applied to the **resident dense** tensor uploads, so those tensors get per-row scales instead of the gs=64 grouped scales → wrong values. The streaming expert-group kernels are unaffected because they carry `gs` through the descriptor.

Full write-up + a fix are on #298 (which passes `gs` as a per-call parameter to `coli_cuda_tensor_upload` / `coli_cuda_matmul` and produces coherent output under the identical `CUDA_DENSE=1` config): https://github.com/JustVugg/colibri/pull/298#issuecomment-5025085943 . Repro on request. Thanks again — this is close, just the resident-dense scale plumbing.

mohamedmastouri2000-boop:
@monotophic — your trace is exactly right, and I can confirm it empirically: I hit precisely the silent-wrong-answer you predicted, running on Blackwell (RTX 5080, sm_120) with the public g64 container.

Your two states map one-to-one to what I saw:

- **Before #451** ("accidentally safe"): every dense fmt=4 tensor logged `... disabled after an error; falling back to CPU` (the `row_bytes(4)==0` upload failure you describe), so it ran on CPU — slow but coherent.
- **After #451, `CUDA_DENSE=1`**: the dense upload now *succeeds* through the plain wrapper, and the output is garbage — `odesk odesk odesk…`. That's the fmt=4 nibbles being decoded through the `fmt==3`/int2 branch exactly as you traced (`gs=0`, `O` scales instead of `O*ng`). It only fires with dense fmt=4 + `CUDA_DENSE=1`, matching your bounding.

Clean isolation on the *same* current-`dev` binary, one env var apart: no `CUDA_DENSE` (dense on CPU, experts on GPU) → coherent; `CUDA_DENSE=1` (dense resident on GPU) → garbage.

And your read that #298 covers the dense path holds up on hardware: my rebased #298 branch, same `CUDA_DENSE=1` config, produces coherent output with the full residency (952 tensors / 12.48 GB dense + 109 experts in VRAM, zero fallbacks) — because it routes the dense upload through the per-call `gs` parameter and copies `O*ng` grouped scales instead of `O`.

So: your static trace + my Blackwell repro land on the same thing from opposite directions — #451 alone arms a silent-wrong-answer on the dense path under `CUDA_DENSE=1`, and #298 disarms it. Details + the rebase are on #298 (https://github.com/JustVugg/colibri/pull/298). Nice catch on the parity comparison — great to have the Metal twin cross-checking this.

monotophic:

@mohamedmastouri2000-boop — thank you for running this down on real hardware; a clean one-env-var
isolation with the coherent-vs-garbage pair is exactly the confirmation that trace needed, and your
before/#451 fallback-log observation nails the "accidentally safe" state we described.

@ZacharyZcR — the expert-group path holding up under that test is good news for the tier itself.
Since #298's combined design covers the dense path properly and is rebasing clean, we see two
sound routes: land #298 soon, or an interim one-line fmt gate at `matmul_qt_ex`'s CUDA dispatch
(mirroring the Metal site's `w->fmt==1||w->fmt==2`) so `dev` isn't shipping the armed state in the
meantime. Your call on which — happy to see this tier moving either way.

PR #454, opened 2026-07-20
Port the coli CLI launcher to Go (dependency-free) — first step for #310

Magnet-js:
Addresses #310 (make the runtime path Python-free by porting the CLI/gateway to Go).

## Scope — CLI process management only (incremental first step)

Ports the `coli` launcher (`c/coli`, 761 lines) to a native Go binary. The gateway and offline tooling **stay in Python** and are invoked as subprocesses, so the Python code remains authoritative and its test suite is untouched.

| Command | Implementation |
|---------|----------------|
| `build`, `info`, `run`, `chat` | native Go (Python-free) |
| `serve`, `web` | spawn `openai_server.py` (gateway stays Python) |
| `plan`, `doctor` | drive `resource_plan.py` / `doctor.py` via `python3 -` |
| `bench`, `convert` | run the Python tools unchanged |

## Design

- **New `c/go/` module, standard library only — zero third-party requires.** Colibri's runtime path is dependency-free by design (CONTRIBUTING.md) and the gateway is stdlib-only for the same reason; the Go binary keeps that property (single static binary). Cobra/Viper were considered and rejected on these grounds.
- **Additive:** no `c/*.py` file is modified. `make check` and the Python tests (63/63) still pass.
- **Delegation over re-porting:** `plan`/`doctor`/`--auto-tier` feed a small driver to `python3 -` (stdin) that imports the existing modules, so no support file needs a new entrypoint.
- Byte-exact engine protocol reproduced: `\x01\x01READY\x01\x01\n` / `\x01\x01END\x01\x01\n` sentinels, `STAT tok tps hit rss` lines, `\x02RESET`/`\x02MORE` control writes, two-stage Ctrl-C in `chat`.

## Build

```sh
make coli-go            # optional target, not part of all/check
# or: cd c/go && go build -o ../coli-go ./...
```

The binary finds the engine and Python support files exactly like `./coli`; `COLI_ENGINE` overrides the engine path.

## Verified parity vs the Python `coli`

- `--help` / `info` / error strings — matches `tests/test_cli_output.py` contract
- `env_for` child environment — byte-identical across `--ram/--temp/--topp/--topk/--ngen/--ctx/--repin/--gpu/--vram/--policy/--auto-tier`
- `plan` / `doctor` JSON + text — identical (same `resource_plan.py`/`doctor.py`)
- `chat` — identical output in both stderr scenarios; multibyte UTF-8 split across engine chunks renders identically
- `serve` — spawns the gateway, `/health` + `/v1/models` answer, SIGINT/SIGTERM shut down cleanly with no orphan
- `bench` / `convert` — argv construction matches (two-pass convert, dataset-skip logic)
- `gofmt`/`go vet` clean; cross-compiles for linux and windows

## Known deltas (documented in `c/go/README.md`)

- Windows `OMP_NUM_THREADS` defaults to the logical CPU count (`runtime.NumCPU()`) rather than physical cores — an overridable `setdefault`.
- The `chat` TTY input box is drawn once per prompt rather than cursor-redrawn; behaviour is otherwise identical.


JustVugg:
Thank you for building this — it's clean work, and the motivation in #310 is a real problem: the runtime path shouldn't need Python when the engine is dependency-free C.

I'm going to be straight with you rather than let it sit: **I don't want to add Go to this project.** Not because of the code, which is fine, but because of what it commits us to. colibri is a C project whose whole promise is one compiler and zero dependencies; adding a second toolchain means every contributor and packager needs it, both launchers have to be kept in step forever, and every future flag lands twice or drifts. That's a permanent cost paid by everyone, for a problem that only affects people who don't have Python.

What I *do* want is the outcome you're after — no interpreter on the runtime path. If we're going to spend a rewrite, the natural answer for this project is **C**: the launcher's job is argument parsing, environment setup and spawning a process, all of which the engine already does natively. Then `coli` is a small binary next to `colibri`, built by the same `make`, shipped in the same archive, with nothing new to install — and the release archives stop carrying Python files at all.

There's a smaller, useful piece hiding in your PR that I'd take today: the **process-management and environment-setup logic**, which you clearly got right. If you wanted to bring that over as C, or to write up what the Go version handles that the current Python launcher doesn't (signal handling, exit codes, Windows quirks), that would go straight in.

I realize this isn't the answer you hoped for after the work you put in, and I'm sorry to give it after the PR has been open a while. That's on me for not saying it sooner. The reasoning is about the project's shape, not the quality of what you wrote.

Magnet-js:
Hi @JustVugg 
Thanks for the clear answer.
Quick clarification: having the CLI in both Python and Go was only meant as a transition until this PR got merged, my plan was to follow up with another PR removing the Python CLI.

That said, I understand your point about C. My main reason for choosing Go was that it's easy to write a simple, readable HTTP server in it. As far as I know, that's not really the case for C (change my mind).

But overall, I agree with you on the CLI. I'd be happy to keep working on it in C as well, though I'm not very familiar with C, my background is more Java/Kotlin and Go.

Maybe next time we should discuss this directly in the related issue.

JustVugg:
Closing to clear the queue — the direction was declined back in July and you accepted it graciously at the time, so this is bookkeeping rather than a new decision. **#310** stays open as the tracker if the dependency-free-launcher question is ever revisited.

Thanks for taking the "no" well; the argument you made for it was a reasonable one.

Issue #452, opened 2026-07-20
fmt=5: a deployable E8-lattice grouped container — int3 that beats shipped int4 quality at 25% fewer bytes (design + plan)


ZacharyZcR:
## Why now

Decode on every measured host is a **bandwidth wall**: expert bytes ÷ effective GB/s is the largest bucket on the critical path (6×5090 forensics in #431: ~47 ms/token of CPU-tier expert sweep at hit 100%). Bytes are speed. The quantization ablations (#81, #255, #347) found the lever:

| container | quality (MMLU-class Δ) | bytes vs int4 |
|---|---|---|
| shipped per-row int4 | −9.3pp | 1.00× |
| uniform int3-g64 | −7.5pp | 0.75× |
| **int3-g64-e8-rot** | **−5.9pp** | **0.75×** |

A 3-bit E8-lattice container measures *better* than the shipped 4-bit while carrying 25% fewer bytes — that is simultaneously: −25% on the CPU-tier wall, +33% experts per GB of VRAM (the whole routed set of GLM-5.2 drops to ~280 GB), and a quality *gain* over what users run today.

**But those numbers are simulated**: `tools/quant_ablation.py` quantizes to the nearest E8 point and stores floats. There is no deployable index encoding, no container format, no decode kernel. This issue is the plan to build them.

## The pedigree (so nobody thinks this is exotic)

E8 codebooks are QuIP#'s core trick (ICML'24) and are **mass-deployed** in llama.cpp's IQ2/IQ3 quant family (E8-derived grids + sign-bit factoring, MIT). Our additions are the rate-scaled ball (#347 — the radius must grow with the bit budget), the g64 outer scale, and the target: a tiered MoE expert stream where bytes convert to tok/s at a measured rate.

## Design: `fmt=5` (E8 grouped container)

1. **Block encoding** — decision point, two candidates:
   - **(a) IQ3-style subset codebook** (adopt llama.cpp's proven construction, MIT): small LUT (256–512 entries) that fits in SIMD registers / GPU shared memory, signs factored out by parity. Deployable today, kernels are a known quantity.
   - **(b) our rate-scaled ball with structured coordinate coding**: matches the ablation exactly but needs entropy-shaped index construction (a 3-bpw E8 ball has ~10⁷ points — naive indexing doesn't close). Research-grade.
[...]

PR #426, opened 2026-07-19
Metal: rebase MTLResidencySet expert residency onto split runtime

rgbkrk:
## Summary

Rebase the `MTLResidencySet` expert-residency work from #392 onto current `dev` after the `c/glm.c` split in #391, following [JustVugg's maintainer guidance](https://github.com/JustVugg/colibri/pull/392#issuecomment-5015917396).

The mechanism remains env-gated behind `COLI_METAL_RESSET=1`. On macOS 15+, it attaches a persistent residency set to the Metal command queue and removes the per-command-buffer `useResource:` walk over expert buffers from `moe_submit`. Gate-off behavior remains unchanged.

## Rebase notes

- Rebased all five commits from #392 onto current `dev`.
- Git's rename handling moved the two coordinator-side hunks from removed `c/glm.c` into `c/colibri.c`.
- `git range-diff` shows the mechanism commits are patch-equivalent apart from that file relocation.
- Updated newly introduced documentation and comments to refer to `colibri.c`, and corrected the E5 comparison table to acknowledge its instrumentation and OOM-unwind touches.

## Attribution

This is a rebased continuation of Christopher Brand's work in #392. All five mechanism commits preserve @monotophic as their original Git author. Additional M5 Max validation and analysis remain documented on #392 by @monotophic and @RDouglasSharp.

## M3 Max validation

Hardware: Apple M3 Max, 128 GB unified memory, macOS 26.4.1.

- `make -C c colibri METAL=1`: passed
- `make -C c colibri METAL=0`: passed
- `make -C c metal-test`: passed all Metal kernel, MoE, GEMM, and fused-attention tests
- Full C test suite: passed
- Python test suite: 77 passed
- Explicit `-Wall -Wextra` Objective-C++ compile: only the pre-existing unused `TG` warning from current `dev`

The real-model comparison used the same prompt and settings as the prior local baseline: `Compare the myths of Lucifer and Prometheus`, `--ram 96 --ngen 1024`, `DIRECT=1`, `IO_THREADS=8`, `MTP=0`, `DRAFT=0`, plus `COLI_METAL_RESSET=1`.

| Run | Decode | Expert hit | RSS |
|---|---:|---:|---:|
| Historical older-main baseline | 1,024 tokens in 1,684.90s, **0.61 tok/s** | 71.8% | 84.88 GB |
| Rebased current-`dev` + residency set | 1,024 tokens in 990.81s, **1.03 tok/s** | 71.7% | 84.88 GB |

[...]

PR #457, opened 2026-07-20
metal: grouped-int4 (fmt=4) GEMV support + two latent fmt=4 fixes

monotophic:
# metal: grouped-int4 (fmt=4) GEMV support — the Metal twin of #298/#451

Cross-references: **#298** and **#451** (CUDA fmt=4 support, read READ-ONLY for
layout/naming conventions — `gs`/`ng` field names, "same packed layout as fmt=2, scale
folded into the accumulation" semantics, and the "new ABI entry rather than break the
existing symbol" caution both echoed below). **#168** (Team A's int3-g64 reconciliation
— Stage 2 of this same effort, not started here; see §7). Campaign context, same framing
as **#447**: no maintainer of this repository has Metal hardware — this lab is the test
surface for the Apple GPU tier of every quantization format the CPU/CUDA tiers gain.

Branch: `metal/grouped-gemv`, based on `origin/dev` @ `e9b3614`. Touches
`c/backend_metal.h`, `c/backend_metal.mm`, `c/colibri.c`, `c/tests/test_backend_metal.mm`.

---

## 1. The problem: two latent defects that mutually masked each other at stock

**Durable claim:** dev's grouped-int4 format (fmt=4 — packed int4 nibbles, same layout
as fmt=2, but one f32 scale per `gs`-element group along I instead of one scale per row)
has a CPU reference (`matmul_i4_grouped`, `c/quant.h:150`) and an active CUDA tier
(#298/#451). The Metal GEMV shader (`mm_gemv` in `c/backend_metal.mm`) only branched on
`fmt == 1|2|3`; anything else — including fmt=4 — fell through to the unconditional
`else` arm and was decoded **as raw f32**. Taken in isolation, that is not a "CPU
fallback, no GPU acceleration" gap — it is a silent-wrong-answer defect: a code path that
reaches the Metal GEMV with an fmt=4 tensor gets a real, non-error, numerically-wrong
result. That structural claim is unchanged from the first revision of this document, and
the code evidence for it (below) is unchanged too. What changed, on review (credit:
review round 1, auditor — see §10), is the REACHABILITY claim that followed it.

**Revised: this defect was NOT live at stock `e9b3614`.** A second, independent defect
masked it, and §2(d) already contained the evidence — this section just failed to draw
the conclusion from it. §2(d) traces the actual allocation site of the fmt=4 scale
buffer: `qt_from_disk`'s fmt=4 branch allocated it with `falloc` — a bare `malloc`, never
page-aligned, never `coli_metal_register`'d — while `resolve()` in `backend_metal.mm`
only finds pointers inside registered slabs. That means at stock, `resolve(t->s, ...)`
on *any* fmt=4 scale buffer always fails, so `resolve_attn`/`bind_gemv` always return
`false` and the caller always falls back to CPU — for every fmt=4 tensor,
unconditionally, before the shader's fmt=4 fallthrough ever gets a chance to execute.
**The two defects mutually cancel at stock: the allocator bug is what kept the shader bug
from ever firing.** Stock was accidentally safe, not correctly handled — an important
distinction, covered next — but that is not the same claim as "this was a live landmine
in stock `e9b3614`," which the first revision of this section overclaimed by tracing only
the shader/dispatch side (`bind_gemv`'s missing fmt gate, quoted below) without also
propagating §2(d)'s own resolve()-fails finding back into this section's reachability
claim.

**Where the shader defect *would* become live, and how this branch's commit ORDER avoids
it.** Fix either defect alone, and depending on which one lands first, the other may stop
providing cover before the real fix arrives. If the allocator fix landed before the
shader fix, the intermediate commit would be ARMED: the scale buffer resolves
successfully, but the shader still doesn't know what fmt=4 means, so `bind_gemv` would
dispatch straight into the f32 fallthrough and silently compute garbage — a state
strictly worse than stock (stock's defect was masked; this one wouldn't be), reachable by
anyone who bisects or cherry-picks that single commit. An earlier revision of this branch
carried exactly that ordering — corrected on review (credit: review round 1, coordinator
— see §10) before this PR was finalized, not left as a residual risk.

This branch's actual, final commit order avoids the armed state by construction rather
than by discipline: **`f0fa5a9`** (the `mm_gemv` shader fix + all host plumbing, §3/§4)
lands FIRST, and **`2ba12d8`** (the `qt_from_disk` allocator fix, §2(d)) lands SECOND. At
`f0fa5a9` alone, the new shader branch is already correct, but the allocator bug still
keeps every real fmt=4 scale buffer from resolving — the capability lands **inert**, not
armed: verified (this branch, this session) that `make glm METAL=0`/`METAL=1` build
clean at that commit and `make metal-test` still passes 27/27, because metal-test's fmt=4
cases construct their own tensors directly via `posix_memalign`+`coli_metal_register` and
never go through `qt_from_disk` — the test suite exercises and proves the new shader
correct at this commit, it just isn't reachable yet from a real loaded model (§6).
`2ba12d8` then activates that already-correct capability by fixing the one thing that was
keeping it unreachable — verified clean/27-27 again at that commit. **No commit in this
three-commit sequence is ever unsafe to check out alone:** capability lands inert first,
the activation commit completes it. The third commit (**`fc0f5a5`**, the attn-test
blind-spot fix, §10) is a test-only strengthening layered on top afterward, with no
bearing on either commit's reachability argument. A reviewer or maintainer bisecting this
branch, or merging a prefix of it, never lands on the armed ordering described above — it
does not exist anywhere in this branch's actual history.

**What doesn't change from the prior framing:** both defects are real (validator
independently confirmed both with quotes — see §10), both are fixed by this PR, and the
trace below (including the QUOTED gate at `colibri.c:1958-1960` showing `attention_rows`
never checked `q_a`/`q_b`/`kv_a`/`o`'s fmt) is accurate as a description of the CODE'S
STRUCTURE — `bind_gemv` genuinely has no fmt validity gate, and the shader genuinely
mishandles fmt=4 if reached:

```c
if(g_metal_enabled && !kvs && S<=4 && (g_absorb==1||(g_absorb<0&&S<=4)) && m->kv_start[layer]==0
   && D==6144 && H==64 && c->q_lora==2048 && c->kv_lora==512 && c->qk_nope==192
   && c->qk_rope==64 && vh==256 && l->kv_b.fmt==2){
```

(the pre-existing gate, `git show e9b3614:c/colibri.c` lines 1958-1960, unmodified in
shape by this PR — only the argument list changed, see §4). No known field model
currently ships an fmt=4 dense/attention weight next to an fmt=2 `kv_b` (fmt=4
checkpoints are new), so even the structural gap was latent, not observed-in-the-wild.
What was wrong in the prior revision was conflating "structurally reachable in general"
with "reachable at stock `e9b3614` specifically" — the allocator bug traced independently
in §2(d) closes that gap at stock, even though the shader-side structural gap is real and
this PR still closes it (§3) for every path that reaches `mm_gemv`, not just the one this
PR's test matrix targets, and regardless of whether any known checkpoint currently
exercises it.

## 2. The trace: how a per-row scale reaches the shader today (quoted before any edit)

Three, and only three, host functions dispatch the `mm_gemv` kernel; all three pass
`scale` as `buffer(1)` and (before this PR) never told the shader anything about grouping:

**(a) `coli_metal_matmul`** (`backend_metal.mm`, the persistent-tensor entry point
metal-test's `run()`/`run_grouped()` harnesses call) — wraps weights and scales into
`MTLBuffer`s on first use:

```c
extern "C" int coli_metal_matmul(ColiMetalTensor **tp, float *y, const float *x,
                                 const void *weights, const float *scales,
                                 int fmt, int S, int I, int O) {
  if (!g_dev || fmt < 0 || fmt > 3) return 0;
  ...
      t->w = wrap(weights, t->wbytes);
      t->s = wrap(scales, (size_t)O * sizeof(float));     // <- hardcoded O floats
```

`fmt > 3` was rejected outright, and the scale buffer was always wrapped at exactly `O`
floats — for fmt=4 the real scale array is `O*ceil(I/gs)` floats, so even lifting the
`fmt>3` guard alone would have UNDER-wrapped the buffer (`wrap()` sizes the `MTLBuffer`
from this byte count) and the shader would read past it.

**(b) `coli_metal_gemm`** (the large-batch prefill GEMM path, reached from
`matmul_qt_ex` in `colibri.c` for `S >= g_metal_gemm_min`) — resolves already-registered
slabs rather than wrapping, so no buffer-size bug here, but the fmt gate excluded 4
outright and no `gs` was threaded through:

```c
extern "C" int coli_metal_gemm(float *y, const float *x, const void *wp, const float *sp,
                               int fmt, int S, int I, int O) {
  if (!g_dev || (fmt!=1 && fmt!=2)) return 0;
```

reached from (`colibri.c:421-424`, `matmul_qt_ex` — the general-purpose matvec dispatch
every dense/attention/shared-expert projection goes through):

```c
if(g_metal_enabled && S>=g_metal_gemm_min && !spec_pinned() && (w->fmt==1||w->fmt==2) && !omp_in_parallel()){
    const void *wp = w->fmt==1 ? (const void*)w->q8 : (const void*)w->q4;
    if(coli_metal_gemm(y,x,wp,w->s,w->fmt,S,w->I,w->O)) return;
}
```

**(c) `bind_gemv`** (the internal helper used by the fused single-command-buffer decode
kernels, `encode_attention` and `coli_metal_layer_decode`'s shared-expert calls) — no fmt
gate at all (§1), and no `gs` parameter to pass even if it wanted to honor one:

```c
static bool bind_gemv(id<MTLComputeCommandEncoder> e, const void* w, const float* s, int fmt, int I, int O,
                      id<MTLBuffer> xin, id<MTLBuffer> yout, int S){
  ...
  [e setBytes:&S ...]; [e setBytes:&I ...]; [e setBytes:&O ...]; [e setBytes:&fmt ...];
  [e setBytes:&NT ...];
```

reached, per weight, from `colibri.c`'s `attention_rows`/`layer_forward_rows` via a
`WP_(q)` macro that already correctly resolves fmt=4 tensors to their `.q4` pointer
(dev's own CPU-side dispatch does the same — `qt_from_disk` stores grouped weights in
`t->q4`, confirmed at `colibri.c:824`):

```c
#define WP_(q) ((q).fmt==1?(const void*)(q).q8:(const void*)(q).q4)
int ok = coli_metal_attn_decode(x,
    WP_(l->q_a), l->q_a.s, l->q_a.fmt, l->q_a_ln, ...
```

**(d) One more link in the chain, found by tracing to the actual allocation site, not
just the dispatch call sites:** `qt_from_disk` (`colibri.c:823-825`, unmodified in shape
by this PR except the one-line fix in §4) allocates the fmt=4 scale array with plain
`falloc` — a bare `malloc`, not page-aligned and never `coli_metal_register`'d — while
every other fmt uses `qsalloc` (`qalloc` under the hood), which page-aligns and registers
under `COLI_METAL`:

```c
else if(fmt==4){ int ng=(I+gs-1)/gs;
    if(t->fmt!=4||!t->q4){ t->fmt=4; t->O=O; t->I=I; t->gs=gs; t->q4=qalloc(nb); t->s=falloc((int64_t)O*ng); }
```

`resolve()` in `backend_metal.mm` only finds pointers inside `coli_metal_register`'d
slabs. A scale buffer allocated with `falloc` can never be found by `resolve()`, so
`bind_gemv`/`coli_metal_gemm` would `return false`/`0` for it — **safe** (CPU fallback,
no wrong answer), but it means fixing (a)-(c) alone would not have mattered in practice:
every real fmt=4 dense/attention tensor's scale buffer would have permanently missed the
GPU path. Found while tracing the per-row-scale path this stage extends to per-group;
fixed alongside it (§4) since it is the literal allocation site of the buffer this whole
PR is about.

## 3. The shader change: per-group scale folded into the accumulation, not applied once

Unlike fmt 1-3 (`y[row] = acc * scale[o]`, one multiply after the full dot product),
fmt=4's group scale must be applied **per group, before summing across groups** — this is
`matmul_i4_grouped`'s actual math (`quant.h:150-184`), not a simplification of it. The new
`fmt==4` branch in `mm_gemv`:

```metal
} else if (fmt == 4) {            // int4 GROUPED: same nibble packing as fmt=2,
                                   // one f32 scale per gsz-element group along I.
  int rb = (I+1)/2; int ng = (I+gsz-1)/gsz;
  device const uchar* wr = w + (long)o * rb;
  device const float* scl = scale + (long)o * ng;
  for (int i = slane*2; i < I; i += 64) {
    uchar b = wr[i>>1];
    int g0 = i / gsz; float sc0 = scl[g0];
    acc += float(int(b&0xF)-8) * xr[i] * sc0;
    if (i+1 < I) {
      int g1 = (i+1) / gsz; float sc1 = (g1==g0) ? sc0 : scl[g1];
      acc += float(int(b>>4)-8) * xr[i+1] * sc1;
    }
  }
} else { ... }
...
acc = simd_sum(acc);
if (slane == 0) y[row] = (fmt == 4) ? acc : acc * scale[o];   // fmt=4: don't scale again
```

Each of the 32 SIMD lanes owns one packed byte (2 elements) per 64-wide stride — memory
access stays coalesced (`wr[i>>1]` is contiguous across lanes), and a group never splits
a byte because `gs` is always even (dev's own loader only accepts `gs` from the candidate
list `{16,32,48,64,96,128,192,256}`, `colibri.c:777`). This is intentionally the
straightforward per-element form, not vectorized like fmt=2's `uchar4`/`float4` path —
see §7 (Not done) for why that tradeoff was made deliberately, not by oversight.

A new `constant int& gsz [[buffer(9)]]` parameter carries the group size; it is ignored
for every other fmt (callers pass the tensor's `QT.gs`, which is 0 for non-grouped
tensors — harmless, since the shader only reads it when `fmt==4`).

## 4. The host-side plumbing: what changed at each traced site

- **`fmt_bytes`** gained an `fmt==4` case (same packed-nibble byte count as fmt=2 — the
  weight layout is identical, only the scale layout differs, matching #298's "row_bytes:
  fmt=4 = same packed int4 layout as fmt=2" note). New `fmt_scale_bytes(fmt,I,O,gs)`
  computes `O*ceil(I/gs)*sizeof(float)` for fmt=4, `O*sizeof(float)` otherwise — this is
  what `coli_metal_matmul` now uses to size the `wrap()` call in §2(a), fixing the
  under-wrap.
- **`coli_metal_matmul`**, **`coli_metal_gemm`**, **`bind_gemv`** all gained a `gs`
  parameter (appended at the call site, inserted right after the paired `fmt` in the
  fused-decode signatures — mirrors #451's convention of adding `gs` fields alongside
  `fmt`), and their fmt gates now accept 4. `coli_metal_matmul`'s doc comment in
  `backend_metal.h` states the new contract explicitly.
- **`AttnW`** (the per-layer weight-pointer bundle used by `encode_attention`) gained
  `qa_gs`/`qb_gs`/`kva_gs`/`o_gs` fields. **`coli_metal_attn_decode`** and
  **`coli_metal_layer_decode`** gained the matching parameters, threaded through to their
  `bind_gemv` calls (attention projections q_a/q_b/kv_a/o, plus shared-expert
  gate/up/down in the layer-decode variant).
- **`colibri.c`** call sites: `matmul_qt_ex`'s Metal-GEMM gate now includes `w->fmt==4`
  and passes `w->gs`; `attention_rows`/`layer_forward_rows` pass `.gs` alongside each
  `.fmt` already being passed (the field defaults to 0 for non-grouped tensors, so this
  is a no-op for every model that isn't using fmt=4 anywhere).
- **`qt_from_disk`**: the one-line allocator fix from §2(d) — `t->s` for fmt=4 now comes
  from `qalloc` (page-aligned + `coli_metal_register`'d) instead of `falloc`.

**`kv_b` is deliberately untouched.** It never flows through `bind_gemv`/`mm_gemv` at
all — the MLA absorption kernels (`a_qabs`, `a_ctx`) dequantize it inline with a
hand-written, per-row-only helper (`a_deqrow`, `backend_metal.mm`), a structurally
separate kernel family from the GEMV this PR extends. The existing `l->kv_b.fmt==2` gate
(§1) is left exactly as-is — not relaxed — specifically so an fmt=4 `kv_b` continues to
CPU-fallback safely rather than hitting `a_deqrow`'s per-row-only math. See §7.

**`coli_metal_moe_block`/`moe_gemv`** (the batched routed-expert path, a third,
independent kernel family from `mm_gemv`) is also untouched. It was already safely gated
to `fmt!=1 && fmt!=2 -> return 0/nil` before this PR — fmt=4 experts already CPU-fall-back
correctly today, with no wrong-answer risk, so this was a coverage gap, not a
correctness bug, and out of scope for this stage. See §7.

## 5. What was verified, and how (RAN, this branch)

`metal-test` follows the existing `run()`/inline-gemm-test pattern the spec named,
extended with an fmt=4-specific CPU reference (`cpu_ref_grouped`) and harness
(`run_grouped`) — `cpu_ref`'s `[O]` per-row scale layout cannot express fmt=4's
`[O,ceil(I/gs)]` layout, so it is a sibling reference, not a branch of the existing one.
`cpu_ref_grouped` mirrors `quant.h`'s `matmul_i4_grouped` nibble-decode and group-index
math exactly, accumulating in **double** — the same construction
`tests/test_i4_grouped.c` (dev's own CPU-side fmt=4 oracle, unmodified by this PR)
already established as the "known exact" reference for this format.

**Tolerance, stated and justified against the existing convention:** the pre-existing
`run()` cases compare `max|Δ| / max|y_ref|` (ymax-relative) at `1e-4`. The new
`run_grouped`/gemm-grouped/attn-grouped cases instead compare, per output element,
`|Δ| / Σ|term_i|` (magnitude-relative, summing `|xs[i]*(nib-8)*scale|` over every term of
that dot product) at the same `1e-4`. Reason: a grouped dot product over many
groups/signed terms can land near zero from cancellation — an ordinary fixed
GPU-vs-double rounding difference then reads as an enormous relative error against a
near-zero **result**, which is the accumulator's precision, not a kernel defect.
Comparing against the summed magnitude of the terms instead means a wrong scale index or
wrong group boundary is still caught as an O(1) relative error (it shifts the result by a
fraction of the terms actually summed), while cancellation noise is not misreported as a
bug. This is exactly the convention `tests/test_i4_grouped.c` uses for the same kernel's
CPU-side oracle (see its `ref_grouped`/`mag` comment, unmodified by this PR) — reused
here for a GPU oracle of the same format, not invented fresh. `1e-4` (vs.
`test_i4_grouped.c`'s `1e-6`) carries the same slack the existing `run()` cases already
accept for GPU-vs-CPU float comparisons in this file, because the GPU sums in a different
order (byte-pair-strided across 32 SIMD lanes, then `simd_sum`) than the scalar double
reference.

**Shape matrix (spec §"Validation"), all via `coli_metal_matmul` unless noted:**

| Case | Shapes | Evidence |
|---|---|---|
| I multiple of gs=64 | gate/up I=6144, down I=2048 (real g64-checkpoint dims), S=1 and S=4 | RAN, ok |
| I NOT a multiple of gs=64 | I=200 (even), I=201 (odd, exercises the nibble tail) | RAN, ok |
| I≤64 degenerate (single group) | I=64 (ng=1, exact), I=40 (<gs, single partial group) | RAN, ok |
| S=1 and S>1 | covered throughout (S∈{1,2,3,4}) | RAN, ok |
| Random + outlier-heavy rows | one activation forced to 50.0 in group 0, I=512 and I=201, verifying grouped reconstruction end-to-end through the GPU path (not just that the kernel runs) | RAN, ok |
| Large-batch GEMM path (`coli_metal_gemm`) | O=2048,I=6144,S=64,gs=64 — the second of the two directly-named GEMV entry points | RAN, ok |
| Fused-decode path (`bind_gemv` via `coli_metal_attn_decode`) | q_a set to fmt=4/gs=64 inside the real fused attention command buffer (S=1 pos=37, S=4 pos=12/MTP — **not** pos=0, see §10: at T=1 softmax is provably independent of q_a). Each case additionally compares q_a's raw pre-RMSNorm GEMV output against the CPU oracle directly (`qraw` metric), closing RMSNorm's blindness to a uniform scale error — proves the plumbing in §4 end-to-end, not just the standalone kernel, and not just modulo RMSNorm's invariances | RAN, ok |
| Negative control: fmt 1/2/3 unchanged | all 8 original `run()` cases + 2 moe_block + 1 gemm(int4) + 4 attn cases, unmodified assertions | RAN, ok (identical nerr values to the pre-PR baseline run, see §6) |

fmt=5 (stage 2): not attempted — blocked on Team A's `int3/g64-rebased` semantics, see §7.

## 6. Gates run on this branch (all RAN, 2026-07-19, macOS 26.5.2, Apple M5 Max 128 GB)

- **`make glm METAL=0`** and **`make glm METAL=1`**: clean, **0 warnings** (`-Wall
  -Wextra`), both before this PR's changes (stock `e9b3614`) and after — verified by
  capturing both build logs and grepping for `warning` (0 hits in all four: stock×2,
  branch×2).
- **`make metal-test`**: **before this PR (stock `e9b3614`): 15/15 cases pass.**
  **After this PR: 27/27 cases pass** — the original 15 unchanged (byte-identical `nerr`
  values, the negative control from §5), plus **12 new fmt=4 cases**: 9
  `run_grouped()` shape-matrix cases, 1 grouped `coli_metal_gemm` case, 2
  `run_attn_grouped()` fused-decode-path cases. **Case COUNT unchanged at 27/27 after
  review round 1** (§10) — the two `run_attn_grouped()` cases were strengthened in place
  (pos_base moved off the degenerate T=1 point; each case gained an internal raw
  pre-RMSNorm comparison against the CPU oracle, printed as `qraw=...`), not added
  alongside as new cases. See §10 for the mutation-detection matrix proving the
  strengthened cases actually catch what the original two could not.
- **`make test-c`** (all suites in `TEST_BINS`, including `tests/test_i4_grouped` — dev's
  own CPU-side fmt=4 oracle, unmodified by this PR): exit 0, 0 failures reported by any
  suite (e.g. `test_topp: 123 cases run, 0 failure(s)`, `test_dsa_select: 129 cases run,
  0 failure(s)`, `test_i4_grouped: ok`).
- **`make test-python`**: **83/83 tests, OK**, exit 0.
- `git diff e9b3614 --numstat` (this branch, after review round 1): 4 files —
  `c/backend_metal.h` (+25/-18), `c/backend_metal.mm` (+70/-36), `c/colibri.c` (+21/-14),
  `c/tests/test_backend_metal.mm` (+245/-5). No file outside `c/backend_metal.{h,mm}`,
  `c/colibri.c`, and its own test file was touched.

## 7. Scope statement: what this PR does NOT do (read before assuming full coverage)

- **`coli_metal_moe_block`/`moe_gemv` (batched routed-expert MoE path) does not gain
  fmt=4 support.** It was already safely gated to CPU-fallback for fmt=4 before this PR
  (§4) and remains so — extending it means a second, independent per-group-scale kernel
  branch (`moe_gemv` is bindless/pointer-array-based, structurally different from
  `mm_gemv`) plus new per-expert `gs` array parameters on `coli_metal_moe_block`/
  `coli_metal_moe_block_begin`. Given fmt=4 is primarily an *expert* format in practice
  (per #298's framing), this is real follow-up work, not a footnote — flagged here rather
  than attempted under this task's time/scope, with the header doc comment in
  `backend_metal.h` updated to say so explicitly.
- **`kv_b` (MLA absorption core: `a_qabs`/`a_ctx`) does not gain fmt=4 support.** These
  hand-written kernels dequantize inline with a per-row-only helper (`a_deqrow`) that
  never touches `mm_gemv`/`bind_gemv` — extending it is a different, larger shader change
  (new `a_deqrow_grouped` plus threading `gs` through `a_qabs`/`a_ctx`'s own buffer
  lists), and the existing `l->kv_b.fmt==2` gate (§1, §4) already keeps this safe by
  construction: an fmt=4 `kv_b` simply never reaches this GPU path. Not relaxed, not
  attempted.
- **Stage 2 (fmt=5, int3-g64) was not started.** Per the governing spec: build stage 1
  fully first if Team A's `int3/g64-rebased` reconciled semantics aren't yet available.
  This worker did not check out or read that branch — a finished stage 1 beats a
  speculative stage 2, and stage 2's addendum is explicitly gated on #168's merge, so
  starting it now would mean guessing at semantics likely to change. See UNCERTAINTIES.
- **The shader arithmetic is not vectorized** the way fmt=2's `uchar4`/`float4` path is.
  fmt=4's per-group scale lookup means a byte's two nibbles can belong to different
  groups (`g0`/`g1` in §3), which doesn't compose cleanly with the 8-wide vector loads
  the other formats use without either restricting `gs` to specific alignments beyond
  what the loader already guarantees, or adding a second code path for the aligned case.
  Given the spec's own framing ("the shader arithmetic is small" — correctness is the
  ask, not throughput) and no model-run performance measurement being permitted under
  this task's constraints anyway, the straightforward per-element form was chosen
  deliberately. A vectorized fast path for `gs`-aligned regions is natural follow-up work
  once real g64-checkpoint throughput numbers exist to justify it.
- **No CUDA files were touched.** #298/#451 were read read-only for convention only.

## 8. DEVIATIONS from the spec

- Spec's validation matrix names shapes generically ("I multiple of 64... I≤64
  degenerate"); this PR's `run_grouped` calls hardcode `gs=64` throughout (the real
  g64-checkpoint value, per #298's title and description) rather than also sweeping
  `gs∈{16,32,...,256}` the loader's `detect_group_size` candidate list supports. The
  shader itself is fully generic over `gs` (nothing in §3 assumes 64 specifically, only
  evenness) — this is a test-coverage choice, not an implementation limitation. Flagged
  as a deviation because the spec's shape matrix could be read as wanting the full
  candidate-list swept; it was not.
- Added a bind_gemv/fused-decode-path integration test (`run_attn_grouped`, §5) that
  the spec did not explicitly request (its validation section names `coli_metal_matmul`-
  style GEMV comparison and the gemm-test pattern). Added because §2/§4's plumbing work
  (`AttnW`, `coli_metal_attn_decode`/`coli_metal_layer_decode` signature changes) would
  otherwise be exercised by nothing — an untested code path in a PR whose entire premise
  is "a wrong PASS is the worst possible output" seemed like the wrong tradeoff to leave
  unaddressed even though not explicitly named.
- Fixed the `qt_from_disk` allocator bug (§2(d), §4) even though it lives outside
  `c/backend_metal.*` and is arguably "dev's pre-existing code, not this PR's shader
  work." Kept in-scope because it is the literal allocation site of the buffer this
  entire PR's plumbing traces and extends — leaving it unfixed would make §3/§4's shader
  and host-side work unreachable by any real fmt=4 dense tensor.

## 9. UNCERTAINTIES

- **Whether any current field checkpoint actually has fmt=4 on q_a/q_b/kv_a/o/shared-
  expert weights (as opposed to only MoE routed experts) is unknown to this worker.**
  §1's landmine and §4's plumbing are real given the loader's generic `qt_resolve_fmt`
  (any tensor loaded via `qt_from_disk` can be detected as fmt=4), but no specific
  checkpoint was inspected — this task ran no model and downloaded nothing. If in
  practice fmt=4 is expert-only today, §7's moe_gemv gap matters more in the field than
  this PR's bind_gemv fix does; both were done regardless, since the spec asked to trace
  "wherever the current per-row scale buffer flows into the shader," not to guess which
  path is hottest in practice.
- **Real-model throughput/quality impact is entirely unmeasured.** No model runs were
  permitted under this task's constraints; every number in §5/§6 is a synthetic
  kernel-vs-reference comparison. Whether fmt=4 dense tensors reaching the Metal GEMV
  produces a measurable decode-speed change (vs. today's CPU fallback) is UNVERIFIED.
- **`gs` values outside `{16,32,48,64,96,128,192,256}` (the loader's candidate list) were
  not tested**, though the shader and host plumbing place no restriction on `gs` beyond
  evenness. Untested does not mean unsupported, but it is UNVERIFIED beyond the tested
  set (which does include 64, the only value #298/#451's own checkpoints use).
- **Stage 2 (fmt=5) semantics, timeline, and #168's state were not checked** beyond
  confirming the branch name `int3/g64-rebased` exists in the spec's own text — this
  worker did not fetch or read it, per §7.

## 10. Review round 1 (credit — house style: the review record is part of the PR's evidence)

Three findings from independent review, all FIX-FIRST, all addressed on this branch. The
final commit sequence — **`f0fa5a9`** (shader + host plumbing + original tests),
**`2ba12d8`** (allocator activation), **`fc0f5a5`** (attn-test blind-spot fix) — already
reflects all three; there was an earlier, different commit sequence during review that
the third finding below corrected, but it was never pushed and left no trace other than
the two SHAs (`142ddd2`/`9315efa`/`5e3acd8`) some of the discussion below still names for
context. `git diff 5e3acd8 HEAD` is empty and the two states' tree hashes are identical
(`95ff3e6f...`) — the reorder changed commit BOUNDARIES and ORDER only, not one byte of
final content.

**Auditor:** traced that §1 (first revision) overclaimed reachability at stock — the
`falloc`/`resolve()` evidence already present in §2(d) shows the allocator bug masks the
shader bug at stock `e9b3614`, so the two defects mutually cancel there rather than the
shader bug being independently live. §1 above is rewritten to reflect this; nothing about
either defect's reality was retracted, only the "live landmine in stock" framing.

**Validator:** found two proven blind spots in the `run_attn_grouped()` fused-decode-path
cases (§5): (a) the original S=1 case used `pos_base=0`, giving `T=pos_base+S=1` — softmax
over a single key is identically `1.0` regardless of the score's value, so that case's
final output was provably independent of q_a's kernel entirely, not merely insensitive to
some subset of defects; (b) both cases fed q_a's output through RMSNorm before comparing
anything downstream, and `RMSNorm(c·v) == RMSNorm(v)` for any positive scalar `c` — a
whole-tensor uniform-scale-calibration bug in q_a's kernel would be completely invisible
to the fused-path comparison (`nerr`/`cache`) no matter how `pos_base` was chosen, since
(b) is independent of (a).

**Fixes applied** (both in `tests/test_backend_metal.mm`, third commit): the S=1 case now
uses `pos_base=37` (mirroring `run_attn`'s own non-grouped S=1 pos=37 case, which exists
for the identical reason); both cases now additionally capture q_a's raw GEMV output
*before* `t_rms` overwrites it in place, and compare that raw output directly against the
CPU oracle (`cpu_ref_grouped`, the same magnitude-relative construction `run_grouped()`
uses) via a standalone `coli_metal_matmul` call on the exact weight/scale/x data the
attention test generated — reported as `qraw=...` in the test's printed line, folded into
that case's overall pass/fail (not a separately counted case; §6).

**Coordinator (final micro-round):** read §1's own rewritten atomicity paragraph (the one
correcting the auditor's finding above) and noticed it proves too much against itself —
§1 stated the armed state ("allocator resolves, shader doesn't handle fmt=4") is reachable
by bisecting or cherry-picking a single commit, and this branch's THEN-current order
(allocator fix landing before the shader fix — the original `142ddd2`→`9315efa`
sequence) put exactly that armed state at the first of the two commits: green build,
15/15 `metal-test`, and armed, checked out alone. Directed a reorder to shader-first so
the hazard §1 itself describes cannot occur anywhere in this branch's history, not just
in the branch's final (squashed-view) state. This is the reorder reflected in the commit
list at the top of this section and in §1's current text.

**Verification of the reorder (RAN, this branch, this session):** each of the three
resulting commits was built and tested standing alone (`git reset`/re-edit to that
commit's exact tree, not merely inspected) before being committed:

| Commit (final SHA) | Content | `METAL=0`/`METAL=1` build | `make metal-test` |
|---|---|---|---|
| `f0fa5a9` (1st) | shader fmt=4 branch + host plumbing + original (blind-spotted) tests | clean, 0 warnings both | **27/27** — fmt=4 cases construct tensors directly (`posix_memalign`+`coli_metal_register`), never through `qt_from_disk`, so none depend on the allocator fix landing first; the new shader is exercised and proven correct here, just not yet reachable from a real loaded model |
| `2ba12d8` (2nd) | + `qt_from_disk` allocator activation | clean, 0 warnings both | **27/27** (unchanged from 1st — metal-test's cases never touched `qt_from_disk`, so this commit doesn't change metal-test's outcome, only real-model reachability) |
| `fc0f5a5` (3rd) | + attn-test blind-spot fix | clean, 0 warnings both | **27/27** (`qraw` metric now present; see the mutation matrix below) |

No test in the 12 new fmt=4 cases needed to become "self-allocating" or move commits —
they already self-allocate (they always have; `qt_from_disk` is a `colibri.c` model-load
function metal-test's binary doesn't even link against — `make metal-test`'s build recipe
compiles only `tests/test_backend_metal.mm` + `backend_metal.mm`, never `colibri.c`).
Tree identity: `git diff 5e3acd8 HEAD` — empty; `git rev-parse 5e3acd8^{tree}` and
`git rev-parse HEAD^{tree}` — identical (`95ff3e6f...`). The reorder is a pure permutation
of the same three diffs across commit boundaries, verified byte-for-byte, not a
re-derivation that could have silently changed something.

**Verification (RAN, this branch, this session):** re-derived the validator's three
mutation classes by temporarily editing `mm_gemv`'s fmt==4 branch in `backend_metal.mm`
(reverted after each run; final `backend_metal.mm` is byte-identical to the pre-mutation
state — confirmed with `diff`), and ran four configurations of a temporary diagnostic
harness (`run_attn_grouped_diag`, not committed) isolating fix (a) from fix (b):

| Mutation | OLD: S=1 pos=0, no qraw check | fix (a) only: S=1 pos=37, no qraw | fix (b) only: S=1 pos=0, qraw | **fix (a+b) — shipped**: S=1 pos=37, qraw |
|---|---|---|---|---|
| Off-by-one group index (`g = i/gs`, biased down by 1 when `>0`) | not caught | **caught** (nerr 4.6e-6→1.34) | **caught** (qraw 1.8e-8→9.4e-2) | **caught** |
| Swapped nibbles (`b&0xF`/`b>>4` swapped) | not caught | **caught** (nerr→2.38) | **caught** (qraw→9.6e-2) | **caught** |
| Uniform 1% scale error (`scale *= 1.01`) | not caught | **not caught** (nerr 5.78e-6, still passes) | **caught** (qraw 1.8e-8→6.4e-4) | **caught** |

Reading the table: the OLD column (validator's claim) caught nothing, confirming the
blind spot was real, not theoretical. Fix (a) alone closes the off-by-one and
swapped-nibble classes but — exactly as the RMSNorm-invariance argument predicts —
**still misses the uniform-scale class even at pos_base=37**, empirically confirming that
(a) and (b) are independent fixes and neither subsumes the other. Fix (b) alone (the
`qraw` check) catches all three classes regardless of `pos_base`, because it inspects
q_a's kernel output directly rather than through the RMSNorm/attention pipeline; fix (a)
remains independently worth keeping because it makes the FUSED-PATH comparison itself
(not just the auxiliary `qraw` probe) sensitive to q_a, which matters for defects in the
integration (e.g. how `qa_gs` is threaded through `bind_gemv`'s buffer indexing inside
`encode_attention`) rather than in the standalone GEMV kernel `qraw` isolates. The shipped
configuration (rightmost column) is what `run_attn_grouped()` now runs unconditionally —
3/3 mutation classes caught, `make metal-test` still exits 0 (27/27) with none of the
three mutations present. Also noted in passing: `run_grouped()`'s own shape-matrix cases
(§5), un-related to this attn-level fix, independently caught the uniform-scale mutation
too (`worst_rel` 7.1e-4 to 3.5e-3 across shapes, above the 1e-4 threshold) — the attn-level
blind spot was specifically about the FUSED attention output hiding the bug via RMSNorm,
not about fmt=4 GEMV correctness being unchecked elsewhere.

## 11. CUDA platform parity (#298 / #451) — added 2026-07-20

A structured comparison against the in-flight CUDA fmt=4 work (PR #298's design +
PR #451's expert-group kernels, both read at head on 2026-07-20) so the two platform
tiers can be reviewed as companions:

- **Format contract: fully aligned.** Same packed-nibble layout (`(I+1)/2` row
  bytes), same `[O, ceil(I/gs)]` f32 scale shape with `scale[o*ng+g]` indexing,
  same `g = i/gs` group math. The one apparent decode divergence (CUDA's
  upload-time XOR-0x88 to two's-complement vs our direct offset-binary `-8`) is
  mathematically equivalent — and deliberately not interchangeable: CUDA mutates a
  device-private copy, while Metal buffers can alias host memory the CPU fallback
  still reads, so an in-place transform is not an option on this backend.
- **Test oracle:** our cases use the magnitude-relative tolerance
  (`|Δ| / Σ|terms|`), matching dev's own `test_i4_grouped.c` and the review
  guidance given on #298 regarding cancellation in grouped dot products.
- **Scope split:** this PR covers the `mm_gemv` family (dense / attention /
  fused-decode). The batched routed-expert path (`moe_submit`/`moe_gemv`) remains
  gated to fmt {1,2} with CPU fallback — the Metal counterpart of #451's
  expert-group work is planned as a follow-up ("stage 3"), for which #451's
  per-tensor `GroupDesc` (per-expert fmt/gs) is the reference shape; Metal's
  current batch path assumes a single scalar fmt per block, which stage 3 must
  make explicit or generalize.
- **End-to-end gate:** all results in this PR are synthetic kernel-vs-oracle
  comparisons (§5–6). A CPU-vs-Metal token-stream identity check on a real
  grouped-int4 container (the determinism gate requested on #298) is planned as a
  follow-up comment once a g64 container is minted from the FP8 source.

---
**Coordinate stamp:** 2026-07-19/20 (session spanned midnight), macOS 26.5.2, Apple M5 Max
128 GB (Mac17,7) — the same machine/lab referenced in #447. All RAN evidence in this
document (§5, §6, §10) was captured on this branch, this session, this machine. No model
runs; no GitHub writes; no pushes.

### DIFF
--- a/c/backend_metal.h
+++ b/c/backend_metal.h
@@ -27,16 +27,19 @@ void coli_metal_stats(size_t *tensor_count, size_t *tensor_bytes);
 int  coli_metal_mem_info(size_t *used_bytes, size_t *total_bytes);
 
 /*
- * y[S,O] = (x[S,I] @ W[O,I]^T) * scale[o].
- * fmt matches QT in glm.c: 0=f32, 1=int8, 2=int4(packed), 3=int2(packed).
- * The first successful call wraps W and its row scales in GPU-visible buffers;
- * later calls reuse them (weights are assumed stable at the same address).
+ * y[S,O] = (x[S,I] @ W[O,I]^T) * scale[o]. fmt=4 (grouped int4) instead folds a
+ * PER-GROUP scale into the accumulation -- see the shader comment in backend_metal.mm.
+ * fmt matches QT in colibri.c: 0=f32, 1=int8, 2=int4(packed), 3=int2(packed),
+ * 4=int4(packed, same layout as 2)+per-group scale (gs = group size along I; scale
+ * array is [O, ceil(I/gs)] floats instead of [O]). gs is ignored for fmt!=4 (pass 0).
+ * The first successful call wraps W and its scales in GPU-visible buffers (sized from
+ * fmt+gs); later calls reuse them (weights are assumed stable at the same address).
  * Returns 1 on success, 0 if Metal is unavailable or fmt is invalid.
  */
 int coli_metal_matmul(ColiMetalTensor **tensor,
                       float *y, const float *x,
                       const void *weights, const float *scales,
-                      int fmt, int S, int I, int O);
+                      int fmt, int S, int I, int O, int gs);
 
 void   coli_metal_tensor_free(ColiMetalTensor *tensor);
 size_t coli_metal_tensor_bytes(const ColiMetalTensor *tensor);
@@ -68,14 +71,14 @@ void coli_metal_unregister(void *base);
  */
 int coli_metal_layer_decode(float *x,
     const float *in_ln, const float *post_ln,
-    const void *qa_w, const float *qa_s, int qa_fmt, const float *qa_ln,
-    const void *qb_w, const float *qb_s, int qb_fmt,
-    const void *kva_w, const float *kva_s, int kva_fmt, const float *kva_ln,
+    const void *qa_w, const float *qa_s, int qa_fmt, int qa_gs, const float *qa_ln,
+    const void *qb_w, const float *qb_s, int qb_fmt, int qb_gs,
+    const void *kva_w, const float *kva_s, int kva_fmt, int kva_gs, const float *kva_ln,
     const void *kvb_w, const float *kvb_s, int kvb_fmt,
-    const void *o_w, const float *o_s, int o_fmt,
-    const void *shg_w, const float *shg_s, int shg_fmt,
-    const void *shu_w, const float *shu_s, int shu_fmt,
-    const void *shd_w, const float *shd_s, int shd_fmt,
+    const void *o_w, const float *o_s, int o_fmt, int o_gs,
+    const void *shg_w, const float *shg_s, int shg_fmt, int shg_gs,
+    const void *shu_w, const float *shu_s, int shu_fmt, int shu_gs,
+    const void *shd_w, const float *shd_s, int shd_fmt, int shd_gs,
     const float *router_w, const float *router_bias,
     int E, int K, int Ksel, float topp, int normk, float rscale,
     float *Lc, float *Rc, int S, int pos_base, int st0,
@@ -83,7 +86,7 @@ int coli_metal_layer_decode(float *x,
     float *inrm_out, float *nrm_out, float *sh_out, int *idx_out, float *w_out, int *keff_out);
 
 int coli_metal_gemm(float *y, const float *x, const void *weights, const float *scales,
-                    int fmt, int S, int I, int O);   /* large-batch sync GEMM; 0 -> CPU */
+                    int fmt, int S, int I, int O, int gs);   /* large-batch sync GEMM; 0 -> CPU. gs: fmt=4 group size (0 otherwise) */
 /* Parallel top-8 expert selection (r_top8_par): run ONE top-8 selection kernel standalone
  * on host arrays — par=0 the serial r_top8, par=1 the parallel exact-match replica gated
  * in the engine by COLI_RTOP8 (default ON; COLI_RTOP8=0 opts out to the serial kernel).
@@ -104,11 +107,11 @@ int coli_metal_rtop8(int par, const float *sig, const float *bias, int S, int E,
 void coli_metal_attn_counts(uint64_t *ok, double *wall, double *kernel);
 void coli_metal_attn_lat(double *ksched, double *gsched);
 int coli_metal_attn_decode(const float *x,
-    const void *qa_w, const float *qa_s, int qa_fmt, const float *qa_ln,
-    const void *qb_w, const float *qb_s, int qb_fmt,
-    const void *kva_w, const float *kva_s, int kva_fmt, const float *kva_ln,
+    const void *qa_w, const float *qa_s, int qa_fmt, int qa_gs, const float *qa_ln,
+    const void *qb_w, const float *qb_s, int qb_fmt, int qb_gs,
+    const void *kva_w, const float *kva_s, int kva_fmt, int kva_gs, const float *kva_ln,
     const void *kvb_w, const float *kvb_s, int kvb_fmt,
-    const void *o_w, const float *o_s, int o_fmt,
+    const void *o_w, const float *o_s, int o_fmt, int o_gs,
     float *Lc, float *Rc, int S, int pos_base, int st0, float eps, float theta, float ascale, float *out);
 
 /* Diagnostics: GPU blocks executed, CPU-fallback blocks, experts run on GPU. */
@@ -129,7 +132,11 @@ int coli_metal_resset_stats(double *flush_s);
  *  D           = hidden size, Iinter = moe intermediate size
  *  g/u/d[e]    = pointers to expert e's gate/up/down quantized weights (in RAM slabs)
  *  gs/us/ds[e] = pointers to expert e's per-row scales
- *  fmt         = quant format (shared across experts)
+ *  fmt         = quant format (shared across experts). NOTE: fmt=4 (grouped int4) is
+ *                NOT yet supported here -- gates to {1,2} and returns 0 (CPU fallback)
+ *                for fmt=4 experts, same as before this stage. Grouped-int4 gained GPU
+ *                support in mm_gemv (coli_metal_matmul/coli_metal_gemm/bind_gemv) only;
+ *                extending the batched routed-expert path is future work (see PR_BODY.md).
  *  xg          = packed activations [total_rows, D]; xoff[e] = row offset of expert e
  *  nr[e]       = rows for expert e; rows[]/rw[] map packed rows back to out positions
  *  out         = [S, D] accumulate target
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -9,17 +9,22 @@
 
 // ---- shader: general quantized GEMV, one threadgroup per output element (o,si) ----
 // y[si,o] = (sum_i dequant(W[o,i]) * x[si,i]) * scale[o]. fmt: 0=f32 1=i8 2=i4 3=i2.
+// fmt=4 (grouped int4, dev e9b3614 matmul_i4_grouped, #298/#451 CUDA twin): same packed
+// nibble layout as fmt=2, but scale is PER-GROUP (gsz elements of I), buffer(1) holds
+// [O, ceil(I/gsz)] floats indexed scale[o*ng+g] -- so unlike fmt 1-3, the group scale is
+// folded into the accumulation itself (see the fmt==4 branch), not applied once at the end.
 static const char *SHADER = R"METAL(
 #include <metal_stdlib>
 using namespace metal;
 
 kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight bytes
-                    device const float* scale  [[buffer(1)]],   // [O]
+                    device const float* scale  [[buffer(1)]],   // [O] (fmt<4) or [O,ceil(I/gsz)] (fmt==4)
                     device const float* x      [[buffer(2)]],   // [S,I]
                     device float*       y      [[buffer(3)]],   // [S,O]
                     constant int& S [[buffer(4)]], constant int& I [[buffer(5)]],
                     constant int& O [[buffer(6)]], constant int& fmt [[buffer(7)]],
                     constant int& NT [[buffer(8)]],
+                    constant int& gsz [[buffer(9)]],             // fmt==4 group size (ignored otherwise)
                     uint tg [[threadgroup_position_in_grid]],
                     uint slane [[thread_index_in_simdgroup]],
                     uint sgid [[simdgroup_index_in_threadgroup]]) {
@@ -53,6 +58,23 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
     for (int i = slane; i < I; i += 32) {
       uchar b = wr[i>>2]; int v = (b >> (2*(i&3))) & 0x3; acc += float(v-2) * xr[i];
     }
+  } else if (fmt == 4) {                            // int4 GROUPED: same nibble packing as fmt=2,
+                                                     // one f32 scale per gsz-element group along I.
+                                                     // Each lane owns one packed byte (2 elements) per
+                                                     // 64-lane stride -> memory access stays coalesced,
+                                                     // and a group never splits a byte (gsz is even).
+    int rb = (I+1)/2; int ng = (I+gsz-1)/gsz;
+    device const uchar* wr = w + (long)o * rb;
+    device const float* scl = scale + (long)o * ng;
+    for (int i = slane*2; i < I; i += 64) {
+      uchar b = wr[i>>1];
+      int g0 = i / gsz; float sc0 = scl[g0];
+      acc += float(int(b&0xF)-8) * xr[i] * sc0;
+      if (i+1 < I) {
+        int g1 = (i+1) / gsz; float sc1 = (g1==g0) ? sc0 : scl[g1];
+        acc += float(int(b>>4)-8) * xr[i+1] * sc1;
+      }
+    }
   } else {                                          // f32
     device const float* wr = (device const float*)(w) + (long)o * I;
     device const float4* w4 = (device const float4*)wr;
@@ -60,7 +82,8 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
     for (int i = I8*8 + slane; i < I; i += 32) acc += wr[i] * xr[i];
   }
   acc = simd_sum(acc);
-  if (slane == 0) y[row] = acc * scale[o];
+  // fmt==4 already folded the per-group scale into acc above -- do not scale again.
+  if (slane == 0) y[row] = (fmt == 4) ? acc : acc * scale[o];
 }
 
 // Batched bindless expert GEMV: each row gr belongs to expert erow[gr], whose weight and
@@ -401,7 +424,13 @@ static size_t fmt_bytes(int fmt, int I, int O) {
   if (fmt == 1) return (size_t)O * I;
   if (fmt == 2) return (size_t)O * ((I+1)/2);
   if (fmt == 3) return (size_t)O * ((I+3)/4);
+  if (fmt == 4) return (size_t)O * ((I+1)/2);   // grouped int4: identical packed-nibble layout to fmt=2
   return (size_t)O * I * sizeof(float);
+}
+// Grouped-int4 (fmt=4) scale-array size: one f32 per gsz-element group, per row -> O*ceil(I/gsz).
+static size_t fmt_scale_bytes(int fmt, int I, int O, int gs) {
+  if (fmt == 4) return (size_t)O * ((I + gs - 1) / gs) * sizeof(float);
+  return (size_t)O * sizeof(float);
 }
 
 // Wrap host memory zero-copy if page-aligned, else copy into a shared buffer.
@@ -560,15 +589,15 @@ extern "C" int  coli_metal_mem_info(size_t *used, size_t *total) {
 
 extern "C" int coli_metal_matmul(ColiMetalTensor **tp, float *y, const float *x,
                                  const void *weights, const float *scales,
-                                 int fmt, int S, int I, int O) {
-  if (!g_dev || fmt < 0 || fmt > 3) return 0;
+                                 int fmt, int S, int I, int O, int gs) {
+  if (!g_dev || fmt < 0 || fmt > 4) return 0;
   @autoreleasepool {
     ColiMetalTensor *t = *tp;
     if (!t) {
       t = new ColiMetalTensor();
       t->fmt = fmt; t->I = I; t->O = O; t->wbytes = fmt_bytes(fmt, I, O);
       t->w = wrap(weights, t->wbytes);
-      t->s = wrap(scales, (size_t)O * sizeof(float));
+      t->s = wrap(scales, fmt_scale_bytes(fmt, I, O, gs));
       *tp = t;
       g_tensor_count++; g_tensor_bytes += t->wbytes;
     }
@@ -582,7 +611,7 @@ extern "C" int coli_metal_matmul(ColiMetalTensor **tp, float *y, const float *x,
     int NT=S*O;
     [e setBytes:&S length:4 atIndex:4]; [e setBytes:&I length:4 atIndex:5];
     [e setBytes:&O length:4 atIndex:6]; [e setBytes:&fmt length:4 atIndex:7];
-    [e setBytes:&NT length:4 atIndex:8];
+    [e setBytes:&NT length:4 atIndex:8]; [e setBytes:&gs length:4 atIndex:9];
     [e dispatchThreadgroups:MTLSizeMake(((size_t)NT+3)/4,1,1) threadsPerThreadgroup:MTLSizeMake(128,1,1)];
     [e endEncoding]; [cb commit]; [cb waitUntilCompleted];
     memcpy(y, [by contents], (size_t)S*O*sizeof(float));
@@ -605,7 +634,9 @@ static void attn_scratch_init(){
 }
 // y[S,O] = quantized-weight(w) applied to xin[S,I]. Weights are registered (page-aligned,
 // zero-copy) at model load; resolve to (buffer,offset). Returns false to fall back to CPU.
-static bool bind_gemv(id<MTLComputeCommandEncoder> e, const void* w, const float* s, int fmt, int I, int O,
+// gs: fmt=4 group size (ignored for fmt!=4; callers pass QT.gs, which is 0 for non-grouped
+// tensors -- harmless, since the shader only reads it when fmt==4).
+static bool bind_gemv(id<MTLComputeCommandEncoder> e, const void* w, const float* s, int fmt, int gs, int I, int O,
                       id<MTLBuffer> xin, id<MTLBuffer> yout, int S){
   uint64_t wa=0,sa=0; id<MTLBuffer> wb=resolve(w,&wa); id<MTLBuffer> sb=resolve(s,&sa);
   if(!wb||!sb) return false;
@@ -616,19 +647,22 @@ static bool bind_gemv(id<MTLComputeCommandEncoder> e, const void* w, const float
   [e setBuffer:xin offset:0 atIndex:2]; [e setBuffer:yout offset:0 atIndex:3];
   int NT=S*O;
   [e setBytes:&S length:4 atIndex:4]; [e setBytes:&I length:4 atIndex:5]; [e setBytes:&O length:4 atIndex:6]; [e setBytes:&fmt length:4 atIndex:7];
-  [e setBytes:&NT length:4 atIndex:8];
+  [e setBytes:&NT length:4 atIndex:8]; [e setBytes:&gs length:4 atIndex:9];
   [e dispatchThreadgroups:MTLSizeMake(((size_t)NT+3)/4,1,1) threadsPerThreadgroup:MTLSizeMake(128,1,1)];
   return true;
 }
 
 // Weight-pointer bundle for one layer's attention (+optional layer tail). All pointers
-// must be inside registered allocations.
+// must be inside registered allocations. *_gs: fmt=4 group size for the corresponding
+// weight (0 if that weight isn't grouped). kv_b has none: it never flows through bind_gemv
+// -- a_qabs/a_ctx dequantize it inline with a PER-ROW-only helper (a_deqrow), so a fmt=4
+// kv_b is out of scope for this stage (see PR_BODY.md UNCERTAINTIES).
 typedef struct {
-  const void *qa_w; const float *qa_s; int qa_fmt; const float *qa_ln;
-  const void *qb_w; const float *qb_s; int qb_fmt;
-  const void *kva_w; const float *kva_s; int kva_fmt; const float *kva_ln;
+  const void *qa_w; const float *qa_s; int qa_fmt; int qa_gs; const float *qa_ln;
+  const void *qb_w; const float *qb_s; int qb_fmt; int qb_gs;
+  const void *kva_w; const float *kva_s; int kva_fmt; int kva_gs; const float *kva_ln;
   const void *kvb_w; const float *kvb_s; int kvb_fmt;
-  const void *o_w;  const float *o_s;  int o_fmt;
+  const void *o_w;  const float *o_s;  int o_fmt; int o_gs;
 } AttnW;
 
 // Encode the fused attention chain into encoder e. Input: ax_ holds the NORMED x [S,AH].
@@ -651,10 +685,10 @@ static bool encode_attention(id<MTLComputeCommandEncoder> e, const AttnW *W,
       [e setBuffer:acomp_ offset:0 atIndex:0]; [e setBytes:&off length:4 atIndex:1]; [e setBytes:&ss length:4 atIndex:2];
       [e setBuffer:dst offset:doff atIndex:3]; [e setBytes:&n length:4 atIndex:4]; [e setBytes:&n length:4 atIndex:5];
       [e dispatchThreads:MTLSizeMake((size_t)S*n,1,1) threadsPerThreadgroup:MTLSizeMake(64,1,1)]; };
-    bind_gemv(e,W->qa_w,W->qa_s,W->qa_fmt,AH,AQLORA,ax_,aqr_,S);
-    bind_gemv(e,W->kva_w,W->kva_s,W->kva_fmt,AH,AKVL+AROPE,ax_,acomp_,S); BAR();
+    bind_gemv(e,W->qa_w,W->qa_s,W->qa_fmt,W->qa_gs,AH,AQLORA,ax_,aqr_,S);
+    bind_gemv(e,W->kva_w,W->kva_s,W->kva_fmt,W->kva_gs,AH,AKVL+AROPE,ax_,acomp_,S); BAR();
     rms(aqr_,0,aqaln_,AQLORA,S); cpy(0,Lb,Loff,AKVL); cpy(AKVL,Rb,Roff,AROPE); BAR();
-    bind_gemv(e,W->qb_w,W->qb_s,W->qb_fmt,AQLORA,AHQH,aqr_,aqf_,S); rms(Lb,Loff,akvaln_,AKVL,S); rope(Rb,Roff,0,AROPE,0,1); BAR();
+    bind_gemv(e,W->qb_w,W->qb_s,W->qb_fmt,W->qb_gs,AQLORA,AHQH,aqr_,aqf_,S); rms(Lb,Loff,akvaln_,AKVL,S); rope(Rb,Roff,0,AROPE,0,1); BAR();
     rope(aqf_,0,ANOPE,AHQH,AQH,AHEADS); BAR();
     [e setComputePipelineState:g_a_qabs]; [e setBuffer:kvbW offset:kvbwoff atIndex:0]; [e setBuffer:kvbS offset:kvbsoff atIndex:1]; [e setBuffer:aqf_ offset:0 atIndex:2]; [e setBuffer:aqabs_ offset:0 atIndex:3];
     [e dispatchThreads:MTLSizeMake((size_t)S*AHEADS*AKVL,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
@@ -667,7 +701,7 @@ static bool encode_attention(id<MTLComputeCommandEncoder> e, const AttnW *W,
     [e dispatchThreads:MTLSizeMake((size_t)S*AHEADS*AKVL,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
     [e setComputePipelineState:g_a_ctx]; [e setBuffer:kvbW offset:kvbwoff atIndex:0]; [e setBuffer:kvbS offset:kvbsoff atIndex:1]; [e setBuffer:aclat_ offset:0 atIndex:2]; [e setBuffer:actx_ offset:0 atIndex:3];
     [e dispatchThreads:MTLSizeMake((size_t)S*AHEADS*AVH,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
-    bind_gemv(e,W->o_w,W->o_s,W->o_fmt,AHVH,AH,actx_,aout_,S);
+    bind_gemv(e,W->o_w,W->o_s,W->o_fmt,W->o_gs,AHVH,AH,actx_,aout_,S);
     return true;
 }
 // Resolve Lc/Rc + kv_b (+pre-check the projection weights). Returns false -> CPU fallback.
@@ -685,18 +719,18 @@ static bool resolve_attn(const AttnW *W, float *Lc, float *Rc,
 }
 
 extern "C" int coli_metal_attn_decode(const float* x,
-    const void* qa_w,const float* qa_s,int qa_fmt,const float* qa_ln,
-    const void* qb_w,const float* qb_s,int qb_fmt,
-    const void* kva_w,const float* kva_s,int kva_fmt,const float* kva_ln,
+    const void* qa_w,const float* qa_s,int qa_fmt,int qa_gs,const float* qa_ln,
+    const void* qb_w,const float* qb_s,int qb_fmt,int qb_gs,
+    const void* kva_w,const float* kva_s,int kva_fmt,int kva_gs,const float* kva_ln,
     const void* kvb_w,const float* kvb_s,int kvb_fmt,
-    const void* o_w,const float* o_s,int o_fmt,
+    const void* o_w,const float* o_s,int o_fmt,int o_gs,
     float* Lc,float* Rc,int S,int pos_base,int st0,float eps,float theta,float ascale,float* out){
   if(!g_dev) return 0;
   if(st0!=0 || S<1 || S>AMAXS) return 0;     // partial-KV / S>4 -> CPU
   int T=pos_base+S;
   @autoreleasepool {
     attn_scratch_init();
-    AttnW W={qa_w,qa_s,qa_fmt,qa_ln,qb_w,qb_s,qb_fmt,kva_w,kva_s,kva_fmt,kva_ln,kvb_w,kvb_s,kvb_fmt,o_w,o_s,o_fmt};
+    AttnW W={qa_w,qa_s,qa_fmt,qa_gs,qa_ln,qb_w,qb_s,qb_fmt,qb_gs,kva_w,kva_s,kva_fmt,kva_gs,kva_ln,kvb_w,kvb_s,kvb_fmt,o_w,o_s,o_fmt,o_gs};
     id<MTLBuffer> Lb,Rb,kvbW,kvbS; size_t loff,roff,kvbwoff,kvbsoff;
     if(!resolve_attn(&W,Lc,Rc,&Lb,&loff,&Rb,&roff,&kvbW,&kvbwoff,&kvbS,&kvbsoff)) return 0;
     ascore_=ensure(ascore_,&ascore_cap,(size_t)S*AHEADS*T*4);
@@ -723,14 +757,14 @@ extern "C" int coli_metal_attn_decode(const float* x,
 // idx/w/keff (routing). Returns 0 -> CPU fallback (whole layer falls back).
 extern "C" int coli_metal_layer_decode(float *x,
     const float *in_ln, const float *post_ln,
-    const void* qa_w,const float* qa_s,int qa_fmt,const float* qa_ln,
-    const void* qb_w,const float* qb_s,int qb_fmt,
-    const void* kva_w,const float* kva_s,int kva_fmt,const float* kva_ln,
+    const void* qa_w,const float* qa_s,int qa_fmt,int qa_gs,const float* qa_ln,
+    const void* qb_w,const float* qb_s,int qb_fmt,int qb_gs,
+    const void* kva_w,const float* kva_s,int kva_fmt,int kva_gs,const float* kva_ln,
     const void* kvb_w,const float* kvb_s,int kvb_fmt,
-    const void* o_w,const float* o_s,int o_fmt,
-    const void* shg_w,const float* shg_s,int shg_fmt,
-    const void* shu_w,const float* shu_s,int shu_fmt,
-    const void* shd_w,const float* shd_s,int shd_fmt,
+    const void* o_w,const float* o_s,int o_fmt,int o_gs,
+    const void* shg_w,const float* shg_s,int shg_fmt,int shg_gs,
+    const void* shu_w,const float* shu_s,int shu_fmt,int shu_gs,
+    const void* shd_w,const float* shd_s,int shd_fmt,int shd_gs,
     const float *router_w, const float *router_bias,
     int E, int K, int Ksel, float topp, int normk, float rscale,
     float *Lc, float *Rc, int S, int pos_base, int st0,
@@ -741,7 +775,7 @@ extern "C" int coli_metal_layer_decode(float *x,
   int T=pos_base+S; const int SI=2048;
   @autoreleasepool {
     attn_scratch_init();
-    AttnW W={qa_w,qa_s,qa_fmt,qa_ln,qb_w,qb_s,qb_fmt,kva_w,kva_s,kva_fmt,kva_ln,kvb_w,kvb_s,kvb_fmt,o_w,o_s,o_fmt};
+    AttnW W={qa_w,qa_s,qa_fmt,qa_gs,qa_ln,qb_w,qb_s,qb_fmt,qb_gs,kva_w,kva_s,kva_fmt,kva_gs,kva_ln,kvb_w,kvb_s,kvb_fmt,o_w,o_s,o_fmt,o_gs};
     id<MTLBuffer> Lb,Rb,kvbW,kvbS; size_t loff,roff,kvbwoff,kvbsoff;
     if(!resolve_attn(&W,Lc,Rc,&Lb,&loff,&Rb,&roff,&kvbW,&kvbwoff,&kvbS,&kvbsoff)) return 0;
     uint64_t ina=0,pna=0,rwa=0,rba=0,d;
@@ -778,8 +812,8 @@ extern "C" int coli_metal_layer_decode(float *x,
     [e dispatchThreads:MTLSizeMake((size_t)S*AH,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
     copyrow(axr_,anrm_,AH); BAR(); rmsw(anrm_,pnB,pnoff,AH,S); BAR();
     // 4) shared expert gate/up + router (all read anrm_, independent)
-    bind_gemv(e,shg_w,shg_s,shg_fmt,AH,SI,anrm_,ash1_,S);
-    bind_gemv(e,shu_w,shu_s,shu_fmt,AH,SI,anrm_,ash2_,S);
+    bind_gemv(e,shg_w,shg_s,shg_fmt,shg_gs,AH,SI,anrm_,ash1_,S);
+    bind_gemv(e,shu_w,shu_s,shu_fmt,shu_gs,AH,SI,anrm_,ash2_,S);
     { int NT=S*E, D=AH; [e setComputePipelineState:g_r_router];
       [e setBuffer:rwB offset:rwoff atIndex:0]; [e setBuffer:anrm_ offset:0 atIndex:1]; [e setBuffer:asig_ offset:0 atIndex:2];
       [e setBytes:&E length:4 atIndex:3]; [e setBytes:&D length:4 atIndex:4]; [e setBytes:&NT length:4 atIndex:5];
@@ -806,7 +840,7 @@ extern "C" int coli_metal_layer_decode(float *x,
       else        [e dispatchThreads:MTLSizeMake(S,1,1) threadsPerThreadgroup:MTLSizeMake(S,1,1)]; }
     BAR();
     // 6) shared down
-    bind_gemv(e,shd_w,shd_s,shd_fmt,SI,AH,ash1_,ashout_,S);
+    bind_gemv(e,shd_w,shd_s,shd_fmt,shd_gs,SI,AH,ash1_,ashout_,S);
     double tc=mnow();
     [e endEncoding]; [cb commit]; [cb waitUntilCompleted];
     if(cb.status==MTLCommandBufferStatusError){ fprintf(stderr,"[metal] layer cmdbuf error: %s\n", cb.error?[[cb.error localizedDescription]UTF8String]:"?"); return 0; }
@@ -827,8 +861,8 @@ extern "C" int coli_metal_layer_decode(float *x,
 // registered (zero-copy); x/y go through grow-only shared scratch. Returns 0 -> CPU fallback.
 static id<MTLBuffer> g_gx, g_gy; static size_t g_gx_cap, g_gy_cap;
 extern "C" int coli_metal_gemm(float *y, const float *x, const void *wp, const float *sp,
-                               int fmt, int S, int I, int O) {
-  if (!g_dev || (fmt!=1 && fmt!=2)) return 0;
+                               int fmt, int S, int I, int O, int gs) {
+  if (!g_dev || (fmt!=1 && fmt!=2 && fmt!=4)) return 0;
   @autoreleasepool {
     uint64_t wa=0,sa=0; id<MTLBuffer> wb=resolve(wp,&wa), sb=resolve(sp,&sa);
     if(!wb||!sb) return 0;
@@ -841,6 +875,7 @@ extern "C" int coli_metal_gemm(float *y, const float *x, const void *wp, const f
     [e setBuffer:wb offset:woff atIndex:0]; [e setBuffer:sb offset:soff atIndex:1];
     [e setBytes:&I length:4 atIndex:5];
     [e setBytes:&O length:4 atIndex:6]; [e setBytes:&fmt length:4 atIndex:7];
+    [e setBytes:&gs length:4 atIndex:9];   /* loop-invariant, like I/O/fmt: group size for fmt=4 (0 = per-row) */
     // Grid-size cap. One dispatch of NT=S*O output elements launches (NT+3)/4 threadgroups; past
     // a device grid limit (observed on M-series: kv_b grid ~3.1e7 tg clean at S=4376, ~5.4e7 tg
     // CORRUPT at S=7478) rows beyond the limit are silently never computed and the output keeps
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -547,9 +547,9 @@ static void matmul_i4_grouped_pair(float *yg, float *yu, const float *x,
  * (~+12% perplexity), measured. Every other prefill matmul keeps IDOT as before. */
 static void matmul_qt_ex(float *y, const float *x, QT *w, int S, int allow_idot){
 #ifdef COLI_METAL
-    if(g_metal_enabled && S>=g_metal_gemm_min && !spec_pinned() && (w->fmt==1||w->fmt==2) && !omp_in_parallel()){
+    if(g_metal_enabled && S>=g_metal_gemm_min && !spec_pinned() && (w->fmt==1||w->fmt==2||w->fmt==4) && !omp_in_parallel()){
         const void *wp = w->fmt==1 ? (const void*)w->q8 : (const void*)w->q4;
-        if(coli_metal_gemm(y,x,wp,w->s,w->fmt,S,w->I,w->O)) return;
+        if(coli_metal_gemm(y,x,wp,w->s,w->fmt,S,w->I,w->O,w->gs)) return;
     }
 #endif
 #ifdef COLI_CUDA
@@ -2374,11 +2374,11 @@ static void attention_rows(Model *m, Layer *l, int layer, float *x, int S, int p
             }
             #define WP_(q) ((q).fmt==1?(const void*)(q).q8:(const void*)(q).q4)
             int ok = coli_metal_attn_decode(x,
-                WP_(l->q_a), l->q_a.s, l->q_a.fmt, l->q_a_ln,
-                WP_(l->q_b), l->q_b.s, l->q_b.fmt,
-                WP_(l->kv_a), l->kv_a.s, l->kv_a.fmt, l->kv_a_ln,
+                WP_(l->q_a), l->q_a.s, l->q_a.fmt, l->q_a.gs, l->q_a_ln,
+                WP_(l->q_b), l->q_b.s, l->q_b.fmt, l->q_b.gs,
+                WP_(l->kv_a), l->kv_a.s, l->kv_a.fmt, l->kv_a.gs, l->kv_a_ln,
                 WP_(l->kv_b), l->kv_b.s, l->kv_b.fmt,
-                WP_(l->o), l->o.s, l->o.fmt,
+                WP_(l->o), l->o.s, l->o.fmt, l->o.gs,
                 m->Lc[layer], m->Rc[layer], S, pos_base, m->kv_start[layer], c->eps, c->theta, c->attn_scale, out);
             #undef WP_
             if(ok){ m->t_attn += now_s()-ta0; return; }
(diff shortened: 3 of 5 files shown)


JustVugg:
Rebase needed — and one substantive note: #298 just merged into `dev`, fixing the CUDA dense/attention kernels to apply fmt=4 scales per-group (`absorb_scale`, `gs`/`ng` params). Worth checking your two latent fmt=4 fixes against it: one of them may be the Metal twin of the same bug (per-row indexing of grouped scales), in which case your GEMV can mirror #298's approach 1:1. The conflict is semantic in the good way — your call, but you clearly know this area.

monotophic:

Rebase: done — Current head `2277004`
sits on `dev` @ `68ac9ff` (post-#298/#192/#421), mergeable, all three commits re-verified on
the new base (43/43 metal-test per commit, zero warning delta, incl. a `COLI_METAL_RESSET=1`
run now that #426 is in the base).

On the substantive question — we checked against `absorb_scale` at source, and the picture is:

**The GEMV scale math already mirrors #298 exactly.** `absorb_scale`:
`wscale[row*ng + k/gs]`, `ng = ceil(K/gs)`. Our `mm_gemv` fmt=4 branch:
`scale[o*ng + i/gsz]`, same `ng`. Only difference is your defensive tail clamp
(`if (g >= ng) g = ng-1`) vs our bound-by-construction (`i < I ⟹ i/gsz ≤ ng-1`, ceiling-division
identity — the review round proved it algebraically and the partial-tail test case exercises it).

**Our two latent fixes are a different defect class than the one #298 fixed, though.** They're
(1) the fmt=4 scale buffer being allocated outside the GPU-visible registration path
(falloc→qalloc — Metal zero-copy machinery, no CUDA analog), and (2) the dispatch falling to the
f32 arm. Neither is scale *indexing*.

**The true Metal twin of your per-row-indexing bug is the attention-absorption path**
(`a_deqrow`/`a_qabs`/`a_ctx`) — and this PR deliberately doesn't extend it: the pre-existing
`l->kv_b.fmt==2` gate keeps grouped kv_b off the Metal absorb fast path entirely, so the bug is
unreachable rather than fixed on our side (disclosed in the PR body's scope notes). When Metal
absorption gets extended to grouped kv_b, `absorb_scale` is exactly the 1:1 blueprint we'd
mirror — happy to take that as a follow-up once this lands, since #298 has now settled the
reference semantics.

JustVugg:

For the grouped-int4 (fmt=4) Metal GEMV: since fmt=4 on GPU is exactly where #298 found the per-group-scale bug, please include a fmt=4 container run showing coherent output on Metal (a gs64 container), token-exact vs CPU if you can. Your two latent fmt=4 fixes are worth landing on that evidence. Thanks @monotophic.

JustVugg:
Hi @monotophic — status on this one, so it isn't sitting silently.

I checked it on current `dev`: **merges clean, no rebase needed**, and I built and ran it here. On the CPU side it's a no-op, which is what I wanted to confirm — token-exactness against the oracle is unchanged (`fp 32/32`, `int4 21/32`, identical to dev baseline). So there's no risk to the non-Metal path.

**What I can't do is validate the Metal path** — I'm on Linux, `backend_metal.mm` never compiles here. That's the last thing standing between this PR and merge, and it's the part that matters, since fmt=4 (grouped-int4) is what we're moving to as the default container.

Could you run the test you included, on your Mac, and paste the output?

```
make metal-test
```

Ideally fail-before / pass-after, so the two latent fmt=4 fixes are visibly doing something. If you can also do a short generation against a real fmt=4 model with `COLI_BACKEND=metal` vs CPU and confirm the tokens agree, that's the strongest evidence and I'll merge on it.

Related: #488 fixes a large-prefill grid-size truncation in the same file (`coli_metal_gemm`). The two don't conflict, but whichever lands second gets a trivial rebase — which we'll do here, not you.

monotophic:
Requested verification is in queue, will be available tomorrow AM. Machine is busy and hot :)

monotophic:

Metal validation data, as requested. All runs on M5 Max 128 GB, macOS 26.5, head
`2277004` (which you verified merges clean on dev), real containers.

## 1. `make metal-test` — full pass, both residency-gate states

```
make metal-test                      → metal backend tests: ok   (43 cases)
COLI_METAL_RESSET=1 make metal-test  → metal backend tests: ok   (43 cases, identical)
```

<details><summary>Full case-by-case output (RESSET off; the RESSET=1 run is line-identical)</summary>

```
clang++ -x objective-c++ -std=gnu++17 -fobjc-arc -O3 tests/test_backend_metal.mm backend_metal.mm -framework Metal -framework Foundation -o backend_metal_test
./backend_metal_test
Metal backend kernel tests:
  int8 gate/up S=1       nerr=2.36e-06  ok
  int4 gate/up S=1       nerr=1.98e-06  ok
  int4 down S=1          nerr=1.03e-06  ok
  int2 gate/up S=1       nerr=1.72e-06  ok
  f32  S=1               nerr=1.77e-06  ok
  int8 gate/up S=4       nerr=2.36e-06  ok
  int4 gate/up S=7 (odd) nerr=1.98e-06  ok
  int4 non-mult-4 dims   nerr=3.78e-06  ok
Metal fmt=4 grouped-int4 tests (coli_metal_matmul vs matmul_i4_grouped semantics):
  grouped gate/up I=6144(mult64) S=1 worst_rel=1.90e-08 (I=6144 gs=64 ng=96 S=1)  ok
  grouped down I=2048(mult64) S=1    worst_rel=2.09e-08 (I=2048 gs=64 ng=32 S=1)  ok
  grouped gate/up I=6144(mult64) S=4 worst_rel=1.83e-08 (I=6144 gs=64 ng=96 S=4)  ok
  grouped I=200 (non-mult-64) S=2    worst_rel=1.29e-08 (I=200 gs=64 ng=4 S=2)  ok
  grouped I=201 (odd, non-mult-64) S=3 worst_rel=2.40e-08 (I=201 gs=64 ng=4 S=3)  ok
  grouped I=64 (degenerate, ng=1) S=1 worst_rel=1.15e-08 (I=64 gs=64 ng=1 S=1)  ok
  grouped I=40 (<gs, single partial group) S=1 worst_rel=2.53e-08 (I=40 gs=64 ng=1 S=1)  ok
  grouped I=512(mult64) outlier-heavy S=3 worst_rel=5.89e-08 (I=512 gs=64 ng=8 S=3)  ok
  grouped I=201(non-mult-64) outlier-heavy S=2 worst_rel=4.91e-08 (I=201 gs=64 ng=4 S=2)  ok
Metal batched moe_block tests:
  moe decode nb=8        R=8 nerr=2.43e-06  ok
  moe ragged nb=6        R=16 nerr=2.49e-06  ok
Metal large-batch gemm test:
  gemm S=64 int4          nerr=2.85e-06  ok
  gemm S=64 int4-grouped   worst_rel=2.42e-08  ok
Metal fused attention tests:
  attn S=1 pos=0           nerr=4.04e-06 cache=1.38e-05  ok
  attn S=1 pos=37          nerr=4.13e-06 cache=1.48e-05  ok
  attn S=4 pos=12 (MTP)    nerr=5.37e-06 cache=1.34e-05  ok
  attn S=3 pos=0           nerr=5.30e-06 cache=1.31e-05  ok
Metal fused attention tests (fmt=4 grouped q_a, proves bind_gemv gs plumbing):
  attn grouped-qa S=1 pos=37 nerr=5.53e-06 cache=7.87e-06 qraw=1.94e-08  ok
  attn grouped-qa S=4 pos=12 (MTP) nerr=5.00e-06 cache=1.38e-05 qraw=2.02e-08  ok
Metal negative control: fmt 1/2/3 unaffected by the fmt=4 shader branch --
  see the runs above (int8/int4/int2/f32/moe/gemm/attn cases): all still ok.
Metal top-8 select serial-vs-parallel tests (exact-match contract, E=256):
  top8 generic S=1                   ok (serial==parallel bitwise, S=1 E=256)
  top8 generic S=4                   ok (serial==parallel bitwise, S=4 E=256)
  top8 ALL-EQUAL ties                ok (serial==parallel bitwise, S=1 E=256)
  top8 massed dup ties S=4           ok (serial==parallel bitwise, S=4 E=256)
  top8 3-level ties rscale           ok (serial==parallel bitwise, S=2 E=256)
  top8 denormal logits               ok (serial==parallel bitwise, S=1 E=256)
  top8 topp=0.01 (Ke=1 edge)         ok (serial==parallel bitwise, S=1 E=256)
  top8 topp=0.6 tied weights         ok (serial==parallel bitwise, S=1 E=256)
  top8 topp=0.999 S=4                ok (serial==parallel bitwise, S=4 E=256)
  top8 topp on ALL-EQUAL             ok (serial==parallel bitwise, S=2 E=256)
Metal top-8 select expert-count-generality tests (E!=256, REAP/#428 motivated):
  top8 E=168 (REAP) generic S=1      ok (serial==parallel bitwise, S=1 E=168)
  top8 E=168 (REAP) massed dup ties S=4 ok (serial==parallel bitwise, S=4 E=168)
  top8 E=24 (<32 lane width) generic ok (serial==parallel bitwise, S=1 E=24)
  top8 E=24 (<32 lane width) ALL-EQUAL ties ok (serial==parallel bitwise, S=1 E=24)
  top8 E=200 (lane straddles E boundary) ok (serial==parallel bitwise, S=4 E=200)
  top8 E=257 (>256, auto-serial-fallback) ok (serial==parallel bitwise, S=1 E=257)
metal backend tests: ok
```
</details>

## 2. Real fmt=4 container generation on Metal — coherent, with a note on token-exactness

Container: a real gs=64 grouped-int4 GLM-5.2 repack (int8 residents + fmt=4-g64
routed experts). Branch binary, `COLI_METAL=1` vs `COLI_METAL=0`, temp 0, same prompt:

- **Both transcripts are fluent and on-topic.** Excerpts (visible decode window; full
  raw logs available on request):

  fmt=4 container, `COLI_METAL=1`:
  > …Christian and Greek mythology. Both figures are archetypal rebels who defy the
  > supreme **authority of the universe**, both are closely associated with the concept
  > of "light," and both suffer catastrophic, eternal punishments…

  fmt=4 container, `COLI_METAL=0`:
  > …Christian and Greek mythology. Both figures are archetypal rebels who defy the
  > supreme **deity of their respective cosmologies**, and both are severely punished
  > for bringing something "forbidden" to humanity…

  fmt=2 CONTROL container (unmodified format), same A/B — diverges from the first
  visible words (Metal: "…share one of the most profound and enduring archetypal
  narratives…" / CPU: "…represent two of the most profound and complex archetypes…"),
  i.e. the control forks at least as hard as fmt=4.
- They agree for the opening span and then diverge textually. **Before attributing
  that to this PR: we ran the identical A/B on an unmodified fmt=2 (per-row int4)
  container as a control, and it diverges the same way.** Cross-backend temp-0
  token-exactness does not currently hold on this engine for ANY container — CPU and
  Metal accumulate in different orders and the argmax forks at some position — which
  our earlier acceptance battery established across five prompt families (baseline
  control included). What DOES hold, and what we'd suggest as the practical merge
  bar for backends: within-backend determinism (repeated Metal runs are
  byte-identical at temp 0 — verified across every battery this month) plus
  coherence, which the transcripts show. Happy to discuss a token-exactness gate as
  a separate engine goal — it would need cross-backend-stable accumulation order,
  which is bigger than this PR.

## 3. Before/after tok/s (same prompt, interleaved, fmt=4 container)

| arm | binary | tok/s |
|---|---|---|
| dev @ 008ecc3 | stock | 0.89 / 0.89 / 0.85 |
| this PR | branch | 0.87 / 0.87 |

**Parity, no regression.** Expected at this stage: `mm_gemv` serves single-row
(resident-tensor) matvecs, and this container's residents are int8 — the routed
fmt=4 experts still take the CPU batch path, which is the stage-3 follow-up in the
PR body's §11 roadmap. The kernel-level fmt=4 speedups are in the metal-test benches;
the decode-level win arrives with the MoE batch path.

## 4. The latent fixes, visibly doing something

The synthetic suite can't reach the `qt_from_disk` allocator defect (tests inject
buffers via `qalloc` directly — a blind spot worth noting), so we demonstrated it at
runtime: reverting only the falloc→qalloc fix (`0c0d9c8`) on an otherwise-identical
build and re-running the same 16-token fmt=4 decode drops the engine's own
Metal-eligibility telemetry from **3,982 GPU expert-blocks to 996** — the scale
buffers silently fail Metal registration and ~4× of the eligible work falls off the
GPU path, exactly the failure the fix closes. (The second fix — the `mm_gemv` fmt=4
dispatch — has no meaningful "before": it lands with the kernel itself.)

Raw logs, full transcripts, and the revert-demo outputs available if needed. Let me know if you have questions or if there is anything else I can do.

JustVugg:
Reviewed properly. The fmt=4 work is correct and the two latent fixes are real — but there's **one concrete blocker** that only appeared an hour ago, and it's my fault for the timing.

## Blocker: this no longer compiles against `dev`

I merged #488 tonight, which added `c/tests/test_gemm_largebatch.mm`. That test calls `coli_metal_gemm` with **8 arguments**:

```c
int ok1 = coli_metal_gemm(yg.data(),  x.data(), W, Sc, fmt, S, I, O);
int ok2 = coli_metal_gemm(yg2.data(), x.data(), W, Sc, fmt, S, I, O);
```

This PR changes the prototype to **9** (adding `gs`) and updates only `test_backend_metal.mm`. After a rebase, `make gemm-test` won't build. One-line fix: add `, 0` to both calls.

Worth knowing **why CI can't catch this**: `METAL ?= 0`, the macOS job runs `make check` = clean + portable + test, and `test` = `test-c` + `test-python`. `metal-test` and `gemm-test` are in neither. **No CI job compiles `backend_metal.mm` at all.** So 8/8 green says nothing about this PR, and the only evidence is your 27/27 on the M5 Max — which I do trust, but you should know the safety net isn't there. If you have an appetite for it afterwards, a macOS CI job that at least *compiles* the Metal backend would stop this class of breakage permanently.

## What I verified

The shader change is additive: a new `else if (fmt == 4)` arm ahead of the f32 `else`, plus the `fmt == 4 ? acc : acc * scale[o]` epilogue. **For fmt 0/1/2/3 both the decode arm and the epilogue are bit-identical to before** — that was my main worry and it's clean. I checked the fmt=4 arm against `matmul_i4_grouped`: same nibble convention (low = even `i`), same group index `i/gs`, scale folded per-term. The lane stride keeps `i` even, so `w[i>>1]` always names the byte holding `i`/`i+1`. Correct.

Both latent fixes check out, and your §1 is honest about them cancelling each other: `bind_gemv` had no fmt gate while the fused-attention gates only checked `kv_b.fmt==2` (so a fmt=4 `q_a`/`q_b`/`kv_a`/`o` would have been decoded as raw f32), and the fmt=4 branch in `qt_from_disk` allocated group scales with `falloc` instead of `qalloc`, so under Metal they were never registered and `resolve()` always fell back. Shader-first, allocator-second is the right commit order.

## One small thing, not a blocker

The `falloc` → `qalloc` change at `c/colibri.c:974` is **outside** `#ifdef COLI_METAL`. On a non-Metal build that trades `falloc`'s overflow/OOM guard (which exits cleanly) for `qalloc`'s unchecked `malloc`. It's the same hole `qsalloc` already has for fmt 1/2/3, so it's consistency rather than a new bug class — but it is a small robustness regression on the CPU path, and worth a line of comment acknowledging it. (Note fmt=5's group scales still use `falloc`, so int3-g64 stays Metal-inert — not a regression, just inconsistent after this.)

Fix the two-argument call, rebase, and I'll merge it.

monotophic:

Fixed and rebased.

- **The #488 break:** `, 0` added to both `coli_metal_gemm` calls in
  `c/tests/test_gemm_largebatch.mm` — `gs=0` selects per-row scales, the
  pre-#457 semantics, so #488's fixtures test exactly what they tested
  before. Rebased onto `dev` @ `f1b0f7e`, head `4d7d3f2` (4 commits: the
  3 you reviewed + one review-fix commit). Range-diff vs the reviewed
  head: commits 2–3 patch-identical; commit 1 differs only where it must —
  your #488 chunk loop is kept verbatim and the `gs` bind is hoisted
  loop-invariant beside I/O/fmt (it doesn't vary per chunk).
- **The `falloc`→`qalloc` note:** acknowledged with a comment at the site
  (c/colibri.c, fmt=4 group-scale alloc): it trades `falloc`'s checked-exit
  for `qalloc`'s unchecked malloc on the CPU path, consistent with what
  `qsalloc` already does for fmt 1/2/3. Also noted there: fmt=5's group
  scales still use `falloc`, so int3-g64 remains Metal-inert — pre-existing,
  now documented rather than silent.
- **Gates on the rebased head** (M5 Max, macOS 26.5): `make check` (the
  CI-equivalent set) passes — all C test binaries + 142 Python tests (10
  platform skips); Metal build zero warnings; `metal-test` ok under both
  `COLI_METAL_RESSET` states; `gemm-test` **all clean, including the
  S=7478 kv_b case** — the original corruption trigger now runs through
  the 9-arg prototype and your chunk loop together, so the merged region
  is exercised by your own regression test. Bisect-safety: each of the 4
  commits builds METAL=1 with zero warnings and passes `metal-test`
  individually.

**Sequencing across our four open PRs** (your ask on #386), so your merge
process never discovers a dependency at merge time:

1. **This one first** — it's one line from your stated merge bar, and it
   settles `backend_metal.mm` for everything behind it.
2. **#386 second** — one r6 rebase then absorbs #525/#526/#532 *and* this
   PR in a single round instead of two.
3. **#528 third**, revised per your review — collision policy inverted
   (reader resolves ambiguity to fmt=1; the writer refuses to emit it).
4. **#529 last, staying stacked on #528** — after the inversion, the stamp
   is the only forward path by which an ambiguous-shape tensor can ever be
   read as fmt=7, so the two PRs define one collision policy between them
   and are best reviewed in order.

**On your #457 × #386 interaction question: no interaction — either order
is safe.** #386's cap-1 default is expert-cache *slot* economics (the #379
residency mechanism) and is format-independent: fmt=4 support changes which
tensors the GEMV can decode, not the cost of caching them. The cap
default's revisit trigger remains the residency-set default flip, exactly
as #386's durable-vs-current-state map already states. We put this PR first
purely for merge-bar reasons, not correctness ones.

On the Metal CI gap you flagged: yes. After this queue clears, we'd like to
contribute the macOS workflow job that at least compiles `backend_metal.mm`
and runs whatever subset of `metal-test`/`gemm-test` the runner's GPU
permits — once merged it runs on GitHub's own macOS runners, so it covers
every contributor's PRs without depending on anyone's hardware. Separately,
my M5 Max can serve as a hands-on Metal bench at your request for changes that
need real-hardware numbers, as it did here; that part is tied to my current
level of activity in the project, and I'll say so on a thread if it changes
rather than leave anyone assuming coverage that isn't there.

JustVugg:
Status: the fmt=4 Metal work here is good, but two things block it now — one is my fault. **(1)** When I merged #488 (Metal grid-truncation fix) it added `tests/test_gemm_largebatch.mm`, which calls `coli_metal_gemm` with **8 args**; your PR changes that prototype to **9** (adds `gs`). So after a rebase, `make gemm-test` won't compile until that call gets `, 0` added. **(2)** It now conflicts with `dev` (today's merges touched nearby code). Rebase onto current `dev` and add the one-line fix to the two calls in `test_gemm_largebatch.mm`, and I'll merge it — the two latent fmt=4 fixes are real and I want them in. Sorry for moving the ground under you with #488.

monotophic:
Rebased and ready for review. Verification on 649b552: git range-diff against r4 shows all four commits patch-identical — a clean replay, no conflict resolutions, no content drift. The two things your review asked for are both present and re-verified on this head: the , 0 (gs=0, per-row scales) on both coli_metal_gemm calls in test_gemm_largebatch.mm, and the #488 chunk-loop reconciliation in coli_metal_gemm (your chunk machinery verbatim, the gs bind hoisted loop-invariant). Gates on this head: make check, METAL=1 build zero warnings, metal-test under both COLI_METAL_RESSET states, and gemm-test all clean including the S=7478 case — your own regression test exercising the merged prototype and chunk loop together. CI on the pushed head is green across the board.

JustVugg:
**Two open PRs now add fmt=4 (grouped-int4) to the Metal path — #457 (@monotophic, the GEMV kernel + two latent fmt=4 fixes) and #587 (@RDouglasSharp, attention + routed-expert decode).** They overlap on the same fmt=4 Metal surface, so we can't merge both blind, and there's a verification gap: **CI builds the Metal target but never executes it on a GPU**, so a green check proves it compiles, not that it's token-correct.

Two things unblock this, and then whichever clears them lands first (the other rebases on top):

1. **Reconcile the overlap.** Between you two: does one supersede the other, or are they complementary (GEMV kernel vs attention path) and mergeable in sequence? If complementary, say which goes first.
2. **Token-exactness from real hardware.** A short run on real Apple Silicon showing fmt=4 (a gs64 container) output is **bit/token-identical to the CPU path**, not just faster — the fmt=4 oracle pattern from #298. Post the diff (ideally the `make metal-test` correctness check on an fmt=4 fixture, or a greedy decode matching CPU).

Once (1) is agreed and (2) is posted for the leading PR, we merge it. This is the last gate — the work itself looks good, we just won't ship an unverified GPU decode kernel.

lBroth:
I have a Mac, so here is the hardware half of the second gate.

I merged `649b552` into `dev` at `7b03ce0` and ran the repo's gates on the merge rather than on the branch, since the branch is 51 commits behind and the merge is what would actually ship. The merge is clean, no conflicts. M5 Pro, 48 GB, macOS 26.5.2, Apple clang 21.

`make metal-test` passes: 43 cases, no failures, and a second run prints the same lines. The fmt=4 part of it:

```
Metal fmt=4 grouped-int4 tests (coli_metal_matmul vs matmul_i4_grouped semantics):
  grouped gate/up I=6144(mult64) S=1 worst_rel=1.90e-08 (I=6144 gs=64 ng=96 S=1)  ok
  grouped down I=2048(mult64) S=1    worst_rel=2.09e-08 (I=2048 gs=64 ng=32 S=1)  ok
  grouped gate/up I=6144(mult64) S=4 worst_rel=1.83e-08 (I=6144 gs=64 ng=96 S=4)  ok
  grouped I=200 (non-mult-64) S=2    worst_rel=1.29e-08 (I=200 gs=64 ng=4 S=2)  ok
  grouped I=201 (odd, non-mult-64) S=3 worst_rel=2.40e-08 (I=201 gs=64 ng=4 S=3)  ok
  grouped I=64 (degenerate, ng=1) S=1 worst_rel=1.15e-08 (I=64 gs=64 ng=1 S=1)  ok
  grouped I=40 (<gs, single partial group) S=1 worst_rel=2.53e-08 (I=40 gs=64 ng=1 S=1)  ok
  grouped I=512(mult64) outlier-heavy S=3 worst_rel=5.89e-08 (I=512 gs=64 ng=8 S=3)  ok
  grouped I=201(non-mult-64) outlier-heavy S=2 worst_rel=4.91e-08 (I=201 gs=64 ng=4 S=2)  ok
  gemm S=64 int4-grouped   worst_rel=2.42e-08  ok
Metal fused attention tests (fmt=4 grouped q_a, proves bind_gemv gs plumbing):
  attn grouped-qa S=1 pos=37 nerr=5.53e-06 cache=7.87e-06 qraw=1.94e-08  ok
  attn grouped-qa S=4 pos=12 (MTP) nerr=5.00e-06 cache=1.38e-05 qraw=2.02e-08  ok
```

`make check` on the same tree is 186 tests, OK with 13 skipped.

Then the container half, since fixtures only prove kernels against references. `tools/convert_fp8_to_int4.py` converts the tiny oracle snapshot locally (`--indir glm_tiny --group-size 64`), so I built a small gs64 container and compared CPU against Metal on stock dev and on this merge, teacher-forced and greedy. This merge behaves exactly like stock. Stock dev already disagrees CPU-versus-Metal at two teacher-forced positions out of 32 on this container, and the merge reproduces the same two positions with the same tokens; greedy is token-identical to the CPU on both trees. So that small prefill disagreement pre-exists on dev, #457 does not add to it, and nothing this PR touches changes a token anywhere I can measure.

Two caveats. The tiny container is pathological, 4-bit quantization of a toy model leaves near-tie logits everywhere, so it amplifies numeric differences a real container would mostly hide. And I ran this PR merged into dev on its own, not combined with #587.

Happy to re-run on any tree you point me at, including a combined one, since the hardware side of this is cheap for me.

monotophic:

Both halves of the gate are now on the record for this PR — collecting them in one place so the decision is easy.

**(1) Reconciliation.** Complementary, this PR first. @RDouglasSharp and we worked this through on #585 on the 24th: #457 is the dense-GEMV kernel + entry-point/allocator fixes, #587 is the fused-attention + routed-expert decode built on the same plumbing; two twin hunks in `colibri.c` drop out of #587 on rebase, and the full hunk-by-hunk disposition map (every #587 hunk: drop-as-duplicate / keep / adapt, with citations) is attached on #585. His merged-457-587 exercise reached the same conclusion from the other direction, and his corrected combined branch is up. So: **#457 lands first; #587 rebases on top with the map** — exactly your sequence rule.

**(2) Token-exactness on real hardware.** Two independent sources:
- Our earlier evidence on this thread: `metal-test` output plus real-container (744B, fmt=4) Metal-vs-CPU agreement from the M5 Max.
- @lBroth's independent run yesterday: `649b552` merged into dev `7b03ce0`, `make metal-test` 43/43, greedy decode **token-identical CPU↔Metal** on a gs64 fixture — and the two teacher-forced positions that disagree are shown to pre-exist on stock dev, unchanged by this PR. Independent hardware, independent merge, same conclusion. (Thank you — that's the cleanest possible form of the check.)

With (1) agreed on-thread and (2) posted twice over, we read this PR as clearing the gate as stated. On #587's side, we've posted a concrete hypothesis + tested guard for the greedy divergence lBroth found there (see #587) — it folds into the post-#457 rebase, so landing #457 unblocks rather than delays that work.

monotophic:

A correction to one piece of the gate-2 evidence we posted here, surfaced by broader testing today for #587 — flagged on this thread directly so the merge decision doesn't rest on a stale datapoint.

**What we claimed:** real-container CPU↔Metal token-exactness from my M5 Max on a 744B g64 container. **What broader testing shows:** that was measured on one prompt, and it was prompt-luck. On a 5-prompt sweep (temp-0 greedy, prefill isolated per #624's protocol, same container), this PR's tree flips greedy top-1 vs CPU on **4 of 5 prompts** — pre-divergence drift ~0.4 logits median, flips landing wherever a decision gap dips under that band. Full matrix, mechanism, and raw dumps are on #587 (where the same protocol also evaluated the follow-on work); the one-line summary is that the drift belongs to the fmt=4 GPU GEMV family generally, at g64 grouping specifically — @lBroth's independent runs (43/43 metal-test, token-identical greedy on his containers) and @RDouglasSharp's flip-free real-g128 runs both stand; real g64 is where exactness breaks, prompt-dependently.

**What we think this does and doesn't change:** the kernels compute correct results (drift is reordering-class, ~1% of logit scale, quality-neutral — both backends are individually deterministic); the dense-path and toy/g128 evidence for this PR is intact; what fails is the unqualified "token-exact on fmt=4/gs64 containers" form of the gate-2 claim at real scale. The precedent from #622/#624 — scope the documented claim to what's measured (there: decode vs prefill; here: by grouping) — seems available if useful, but the call is the maintainer's. We wanted to correct our evidence (again) before a merge decision is made based upon incomplete verification.

RDouglasSharp:
Extended `COLI_METAL_MOE_EXACT` to actually cover all fmt=4 GPU paths — thanks for catching that the moe_submit-only version left g64 open.

**Root of the 4/5.** The flag gated only `moe_submit` (batched routed experts). fmt=4 reaches Metal through four doors, and three were ungated: the dense `matmul_qt`→`coli_metal_gemm` path, `coli_metal_attn_decode`, and — the one that matters for decode — `coli_metal_layer_decode`, which runs the **shared expert** (`sh_gate/sh_up/sh_down`, fmt=4) **and** attention on the GPU. That fused kernel ran in both my g128 and your g64 flag runs; g128 just stayed under the gap threshold. So the blocchi-75-vs-0 you flagged isn't config-dependent path selection — it's that ungated `coli_metal_layer_decode` (the `METAL-ATTN: layer GPU` line), not a serial routed-expert fallback.

**Fix.** `g_moe_exact` now also excludes fmt=4 from those three guards, so with the flag set, attention, shared expert, routed experts, and dense all fall to the exact CPU kernels. fmt=1/2 and the default (flag off, fast batched `float4`) are byte-unchanged.

**Verified on a real g128 container** (temp-0 greedy): with `COLI_METAL_MOE_EXACT=1` the `METAL-ATTN` line is now **gone** (attention off the GPU) alongside `expert su GPU 0` — previously `layer GPU 4914` persisted with the flag on. Default run unchanged (`layer GPU 4914`, `expert su GPU 41455`). g128 can't exercise a flip, so this only confirms the gate is now complete, not the flip-fix.

Branch (fetch-and-run, no PR):
**`RDouglasSharp/colibri:metal-moe-exact-allpaths`**
```
git fetch https://github.com/RDouglasSharp/colibri metal-moe-exact-allpaths
```

@monotophic / @lBroth — this is the one to re-run your g64 5-prompt sweep against with the flag on. Prediction: 0/5 now that the fused attention/shared path is also on CPU. If it closes, the flag is a genuine gate-2 verification mode with the fast batched default preserved; if anything still flips, there's a fifth fmt=4 GPU path I've missed and it narrows the search further.

lBroth:
Fixture-side confirmation on metal-moe-exact-allpaths at d7ca0ed, since the real g64 sweep is @monotophic's to run. But first a scope note on my own evidence, since the correction above keeps my greedy datapoints standing: they were single-prompt too. Tonight's self-check on #452 narrows them, 1 of 10 synthetic prompts flips on the int3 container through the fmt=2 shared path, pre-existing on stock; and on the pure-#457 tree the toy fixture runs no fmt=4 GPU path at all in greedy, counters read zero, so that greedy identity never exercised these kernels in decode. The 43/43 metal-test half and the TF prefill attribution stand as-is.

On the gs64 toy fixture, where the batched-path divergence was first caught: flag off, the harness prompt diverges at token 5 as before and 2 of 10 synthetic seeded prompts flip CPU vs Metal; flag on, all 11 prompts come out token-identical, and the counters read what the gate promises, zero blocks and zero experts (METAL-ATTN is absent on toy dims regardless, so the fixture cannot exercise that part of the gate; your g128 counter run already covers it). So on the surface I can exercise, the extended gate closes everything the ungated version left open, consistent with your prediction for the real container.

One boundary datapoint, expected by design but worth having on the record for gate-2 semantics: on the int3 fixture, where the flip comes through the per-row int4 shared expert (the fmt=2 case I posted on #452 tonight), the flag leaves it on the GPU (counters still read 2 blocks, 2 experts with the flag set) and the flip persists. That is the documented scope, fmt=4 only, doing exactly what it says. It just means the flag is a verification mode for the fmt=4 family specifically, not for the batched accumulation drift class as a whole, which #452 shows also lives in the fmt=2 branch at fixture scale.

Runs were deterministic across repeats, both settings, and the seed harness is the same one from the #452 self-check if anyone wants the recipe.

monotophic:
(We/our = the joint analysis; me/my = @Monotophic's own hardware and containers).*

@RDouglasSharp — we ran your ask: the real-g64 5-prompt sweep against
`metal-moe-exact-allpaths` @ d7ca0ed, flag on, same protocol as our matrix
(temp-0 greedy ngen-64, `COLI_METAL_GEMM_MIN=100000`, 744B g64 container,
logit-gap dumps on). Two results — the second one is the good news.

## Run 1: prediction refuted informatively

**Flag-on still flips 4/5** (first-flip positions: p1 —, p2@7, p3@62, p4@3,
p5@35). Flag-off control on p2 flips as always (Metal engaged; not a build
artifact). But the engagement counters explain everything:

```
METAL: blocchi GPU 75 | fallback CPU 0 | expert su GPU 75
```

75 expert blocks stay on the GPU with the flag on. The cause is container
tier structure, not your gates: our container is mixed-tier (v1 recipe —
fmt=1 int8 resident experts + fmt=4 int4-g64 disk experts). The batch builder
submits the resident subset with `mfmt=1`, which passes a `fmt == 4`-scoped
gate untouched — so the fmt=1 resident tier keeps riding the batched kernel.
A uniform-format fixture can't see this, which is why the fixture closed
11/11 while the real container didn't.

The flip signature (—, 7, 62, 3, 35) is identical to our matrix's `attnoff`
and pure-#457 conditions — the shared element across all three is exactly
this fmt≠4 resident tier on GPU. Which revises our earlier attribution
(the "fmt=4 GEMV family" framing from the sweep-matrix comment): your gates
DO remove the fmt=4 paths' contribution; the residual g64 drift is the
expert-tier accumulation class, format-general, riding the residents.

## Run 2: one-line change → token-exact, pre-registered

We patched the expert gate fmt-agnostic (`if (g_moe_exact) return nil;` — all
routed experts to CPU under the flag), pre-registered 0/5, and re-ran:

**0/5 — all five prompts token-identical CPU↔Metal**, counters
`blocchi GPU 0 | expert su GPU 0`. To our knowledge this is the first
GPU-enabled configuration that is token-exact on a real g64 container in any
tree or condition measured on this thread family.

So your prediction was right about the mechanism you gated and wrong only
about a tier your fixture couldn't represent — and the fix is the one-liner:
scope `COLI_METAL_MOE_EXACT` by path (all routed experts), not by fmt. Cost:
flag-on also moves resident-tier experts to CPU, which seems consistent with
the flag's opt-in-exactness contract; the fast default is untouched.

Full tables + provenance (engine md5s, raw sweep dirs, exact positions):
happy to attach the results doc on request. Evidence file of record on our
side: `L10_ALLPATHS_RESULTS_2026-07-28.md`.

*Durable vs current-state: the mixed-tier gate-scoping finding and the
path-vs-fmt fix direction are durable; flip positions and counters are
calibrations at (d7ca0ed, glm52_v1 g64 container + kv_b overlay, M5 Max,
2026-07-28) and re-derive with tree or container.*

JustVugg:
Heads-up, and I want to be transparent rather than pick a winner quietly: **#457 (@monotophic) and #587 (@RDouglasSharp) implement the same thing** — Metal grouped-int4 (`fmt=4`) — and they conflict with each other in `c/backend_metal.h`, `c/backend_metal.mm` and `c/colibri.c` (4 conflicts). Each is clean against `dev` on its own; merging either one forces the other into a rebase or makes it redundant.

Both of you have put real work in and I would rather you two decide how to combine it than have me choose based on merge order. It worked well last week: @steve-m and @ZacharyZcR hit the same overlap on Kimi and resolved it by one absorbing the other's commit with authorship preserved (#676/#684) — clean history, credit intact, no duplicated effort.

Whatever you agree on, I will merge. If it helps, tell me which one should be the base and I will hold the other until it rebases on top.

monotophic:
@JustVugg we have worked together to coordinate our two PRs and determined that they complement each other if #457 merges first and then #587 follows. There has been excellent collaboration and these two PRs have been tested extensively now. This is a great community building around this project, thanks to both @RDouglasSharp and @lBroth.

Anything to add @RDouglasSharp?

Issue #467, opened 2026-07-20
[Performance]: RTX 5080 / 128 GB resource matrix (drive x compute x RAM) ΓÇö RAM dominates; COLI_CUDA=0 Windows quirk; PR-ready N-drive shard split


mohamedmastouri2000-boop:
## Summary

@JustVugg — after the #298 confirmation I kept going on the same box and ran a controlled 12-cell resource matrix (drive x compute x RAM, everything else held fixed) plus day-one tests of #421. Full results below, then four concrete findings/proposals — two are cheap doc/UX wins, one is a suggested experiment, one is a small feature I already have working and can PR if you want it.

## Setup

- **Host:** RTX 5080 (sm_120, 16 GB) / Core Ultra 9 285K / 128 GB DDR5 / Windows 11 native / 2x identical SK Hynix 2 TB NVMe (one is the OS drive)
- **Model:** g64 grouped-int4 container (fmt=4, 399.8 GB, 142 shards), byte-identical copy on each drive
- **Engine:** dev @ `68ac9ff` (post #298/#421)
- **Protocol:** full K=8, greedy `TEMP=0`, same prompt, 64-token decode, one cell at a time. All matrix cells with `PIPE=1 DIRECT=1` (see finding 1).

## Results (decode tok/s, expert hit)

| RAM | drive | GPU (dense+expert VRAM) | GPU-experts-only* | CPU (`--gpu none`) |
|---|---|---:|---:|---:|
| full (~121 GB) | data NVMe | **0.89** (84%) | 0.72 (86%) | 0.62 (77%) |
| full (~121 GB) | OS NVMe | 0.84 (83%) | 0.69 (85%) | 0.60 (76%) |
| full (~121 GB) | 2-drive shard split | **0.92** (84%) | — | — |
| 25 GB (`RAM_GB=25`) | data NVMe | 0.37 (32%) | 0.38 (33%) | 0.31 (16%) |
| 25 GB | OS NVMe | 0.35 (29%) | 0.36 (31%) | ~0.29 est. |
| *reference, no `PIPE`/`DIRECT`* | data NVMe | 0.67 (84%) | — | — |
| *reference, no `PIPE`/`DIRECT`* | #421 mirror | 0.47 (83%) | — | — |

\* the "GPU-experts-only" column exists because of finding 2 — it's what `COLI_CUDA=0` actually runs on Windows.

[...]

Issue #389, opened 2026-07-18
[Performance]: 5.42 tok/s — dual-socket Xeon Gold 6430 AVX-512 VNNI, 1TB DDR5 RAM


biemster:
### Commit

72d3d37231e922a6fa9afca16e08fa45842d5eb4

### Hardware and storage

dual-socket Xeon Gold 6430 AVX-512 VNNI, 1TB DDR5 RAM

### Software environment

Ubuntu 22.04, GCC 11.4.0

### Build and benchmark commands

```shell
./iobench ../../glm52/out-00069.safetensors 19 64 8 0

and

OMP_NUM_THREADS=64 OMP_PROC_BIND=spread OMP_PLACES=cores COLI_NO_OMP_TUNE=1 PIN=/home/user/temp/glm52/.coli_usage PIN_GB=340 numactl --interleave=all ./coli run --gpu none --ngen 250 --temp 0 --model /home/user/temp/glm52 "explain the fine structure constant in 10 lines"
```

### Results

```text
This performance report is very similar to the hardware of #231, but run bare metal and with 1TB DDR5 RAM. The instructions for running a benchmark are a bit confusing, but this is what I did:

$ ./iobench ../../glm52/out-00069.safetensors 19 64 8 0
buffered x8 threads: 64 reads x 19MB = 1.3 GB in 0.08s -> 16.27 GB/s (1.2 effective ms/block)

and

$ OMP_NUM_THREADS=64 OMP_PROC_BIND=spread OMP_PLACES=cores COLI_NO_OMP_TUNE=1 PIN=/home/user/temp/glm52/.coli_usage PIN_GB=340 numactl --interleave=all ./coli run --gpu none --ngen 250 --temp 0 --model /home/user/temp/glm52 "explain the fine structure constant in 10 lines"

     ▄▀▀▀▄  ▄        colibrì v1.0
  ▄▄▄▄▀▀▀▀▄▀▀        tiny engine, immense model
      ▀▀▀▀▀▀▀        GLM-5.2 · 744B MoE · int4 · streaming CPU
        ▀▀▀▀         run
          ▀          
  ──────────────────────────────────────────────────────────
== GLM C engine (glm_moe_dsa), cache=8 experts/layer | experts@8-bit dense@8-bit | idot: avx512-vnni ==
loaded in 6.63s | resident dense: 9912.75 MB | layers=78 experts=256 | MTP ACTIVE (draft=3)
[MTP] active: native speculative decoding (draft=3)
[...]

PR #470, opened 2026-07-20
fix(plan): respect explicit COLI_CUDA_MTP=1 - skip the planner's DRAFT=0 export (opt-in was silently inert)

mohamedmastouri2000-boop:
## What

Per the request in #467: when `COLI_CUDA_MTP=1` is explicitly in the environment, `_auto_tune` skips its `DRAFT=0` export so the engine's `DRAFT=-1` auto path (`colibri.c` — `g_draft = (has_mtp && (!cuda || cuda_mtp)) ? 3 : 0`) can actually see the opt-in. Plus the ENVIRONMENT.md line next to `COLI_CUDA_MTP`.

## Why

The documented opt-in was silently inert on the Windows bare-run and `--auto-tier` flows: `resource_plan.py` `_auto_tune` exported `DRAFT=0` (compute-bound, or disk-bound with projected hit < 0.90), which preempts the engine's auto path — the only place `COLI_CUDA_MTP` is consulted. Anyone A/B-ing speculation under CUDA got a silent no-op unless they also knew to set `DRAFT=3` (my first A/B pass in #467 measured exactly baseline-vs-baseline noise because of this).

## Behaviour

- `COLI_CUDA_MTP` unset (default): unchanged — planner still exports `DRAFT=0` on compute-bound / low-hit-disk plans; MTP stays off under CUDA. This is the measured-correct default (#389 compute-bound −42%; #467 streaming-bound −32%).
- `COLI_CUDA_MTP=1` explicit: planner leaves `DRAFT` alone → engine auto path resolves draft=3. An explicit user `DRAFT=n` continues to win over everything (`environment_for_plan` is `setdefault`-based).

## Verification

On the #467 box: `COLI_CUDA_MTP=1` alone (no `DRAFT`) through the bare-run Windows flow now loads with `MTP ACTIVE (draft=3)` and speculates (`2.00 tok/forward` on a 2-token smoke); without the var, `draft=0` as before. The doc line also records the measured trade-off so nobody reads the opt-in as free speed: acceptance holds (49% on sm_120) but at ~85% hit the widened S=4 expert union costs −32% net; the opt-in pays near full residency.

## Scope

`c/resource_plan.py` (one guard + comment in `_auto_tune`), `docs/ENVIRONMENT.md` (one row). No engine changes.

### DIFF
--- a/c/resource_plan.py
+++ b/c/resource_plan.py
@@ -296,7 +296,16 @@ def _auto_tune(bottleneck_class, projected_hit, gpus, cpu_sockets, plan_has_meta
     n_gpu = len(gpus)
 
     # MTP: costs more than it saves when compute-bound (#389 measured 42% loss)
-    if bottleneck_class == "compute":
+    # or streaming-bound (#467 measured 32% loss under CUDA at 85% hit).
+    # EXCEPTION: an explicit COLI_CUDA_MTP=1 in the environment is a documented
+    # opt-in to test speculation under CUDA (glm.c resolves DRAFT=-1 -> 3 only
+    # when it sees the var). Exporting DRAFT=0 here preempted that auto path,
+    # so the opt-in was silently inert on the Windows bare-run/auto-tier flows
+    # (#467): respect it and let the engine's auto path take over. Unset still
+    # gets DRAFT=0 -> MTP off, which is the measured-correct default.
+    if os.environ.get("COLI_CUDA_MTP") == "1":
+        pass  # explicit opt-in: leave DRAFT to the engine's auto resolution
+    elif bottleneck_class == "compute":
         tune["DRAFT"] = {"value": "0",
                          "reason": "compute-bound: MTP batch overhead exceeds yield"}
     elif bottleneck_class == "disk" and projected_hit < 0.90:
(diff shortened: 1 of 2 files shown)


JustVugg:
Merged into `dev` — thank you @mohamedmastouri2000-boop. Verified locally beyond CI: unit-tested `_auto_tune` both ways — `COLI_CUDA_MTP=1` at disk/0.85 hit now leaves `DRAFT` unset (opt-in engages draft=3), and unset still exports `DRAFT=0` (the measured-correct default). `import os` present, CI 8/8.

This closes the third-and-last wrapper gotcha from #467 — the documented opt-in now actually works, so anyone can reproduce your MTP-under-CUDA A/B (49% acceptance, −32% at 85% hit) without the `DRAFT=3` incantation. One tiny non-blocking note for a future pass: `COLI_CUDA_MTP=1` set on a *no-GPU* box would now also skip the `DRAFT=0` export (it's a CUDA flag, so that's user-error territory) — if you ever want to tighten it, gating on `has_gpu` would scope it to where it's meant to apply. Not worth a PR on its own; just noting it.

With this, **#467 is fully closed**: findings 1/2/4 shipped (#468/#469), finding 3 measured with the default confirmed, and the opt-in now honoured (#470). Genuinely one of the best-run issues in the repo.

Issue #472, opened 2026-07-20
Datapoints: GLM-5.2 int4 fully RAM-resident on 2-socket Ice Lake (48C, 661 GiB) — 3.85 tok/s; DRAFT=0 beats MTP at full residency


veerareddyvishal144:
Benchmark datapoints from a 2-socket Ice Lake host with the whole int4 container RAM-resident — posting per the README's request for real-machine numbers. Also independently confirms two existing findings (#82 NUMA gains; the 6×5090 experiment log's "speculation loses at full residency") on pure-CPU hardware.

### Machine

- Azure Standard_E96d_v5: 2× Intel Xeon Platinum 8370C (Ice Lake), 48 physical cores / 96 SMT, 2 NUMA nodes, 661 GiB RAM, avx512-vnni IDOT kernel
- Local NVMe: 4.18 GB/s O_DIRECT (iobench 19 MB × 64, 8 threads); buffered 26.9 GB/s (page cache)
- colibri `main @ 00ed8d5`, gcc 13.3, `setup.sh`, self-test 32/32
- Model: `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp` (int8 MTP heads verified by size)

### Method

Greedy (`TEMP=0`), `--ngen 256`, single request, `KVSAVE=0`, fresh process per run. Full residency via `PIN_FILL=1` + `PIN_GB=380` (all 19,456 experts pinned at startup, ~47 s warm / ~2 min cold page cache; expert hit 100%, disk service 0.000 s in every run below). 48 physical cores, `OMP_PROC_BIND=spread OMP_PLACES=cores` with the `COLI_OMP_TUNED=1` workaround for the re-exec affinity issue (#471). Run-to-run noise observed ±0.1 tok/s.

### Results (each row changes one thing vs the row above)

| config | tok/s | notes |
|---|---:|---|
| baseline (defaults + full pin) | 2.14 | expert-matmul 58% of wall, 48.9 GB/s effective |
| + `COLI_NUMA=1` | **2.65** | +24% — matches the 2-socket selective-interleave gains in #82 |
| + `DRAFT=0` (MTP off) | **3.65** | see below |
| + `I4S=1` (int4 IDOT at S=1) | **3.85** | +5.5%; expert-matmul 67.8→89.5 GB/s |
| + `TOPP=0.7` (lossy) | 4.85 | for completeness |

[...]

PR #473, opened 2026-07-20
quant: int4 IDOT at S=1 on AVX-512 VNNI (g_i4s=1), like ARM SDOT (+5.5% measured)

veerareddyvishal144:
### What

Set `g_i4s=1` when `__AVX512VNNI__ && __AVX512BW__` — the same S=1 carve-out the tree already has for ARM SDOT and VSX. The original `g_i4s=2` threshold came from an AVX2 measurement; with VNNI the int4 IDOT kernel wins at S=1 as well.

### Measured (2× Xeon Platinum 8370C, Ice Lake, 48 physical cores / 2 NUMA nodes, 661 GiB RAM)

GLM-5.2 int4 (int8 MTP heads), all 19,456 experts RAM-resident (`PIN_FILL`, hit 100%, disk 0%), `TEMP=0 DRAFT=0 COLI_NUMA=1`, 256-token greedy decode, single request:

| | tok/s | expert-matmul effective |
|---|---:|---:|
| `I4S=2` (current default) | 3.65 | 67.8 GB/s |
| `I4S=1` (this PR's default) | **3.85** | **89.5 GB/s** |

Output stays coherent under greedy decode (same IDOT family every S≥2 forward already uses; the usual ~0.3% RMS activation-quantization note applies to the S=1 down-proj that switches family).

### Validation

- Clean build, 0 warnings (gcc 13.3, `ARCH=native`)
- Oracle: 32/32 teacher-forcing + 20/20 greedy (note: the tiny oracle is bf16, so these gate the default path; the int4 S=1 effect is covered by the real-model measurement above — same validation footprint as the SDOT/VSX carve-outs)
- `SPEC_PIN` interaction: with `g_i4s=1` the `spec_pinned() ? g_i4s<=1 : S>=g_i4s` gate makes draft and verify agree on the IDOT family by construction, same as on ARM.

Happy to run any additional A/B on this host (it's a 48-core AVX-512 VNNI box with the model resident).

Full config ladder and profile context for this host: #472.


JustVugg:
Just a rebase heads-up: `dev` moved (a batch of CPU perf/fix PRs landed — #481/#476 among them), and this now conflicts. It's a clean, high-value change (int4 IDOT at S=1 on AVX-512 VNNI) — could you rebase on current `dev`? Once green I'll verify byte-identity and merge. Happy to help if the conflict is mechanical. Thanks!

PR #475, opened 2026-07-21
moe: opt-in XEXP=1 — one OpenMP region per expert block at S=1 full residency (+11.6% ABAB on 2-socket 48C, byte-identical)

veerareddyvishal144:
### What

`XEXP=1` (opt-in, default off): at S==1 decode, when every expert of a batch-union block is RAM-resident and int4 and the S=1 int4-IDOT family is active (`g_i4s<=1`), compute the whole block in **one OpenMP parallel region** — gate+up row-chunks with fused silu·up, a per-expert intermediate requant, then down-projection row-chunks — instead of ~2 fork/join regions per expert (16 per layer at topk=8).

### Why

On a 2-socket Ice Lake (48 physical cores), `PROF=1` at full residency shows expert-matmul as the top phase but running at only ~68–89 GB/s effective vs ~350+ GB/s platform DRAM bandwidth, with near-zero matmul scaling from 24→48 cores (61.5 → 67.8 GB/s). The per-expert region cadence gives each worker ~200 KB between barriers; wide hosts burn the bandwidth headroom on fork/join latency and stream restarts. One region per block gives ~1 MB contiguous chunks per worker and 2 barriers per block instead of 16.

### Measured

2× Xeon Platinum 8370C (Ice Lake, 48C, 2 NUMA), 661 GiB RAM; GLM-5.2 int4 fully resident (`PIN_FILL`, hit 100%, disk 0%); `TEMP=0 DRAFT=0 COLI_NUMA=1 I4S=1`, 256-token greedy, single request, fresh process per run, **ABAB-interleaved**, 3 prompts × 2 repeats:

| prompt | XEXP=1 | XEXP=0 | gain |
| --- | ---: | ---: | ---: |
| sky | 4.70 / 4.68 | 4.22 / 4.21 | +11.3% |
| water cycle | 4.74 / 4.66 | 4.20 / 4.21 | +11.8% |
| rivers | 4.64 / 4.67 | 4.13 / 4.20 | +11.8% |
| **mean** | **4.68** | **4.20** | **+11.6%** |

(A non-interleaved early pair suggested +24%; the ABAB battery supersedes it — baseline sessions drift a few % with the accumulated `.coli_usage` pin ordering, which is exactly what interleaving controls for.) `PROF` phase shift: expert-matmul 49% → 41% of wall, effective expert bandwidth 89.5 → 131.9 GB/s; attention becomes the top phase.

**Where it should NOT help:** a similar cross-expert restructuring was neutral/negative on the 24-core lab box (docs/experiments/glm52-6x5090-2026-07-12.md, "OpenMP restructuring of RAM experts"), and this path never engages under CUDA/Metal, on misses, during speculation, or at S≥2 — hence opt-in; candidates are wide multi-socket CPU-only hosts at full residency.

### Correctness

The math per row is `dot_i4i8` — the same function `matmul_i4_idot` uses at S=1 — with silu, requant (`qrow_i8`) and the j-ordered accumulation into `out` preserved, so the path is **byte-identical to the stock engine on the same kernel family**:

- GLM-5.2 int4 (real model): 256-token greedy output byte-identical (`cmp` clean, 1,695 chars), `XEXP=1` vs `XEXP=0 I4S=1 COLI_NO_FUSED_PAIR=1`.
- int4 tiny oracle (`make_glm_oracle.py --fp8` + `convert_fp8_to_int4.py`): greedy tokens identical with XEXP on/off (verified on this branch's binary).
- bf16 tiny oracle (default path untouched): 32/32 teacher-forcing + 20/20 greedy on this branch's binary.
- Build: 0 warnings (gcc 13.3, `ARCH=native`).

Note on SCORE: not applicable as a quality gate — SCORE runs batched teacher-forcing (S=T) and never enters this S=1-only path; byte-identity is the stronger statement (quality delta is exactly zero relative to the `g_i4s<=1` family, whose S=1 default on VNNI is proposed separately in #473).

### Gating

`g_xexp && S==1 && !nmiss && !spec_pinned() && g_idot && g_i4s<=1 && !cuda && !metal_done`, plus an all-int4/shape check per block; any failure falls through to the stock loop unchanged. `spec_pinned()` keeps draft/verify on one kernel family per #163's rule.

Full host context and the config ladder that led here: #472.


JustVugg:
Rebase heads-up: `dev` advanced (perf/fix batch incl. #481/#476), and this conflicts now. XEXP=1 as an opt-in with byte-identical output is exactly the kind of change we want — could you rebase on current `dev`? I'll verify (opt-in default off, token-exact) and merge once green. Shout if you want a hand with the conflict.

Issue #442, opened 2026-07-19
[Performance]: hot-path float reductions are scalar and don't auto-vectorize (router matmul + MLA-absorb attention)


cdhdt:
### Observation

The build is `-O3 -fopenmp` with **no `-ffast-math`, no `restrict`, no `#pragma
omp simd`** (0 occurrences of each in the tree). That's a fine correctness default,
but it means the compiler may **not** reassociate floating-point reductions — so
any hand-scalar `a += x[i]*w[i]` reduction in a hot path stays scalar (no
auto-vectorisation), running at ~1/8–1/16 of the available SIMD width.

Two such reductions sit in the per-token decode path (`main` @ 72874f3):

1. **Router logits — `matmul` (`c/glm.c`, ~line 379).** A fully scalar f32
   `y[o] += x[i]*W[o][i]`, used every MoE layer to produce the router logits
   (`logits_all`, ~line 2919): D≈6144 × E=256 ≈ 1.5M scalar MACs per token per
   layer, no SIMD.

2. **MLA-absorb attention core (`c/glm.c`, ~lines 2792–2803).** The per-`(s,h)`
   inner loops `a += qabs[i]*Lt[i]` (scores) and `clat[i] += a*Lt[i]` (value mix)
   are scalar and walk the latent KV twice per head per token; cost grows with
   context length.

### Proposal

Add SIMD FMA inner loops (AVX2 / AVX-512 / NEON), or `#pragma omp simd reduction`,
for these reductions — the file already has the vector kernels to reuse.

### Important caveat (why this is an issue, not yet a PR)

SIMD reduction sums in a different order than the scalar loop, so results are **not
bit-identical**. For the router that is load-bearing: a tiny rounding change can
flip a top-k routing decision and change the emitted token. So any such PR **must**
be gated by the end-to-end token-exact oracle (`SNAP=./glm_tiny TF=1 ./glm …`,
32/32 + 20/20). I can't run that locally yet — `glm_tiny/` isn't in the repo and I
don't have the full checkpoint on this box — so I'm filing the analysis rather than
a PR I can't validate to the project's bar. Wave-2 PR once the oracle is runnable
here.

For the attention core, a fixed-order SIMD reduction (e.g. tree/pairwise) plus the
oracle A/B is the safe route; the router likely wants the exact same summation
order the scalar loop uses (or an integer/quantised router path) to stay
token-exact.

Cross-ref: #153 (attention-projection precision on x86), #292 (token-exactness
[...]

PR #477, opened 2026-07-21
attn: vectorize the MLA absorption helpers — +13% decode on AVX-512 hosts, byte-identical output

veerareddyvishal144:
### What

The S=1 absorbed-attention path dequantizes `kv_b` row-by-row through **scalar** nibble loops: `qt_addrow` (q_nope absorption) and `qt_matvec_rows` (value projection). At H=64, `kv_lora=512`, 78 layers that is ~1.15G scalar dequant+MAC ops per decoded token — on a 48-core Ice Lake at full residency it measures **44 ms/token, roughly two thirds of the whole attention phase** (`score-softmax-value` in the PROF breakdown).

This PR vectorizes both:

- `qt_addrow` (int4) → new `axpy_i4f_avx512`, the axpy twin of the existing `dot_i4f_avx512`. Each `acc[i]` receives exactly **one fma per call** — no cross-element accumulation — so it is **bit-identical** to the scalar loop (gcc `-O3` already contracts the scalar body to fma).
- `qt_matvec_rows` (int4) → calls the existing, already-qualified `dot_i4f_avx512`, under the **same `g_i4_acc512` gate** and with the same documented accumulation-order tradeoff as `matmul_i4`'s I4_ACC512 path. `I4_ACC512=0` restores the scalar order exactly as it does elsewhere.

Non-AVX-512 builds and all other formats (fp32/int8/int2/grouped) are untouched.

### Measured (2× Xeon Platinum 8370C, 48C/2 NUMA; GLM-5.2 int4 fully resident via `PIN_FILL`, hit 100%, disk 0%; `TEMP=0 DRAFT=0 COLI_NUMA=1`, 256-token greedy, stock kernels otherwise; this branch vs clean `dev`, same host, back-to-back)

| | clean dev | this PR |
|---|---:|---:|
| decode | 4.23 tok/s | **4.79 tok/s (+13.2%)** |
| score-softmax-value | 11.54 s | **5.35 s** |
| 256-token greedy output | — | **byte-identical** (`cmp` clean) |

The gain is config-independent (the absorption cost does not depend on expert-path settings); combined with #473 + #475 this host reaches 5.4–5.5 tok/s.

### Validation

- `I4_ACC512_TEST=1` selftest: ok
- bf16 tiny oracle: 32/32 teacher-forcing + 20/20 greedy on this branch's binary
- Build: 0 warnings (gcc 13.3, `ARCH=native`)

Host context: #472.


JustVugg:
Heads-up: #481 (SIMD MLA-absorb score + value-mix reductions) just merged into `dev` — it also vectorizes the MLA-absorb path (#442), so this now conflicts with it. Could you check whether your changes are **complementary** to #481 (rebase on top — e.g. if you vectorize a different helper or a wider ISA) or **superseded** by it? If complementary, a rebase on current dev and a note on what's additive vs #481 and I'll merge. Both target the same measured win, so I want to land the best of both without double work. Thanks @veerareddyvishal144 — the AVX-512 numbers are great.

Issue #490, opened 2026-07-21
[Performance]: LRU hit share dropped 27-38% to 15% on Windows/sm_120 between 68ac9ff and 4aca059 (0.95 to 0.82 tok/s) - confounded observation, bisect offered


mohamedmastouri2000-boop:
### What happened

While running the CACHE_ROUTE ladder on my usual box (RTX 5080 sm_120 / 128 GB DDR5 / Windows 11, g64 grouped-int4, `CUDA_DENSE=1` + 4 GB tier, `PIPE=1 DIRECT=1`, full K=8, greedy, 64-tok decode — same protocol as #467), the **stock-routing baseline dropped from my long-standing ~0.95 tok/s to 0.82**, and the composition of the expert hit rate changed shape:

| binary | decode tok/s | hit | pin share | LRU share |
|---|---:|---:|---:|---:|
| dev @ `68ac9ff` (many runs, #467) | 0.89-0.95 | 84-86% | 47-59% | **27-38%** |
| dev @ `4aca059` (2 runs) | 0.82 | 74.5% | 59.4% | **15.0%** |

The pin share is stable; it's specifically the **LRU contribution that collapsed** (27-38% → 15%). Prefill unchanged; coherent output; no fallbacks.

### Why I'm asking rather than asserting

Two changes landed between my binaries in the LRU/prefetch area — #474 (LFRU-aware pilot eviction guard) is the obvious candidate since it changes eviction decisions on exactly this path — but my observation is **confounded**: my `.coli_usage` grew ~500k selections between sessions (pin set drifted), and page-cache state differed. So this is a "does this reproduce for anyone else / does it look plausible from the diff" question, not a regression claim.

### Offer

I can run a controlled bisect on this box: rebuild `68ac9ff`, `1b8f307` (pre-#474) and `4aca059` (post-#474) against a frozen copy of `.coli_usage`, 2 runs each, same session. ~1 hour of machine time — say the word and I'll attach the table. If it's real, the interesting part is that the guard may be *over*-protecting demand experts into staleness on hosts where PILOT_REAL issues a lot of cross-layer loads (6-7k per 64 tokens here, per the footer).

woolcoxm:
## Root cause found — it's real, and it's #474. Full credit to @mohamedmastouri2000-boop

[...]

PR #495, opened 2026-07-21
fix(cuda): honor explicit CUDA_EXPERT_GB under CUDA_DENSE=1 (#491)

woolcoxm:
Closes #491.

## Root cause

Reported by @mohamedmastouri2000-boop: with `CUDA_DENSE=1`, the CUDA hot expert tier caps at 109 experts (2.31 GB) regardless of the `CUDA_EXPERT_GB` budget — `=4` and `=6` produce byte-identical tiers.

In `pin_load()` the effective expert budget was computed as:

```c
double budget=g_cuda_expert_gb*1e9, safe_total=0;
... remaining[i] = free_b - g_cuda_dense_projected[i] - 2e9;   // -2GB reserve
if(g_cuda_expert_auto || budget>safe_total) budget=safe_total; // <-- silent clamp
```

With `CUDA_DENSE=1`, `g_cuda_dense_projected` accounts for the dense tensors already claimed during model load. On a 16 GB card (17.1 GB usable) that leaves `safe_total ≈ 2.3 GB` after the 2 GB reserve, regardless of the configured budget. Because `budget > safe_total`, the explicit `CUDA_EXPERT_GB=4` got clamped down to 2.3 GB — and `=6` got clamped to the same value. Hence identical tiers.

The tier log then printed `g_cuda_expert_gb` (the configured value) instead of the effective `budget`, masking the clamp:

```
[CUDA] hot expert tier: 109/2431 experts, VRAM 2.31 GB (total budget 4.0 GB)
                                                                   ^^^^^ lies
```

## The fix (all in `c/colibri.c`, ~14 lines)

1. **Stop clamping explicit budgets.** `if(g_cuda_expert_auto) budget=safe_total;` — only auto mode fills to measured headroom. An explicit `CUDA_EXPERT_GB` is honored as-is. This is safe because the existing per-expert upload-failure path (`remaining[best]=0`, continue to next expert) already degrades gracefully on OOM rather than crashing.
2. **Tunable reserve.** The hardcoded `2e9` becomes `CUDA_RESERVE_GB` (default 2.0 GB), parsed alongside `CUDA_EXPERT_GB`. Lets a 16 GB card owner tighten the margin deliberately to reclaim headroom beyond just honoring the explicit budget.
3. **Honest log.** The `hot expert tier` line now prints the *effective* budget (`safe_total` in auto mode, the explicit budget otherwise), flags auto mode, and surfaces the reserve.

[...]

PR #497, opened 2026-07-21
fix(pilot): narrow LFRU eviction guard so speculations don't over-drop (#490)

woolcoxm:
## What

Fixes the LRU hit-share collapse reported in #490 (27-38% → 15% on Windows/sm_120, 0.95 → 0.82 tok/s between `68ac9ff` and `4aca059`). The regression is real, it is #474, and the confounders the reporter flagged (`.coli_usage` drift, page-cache state) did **not** contribute — the bug is deterministic from the source.

## Root cause

`#474` (`eaa68f4`) added an eviction guard to the two PILOT_REAL victim-selection sites — `pilot_realload` (the blocking path, the one Windows hits) and `pilot_uring_batch` (Linux). The intent was right: stop a speculation from evicting a just-demand-loaded, still-warm expert (the #441 thrash). The shipped comparison was inverted for a speculation:

```c
if(cs <= vs+(vs>>2)+(4u<<8)){ drop; return; }   // cs=speculation, vs=victim
```

`tier_lfru_score = (heat<<8)|recent`, so one frequency count = 256 and recency ≤ 255. The test demanded the **speculation** beat a just-used **demand victim** by 25% + 4 full frequency counts. A correct speculation is, by definition, historically colder than the resident it would evict — so the condition is almost never met. Once the cache is full, the guard drops ~100% of speculations (verified on the score math, 10⁶ samples). PILOT_REAL collapses to a willneed-only hinter: no real cross-layer loads, no LRU churn, the LRU share evaporates.

That is precisely the shape change measured: **pin share stable (47-59% → 59%), LRU share 27-38% → 15%**, hit 84-86% → 74.5%, 0.95 → 0.82 tok/s. No fallbacks, coherent output — the guard is cache-placement only, a dropped speculation is demand-loaded later to the same value, so the byte-identical contract held and **only the performance shape regressed** (the sneakiest kind).

## The fix

Protect the victim **only when it is genuinely warm** (`eheat >= 2` demand accesses) **and** clearly hotter than the speculation by the same 25%+4-freq hysteresis — i.e. test the *victim* against speculation+margin (protect the resident) instead of the *speculation* against victim+margin. This matches the `#474` commit message's stated intent ("never evict a **warm** demand-loaded expert") rather than the over-broad formula that shipped.

```c
if(g_pilot_evict_guard && m->eheat && m->elast && Sl[lru].eid>=0){
    int vid=Sl[lru].eid; uint32_t vh=m->eheat[layer][vid];
    if(vh>=2){
        uint64_t vs=tier_lfru_score(vh, m->elast[layer][vid], m->eaccess_clock);
        uint64_t cs=tier_lfru_score(m->eheat[layer][eid], m->elast[layer][eid], m->eaccess_clock);
        if(vs+(vs>>2)+(4u<<8) > cs){ drop; return; }
    }
}
```

Score math after the fix (10⁶ samples):

| victim regime | guard drops | guard allows (plain-LRU evict) |
|---|---|---|
| warm demand victim (heat ≥ 2) | 100% | 0% |
| cold victim (heat ≤ 1) | 0% | 100% |

The #441 thrash protection is preserved; the plain-LRU churn that produces the 27-38% LRU hit share is restored. Same correction applied to `pilot_uring_batch`. `PILOT_EVICT_GUARD=0` still restores plain-LRU for a single-binary A/B.

## Verification

- `make check` **111/111 OK** (12 skipped — GPU/optional), same as the `#474` baseline.
- Cache-placement only → **output byte-identical** (a dropped speculation is demand-loaded later, same value; the `moe()` barrier still fences every layer).

## Field diagnostic (no rebuild needed)

The footer already prints the counters:
```
PILOT_REAL: <N> load cross-layer completati, <M> scartati (main gia' sul layer)
```
On the broken `4aca059` build, **`M` ≫ `N`** (drops dominate). On this fix, a healthy split returns. One-line confirmation for anyone holding a `4aca059` binary.

## Credit

**@mohamedmastouri2000-boop** — the regression report, the correct first look at the diff (#474), the offered bisect, and the measured shape change that isolated LRU share as the collapsed component. Filed as a conservative "does this reproduce for anyone else" question; the report was exactly right and the bisect turned out not to be needed.

cc @cdhdt (author of #474) — the #441 thrash is real and this still guards against it; the change just narrows "block almost everything" to "block only when the victim is genuinely warm."

Refs #490, #474, #441.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -734,10 +734,10 @@ static int g_pilot_two=0; /* PILOT_TWO=1: two-step prefetch — before running L
                           * approximate MoE(L) using only the shared expert (resident, no disk)
                           * and add it to the state. Trades 3 small matmuls for +2.3% recall. */
 static int g_pilot_evict_guard=1;/* PILOT_EVICT_GUARD=0 -> old behavior (a speculation evicts the plain LRU).
-                          * Default ON: a speculative pilot load may evict a RESIDENT expert only if the
-                          * predicted expert is historically HOTTER than the victim (same LFRU hysteresis as
-                          * tier_pick_lfru); otherwise it drops the speculation rather than thrash a warm
-                          * demand-loaded expert. Cache placement only -> output byte-identical. (#441) */
+                          * Default ON: protect a RESIDENT demand-loaded expert from a speculation only when
+                          * it is genuinely WARM (>=2 accesses) AND clearly hotter than the predicted expert
+                          * by tier_pick_lfru's 25%+4-freq hysteresis; otherwise the speculation may evict the
+                          * plain-LRU victim as before. Cache placement only -> output byte-identical. (#441, #490) */
 /* Handshake main<->pilota per il load-vero cross-layer. Invariante di sicurezza in DUE parti:
  *  1) Percorso MATMUL (moe): il pilota scrive SOLO ecache[layer] con layer > g_cur_moe_layer;
  *     il matmul in moe() legge SOLO ecache[layer]==g_cur_moe_layer, e la barriera a inizio moe()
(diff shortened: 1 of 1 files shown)


JustVugg:
@mohamedmastouri2000-boop can you test it too?if ok i merge it!

mohamedmastouri2000-boop:
**Validation on the #490 box (RTX 5080 / Windows, g64, `CUDA_DENSE=1 CUDA_EXPERT_GB=4 CUDA_RESERVE_GB=1 PIPE=1 DIRECT=1`, full K=8, greedy, 64-tok decode). Built as this PR's commit cherry-picked onto current `dev` (`5e02064`, so #495 + #477 included). Fix confirmed directionally; one honest caveat on magnitude.**

| binary | stock tok/s | hit | LRU share | CR J3/M16 | CR M24 |
| pre-#474 (`68ac9ff`, many runs) | 0.89–0.95 | 84–86% | 27–38% | — | — |
| regressed (`4aca059`) | 0.82 | 74.5% | **15.0%** | 1.10 | 1.18 |
| **this PR on dev** | **0.89** | 79.4% | **18.7%** | **1.18** | **1.27** |

**The guard-narrowing works**: speculations evict again, LRU share recovers 15.0 → 18.7% stock (and 23.3–23.4% under CACHE_ROUTE), stock decode 0.82 → 0.89.

**The honest caveat**: LRU share does not fully return to the pre-#474 27–38% band on this box — and since this build also carries #477's +13% decode, a *complete* restoration should arguably land stock above the old 0.95, not at 0.89. So the narrowed guard still drops meaningfully more speculations than pre-#474 behavior. That may be exactly the intended trade (the #441 protection is worth something), but flagging it so the residual isn't discovered later as "another regression": on this workload the guard still costs an estimated ~5–10% vs pre-#474 eviction behavior. If useful I can run the same two cells with `PILOT_EVICT_GUARD=0` (if the kill-switch exists in this shape) to size the residual exactly.

**Side effect worth knowing**: a healthy LRU compounds with CACHE_ROUTE — the balanced J3/M16 config went 1.10 → 1.18 *with improved fidelity* (swap 17.9%, agree 82.1%, KL 2.93 vs 3.42), and M24 sets a new box record at **1.27 tok/s / 96.4% hit**. The regression's biggest victim was routing-assisted configs, not stock.

Raw log available. Thanks for the fast root-cause — the 256-points-per-frequency-count analysis matched the observed collapse exactly.

JustVugg:
Merged. Thank you both — @woolcoxm for the root-cause (the inverted comparison is exactly the kind of bug that hides behind a correct-looking guard) and @mohamedmastouri2000-boop for validating it on the very box that reported the regression.

**My mistake to own:** I merged #474 after verifying it was byte-identical — which it was, since cache placement never changes tokens. But byte-identity is blind to *performance shape*: a guard that drops 100% of speculations still produces identical output, just slower. The lesson for cache/placement PRs is that the correctness gate (token-exact) must be paired with a **behaviour** gate (hit-rate / LRU-share / bytes-per-token), or exactly this class of regression sails through. I'll apply that going forward.

**On your caveat, @mohamedmastouri2000-boop — yes please, run it.** `PILOT_EVICT_GUARD=0` still exists in this shape, so the A/B (guard on vs off, same binary) sizes the residual exactly. Your reasoning is right: with #477's +13% in the same build, a *full* restoration should land above the old 0.95, and 0.89 means the narrowed guard still costs ~5-10% of speculation opportunity. That may well be the correct trade for the #441 thrash protection — but it should be a *measured* trade, not an assumed one. If the residual is large, the next refinement is obvious: make the warmth threshold (`eheat>=2`) tunable, or scale it by cache pressure.

Post that A/B and we'll decide whether the current threshold is the right default.

PR #500, opened 2026-07-21
tests: remove 12 stray test binaries from dev — every PR's CI dies at test_e8_kernel (EACCES)

ZacharyZcR:
#368's conflict-resolution merge (cc7461f) accidentally committed 12 built test binaries under `c/tests/`. They land in every fresh checkout as mode-100644 files (no exec bit), make sees an up-to-date target and **skips compiling them**, and `run_tests.py` then dies with `PermissionError: [Errno 13] ... 'tests/test_e8_kernel'` — currently failing the Engine/linux/macos jobs on every PR that merges with dev (see #434's latest run for an example).

- deletes the 12 binaries (they regenerate with `make check`)
- replaces the one-by-one test-binary entries in `.gitignore` — which had fallen behind the test suite, which is how these slipped in — with a pattern that ignores the built outputs and keeps the `.c/.cu/.mm/.py` sources


JustVugg:
Merged — thank you, and sorry: those binaries came in through merges I made (test_e8_kernel arrived with #465), and I fixed the same class of problem for #421 without generalising it. Your pattern-based ignore is the right fix precisely because it can't be forgotten next time. This was blocking every open PR's CI, so it goes in ahead of the release.

Issue #524, opened 2026-07-22
Feature request: fp8-e4m3-b128 passthrough weight format — zero loss at int8 cost (repack, not requantize)


monotophic:
# Feature request: format `fp8-e4m3` passthrough

*Authored by Fable 5 in Claude Code, analysis in partnership with @monotophic.*

**Ordinal to be assigned by the maintainer at merge.** During development we
use a private, non-public 100-block ordinal (`fmt=100` in our branch) purely
as an internal enum value — it is never written to a container and never
claimed as the public identity of this format. See the companion registry
draft, `FORMATS_registry_draft.md`, for why: two proposals (this one and
[#465](https://github.com/JustVugg/colibri/pull/465)) independently reached
for the same literal number this week, which is what prompted both that
draft and this naming convention. (#465 has since **merged** — fmt=6 is now
the E8/IQ3 container; the next free public ordinal is 7, but per the
convention above we leave the assignment to the maintainer.)

## TL;DR

Add a format that stores Z.ai's own public GLM-5.2-FP8 checkpoint bytes
**as-is** — no dequant/requantize round trip — at the same 8-bits/weight
streaming cost as our existing int8 format, with zero additional
quantization loss. This arrives **with working, tested code** on a
validated branch, not as a design-only proposal.

## Motivation

- **Zero loss at int8 cost — because it's a repack, not a requantize.** The
  FP8 (e4m3) bytes in Z.ai's published GLM-5.2-FP8 checkpoint are already
  the highest-fidelity form the model ships in. Our existing `int8-row`
  format quantizes fresh from a dequantized source at the same 1 byte/weight
  cost; this format instead copies the FP8 bytes as-is and only renames/
  reshapes the scale sidecar — no requantization step, no requantization
  loss, and no extra quantization compute spent producing the container.
  Verified byte-identical against `torch.float8_e4m3fn`'s own storage (RAN,
  `tests/test_fp8_repack.py`).
- **Precedent.** [#399](https://github.com/JustVugg/colibri/pull/399) (FP8
  KV-cache) already established FP8 as a format family this project is
  willing to support; this proposal is the adjacent case for FP8 *weights*.
- **Role pairing, backed by measurement.** We ran a full-corpus per-kind
  int4-sensitivity census against real GLM-5.2 weights (flat-int4 relative
  RMS, ranked worst-to-best across all 14 tensor kinds). The result: shared-
  expert gate/up, `o_proj`, dense-MLP, and general attention projections
[...]

PR #488, opened 2026-07-21
Metal: fix grid-size truncation in coli_metal_gemm at large prefill (long-context corruption)

RDouglasSharp:
# Metal: fix grid-size truncation in `coli_metal_gemm` at large prefill (long-context corruption)

## Summary

On the Metal backend, dense GEMMs with a very large output (`S × O` elements) are dispatched as a
**single** grid. Past a device threadgroup-grid limit the grid is silently truncated — the tail
output rows are **never computed** and keep whatever was in the buffer. In the engine that buffer is
reused scratch, so the stale contents are nondeterministic garbage → **corrupted generation at long
context**. This fixes it by chunking the dispatch so no single launch exceeds a safe grid size. No
kernel change; output is bit-exact vs. CPU.

## Motivation — this is what makes Metal + OpenCode/agent backends usable

colibri on Metal (`COLI_METAL=1`) is currently **unusable as an OpenCode / coding-agent backend**
because of this bug. Agent front-ends prepend a large fixed system header — tool declarations, rules,
worked examples — that is **several thousand tokens before the user types anything** and cannot be
trimmed. That header alone requires a large context (we run `CTX=40960` just to fit it) and drives a
big prefill on the *first* message of *every* session. That prefill is exactly the regime that
crosses the grid limit, so the GPU path returns garbage from the first token, while the CPU path is
correct but far slower. In other words: without this fix, Metal + OpenCode simply does not work; with
it, the GPU path produces coherent output and well-formed tool calls at real agentic context sizes.
(Nobody upstream hit this because Metal validation uses tiny 8–13-token prompts — see "Why it evaded
detection".)

## Symptom / impact

- Long prompts on `COLI_METAL=1` produce corrupted output (dropped/blank tokens, degrading to token
  salad); the **identical prompt is correct on CPU** (`COLI_METAL=0`) and correct on Metal at short
  lengths.
- Because the trigger is the fixed agent header, it hits the **first turn of every OpenCode session**
  on Metal — i.e. the GPU backend is effectively dead for coding-agent workloads until this lands.

## Root cause

`matmul_qt_ex` routes dense matmuls with `S ≥ COLI_METAL_GEMM_MIN` to `coli_metal_gemm`, which
dispatches `NT = S*O` output elements as one grid of `(NT+3)/4` threadgroups. The largest projection
(`kv_b`, O≈28672) crosses the device grid limit at large `S`: on M-series, `kv_b` at S=4376
[...]

PR #528, opened 2026-07-22
fp8-e4m3-b128 passthrough weight format — CPU + Metal + repack tool (#524)

monotophic:
*Authored by Fable 5 in Claude Code (We/our), analysis in partnership with @monotophic (me/my).*

Revised per review (2026-07-23): collision policy inverted; see the comment of this date for the full revision + evidence.

# fmt=7: native FP8-e4m3 passthrough — CPU path, Metal kernel, repack tool

Branch: `fmt7/fp8-passthrough`. Base: dev `9baae9b`. Closes the FP8-passthrough
Feature Request; implements the format the maintainer assigned `fmt=7` on
#524. Companion PR (`fmt7/stamp-registry`, stacked on this one) adds the
container metadata stamp and the `docs/FORMATS.md` registry — split out per
the maintainer's own scoping request, so this PR is CPU path + repack tool +
collision/refusal logic + the Metal kernel only.

## What / why

Z.ai ships GLM-5.2's release-native checkpoint in FP8-e4m3
(`zai-org/GLM-5.2-FP8`) — to our knowledge there is no public bf16/fp16
upstream of it. Every existing path into this engine from that checkpoint
first **dequantizes to f32 and requantizes** to a lossy format (int4/int3),
paying a real second quantization error on top of whatever the FP8 already
cost. This PR adds a different path for tensors where the extra fidelity is
worth the extra 8 bits/weight over int4: **repack, not requantize**. The
repack tool copies the fp8 weight bytes AS-IS (`.view(torch.uint8)`, a pure
bit-reinterpret, verified byte-identical against `torch.float8_e4m3fn`'s own
storage) and only renames/reshapes the `weight_scale_inv` sidecar to the
engine's `.qs` convention. Same 8 bits/weight streaming cost as the FP8
source, **zero additional quantization loss** — a resident/quality-core tier
sitting above int4/int3 in fidelity, below f32 in size.

## Selection: which tensor roles get fmt=7

`tools/repack_fp8_passthrough.py` reuses `convert_fp8_to_int4.classify()` —
the engine's existing tensor-role taxonomy — rather than inventing a new one:

| role (`classify()` kind) | selected for fmt=7? | why |
|---|---|---|
| shared expert (`sh`) | yes | resident, always-active — fidelity matters every token |
| `o_proj` (`o`) | yes | resident attention output projection |
| other attention projections (`attn`, incl. `q_a_proj`) | yes | resident |
| dense-MLP first layers (`dmlp`) | yes | resident |
| generic resident fallback (`q`) | yes | catches anything else resident-shaped |
| `kv_b_proj` (`kvb`) | **no** | resident, but no fmt=7 case in the CPU/CUDA MLA-absorb path — see below |
| routed experts (`x`) | no | stays on the existing int4-g64 streaming path (58k+ tensors — the streaming-cost tradeoff is different for a per-token top-K subset) |
| routers/norms/biases (`f32`) | no | never FP8 in the source, always f32 |
| embed/lm_head (`io`) | no | BF16 in source, never FP8 |

Blending selected fmt=7 residents with int4-g64 routed experts in the same
loadout directory is separate, deferred "loadout index-rewrite blending"
work — out of scope here; the tool only ever emits a `fmt=7` container for
the roles in the table.

## Design

**Decode LUT** (`quant.h`'s `E4M3_LUT`, 256 entries): a static compile-time
table, cross-checked byte-for-byte against `torch.float8_e4m3fn` for all 256
codes including both signs of zero, subnormals, and the OCP E4M3-FN NaN
encoding (`exp==0xF && mant==0x7` only — `exp==0xF` is NOT reserved for
infinity in this format). Static rather than lazily initialized because
`matmul_fp8` runs inside `#pragma omp parallel for` and a lazy-init global
would race under concurrent first use. The Metal `mm_gemv` shader's `fmt==7`
branch does the identical bit-decode in-kernel (no LUT texture — BW-bound
kernel, decode ALU is free) and is verified to byte-match the CPU LUT via an
exhaustive 256/256-code GPU test (see Metal validation below).

**Block scales**: one `f32` per 128×128 block of the `[O,I]` weight matrix
(`ceil(O/128)*ceil(I/128)` entries, row-major block order), not one per
output row. Dequant is `w[o,i] = e4m3_decode(byte) * scale[o/128, i/128]`
(multiply, not divide) — matches `tools/convert_fp8_to_int4.py`'s own
`dequant()`, the authoritative reference for how Z.ai's checkpoints read.

**Scale encoding is a declared property of the format, not a hardcoded
constant.** This was an open design question we'd have gotten wrong by
defaulting to "f32, full stop" — the maintainer's own finding, surfaced on
#524, is what settled it: **DeepSeek-V4 ships this identical weight geometry
(FP8 E4M3, 128×128 blocks) with UE8M0 scales instead of f32** — one byte per
block, a power-of-two exponent, dtype `F8_E8M0`. Concretely, in this PR:

- fmt=7's scale encoding is named explicitly in the QT struct comment and
  `qt_resolve_fmt`'s own documentation; **f32 is the implemented value**
  (what Z.ai's GLM-5.2-FP8 checkpoints ship, and what this PR's CPU kernel,
  Metal kernel, and repack tool all read/write).
- The resolver **recognizes the UE8M0 byte signature and refuses by name**
  rather than misreading it: the two encodings differ cleanly in
  scale-sidecar arithmetic (`ceil(O/128)*ceil(I/128)*4` bytes for f32 vs
  `*1` for UE8M0), a real, distinct discriminator (never coincidentally
  equal for any `O,I >= 1`, since one is exactly 4× the other). An
  unstamped container with a 1-byte-per-block sidecar gets *"fp8-e4m3-b128
  with ue8m0 scales recognized but not implemented; only f32 block scales
  are supported in this build"* instead of garbage.
- UE8M0 decode itself is invited follow-up work, same posture this project
  already applies to the CUDA tier: whoever reaches DeepSeek-V4 first (that
  may be us, later) lands it into a seam that already exists rather than a
  near-duplicate format. The MXFP4 routed-expert question (E2M1 + UE8M0 g32)
  raised alongside this finding is a separate format and a separate
  conversation — but the same scale-encoding-as-property mechanism serves
  it too.

## The collision/refusal analysis

`qt_resolve_fmt` infers a tensor's format from byte arithmetic alone — the
container carries no format ordinal. fmt=7's weight bytes are
byte-identical to fmt=1 (int8): `O*I` raw bytes either way. That creates
**three** genuine byte-collision shapes this PR's resolver must catch and
refuse rather than silently misread:

1. **fmt=1 vs fmt=7 (THE DESIGN LANDMINE).** For small `O` (≤128) and/or
   small `I`, `ceil(O/128)*ceil(I/128)` can equal `O` exactly (e.g. `O=1,
   I≤128`; `O=2, I∈[129,256]`; `O=256, I∈(16256,16384]`) — a scale array
   that satisfies BOTH fmt=1's per-row convention (`O*4` bytes) and fmt=7's
   per-block convention (`ceil(O/128)*ceil(I/128)*4` bytes) at once.
   Guessing either way silently corrupts every weight. Refused.
2. **fmt=6 vs fmt=7 at I=98 (SECOND DESIGN LANDMINE).** Upstream's own
   fmt=6 (E8/IQ3, #465) tags itself with `ns==4 && nb==O*e8_rowbytes(I)`,
   checked unconditionally before any fmt=7-aware logic runs.
   `e8_rowbytes(I) = ceil(I/256)*98` is the constant 98 for every `I` in
   `(0,256]`, so that check collapses to `nb==O*98` — which is ALSO fp8's
   raw weight-byte count (`O*I`) at exactly `I=98`. At that same `I=98`, a
   single-block fp8 tensor (`O≤128`, f32 scales) carries exactly one
   `f32` block scale (`ns==4`, the same tag fmt=6 uses); a **four-block**
   fp8 tensor (`O∈(384,512]`, UE8M0 scales) also lands at `ns==4` (4
   blocks × 1 byte/block). A straight, un-reconciled rebase onto post-#465
   dev would silently misread either as E8/IQ3-lattice-decoded garbage —
   demonstrated fail-before during development (a standalone probe at
   `O=64, I=98` returned `fmt=6` with no error before the fix). Refused
   after the fix, for both the f32-scaled and UE8M0-scaled variants.
3. **`O==1` triple collision.** At `O=1, I=98`, fmt=1's own per-row tag
   (`O*4`) is ALSO 4 — a genuine three-way ambiguity (fmt=1 / fmt=6 /
   fmt=7) at one shape. Refused.

**This PR's refusal posture is unconditional**: it carries no container
metadata stamp (that's the companion PR), so every one of these three
collisions refuses regardless of the tensor's true identity — "nothing real
is lost: no real checkpoint tensor has these shapes" (the maintainer's own
framing, #524). The companion PR's stamp is the only thing that can turn a
refusal into a resolution, and even then only for the f32-scaled candidates
— a stamp can confirm a WEIGHT format, it can't grant a decoder this build
doesn't have, so the UE8M0 corners of collisions 2 and 3 stay refused even
under a (correct) stamp.

Test coverage: `tests/test_fp8_load.c`'s Part A (15 cases: golden paths, 6
degenerate collisions, 4 boundary-adjacent non-collisions) + Part A2 (the
fmt=6 collision at `I=98`, both boundary shapes, the three-way `O=1`
case, plus 2 non-colliding fmt=6 regression fixtures proving dev's own
E8/IQ3 inference stays unbroken) + Part A3 (2 UE8M0-signature refusal
cases, including the small-O collision variant).

## `kv_b_proj` exclusion

`kv_b_proj` (kind `kvb`) is a valid resident tensor `classify()` would
otherwise select, and it's deliberately excluded anyway: colibri.c's
MLA-absorption CPU path (`qt_addrow`/`qt_matvec_rows`, the always-available
fallback whenever the Metal fused decode kernel isn't used) has no fmt=7
case and would silently misread fp8 bytes as int2-packed data; the CUDA
absorb kernels (`coli_cuda_attention_absorb`/`_kvdev` in `backend_cuda.cu`)
are similarly int4-specific. Repacking `kv_b_proj` to fmt=7 needs that
CPU+CUDA absorb-path work first — out of scope for this PR, so the tool
refuses to produce a container the engine cannot safely read
(`tests/test_fp8_repack.py::test_kv_b_proj_excluded` is the regression
guard).

## Metal validation record

**This is the evidence we'd ask you to merge on, absent Apple hardware on
your side** — every fmt=7 code path exercised on real M-series Metal, not
just described:

- **LUT-through-GPU exactness: 256/256 byte codes**, run through the
  ACTUAL compiled `mm_gemv` shader (not a CPU-only spot check) — proves the
  in-kernel bit-decode agrees with `quant.h`'s reference LUT for every
  possible byte value, including both zeros, all subnormals, and the NaN
  encoding.
- **Adversarial shape suite**: block edges (`O,I` both non-multiples of
  128), degenerate 1×1 (single sub-block), a non-square block grid
  (`nblkO=3, nblkI=48` — a stride-audit case that would surface a
  transposed or swapped block index), worst observed relative error
  `3.0e-8` to `7.0e-8` across all cases (tolerance `1e-4`).
- **`S>1` gate test**: the spec-shaped `[2048,6144]` case run at `S=4`
  (batched decode), not just `S=1`.
- **Two independent negative-gate tests**: `coli_metal_gemm` explicitly
  refuses fmt=7 (`rc=0`, CPU-fallback signal) — the large-batch sync GEMM
  path was never in scope for this format. `coli_metal_moe_block`/
  `moe_submit`'s existing `fmt != 1 && fmt != 2` allowlist already refuses
  fmt=7 before any pointer is dereferenced — proven BY TEST with
  deliberately-invalid `0xdeadbeef` weight/scale pointers that must never
  be read, not left as an artifact of the allowlist happening not to
  include 7 (yet).
- **Both `COLI_METAL_RESSET` states**: `make metal-test` full green under
  `COLI_METAL_RESSET=0` AND `COLI_METAL_RESSET=1` (residency-set path
  confirmed active via `"[METAL] residency-set: on"`), all 11 fp8 GPU
  lines identical in both.

Full case-by-case output (RESSET=0; RESSET=1 is line-identical apart from
the residency-set banner):

```
Metal fmt=7 native FP8-e4m3 passthrough tests:
  fp8 LUT exactness (256/256 codes via GPU kernel) 0/256 mismatches  ok
  fp8 gate/up-shaped O=2048 I=6144 (spec example) S=1 worst_rel=3.01e-08 (I=6144 O=2048 nblkO=16 nblkI=48 S=1)  ok
  fp8 down-shaped O=6144 I=2048 S=1          worst_rel=3.63e-08 (I=2048 O=6144 nblkO=48 nblkI=16 S=1)  ok
  fp8 gate/up-shaped O=2048 I=6144 S=4       worst_rel=4.05e-08 (I=6144 O=2048 nblkO=16 nblkI=48 S=4)  ok
  fp8 block edges: O,I both non-mult-128     worst_rel=5.95e-08 (I=200 O=130 nblkO=2 nblkI=2 S=2)  ok
  fp8 block edges: O just over 128, I exact  worst_rel=4.65e-08 (I=128 O=129 nblkO=2 nblkI=1 S=1)  ok
  fp8 block edges: O exact, I just over 128  worst_rel=6.93e-08 (I=129 O=128 nblkO=1 nblkI=2 S=1)  ok
  fp8 degenerate 1x1 (single sub-block)      worst_rel=5.87e-08 (I=1 O=1 nblkO=1 nblkI=1 S=1)  ok
  fp8 non-square block grid nblkO=3 nblkI=48 (stride audit) worst_rel=3.33e-08 (I=6144 O=384 nblkO=3 nblkI=48 S=3)  ok
  fp8 GEMM entry explicitly gated off (coli_metal_gemm refuses) rc=0 (expect 0/CPU-fallback)  ok
  fp8 MB_BUILD/moe_submit entry gated off (shared-expert fmt=7 hazard) rc=0 (expect 0/CPU-fallback)  ok
```

Every pre-existing Metal case (int8/int4/int2/f32, moe_block, gemm S=64,
fused attention, top-8 select) unaffected, in the same run.

## Durable vs. current-state (review map)

Per this project's own convention for delineating what's structural from
what has a foreseeable resolution:

| | claim | durable / current-state | revisit trigger |
|---|---|---|---|
| 1 | fmt=7 weight-byte layout (`O*I` raw e4m3), E4M3 decode LUT, block-scale indexing | **durable** | none — fixed by the OCP E4M3-FN spec and this format's own definition |
| 2 | fmt=1-vs-fmt=7 and fmt=6-vs-fmt=7 collision/refusal logic | **durable** | none — the ambiguous shapes are structural (byte-arithmetic facts), not tunable |
| 3 | UE8M0 scale encoding: recognized, refused, not decoded | **current-state** | a UE8M0 CPU+Metal decoder landing into this seam (`qt_resolve_fmt`'s fmt=1 candidate branch + a new dispatch case) — invited follow-up, may be us |
| 4 | `kv_b_proj` excluded from fmt=7 selection | **current-state** | CPU (`qt_addrow`/`qt_matvec_rows`) + CUDA MLA-absorb kernels gaining an fmt=7 case |
| 5 | "refuse unconditionally" at all three collision shapes (no stamp) | **current-state**, resolved by the companion PR | `fmt7/stamp-registry` merging — a stamped, unambiguous container resolves instead of refusing (except the UE8M0 corners, which stay current-state per row 3) |
| 6 | Metal validation record below | checkable coordinates | measured 2026-07-22, macOS 26.5.2 (build 25F84), Apple M5 Max (MacBook Pro, Mac17,7), engine base dev `9baae9b` |

## Behavioral contract

This PR was revised against a written contract; every invariant below is a
registered claim (stated before the revision was built), each with its
decisive evidence:

| Invariant | Capstone evidence |
|---|---|
| Every container that loads on dev loads identically here (resolutions, notices, output) | Differential matrix vs. current dev: **byte-identical temp-0 generated text** on the stock int8-MTP container and an int8-resident mint |
| No byte pattern readable today changes meaning; ambiguity favors the incumbent | `[6144,16384]` int8-row (GLM-5.2's real `o_proj`): refused by the previous head (`exit 1`, quoted in-thread), **loads as fmt=1 here** |
| The writer can never emit what the reader would mis-resolve | Repack refusal demonstrated at a real ambiguous shape (`[2048,16384]`); round-trip load verified on non-ambiguous output |
| fmt=7 decode is exact OCP E4M3-FN — passthrough, no approximation | Independently derived oracle (from the OFP8 spec text): **bit-exact at all 16,900 positions** of a partial-block (130×130) tensor |
| Failure is loud: every refusal names its condition | New: `qt_addrow`/`qt_matvec_rows` refuse unhandled formats by name (previously: SIGSEGV for fmt=7, silent misread for fmt=6 — found by the acceptance suite, fixed here) |
| Collision family is enumerated, not sampled | In-tree census over every real tensor shape: `o_proj` is the **only** ambiguous family (79 instances/container) |

<details>
<summary>Full evidence matrix (instruments and results per requirement)</summary>

- **Compatibility (I1)**: old-binary/new-binary differential over both real
  containers, per-run usage-state control, telemetry-excised output
  comparison; plus same-binary control run proving the method's
  discrimination. Result: identical loads, per-tensor resolutions, notices,
  and generated text, all four comparisons.
- **Collision policy (E4/I2)**: reader inversion verified at source and
  behaviorally (collision-family fixtures at 4 real shapes + 1
  non-collision control resolve fmt=1); the three previously-refusing test
  shapes plus four degenerate cases now assert fmt=1; census tool
  regenerable upstream (`c/tools/fp8_collision_census.py`).
- **Exact decode (I4)**: spec-blind acceptance suite with an independently
  derived E4M3-FN oracle (derivation caught an internal contradiction in
  the OFP8 spec's own bias statement, resolved via its worked examples);
  bit-exact on NaN-free data; NaN weights verified to poison their row
  through the dot product (IEEE-correct propagation, 130/130 rows).
- **Wire-path sizing**: `qt_scale_bytes()` wired via shared
  `qt_wire_split()`; seam test intercepts actual `mlock`/`munlock` calls —
  a sites-only regression to `O*4` fails 4 named assertions; fmt=6's
  4-byte-tag case corrected while making the helper authoritative.
- **Loud failure (I6)**: refusal tests for fmt=6/7 through both absorb-path
  functions (previously crash/silent); repack tool refuses on zero
  repack-target tensors (previously silent success).
- **Gates**: `make check`; METAL=1 zero warnings; `metal-test` under both
  `COLI_METAL_RESSET` states; per-commit bisect builds; suppression-
  stripped `-Wunused-function` build; POSIX-construct check (Windows CI
  pattern) on all new tests.

Deeper artifacts (specific transcripts, census tables, suite design)
available on request.
</details>

## Commits in this PR

1. `fp8: CPU read path + matmul_fp8, as fmt=7 (public ordinal, #524)` —
   quant.h's LUT/decode/kernel, colibri.c's byte-accounting and THE DESIGN
   LANDMINE refusal, the UE8M0 recognition seam, `qt_resolve_fmt` in its
   plain 6-argument (no stamp) form.
2. `fp8: Metal mm_gemv kernel for fmt=7 native FP8-e4m3 passthrough` — the
   GPU kernel and its full adversarial test suite, including the
   `moe_submit` gate regression test.
3. `fp8: repack tool (Z.ai FP8 shards -> byte-preserved fmt=7 container)` —
   the writer side, no metadata stamp.
4. `fp8: end-to-end test, real repack tool -> real C loader` — the real
   round trip, closing the last coverage gap.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -214,7 +214,7 @@ else
 PYTHON    ?= python3
 endif
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_mirror$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_temp_env$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_int3$(EXE) tests/test_int3_load$(EXE) tests/test_fp8_passthrough$(EXE) tests/test_fp8_load$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_router_nan$(EXE) tests/test_pipe_block$(EXE) tests/test_e8_kernel$(EXE) tests/test_rans$(EXE) tests/test_corpus_draft$(EXE) tests/test_pilot_ring$(EXE)
+TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_mirror$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_temp_env$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_int3$(EXE) tests/test_int3_load$(EXE) tests/test_fp8_passthrough$(EXE) tests/test_fp8_load$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_router_nan$(EXE) tests/test_pipe_block$(EXE) tests/test_e8_kernel$(EXE) tests/test_qt_addrow$(EXE) tests/test_rans$(EXE) tests/test_corpus_draft$(EXE) tests/test_pilot_ring$(EXE)
 ifneq (,$(LINUX))
 TEST_BINS += tests/test_uring$(EXE)
 endif
@@ -546,6 +546,9 @@ tests/test_fp8_passthrough$(EXE): tests/test_fp8_passthrough.c quant.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_fp8_load$(EXE): tests/test_fp8_load.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_qt_addrow$(EXE): tests/test_qt_addrow.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_logit_nan$(EXE): tests/test_logit_nan.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
--- a/c/Makefile
+++ b/c/Makefile
@@ -214,7 +214,7 @@ else
 PYTHON    ?= python3
 endif
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_mirror$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_temp_env$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_int3$(EXE) tests/test_int3_load$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_router_nan$(EXE) tests/test_pipe_block$(EXE) tests/test_e8_kernel$(EXE) tests/test_rans$(EXE) tests/test_corpus_draft$(EXE) tests/test_pilot_ring$(EXE)
+TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_mirror$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_temp_env$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_int3$(EXE) tests/test_int3_load$(EXE) tests/test_fp8_passthrough$(EXE) tests/test_fp8_load$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_router_nan$(EXE) tests/test_pipe_block$(EXE) tests/test_e8_kernel$(EXE) tests/test_rans$(EXE) tests/test_corpus_draft$(EXE) tests/test_pilot_ring$(EXE)
 ifneq (,$(LINUX))
 TEST_BINS += tests/test_uring$(EXE)
 endif
@@ -540,6 +540,12 @@ tests/test_int3$(EXE): tests/test_int3.c colibri.c st.h uring.h json.h tok.h tok
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_int3_load$(EXE): tests/test_int3_load.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_fp8_passthrough$(EXE): tests/test_fp8_passthrough.c quant.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
+tests/test_fp8_load$(EXE): tests/test_fp8_load.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_logit_nan$(EXE): tests/test_logit_nan.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h
--- a/c/backend_metal.h
+++ b/c/backend_metal.h
@@ -29,9 +29,15 @@ int  coli_metal_mem_info(size_t *used_bytes, size_t *total_bytes);
 /*
  * y[S,O] = (x[S,I] @ W[O,I]^T) * scale[o]. fmt=4 (grouped int4) instead folds a
  * PER-GROUP scale into the accumulation -- see the shader comment in backend_metal.mm.
+ * fmt=7 (fp8 passthrough -- see colibri.c) likewise folds its per-128x128-block
+ * scale into the accumulation.
  * fmt matches QT in colibri.c: 0=f32, 1=int8, 2=int4(packed), 3=int2(packed),
  * 4=int4(packed, same layout as 2)+per-group scale (gs = group size along I; scale
- * array is [O, ceil(I/gs)] floats instead of [O]). gs is ignored for fmt!=4 (pass 0).
+ * array is [O, ceil(I/gs)] floats instead of [O]),
+ * 7=fp8-e4m3 (one raw byte/element, same layout as fmt=1) + per-128x128-
+ * block scale (scale array is [ceil(O/128),ceil(I/128)] floats; no group-size
+ * parameter needed -- the block is a fixed 128x128, not caller-configurable).
+ * gs is ignored for fmt!=4 (pass 0).
  * The first successful call wraps W and its scales in GPU-visible buffers (sized from
  * fmt+gs); later calls reuse them (weights are assumed stable at the same address).
  * Returns 1 on success, 0 if Metal is unavailable or fmt is invalid.
--- a/c/backend_metal.h
+++ b/c/backend_metal.h
@@ -29,12 +29,12 @@ int  coli_metal_mem_info(size_t *used_bytes, size_t *total_bytes);
 /*
  * y[S,O] = (x[S,I] @ W[O,I]^T) * scale[o]. fmt=4 (grouped int4) instead folds a
  * PER-GROUP scale into the accumulation -- see the shader comment in backend_metal.mm.
- * fmt=7 (fp8 passthrough -- see colibri.c) likewise folds its per-128x128-block
+ * fmt=8 (fp8 passthrough -- see colibri.c) likewise folds its per-128x128-block
  * scale into the accumulation.
  * fmt matches QT in colibri.c: 0=f32, 1=int8, 2=int4(packed), 3=int2(packed),
  * 4=int4(packed, same layout as 2)+per-group scale (gs = group size along I; scale
  * array is [O, ceil(I/gs)] floats instead of [O]),
- * 7=fp8-e4m3 (one raw byte/element, same layout as fmt=1) + per-128x128-
+ * 8=fp8-e4m3 (one raw byte/element, same layout as fmt=1) + per-128x128-
  * block scale (scale array is [ceil(O/128),ceil(I/128)] floats; no group-size
  * parameter needed -- the block is a fixed 128x128, not caller-configurable).
  * gs is ignored for fmt!=4 (pass 0).
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -13,7 +13,7 @@
 // nibble layout as fmt=2, but scale is PER-GROUP (gsz elements of I), buffer(1) holds
 // [O, ceil(I/gsz)] floats indexed scale[o*ng+g] -- so unlike fmt 1-3, the group scale is
 // folded into the accumulation itself (see the fmt==4 branch), not applied once at the end.
-// fmt=7 (native FP8-e4m3 passthrough -- see colibri.c): raw byte
+// fmt=8 (native FP8-e4m3 passthrough -- see colibri.c): raw byte
 // layout identical to fmt=1 (one e4m3 byte per element, no packing), but the scale is
 // per-128x128 BLOCK of [O,I] -- buffer(1) holds [ceil(O/128),ceil(I/128)] floats indexed
 // scale[(o/128)*nblkI+i/128]. Folded into acc like fmt=4's per-group scale, for the
@@ -85,7 +85,7 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
         acc += float(int(b>>4)-8) * xr[i+1] * sc1;
       }
     }
-  } else if (fmt == 7) {                            // fp8 e4m3 passthrough: one raw byte per
+  } else if (fmt == 8) {                            // fp8 e4m3 passthrough: one raw byte per
                                                      // element (like fmt=1), scale per 128x128
                                                      // block folded into acc (like a grouped fmt).
     int nblkI = (I + 127) / 128;
@@ -111,9 +111,9 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
     for (int i = I8*8 + slane; i < I; i += 32) acc += wr[i] * xr[i];
   }
   acc = simd_sum(acc);
-  // fmt==4 (per-group) and fmt==7 (per-block) already folded their scale into acc
+  // fmt==4 (per-group) and fmt==8 (per-block) already folded their scale into acc
   // above -- do not scale again.
-  if (slane == 0) y[row] = (fmt == 4 || fmt == 7) ? acc : acc * scale[o];
+  if (slane == 0) y[row] = (fmt == 4 || fmt == 8) ? acc : acc * scale[o];
 }
 
 // Batched bindless expert GEMV: each row gr belongs to expert erow[gr], whose weight and
@@ -477,20 +477,20 @@ static size_t fmt_bytes(int fmt, int I, int O) {
   if (fmt == 2) return (size_t)O * ((I+1)/2);
   if (fmt == 3) return (size_t)O * ((I+3)/4);
   if (fmt == 4) return (size_t)O * ((I+1)/2);   // grouped int4: identical packed-nibble layout to fmt=2
-  if (fmt == 7) return (size_t)O * I;           // fp8 e4m3: one raw byte/element, same as fmt=1
+  if (fmt == 8) return (size_t)O * I;           // fp8 e4m3: one raw byte/element, same as fmt=1
   return (size_t)O * I * sizeof(float);
 }
 // Grouped-int4 (fmt=4) scale-array size: one f32 per gsz-element group, per row -> O*ceil(I/gsz).
-// fp8 (fmt=7) scale-array size: one f32 per 128x128 BLOCK -> ceil(O/128)*ceil(I/128) (2D,
+// fp8 (fmt=8) scale-array size: one f32 per 128x128 BLOCK -> ceil(O/128)*ceil(I/128) (2D,
 // not per-row -- quant.h isn't included here, so the ceil-div is inlined rather than sharing
 // colibri.c's qt_scale_bytes/quant.h's fp8_nblk). The block is a fixed 128x128, so gs is
-// ignored for fmt==7. f32 is this build's implemented scale
-// ENCODING for fmt=7 (see quant.h/colibri.c) -- this file has no reason to know that a
+// ignored for fmt==8. f32 is this build's implemented scale
+// ENCODING for fmt=8 (see quant.h/colibri.c) -- this file has no reason to know that a
 // UE8M0 encoding exists at all: qt_resolve_fmt refuses it on the CPU read path before any
 // tensor in that encoding could ever reach this Metal-side sizing helper.
 static size_t fmt_scale_bytes(int fmt, int I, int O, int gs) {
   if (fmt == 4) return (size_t)O * ((I + gs - 1) / gs) * sizeof(float);
-  if (fmt == 7) return (size_t)((O + 127) / 128) * (size_t)((I + 127) / 128) * sizeof(float);
+  if (fmt == 8) return (size_t)((O + 127) / 128) * (size_t)((I + 127) / 128) * sizeof(float);
   return (size_t)O * sizeof(float);
 }
 
@@ -658,9 +658,9 @@ extern "C" int  coli_metal_mem_info(size_t *used, size_t *total) {
 extern "C" int coli_metal_matmul(ColiMetalTensor **tp, float *y, const float *x,
                                  const void *weights, const float *scales,
                                  int fmt, int S, int I, int O, int gs) {
-  /* fmt==7 (fp8 passthrough) is an explicit allow-list entry, not folded into the 0..4
+  /* fmt==8 (fp8 passthrough) is an explicit allow-list entry, not folded into the 0..4
    * contiguous range check below: it is not adjacent to it. */
-  if (!g_dev || fmt < 0 || (fmt > 4 && fmt != 7)) return 0;
+  if (!g_dev || fmt < 0 || (fmt > 4 && fmt != 8)) return 0;
   @autoreleasepool {
     ColiMetalTensor *t = *tp;
     if (!t) {
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -13,6 +13,16 @@
 // nibble layout as fmt=2, but scale is PER-GROUP (gsz elements of I), buffer(1) holds
 // [O, ceil(I/gsz)] floats indexed scale[o*ng+g] -- so unlike fmt 1-3, the group scale is
 // folded into the accumulation itself (see the fmt==4 branch), not applied once at the end.
+// fmt=7 (native FP8-e4m3 passthrough -- see colibri.c): raw byte
+// layout identical to fmt=1 (one e4m3 byte per element, no packing), but the scale is
+// per-128x128 BLOCK of [O,I] -- buffer(1) holds [ceil(O/128),ceil(I/128)] floats indexed
+// scale[(o/128)*nblkI+i/128]. Folded into acc like fmt=4's per-group scale, for the
+// same reason: it is not constant across one output row. e4m3->float is bit manipulation
+// in-kernel (sign/exp/mantissa, OCP E4M3-FN: exp==0 subnormal, exp==0xF&&mant==0x7 is the
+// only NaN code) -- BW-bound kernel, decode ALU is free, no LUT texture. Must byte-match
+// quant.h's E4M3_LUT/e4m3_decode (the CPU reference) including the NaN policy, which is
+// why the metal-test suite runs all 256 byte codes through this exact kernel path (not
+// just spot values) -- see run_fp8_lut().
 static const char *SHADER = R"METAL(
 #include <metal_stdlib>
 using namespace metal;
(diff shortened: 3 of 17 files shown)


monotophic:
Working on the Windows / MinGW fail. Will update with a fix shortly.

monotophic:
Fixed and ready for review

JustVugg:
I have to block this one, and I'm sorry to do it on work this careful — but the refusal logic fires on a shape **GLM-5.2 actually has**, so merging it can stop the engine from loading real containers.

## The collision includes GLM-5.2's own `o_proj`

`self_attn.o_proj.weight` loads as `[D, H*v_head]` = **`[6144, 16384]`** on GLM-5.2. Verified independently:

```
per-row int8 scale bytes : O*4              = 6144*4      = 24576
per-128x128-block bytes  : ceil(6144/128) * ceil(16384/128) * 4
                         = 48 * 128 * 4                   = 24576   <-- identical
```

`48 * 128 == 6144 == O` exactly, so a **valid, pre-existing int8-row** `o_proj` satisfies both geometries and hits THE DESIGN LANDMINE — `exit(1)` at model load. The general family is any int8 tensor where `O == ceil(O/128)*ceil(I/128)`; `[6144,16384]` is not a contrived corner, it is the shape in production.

This is reachable through this repo's own tooling: `--ebits` is documented as *"default 4; 8 per --mtp/--indexer"* and `--io-bits` defaults to 8, so int8 residents are written by our recommended configuration (the int8-MTP-head setup from #8). Please double-check the exact tensors your local converter emits at those settings — if `o_proj` lands as int8 in any documented configuration, this merge turns a working snapshot into one that refuses to start.

The reason CI is green: every C fixture uses `hidden_size=128`, where `o_proj` is `[128,128]` and cannot collide. **No test in the suite uses a real GLM-5.2 shape.**

## The fix I'd like, and why it's strictly safer

**Invert the collision policy.** On `is_row && is_blk`, resolve to **fmt=1** — the incumbent, already-on-disk, decodable format — and move the refusal to the **writer**: `_check_geometry()` in `repack_fp8_passthrough.py` refuses to emit an fmt=7 container at an ambiguous shape. Nothing existing breaks, and nothing this toolchain can produce is ever misread. Today the trade runs the other way: a decodable int8 tensor is refused to protect an encoding the reader can't decode anyway.

Two supporting notes, both verifiable in the branch:

- **`_check_geometry()` has no collision check at all.** Its docstring promises *"never anything that could coincide with O*4 by accident… refuses rather than silently emitting a container the engine might misread"*, but it only asserts the block counts. A `[6144,16384]` fp8 `o_proj` is repacked happily, then refused by the loader.
- **`test_fp8_load.c` enshrines the regression**: `expect_refuse(128,16384,…,512)`, `(256,16384,…,1024)`, `(384,16384,…,1536)` are all valid int8-row tensors (`ns == O*4`) asserted to be rejected. I'd want `[6144,16384]` added as an explicit **non**-refusal test.

## Three smaller things

1. **`qt_scale_bytes()` has no production callers.** The ~25-line comment says the two `qt_wire_mmap`/`qt_unwire_mmap` sites were fixed; both still compute `scale_b=(int64_t)t->O*4`. It only compiles clean because `-Wno-unused-function` is in CFLAGS. So the claimed fmt=4 expert-pinning fix isn't applied, and an fmt=7 tensor reaching `qt_wire_mmap` gets `weight_b` overshooting the allocation whenever `nblk > O`. Wire it or drop it, and fix the test comment that asserts otherwise.
2. **fmt=7 has no live GPU path.** The shader branch is correct — I checked the decode against the LUT, including subnormals, the single NaN code and `-0.0` — but the only caller of `coli_metal_matmul` in the tree is a test. Production dispatch (`matmul_qt_ex`, `coli_metal_gemm`, `moe_submit`) all exclude 7, and this PR adds `fmt!=7` guards to the fused paths. So adopting fmt=7 on Apple silicon means those tensors run the **scalar** `matmul_fp8` on CPU *and* disable the fused attention + shared-expert Metal kernel for every layer they appear in. That should be stated plainly, or `bind_gemv` wired to the new branch.
3. The guard comments say the shader *"would silently misread fp8 as f32"* — stale now that this same PR adds the `fmt==7` branch.

## What's right

The E4M3 LUT is correct OCP E4M3-FN (spot-checked `0x01`, `0x78`, `0x7E`, `0x7F`/`0xFF`, `0x80`). No OOB in `matmul_fp8`. `qt_bytes` fmt=7 prevents the AUTOPIN undercount. The fmt=6 hardening only alters behaviour at `I==98` and actually **fixes** a pre-existing misread of an int8 `[1,98]`. And the instinct throughout — refuse rather than guess — is the right one; the bug is which side of the ambiguity gets the benefit of the doubt.

Invert the collision policy, add the `[6144,16384]` regression test, and I'll take this. The ordinal `fmt=7` is yours.

monotophic:

You were right, and the original wasn't up to the standard we want our
contributions to meet — we've tightened our process around exactly the gap
you exposed, and re-verified this from written requirements rather than
implementation intent. Thank you for the catch; the `o_proj` analysis was
exact (confirmed on real containers on my side, including one where the
pushed head refused to load).

**The revision** (head `d1a6897`, rebased onto current `dev` @ `e4b5bf3` —
clean replay over #488 and the Inkling merge):

1. **Collision policy inverted as requested.** Reader: unstamped
   `is_row && is_blk` resolves to fmt=1, the incumbent. Writer:
   `_check_geometry()` refuses to emit fmt=7 at any shape where
   `O == ceil(O/128)*ceil(I/128)`. `[6144,16384]` is now an explicit
   non-refusal regression test; the three `expect_refuse` cases you flagged
   (plus four degenerate cases on the same inverted branch) assert fmt=1.
2. **The family is enumerated, not sampled**: a census tool
   (`c/tools/fp8_collision_census.py`) sweeps the predicate over every real
   tensor shape; on GLM-5.2, `self_attn.o_proj` `[6144,16384]` is the
   *only* ambiguous family (79 instances/container — every real O here is a
   multiple of 128, so the family reduces to real-O × I=16384).
3. **`qt_scale_bytes()` wired at both `qt_wire_mmap`/`qt_unwire_mmap` sites**
   via a shared `qt_wire_split()`, with a seam test that intercepts the
   actual `mlock`/`munlock` calls — a sites-only regression to `O*4`
   fails 4 named assertions. Two adjacent gaps closed while making the
   helper authoritative: an fmt=6 branch (fixed 4-byte tag, not `O*4`),
   and a loud named refusal in `qt_addrow` for formats its fall-through
   can't decode (fmt=6/7 there previously walked into the int2 arm —
   the same class your fmt=4 comment in that function documents).
4. **GPU-path status stated plainly** (PR body): fmt=7 runs scalar
   `matmul_fp8` on CPU and fmt=7-carrying layers are excluded from the
   fused Metal kernels by this PR's guards; wiring `bind_gemv` to the
   (tested) shader branch is a follow-up we want to take on my hardware.
5. Stale guard comments and the FORMATS.md narrative corrected to the
   post-inversion policy.

**How it was verified** — the part that changed most on our side:

- A **spec-blind acceptance suite**: built from a written requirements spec
  by a process that never saw this implementation, with an independently
  derived OCP E4M3-FN oracle (from the OFP8 spec text — which, incidentally,
  contradicts itself on E4M3's exponent bias between §4.2 and Table 1;
  resolved via Table 2's worked examples). Against this branch it verified
  **bit-exact decode at all 16,900 positions** of a partial-block tensor
  (130×130 — shapes no real GLM tensor exercises), NaN row-propagation
  semantics, the collision inversion, and the writer refusal on a real
  ambiguous shape. It also found the `qt_addrow` crash above — a defect
  every earlier review layer missed.
- **Differential container matrix**: current-dev binary vs this branch's
  head, over the stock int8-MTP container and an int8-resident mint, with
  per-run usage-state control — identical loads, per-tensor resolutions,
  notices, and **byte-identical temp-0 generated text** on both containers.
  The before/after in one pair of lines: the previously-pushed head refuses
  the int8-resident container at load (`model.layers.0.self_attn.
  o_proj.weight: [6144,16384] scale array is 24576 bytes — matches BOTH …
  refusing`, exit 1); this head loads it, resolves `o_proj` as fmt=1, and
  generates the same text as dev.
- Blind validation + an independent hazard audit (their own mutation tests
  and negative controls, including a reproduction of your exact `nblk > O`
  overshoot scenario), then full gates on the final head: `make check`,
  METAL=1 zero warnings, `metal-test` both `COLI_METAL_RESSET` states,
  per-commit bisect builds, suppression-stripped `-Wunused-function` build.
  
  ## Behavioral contract

This PR was revised against a written contract; every invariant below is a
registered claim (stated before the revision was built), each with its
decisive evidence:

| Invariant | Capstone evidence |
|---|---|
| Every container that loads on dev loads identically here (resolutions, notices, output) | Differential matrix vs. current dev: **byte-identical temp-0 generated text** on the stock int8-MTP container and an int8-resident mint |
| No byte pattern readable today changes meaning; ambiguity favors the incumbent | `[6144,16384]` int8-row (GLM-5.2's real `o_proj`): refused by the previous head (`exit 1`, quoted in-thread), **loads as fmt=1 here** |
| The writer can never emit what the reader would mis-resolve | Repack refusal demonstrated at a real ambiguous shape (`[2048,16384]`); round-trip load verified on non-ambiguous output |
| fmt=7 decode is exact OCP E4M3-FN — passthrough, no approximation | Independently derived oracle (from the OFP8 spec text): **bit-exact at all 16,900 positions** of a partial-block (130×130) tensor |
| Failure is loud: every refusal names its condition | New: `qt_addrow`/`qt_matvec_rows` refuse unhandled formats by name (previously: SIGSEGV for fmt=7, silent misread for fmt=6 — found by the acceptance suite, fixed here) |
| Collision family is enumerated, not sampled | In-tree census over every real tensor shape: `o_proj` is the **only** ambiguous family (79 instances/container) |

<details>
<summary>Full evidence matrix (instruments and results per requirement)</summary>

- **Compatibility (I1)**: old-binary/new-binary differential over both real
  containers, per-run usage-state control, telemetry-excised output
  comparison; plus same-binary control run proving the method's
  discrimination. Result: identical loads, per-tensor resolutions, notices,
  and generated text, all four comparisons.
- **Collision policy (E4/I2)**: reader inversion verified at source and
  behaviorally (collision-family fixtures at 4 real shapes + 1
  non-collision control resolve fmt=1); the three previously-refusing test
  shapes plus four degenerate cases now assert fmt=1; census tool
  regenerable upstream (`c/tools/fp8_collision_census.py`).
- **Exact decode (I4)**: spec-blind acceptance suite with an independently
  derived E4M3-FN oracle (derivation caught an internal contradiction in
  the OFP8 spec's own bias statement, resolved via its worked examples);
  bit-exact on NaN-free data; NaN weights verified to poison their row
  through the dot product (IEEE-correct propagation, 130/130 rows).
- **Wire-path sizing**: `qt_scale_bytes()` wired via shared
  `qt_wire_split()`; seam test intercepts actual `mlock`/`munlock` calls —
  a sites-only regression to `O*4` fails 4 named assertions; fmt=6's
  4-byte-tag case corrected while making the helper authoritative.
- **Loud failure (I6)**: refusal tests for fmt=6/7 through both absorb-path
  functions (previously crash/silent); repack tool refuses on zero
  repack-target tensors (previously silent success).
- **Gates**: `make check`; METAL=1 zero warnings; `metal-test` under both
  `COLI_METAL_RESSET` states; per-commit bisect builds; suppression-
  stripped `-Wunused-function` build; POSIX-construct check (Windows CI
  pattern) on all new tests.

Deeper artifacts (specific transcripts, census tables, suite design)
available on request.
</details>

Thanks again for the review feedback, this was a valuable reset.

monotophic:
CI caught a Linux-only short-read in our test harness's stderr capture (the refusal fired; the test's read was truncated) — fixed, head now 9f6b4ab.

monotophic:

Rebased over today's wave: rev3, head `6313f32` on `dev` @ `75a7374`
(TEST_BINS superset resolutions only; product code patch-identical to
`9f6b4ab`, which your review + our revision comment describe). Full
verification re-run on this head: gates green, the spec-blind acceptance
suite passes (bit-exact fmt=7 decode, collision inversion, writer refusal),
and the differential container matrix vs current dev shows byte-identical
temp-0 output on both real containers.

monotophic:

Rebased onto current dev `c031307` — new head `c635b8d`, rebase-only, no semantic change; awaiting re-review exactly as before.

One conflict, one content difference: #627's new CUDA fmt=6 gating in `qt_cuda_upload` (the split fmt=5 line + the E8 codebook-readiness check) collided with this PR's fmt=7 early-return in the same gate. Resolved by stacking ours under the new upstream lines — all three formats keep their documented CPU-side behavior. The fmt=7 comment's claim that `backend_cuda.cu` has no fmt=7 case was re-verified against the post-#627 sources (zero occurrences; fmt=6 cases now present, as expected).

Verification, one decisive check per claim:

| claim | evidence |
|---|---|
| commits other than the conflicted one are byte-identical to the reviewed rev3 | `git range-diff`: 11/12 `=`; the twelfth differs only in the stacked gate hunk |
| the format machinery works end-to-end on the rebased stack | fmt-7 acceptance suite (blind-built from the requirements spec): **OVERALL: PASS** on the rebased tree |
| nothing else regressed | `make check` 211 OK · `METAL=1` build zero warnings · `metal-test` both `COLI_METAL_RESSET` states · per-commit bisect builds |

monotophic:

Rebased onto current dev (`a3a5a75`) — rev5, head `dbf8d00`. The only conflict surface across the whole 12-commit series was the `c/Makefile` `TEST_BINS` list (twice), colliding with `test_rans` from #671 — resolved as the union, both sides' tests kept. Range-diff: 10/12 commits byte-identical, and the interdiff on the two changed commits contains zero lines beyond the `TEST_BINS` unions. No source content moved.

Verification on `dbf8d00` (all re-run on this head, macOS 26.5 / M5 Max):

| gate | result |
|---|---|
| `make check` | OK (13 skipped) |
| per-commit bisect builds | 12/12, 0 warnings each |
| fmt-7 acceptance suite (collision/refusal battery) | OVERALL PASS — F4 collision member resolves fmt=1 incumbent; non-collision real shape discriminates; fmt=7 fixture refuses as expected on dev |
| Windows-CI safety (`fork`/`pipe`/`waitpid` in tests) | comment-aware scan of every changed test file: all code-level POSIX constructs inside `#ifndef _WIN32` (same content passed the Windows leg at rev4) |

monotophic:

## rev6: re-derive onto current dev + renumber to fmt=8

@JustVugg, this revision moved for two distinct reasons, kept separable below: (1) the
mechanical re-derive dev's motion required, and (2) a full re-pass of our
validation process over the rebased code, which surfaced and fixed five
pre-existing issues that rev5 would have shipped and hardened two adjacent
seams. Head: `bda0ed0` (24 commits over dev `292ed4c`, every commit
bisect-built at zero warnings).

### 1) The re-derive and the renumber (caused by dev motion)

- Rebased onto `292ed4c` (post #457/#418/#676/#705). Six files carried
  conflicts, all resolved as additive unions; the one semantic adaptation is
  to #457's merged `coli_metal_gemm`/`coli_metal_matmul` signatures.
- **Our format is now `fmt=8` (was 7).** Merged #705 uses `fmt=7` for MXFP4
  in the Vulkan tier, so the ordinal this PR had claimed while mergeable was
  taken by first-merge — 191 references moved across 17 files. The renumber
  is a pure retag: **the rev5 and rev6 repack tools emit byte-identical
  tensor payloads** (verified offset-by-offset); no ordinal is ever persisted
  (the stamp in #529 stores the format *name*), so nothing on disk changes
  meaning.
- **Process note, factual:** this is the second time this format has been
  renumbered before merge for the same reason — format IDs are claimed by
  first merge into dev, and there is no reservation mechanism or registry, so
  every in-flight format PR must rescan dev and open PRs to keep its ordinal.
  That coordination gap is what #529's `docs/FORMATS.md` registry (now with
  an explicit ID-assignment section) exists to close; anyone building
  container formats for this ecosystem hits it. Cost this round: the 191-site
  renumber plus a full re-review cycle.
- **Before claiming 8 we ran the scan the registry now prescribes:** dev HEAD
  (`292ed4c`) carries no `fmt=8` in the resolver, dispatch, registry, or any
  backend, and a grep of all 36 open PR diffs (2026-07-30, patterns covering
  code, shader, and registry claim styles — the same notations #705 used for
  7) found zero fmt=8 claims; the only hits are pre-existing resolver
  call-sites in #711/#587/#544. To the best of our ability at scan time,
  fmt=8 is unclaimed; if this collides anyway with in-flight unpushed work,
  we renumber again without complaint — the registry is the fix, not the
  ordinal.

### 2) Review findings fixed in this revision (pre-existing in rev5)

Re-running our two-reviewer battery over the rebased code found five issues
that were present in rev5 (none corrupts data; classes are test integrity,
documentation integrity, and diagnostic quality): two test assertions whose
labels claimed more than their arguments tested; a registry/name-table
divergence for fmt=6 (`e8-iq3` vs the loader's `e8-iq3-lattice`); an
overstated hazard comment; a refusal message that lacked refusal wording
(and the test harness pattern that let it pass); and the undocumented
breadth of the unstamped-collision refusal family — now documented with its
precise membership condition and pinned by tests in both directions
(family-shaped artifacts: named refusal on this head, fmt=1 load on dev —
a knowing strictness trade, stamp = the designed escape hatch; zero GLM-5.2
role shapes are members, re-verified by census arithmetic).

### 3) Behavior changes vs dev (complete list)

- **Fused Metal decode gates: negative guard → positive allowlist.** The two
  eligibility gates now admit only fmt 1/2/3/4 — the mm_gemv shader's
  explicit integer-format branches, the exact set both admitted and computed
  correctly before — via one named predicate (`metal_fused_fmt_ok`). Changes
  behavior only for fmt 0/5/6 and future formats: fmt=5/6 move from a silent
  GPU f32 misread of their q4 bytes to a correct, slower CPU fallback; fmt=0's
  exclusion is a no-op formalization (it never actually fused — NULL q4);
  future formats fail closed. No change for any real GLM-5.2 loadout; the
  kv_b fmt==2 pin is untouched.
- **Unstamped collision shapes refuse by name** (family documented per §2 —
  present since rev5's maintainer-agreed collision-policy inversion, now
  precisely documented and test-pinned).
- (In #529: conflicting duplicate stamps refuse at ingest — see that PR's
  comment.)

### Capstone matrix (one decisive falsifier per key claim)

| Claim | Decisive evidence |
|---|---|
| Nothing regresses for existing containers | Differential resolution matrix: 176,232 real-tensor resolutions across three full GLM-5.2 containers (per-row int4, int8+g64, int3-g64), dev binary vs this head — **zero deltas, zero refusals** |
| Renumber is a pure retag | rev5 vs rev6 repack outputs byte-identical; old-ordinal readers load the artifact unchanged |
| fmt=7 (MXFP4) cannot be misread by this PR | 13,200-probe property sweep: `qt_resolve_fmt` never returns 7; producer enumeration shows no QT constructor can yield it; Metal/CPU dispatch refuse it by allowlist |
| The fused-path allowlist routes correctly | Truth-table test over fmt −1..100 + in-tree pin of all 11 gate sites; mutation (re-admitting 5/6) fails the suite |
| Changed tests actually bite | Every behavioral test run against the pre-rebase head with pasted failures; two rev5 assertions that could NOT bite were found by review and fixed |

<details><summary>Full gate record (collapsed)</summary>

make check (246/250 py tests OK) · METAL=1 zero warnings · metal-test both
COLI_METAL_RESSET states · gemm-test · fp8 C+python battery · per-commit
bisect builds, all 24 commits · -Wunused-function differential vs dev
(identical sets) · POSIX-construct check · writer round-trip at real shapes
(new reader accepts; format-less dev reader refuses with named error) ·
collision-family artifacts both-directions (refuse here / load on dev,
outputs on file) · A1 resolution matrix as above. Coordinates: measured
2026-07-30 on dev base `292ed4c`, macOS 26.5 / M5 Max; Windows/CUDA/Vulkan
builds not run here (CI covers Windows; Vulkan paths untouched by diff).
</details>

Footprint note: the renumber commit touches `c/tools/rans_format.py` and
`docs/int4-rans256-g0.md` — dev-owned files whose prose references this
format's ordinal; changes are comment/doc-only consistency updates, zero
behavior. Series commit messages retain historical "fmt=7" wording (reviewed
history preserved; the renumber commit on top records the retag).

PR #529, opened 2026-07-22
formats: self-describing container stamp (TRUST-VERIFY-REFUSE) + FORMATS.md registry (#524)

monotophic:
*Authored by Fable 5 in Claude Code (We/our), analysis in partnership with @monotophic (me/my).*

Revised per review (2026-07-23): restacked on [#528](https://github.com/JustVugg/colibri/issues/528)'s inverted collision policy — see the comment of this date for the revision + evidence. Where this description says PR 1 "refuses unconditionally" at ambiguous shapes: that predates the inversion. Current behavior: unstamped fmt=1-vs-fmt=7 ambiguity resolves to the incumbent (fmt=1); the stamp lets a stamped fmt=7 at such a shape be read as fmt=7. The fmt=6-vs-fmt=7 (I=98) collision still refuses unstamped, and a stamp there still resolves it.

# fmt=7 registry + metadata stamp (TRUST-VERIFY-REFUSE)

Branch: `fmt7/stamp-registry`, stacked on `fmt7/fp8-passthrough` (PR 1 — CPU
path, repack tool, Metal kernel, collision/refusal logic; merge that one
first). This PR is the **convention discussion** the maintainer asked to keep
separate from the implementation: a self-describing container stamp, and the
in-repo format registry it's documented against. Nothing in PR 1's behavior
changes if this PR is never merged — PR 1 stands on its own: unstamped fmt=1-vs-fmt=7 ambiguity resolves to the incumbent, and the narrower fmt=6-vs-fmt=7 collision refuses. This PR's stamp confirms identity where arithmetic already decides, upgrades an ambiguous-shape fmt=7 from incumbent-resolution to its true format, and turns the remaining refusal into a resolution — it never turns a resolution into a refusal, and it never grants a decoder PR 1 doesn't have.

## Why a stamp, and why now

`QT.fmt` is a plain `int` with no enum declaration and no in-repo list of
what's taken — the process gap that let #465 (E8/IQ3) and this project's own
FP8 proposal both mint `fmt=6` independently, in the same week, with neither
author wrong. `docs/FORMATS.md` (this PR) is the lightweight fix for the
*ordinal* half of that gap: an in-repo table of what's assigned. The stamp is
the fix for the other half — a container's format *identity* shouldn't have
to depend on getting the *number* right by social coordination alone, and PR
1's own collision analysis shows byte arithmetic genuinely can't always
decide identity on its own (three structural ambiguities — one of them, 
fmt=1-vs-fmt=7, at a shape GLM-5.2 actually ships: o_proj [6144,16384], 
the very case the #528 review surfaced; the other two at shapes no real GLM 
tensor has.). A NAME-based stamp is the thing that can actually check a NAME 
against the bytes.

## TRUST-VERIFY-REFUSE

**Writer** (`tools/repack_fp8_passthrough.py`): every tensor it repacks
into the fp8-e4m3-b128 container gets an entry in the output shard's
safetensors `__metadata__["colibri.fmt"]` — a JSON-encoded map of exact
tensor name → format NAME string (`"fp8-e4m3-b128"`, never the internal
`fmt=7` ordinal, which stays a colibri.c-only enum value the container never
depends on). The stamp names the WEIGHT format only, deliberately — not a
scale encoding, since fmt=7 can legitimately carry more than one (see PR 1's
"scale encoding is a declared property"; more below).

**Reader** (`qt_verify_fmt_stamp` + `qt_resolve_fmt`'s new `stamped_name`
parameter, `colibri.c`):

- **Agree** → silent no-op. A stamped, unambiguous tensor loads exactly as
  it would unstamped.
- **Disagree**, or the stamp names a format this build doesn't recognize →
  refuse loudly (tensor name + both identities printed, `exit(1)`) — same
  "untrusted container" discipline `qt_resolve_fmt` already applies
  everywhere else.
- **Absent** → zero behavior change. Byte-arithmetic inference alone
  decides, exactly as before this feature existed — every container that
  predates this PR, or comes from a tool that doesn't stamp, is unaffected.
- **Resolves an ambiguity** (the bonus case): for PR 1's two collision
  points — fmt=1-vs-fmt=7 (THE DESIGN LANDMINE) and fmt=6-vs-fmt=7 at
  `I=98` (SECOND DESIGN LANDMINE) — a stamp naming exactly one live,
  *decodable* candidate resolves it directly instead of refusing. This is
  the one case where the stamp becomes load-bearing rather than a
  post-hoc cross-check: bytes alone are fundamentally ambiguous at these
  shapes, so the stamp is the only thing that CAN decide.

**What the stamp cannot do**: grant a decoder this build doesn't have. A
tensor whose scale sidecar is UE8M0-shaped still refuses even when
correctly stamped `"fp8-e4m3-b128"` — in the clean, non-colliding case (PR
1's straightforward "recognized but not implemented" refusal) and in both
places a stamp could otherwise resolve a collision (the small-`O`
fmt=1-vs-UE8M0 corner, and the `O∈(384,512], I=98` fmt=6-vs-UE8M0 corner).
The stamp confirms the WEIGHT format; the scale ENCODING is a separate,
declared property PR 1 only implements one value of. `qt_resolve_fmt`'s own
comments carry the exact case-by-case rule.

## The registry: `docs/FORMATS.md`

Adapted from the Feature Request draft already referenced on #524
(`upstream_contribution/FORMATS_registry_draft.md` on our side) into an
in-repo reference document:

- Every format 0–7 in one table: ordinal, name, weight-byte layout,
  scale-array layout, status, and the commit/PR it landed in.
- fmt=7's row status reads **"this PR pair"** — the ordinal is public and
  the implementation is what's under review, not a proposal anymore.
- A dedicated **"Scale encoding is a declared property (fmt=7)"** section
  documenting both values explicitly: f32 (implemented) and UE8M0
  (recognized, refused by name, not implemented — the DeepSeek-V4 finding
  credited to the maintainer's own #524 comment).
- The **PRIVATE ORDINAL BLOCK convention** (100+ scratch-ordinal range for
  in-flight branches) is kept intact, updated to cite fmt=7 itself as the
  convention's own worked example: minted `fmt=6` → collided with #465 →
  re-minted `fmt=100` → graduated to the maintainer-assigned `fmt=7`, a
  pure find-and-replace at every step, zero on-disk impact.
- **"How to add a format"** gains one bullet: a Feature Request should
  declare which scale/secondary-array *encoding(s)* it implements, not
  just its byte layout — the precedent this PR sets.
- **Open questions for maintainer review**, carried forward: is `100+` the
  right private-block convention (or would a different reserved range be
  preferred)? Should `colibri.fmt` become the project's standard stamp key,
  or is a different shape preferred? Two questions from earlier drafts are
  now resolved-by-events and marked as such (the #465 sequencing question;
  the scale-encoding-as-property question, decided by the maintainer on
  #524).

## Testing

`tests/test_fp8_load.c` Part D (four stamp outcomes: agreeing, mismatching,
unrecognized-name, absent — fork+waitpid for the two refusing cases) plus
stamped variants added to Part A2 (both collision directions resolve to the
stamped format; an unrelated stamp does NOT resolve; the UE8M0-scaled
corners of both collisions stay refused even when stamped
`"fp8-e4m3-b128"`) and Part A3 (the clean UE8M0 case stays refused when
stamped; the small-`O` UE8M0-vs-fmt=1 collision DOES resolve when stamped
`"int8-row"`, since that candidate is real and decodable).
`tests/test_fp8_repack.py` gains two cases (the stamp's tensor-name set
matches selection exactly; the real CLI's output carries a correct,
parseable stamp). `tests/test_fp8_e2e_repack_load.py` (inherited from PR 1)
now also asserts the real tool's stamp end-to-end against the real C loader.

- `make glm METAL=1` (clean rebuild): zero warnings, at both commits in
  this PR individually.
- `make test-c`: full C suite green, exit 0, at both commits — including
  the expanded `test_fp8_load` (Parts A–D) and `test_int3` (its 5
  `qt_resolve_fmt` call sites updated for the new `stamped_name`
  parameter, `NULL` throughout — this suite predates and is orthogonal to
  the stamp feature).
- `make test-python`: 154 tests, 10 skipped, 0 failures, exit 0.
- `make metal-test` under `COLI_METAL_RESSET=0` AND `=1`: both full
  green, exit 0 — unaffected by this PR (the stamp is a CPU-side,
  pre-Metal-dispatch concern; `qt_from_disk` resolves `fmt` before any
  Metal call, so the GPU kernel never sees a stamp either way).
- Per-commit gates verified in isolation, same discipline as PR 1.

## Durable vs. current-state (review map)

| | claim | durable / current-state | revisit trigger |
|---|---|---|---|
| 1 | TRUST-VERIFY-REFUSE discipline (agree=no-op, disagree=refuse, absent=unaffected) | **durable** | none — this is a correctness posture, not a calibration |
| 2 | A stamp resolves THE DESIGN LANDMINE and the SECOND DESIGN LANDMINE (f32-scaled candidates only) | **durable** | none — follows directly from PR 1's byte-arithmetic facts |
| 3 | `colibri.fmt` as the stamp's `__metadata__` key, JSON-map-of-name as its shape | **current-state / provisional** — explicitly flagged as this PR's proposal, open to revision, in `docs/FORMATS.md`'s own "Open questions" section | the maintainer (or a future project-wide stamp standard) choosing a different key/shape |
| 4 | `100+` PRIVATE ORDINAL BLOCK convention | **current-state / provisional** | maintainer endorsement, or a preferred alternative (negative values, a separate `fmt_ext` field) |
| 5 | A stamp cannot resolve any UE8M0-scaled collision (no decoder exists) | **current-state**, tied to PR 1's row 3 | the same UE8M0 decoder that would flip PR 1's refusal flips this PR's stamp-resolution too, once it exists |
| 6 | measurement coordinates | checkable | measured 2026-07-22, macOS 26.5.2 (build 25F84), Apple M5 Max (MacBook Pro, Mac17,7), engine base = PR 1 tip |

## Commits in this PR

1. `fp8: self-describing __metadata__ stamp, trust-verify-refuse reader` —
   `st.h`'s stamp ingest/lookup, `colibri.c`'s FMT_NAMES table and
   `qt_verify_fmt_stamp` (post-hoc verify only, not yet resolving), the
   repack tool's writer side, the two new `test_fp8_repack.py` cases.
2. `fp8: qt_resolve_fmt lets a metadata stamp resolve byte collisions` —
   threads `stamped_name` into `qt_resolve_fmt` itself, making both
   collision points stamp-resolvable (except the UE8M0 corners); updates
   the three routed-expert call sites and `test_int3.c`'s five direct
   calls for the new signature; `test_fp8_load.c` Parts A2/A3 gain
   stamped cases, new Part D.
3. `docs: add FORMATS.md, the quantized-format registry (fmt=6/fmt=7 +
   convention)` — this PR's own documentation, no code changes.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -1341,14 +1341,19 @@ static const char *qt_name_by_fmt(int fmt){
  * 4/byte). Qui i byte del peso devono corrispondere a un layout noto e i byte
  * della scala alla cardinalita' attesa (O per-row, O*ng per-gruppo) — altrimenti
  * si termina invece di sforare. Ritorna fmt (1/2/3/4/5/6/8) e scrive *gs.
- * No container-level format stamp is consulted INSIDE this function yet in
- * this commit -- qt_verify_fmt_stamp (below) checks one AFTER the fact, a
- * pure post-hoc cross-check that cannot yet resolve a genuine byte-collision
- * below (every ambiguous shape this function can see still refuses
- * unconditionally here). A follow-up commit threads a `stamped_name`
- * parameter into this function itself so a stamp can resolve those
- * collisions instead of merely confirming an already-unambiguous result. */
-static int qt_resolve_fmt(const char *name, int O, int I, int64_t nb, int64_t ns, int *gs){
+ * `stamped_name`: the tensor's __metadata__ format-NAME stamp if the caller
+ * looked one up (st_fmt_stamp, st.h), else NULL -- used ONLY to break a
+ * genuine byte-count collision (see THE DESIGN LANDMINE and the SECOND
+ * DESIGN LANDMINE below); every other decision in this function is
+ * byte-arithmetic alone, unchanged by whether a stamp is present. A stamp
+ * can never grant this build a decoder it doesn't have: the UE8M0
+ * scale-encoding refusals below stay refusals even when a stamp names
+ * "fp8-e4m3-b128" correctly, because that name confirms the WEIGHT format,
+ * not a scale encoding this build can read. Routed-expert callers
+ * (expert_load_impl and friends) always pass NULL: this branch's repack
+ * tool never stamps routed experts, so there is nothing for those paths to
+ * consult. */
+static int qt_resolve_fmt(const char *name, int O, int I, int64_t nb, int64_t ns, int *gs, const char *stamped_name){
     int64_t exp_i8=(int64_t)O*I, exp_i4=(int64_t)O*((I+1)/2), exp_i2=(int64_t)O*((I+3)/4);
     int64_t exp_i3=(int64_t)O*i3_rowbytes(I);   /* int3-g64 (fmt=5): 24B per 64-input group */
     /* fmt=6 (E8/IQ3, #452): scales live inside the 98B super-blocks, so the .qs
@@ -1371,24 +1376,36 @@ static int qt_resolve_fmt(const char *name, int O, int I, int64_t nb, int64_t ns
      *     with UE8M0, just against fmt=6's tag instead of fmt=1's per-row
      *     count; see that landmine's own comment for the general shape.
      *   - at O==1 specifically, fmt=1's own per-row tag (O*4) is also 4.
-     * Refuse unconditionally when any of these coincide with the E8/IQ3 tag --
-     * this build has no stamp to break the tie (see the function's own header
-     * comment) -- rather than let dev's own unconditional `return 6` silently
-     * misread an fp8-e4m3-b128 (or, at O==1, plain int8) tensor as E8/IQ3-
-     * lattice-decoded garbage with no error. No real GLM tensor has these
-     * shapes; the discipline exists for untrusted containers. */
+     * A stamp naming exactly one live candidate resolves the ambiguity --
+     * EXCEPT a "fp8-e4m3-b128" stamp on the UE8M0-shaped candidate: the stamp
+     * confirms the WEIGHT format, not a scale encoding this build can decode,
+     * so that specific case still refuses (same "recognized but not
+     * implemented" discipline as the landmine below). Everything else
+     * (unstamped, or a stamp that doesn't resolve) refuses rather than let
+     * dev's own unconditional `return 6` silently misread an fp8-e4m3-b128
+     * (or, at O==1, plain int8) tensor as E8/IQ3-lattice-decoded garbage with
+     * no error. No real GLM tensor has these shapes; the discipline exists
+     * for untrusted containers. */
     if(ns==4 && nb==(int64_t)O*e8_rowbytes(I)){
         int fp8_blk_f32_also   = (nb==(int64_t)O*I) && (fp8_nblk(O)*fp8_nblk(I)==1);
         int fp8_blk_ue8m0_also = (nb==(int64_t)O*I) && (fp8_nblk(O)*fp8_nblk(I)==4);
         int i8_row_also        = (nb==(int64_t)O*I) && (O==1);   /* ns==O*4==4 iff O==1 */
         if(fp8_blk_f32_also || fp8_blk_ue8m0_also || i8_row_also){
+            int sf = stamped_name ? qt_fmt_by_name(stamped_name) : -1;
+            if(sf==6){ *gs=0; return 6; }
+            if(sf==8 && fp8_blk_f32_also){ *gs=0; return 8; }
+            if(sf==1 && i8_row_also){ *gs=0; return 1; }
+            /* sf==8 with only fp8_blk_ue8m0_also true falls through here too --
+             * see the comment above this block. */
             fprintf(stderr,"%s: [%d,%d] byte layout (nb=%lld ns=%lld) matches E8/IQ3 "
                 "(fmt=6, 4-byte tag)%s%s%s; refusing rather than guessing (untrusted "
-                "container, fmt=6 collision at I=98)\n",
+                "container, fmt=6 collision at I=98)%s\n",
                 name,O,I,(long long)nb,(long long)ns,
                 fp8_blk_f32_also   ? " AND per-128x128-block FP8 f32 scales (fmt=8, single block)" : "",
                 fp8_blk_ue8m0_also ? " AND per-128x128-block FP8 ue8m0 scales (fmt=8, 4 blocks, recognized-not-implemented)" : "",
-                i8_row_also        ? " AND plain int8 per-row (fmt=1, O=1)" : "");
+                i8_row_also        ? " AND plain int8 per-row (fmt=1, O=1)" : "",
+                stamped_name ? " -- metadata stamp present but names a format/encoding that doesn't resolve the ambiguity"
+                             : "");
             exit(1);
         }
         *gs=0; return 6;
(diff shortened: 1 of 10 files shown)


monotophic:
Working on the Windows / MinGW fail. Will update with a fix shortly.

monotophic:
Fixed and ready for review

JustVugg:
Reviewed alongside #528. Answering the question you asked in #524 first, because it's the one that decides whether this convention is safe:

**Does the stamp ever override byte-arithmetic inference? No — verified at every call site.** Each stamp-consulting branch is gated on a candidate that is *already* byte-consistent: the fmt=6 block sits inside `ns==4 && nb==O*e8_rowbytes(I)`; `fp8_blk_f32_also` and `i8_row_also` each require `nb==O*I` plus the matching block count; the landmine block only consults the stamp when both candidates are byte-valid; the UE8M0 branch requires `is_row`. A stamp disagreeing with an unambiguous inference hits `qt_verify_fmt_stamp` and refuses; an unrecognised name refuses; absent is a no-op. `fp8_blk_f32_also` (nblk==1) and `fp8_blk_ue8m0_also` (nblk==4) are mutually exclusive, so there's no double-resolve. **TRUST-VERIFY-REFUSE holds as specified.** Nice work — this is the right shape for the convention.

Four notes, none fatal:

1. **It does not rescue #528's regression, and shouldn't be relied on to.** Every container written before this convention is unstamped by definition, so the GLM-5.2 `o_proj` `[6144,16384]` collision I flagged on #528 still refuses. The registry is not a migration path for existing files — worth saying so explicitly in FORMATS.md, so nobody reads the stamp as retroactive protection.
2. **`st_fmt_stamp()` is an O(n) linear `strcmp` scan** per `qt_from_disk`. Fine today (residents only; routed experts pass NULL), but an untrusted container stamping 58k+ expert names makes every resident load walk the whole array. Bounded, not a crash — a hash or a cap on stamp-map size would close it.
3. **`st_fmt_stamp_ingest` adds a new `exit(1)` surface at `st_init_multi` time.** Fail-closed is defensible and unstamped containers are untouched, but note it now aborts model *discovery*, not just tensor load — worth documenting.
4. **Stamps on non-`.qs` tensors are silently ignored** (`qt_verify_fmt_stamp` is only called inside `if(st_has(...))`). Harmless, but "TRUST-VERIFY-REFUSE" reads as covering everything stamped; either verify them or scope the claim.

Test coverage is genuinely there and tests what it claims — the fork-based stderr assertions cover both stamp directions, the unrelated-stamp case and the UE8M0-stays-refused-when-stamped cases. The one false claim I found is shared with #528: the comment asserting `qt_wire_mmap`/`qt_unwire_mmap` now use `qt_scale_bytes()`. They still compute `t->O*4` on both sides.

**Verdict: good to merge on its own merits, but it's stacked on #528**, so it goes in after the collision policy there is inverted. If you'd rather land the registry sooner, rebasing this onto `dev` independently of the fmt=7 work would let it merge on its own — the convention is valuable even before there's a seventh format to describe.

monotophic:

All four notes taken; restacked on the revised #528. Head `fe9ac93` on
`9f6b4ab` (#528's revised head, both rebased onto dev `e4b5bf3`).

1. **Non-retroactivity is explicit in FORMATS.md**: the stamp is a forward
   convention; pre-convention containers resolve by byte arithmetic alone —
   the registry is not a migration path and gives no retroactive protection.
2. **Stamp-map lookup bounded**: `ST_FMT_STAMP_MAX=4096`, refuse-on-exceed
   with a named error (a container stamping tens of thousands of expert
   names is malformed by the convention's own resident-tensor scope);
   boundary tested on both sides.
3. **The discovery-time `exit(1)` documented** at the site and in
   FORMATS.md — fail-closed at model discovery, not just tensor load.
4. **Stamp scope stated**: the convention covers `.qs`-backed tensors;
   stamps on other tensors are ignored by design, and the docs now say so
   (verified against the actual lookup gating before writing it).

Also corrected here: two FORMATS.md passages still described the
pre-inversion collision policy; both now state the actual rule (unstamped
ambiguity → incumbent at the reader; writer refuses; a stamp is what makes
an ambiguous-shape fmt=7 readable as fmt=7 going forward).

On your unstacking offer — keeping it stacked, deliberately: after the
inversion, the stamp is the only path by which an ambiguous-shape tensor
can ever legitimately read as fmt=7, so the two PRs define one collision
policy and are best reviewed in order. If your merge window prefers the
registry sooner, say the word and we'll re-derive it standalone.

Verification rode the same upgraded process as #528: the spec-blind
acceptance suite passes on this head (incl. bit-exact fmt=7 decode and the
collision-inversion invariants), the differential container matrix shows
byte-identical temp-0 output vs current dev on both real containers, and
the full gate set is green on `fe9ac93` (make check, METAL=1 zero warnings,
metal-test both RESSET states, per-commit builds).

monotophic:

Restacked on #528's rev3: head `24de8b1`. Same re-verification as #528
(gates, acceptance suite, container matrix — all green on this head).
Content unchanged from what your four notes reviewed, per the revision
comment above.

monotophic:

Restacked onto #528's rebased head (`c635b8d`, itself on dev `c031307`) — new head `8374a43`. No conflicts and no content change: `git range-diff` reports **7/7 commits byte-identical** to the reviewed rev3; this is the same patch series on a moved base.

Verification on the new head: fmt-7 acceptance suite (blind-built from the requirements spec) **OVERALL: PASS** on this tree · `make check` 211 OK · `METAL=1` build zero warnings · `metal-test` under both `COLI_METAL_RESSET` states · per-commit bisect builds. Awaiting re-review as before; on merge, `docs/FORMATS.md` registry goes live unchanged.

monotophic:

Restacked onto #528 rev5 (`dbf8d00`, on dev `a3a5a75`) — rev5, head `e0874b2`. Clean replay: all 7 commits byte-identical to rev4 (range-diff 7/7 `=`), zero conflicts.

Verification on `e0874b2` (all re-run on this head, macOS 26.5 / M5 Max):

| gate | result |
|---|---|
| `make check` | OK (13 skipped) |
| per-commit bisect builds | 7/7, 0 warnings each |
| fmt-7 acceptance suite (collision/refusal battery) | OVERALL PASS — stamp-resolution and refusal paths exercised on the stacked tree |
| Windows-CI safety (`fork`/`pipe`/`waitpid` in tests) | comment-aware scan of this branch's changed test files: all code-level POSIX constructs inside `#ifndef _WIN32` (same content passed the Windows leg at rev4) |

monotophic:

## rev6: restacked on #528 rev6 + registry corrections + one new refusal

@JustVugg this is now rebased with enhancements/fixes. Head: `38c2185`, 
stacked on #528 rev6 (`bda0ed0`); merge #528 first, asbefore. Three 
things changed beyond the restack:

1. **Registry corrections.** The fmt=6 name now matches the loader's own
   table byte-for-byte (`e8-iq3-lattice`; the rev5 doc said `e8-iq3` — a
   stamping tool following the registry would have emitted names the loader
   refuses, in the artifact whose purpose is preventing exactly that). The
   registry's name column is pinned to `FMT_NAMES` by a new test, so this
   class cannot recur. The registry also gains a factual **fmt=7 = MXFP4**
   row (merged #676/#705, Vulkan tier — container carries UE8M0 u8 exponents
   per 32-group, expanded to f32 at upload) and an **ID-assignment section**:
   ordinals are claimed by first merge into dev; there is no reservation
   mechanism; this registry is the coordination point. Our row moves to
   **fmt=8** accordingly (see the #528 comment for the renumber record).

2. **Conflicting duplicate stamps now refuse at ingest.** Previously the
   stamp lookup was first-wins across shards: a container whose shards
   disagreed about a tensor's format got shard-enumeration-order-dependent
   behavior — demonstrated on the pre-fix build, where the same
   self-contradicting container refused with a wrong diagnosis in one
   readdir order and loaded silently in the other. Now the contradiction is
   named at discovery: agreeing duplicate stamps stay legal and idempotent
   (centralized manifest writers remain supported — no shard-locality
   constraint), and coverage remains optional at load (unstamped tensors
   and containers load exactly as before; completeness is a writer-side
   guarantee). Documented in the registry's stamp rules.

3. **Stamp-cap comment corrected** to state precisely what the cap bounds
   (the persistent name/value arrays and lookup scans — not the header-parse
   allocation, which the existing `ST_MAX_HEADER` bounds).

| Claim | Decisive evidence |
|---|---|
| Registry names are loader-canonical | Parity test parses the registry table and diffs it against `FMT_NAMES`; run against the rev5 defect, it fails with exactly that divergence |
| Conflict refusal works, order-independently | Two-shard conflicting fixture refuses naming both formats in either enumeration order; same fixture on the pre-fix build survives ingest (pasted); agreeing-duplicate fixture loads |
| Stamp semantics otherwise unchanged | TRUST-VERIFY-REFUSE suites green and mutation-proven (polarity reversion caught); unstamped behavior byte-identical (A1 matrix in the #528 comment) |

Gates: identical record to the #528 comment (this head), plus the stamped
arm of the 13,200-probe resolver sweep and the registry parity suite.

PR #542, opened 2026-07-23
feat: interactive chat mode for olmoe (multi-turn conversations)

opxyc:
## Summary

`olmoe.c` was benchmark-harness-only (ref.json prompt/full ids in, token ids out) — no way to actually converse with it. Adds a `CHAT=1` interactive mode: tokenizer wiring (tok.h/sample.h), a KV cache allocated once for the context window and reused across turns (real multi-turn context, not a fresh `generate()` per message), OLMoE-Instruct's actual `<|user|>`/`<|assistant|>` chat template, and `/reset`. Also adds `chat_olmoe.sh`, a launcher wrapping the process in a `systemd-run` cgroup (`MemoryMax` + `MemorySwapMax=0`) so a misbehaving run degrades to a clean self-kill instead of taking the whole machine down — learned that the hard way on a 21GB-RAM box before adding it.

## Validation

- [x] `make -C c check` — same result as the sibling AVX2 PR: portable build + C unit tests pass; one pre-existing, unrelated Python test fails on this dev machine from local disk-quota pressure, not from this change.
- [ ] CUDA changes were tested with `make -C c cuda-test` (if applicable) — N/A.
- [x] Performance claims include hardware, commands, and repeatable measurements — no performance claim made here; this is a feature addition, not an optimization.

Validated with a real multi-turn conversation (two follow-up coding questions in one session) — the second turn correctly used context from the first without re-explaining itself, confirming the KV-cache carry-over actually works rather than silently starting fresh each turn.

## Compatibility

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -341,7 +341,7 @@ cuda-bench: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/bench_tens
 	"$(GPUCC)" $(GPUFLAGS) backend_cuda.cu tests/bench_tensor_core.cu -o backend_cuda_bench$(EXE)
 	./backend_cuda_bench$(EXE)
 
-olmoe$(EXE): olmoe.c st.h json.h compat.h
+olmoe$(EXE): olmoe.c st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h
 	$(CC) $(CFLAGS) olmoe.c -o olmoe$(EXE) $(LDFLAGS)
 
 inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ)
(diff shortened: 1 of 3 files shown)


vaibDan:
Hi @opxyc, this looks great — exactly what I was hoping for. I want to try this on WSL2 (Ubuntu) with a laptop that has 8-16GB RAM and no dedicated GPU.

A couple of questions before I dive in:
1. Does `chat_olmoe.sh` expect specific env vars (model path, memory cap value) or does it auto-detect?
2. The `systemd-run` cgroup wrapper — does this require systemd enabled in WSL2 (`systemd=true` in `/etc/wsl.conf`), or does it fall back gracefully if systemd isn't available?

Happy to test on my setup and report back once I know what to point it at.

PR #544, opened 2026-07-23
feat: add Qwen3-30B-A3B engine

opxyc:
## Summary

Adds `qwen3moe.c`, a new engine for Qwen/Qwen3-30B-A3B (30B total / ~3B active params), forked from `olmoe.c` since its plain softmax-router/no-shared-expert shape is the closer starting point than GLM's MLA+DSA+shared-expert design. Two real architectural gaps needed fixing, not just config plumbing:

- **GQA attention**: 32 query heads share 4 KV heads, and `head_dim=128` is explicit in `config.json` — not derivable as `hidden/n_heads` the way `olmoe.c` assumed (32×128=4096 ≠ hidden_size 2048). Rewrote `attention()` with separate `Dq`/`Dkv` widths, a query-head-to-KV-group mapping, and a KV cache sized by `n_kv_heads`.
- **Per-head QK-RMSNorm**: Qwen3's `q_norm`/`k_norm` apply per attention head (weight length = `head_dim`) before RoPE — `olmoe.c`'s existing norm wiring is the right mechanism but the wrong shape/granularity.

Everything else (`st.h` tensor loading, int8 quantize/dequant, the per-layer expert LRU cache, PILOT prefetch, the router shape itself) ports over unchanged. Also adds `tools/make_qwen3moe_bench_model.py` (tiny synthetic Qwen3MoE fixture for a token-exact correctness gate, deliberately keeping `Dq`/`Dkv`/`hidden` distinct so it actually exercises the GQA reshape) and `tools/convert_qwen3moe.py` (same disk-safe streaming/resumable design as this fork's OLMoE converter PR).

**Scope**: correctness-validated engine + converter only. No chat/serve mode, OpenAI-server integration, or `coli` CLI wiring yet — planned as follow-up once this lands, mirroring the split already done for the OLMoE side of this work.

## Validation

- [x] `make -C c check` — same disk-quota caveat as the sibling PRs (pre-existing, unrelated test); portable build and C unit tests pass.
- [ ] CUDA changes were tested with `make -C c cuda-test` (if applicable) — N/A.
- [x] Performance claims include hardware, commands, and repeatable measurements — no performance claim; this PR is about correctness, not speed.

Token-exact validation:
```
python tools/make_qwen3moe_bench_model.py --output qwen3moe_bench
python tools/convert_qwen3moe.py --model qwen3moe_bench --out qwen3moe_bench_i8 --bits 8
SNAP=qwen3moe_bench_i8 ./qwen3moe 16 8 qwen3moe_bench/ref_qwen3moe.json
```
8/8 greedy tokens match the `transformers` reference exactly — under plain decode, `PILOT=1` prefetch, and a forced-small expert cache (cap=4, evicting every access). Also converted and ran the real `Qwen/Qwen3-30B-A3B` checkpoint end-to-end (int8, all 6144 experts present) — coherent, correctly-formatted, deterministic output (byte-identical across two independent runs of the same prompt).

## Compatibility

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@ c/glm
 c/glm.exe
 c/olmoe
 c/inkling
+c/qwen3moe
+c/qwen3moe.exe
 c/olmoe.exe
 c/iobench
 c/iobench.exe
--- a/c/Makefile
+++ b/c/Makefile
@@ -347,6 +347,9 @@ olmoe$(EXE): olmoe.c st.h json.h compat.h
 inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ)
 	$(CC) $(CFLAGS) inkling.c $(INK_CUDA_OBJ) -o inkling$(EXE) $(LDFLAGS)
 
+qwen3moe$(EXE): qwen3moe.c st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h
+	$(CC) $(CFLAGS) qwen3moe.c -o qwen3moe$(EXE) $(LDFLAGS)
+
 # Use a baseline that matches the compiler target. macOS already targets a
 # portable baseline when ARCH is empty; forcing the x86 value there breaks
 # Apple Silicon. Unknown targets use native rather than an invalid x86 flag.
--- a/c/qwen3moe.c
+++ b/c/qwen3moe.c
@@ -1,0 +1,1178 @@
+/* Qwen3-30B-A3B inference engine in pure C, with EXPERT-STREAMING from disk.
+ * Forked from olmoe.c: same disk-streaming/quant/LRU-cache skeleton, but with
+ * GQA attention (32 query heads / 4 KV heads, explicit head_dim=128 that does
+ * NOT equal hidden/n_heads) and per-head QK-RMSNorm before RoPE, instead of
+ * OLMoE's plain MHA-shaped attention and whole-vector QK-norm.
+ *
+ * Densa (embed, attn, router, norme, lm_head) residente in RAM (float32).
+ * Expert letti dal disco on-demand via pread+fadvise(DONTNEED), cache LRU per-layer.
+ * Matmul multi-thread con OpenMP (niente BLAS).
+ *
+ * ENV VARS:
+ *   PILOT=0/1/2/3 : 0=no prefetch, 1=1-layer lookahead, 2=2-layer, 3=3-layer lookahead
+ *   HOT=N         : pin top-N hot experts per layer permanently (never evict)
+ *   WARMUP=N      : tokens before hot pinning activates (default 5)
+ *   WIDE=N        : prefetch top-K*N candidates (default 1, try 2 or 3)
+ *   SMOOTH=F      : EMA coefficient for routing momentum (default 0.3, range 0.0-0.95)
+ *   CONF_LIMIT=F  : cumulative gate probability threshold for prefetch cutoff (default 0.92)
+ *   (expert queue is sorted by eid for SSD read locality)
+ */
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+#include <time.h>
+#include <pthread.h>
+#if defined(__APPLE__) || defined(__linux__) || defined(__FreeBSD__)
+#include <sys/resource.h>
+#include <unistd.h>
+#endif
+#include "st.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#define sleep_ms(ms) Sleep(ms)
+#else
+#define sleep_ms(ms) usleep((ms) * 1000)
+#endif
+
+
+
+/* ---------- config ---------- */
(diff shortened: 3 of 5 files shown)


JustVugg:
Thank you for this — a Qwen3-30B-A3B engine is real work, and the separate-engine structure is exactly right.

I'm leaving this **open**, but being honest about timing: we have a lot of model engines in flight right now that need to actually work end-to-end (GLM-5.2, plus Inkling #312 and DeepSeek V4 #165 mid-integration), and the focus at the moment is getting those running well rather than adding more families to test. So this isn't rejected — it just isn't where the attention is going this week.

The structure is sound and it stays on the table. When the current engines are solid and there's room to bring in another, this is a strong starting point. Thanks for the clean work, and for your patience while we get the in-flight ones over the line.

opxyc:
I did this so that I have something that fits my hardware. Thanks for looking into it. @JustVugg

rofl0r:
out of curiosity: is there a specific reason you target the less capable qwen3 over qwen3.6?

techfury90:
I was wondering the same thing- I feel like Qwen 3.6 35B A3B is the meta target that everyone's overlooking. I am _extremely_ tempted to try to hack up something for that because my intuition leads me to think this is the model target that unlocks practical local agentic coding workflows on 12-16GB consumer GPUs when combined with expert streaming. 

IMO, that is **far** more interesting than trying to get any of the larger models working here. Unlock Qwen 3.6 and you massively alter the fundamental economics of local agentic coding workflows.

rofl0r:
> Qwen 3.6 35B A3B [snip] ... the model target that unlocks practical local agentic coding workflows on 12-16GB consumer GPUs when combined with expert streaming.

i'm already running the model with ik_llama.cpp on a 6-core cpu (no GPU at all) at 15-10 tok/s (15 with MTP, but gets slower as context window increases). that's already quite useable if you use a harness with a slim system prompt. if you have >24 GB RAM you can already do that too. colibri would unlock using it on machines with even less RAM (and maybe make it even faster).

techfury90:
I think it can be made much faster than that with careful GPU expert streaming and "cheapish" GPUs. 

Let's do some quick back of the envelope math here. 

I've got a box with a 12GB RTX5070 and 32 GB of RAM right here. On Qwen 3.5 9B, I get around 80 tok/s with 4 bit quantization and ~200k context (8 bit KV cache quantization). That's 12.5ms per token. 

Now, I also have a M3 Pro MBP with 36 GB of unified memory. It's a decent bit slower at 3.6 35B A3B than the RTX5070 on 3.5 9B. However, I find that the tok/s for 35B A3B is ~40 versus ~15-20 for 9B. Naively extrapolating that to the 5070 suggests a possible burst tok/s rate on the order of 160-240 tok/s. If the hot set of experts is on the order of 20% of the model's total parameters... then it will almost certainly run faster than the 80 tok/s on 9B. Even if it takes on the order of 10ms to DMA a few experts in... the high burst rate could make this _extremely_ compelling.

opxyc:
> out of curiosity: is there a specific reason you target the less capable qwen3 over qwen3.6?

Different architectures, not a capability call. Qwen3-30B-A3B is plain GQA + softmax router, no shared expert which is why it forked cleanly onto the existing OLMoE engine and matches the HF reference token-for-token.

Qwen3.6-35B-A3B is the Gated DeltaNet hybrid: 3:1 DeltaNet-to-full-attention, shared expert, vision encoder. None of that ports well (in my understanding). The recurrent-state path alone means there's no KV cache to reuse, so it's a separate engine work.

Glad to look into a 3.6 engine as its own PR if that's what everyone's interested in.

rofl0r:
> Glad to look into a 3.6 engine as its own PR if that's what everyone's interested in.

there's already #602 but its author didn't respond to PR review so far (and afaict also uses the "dumb int4" quant, rather than the better group-based options colibri now supports)

JustVugg:

JustVugg:
@opxyc queue-reduction pass: #712 implements the Qwen3.6-35B-A3B engine (hybrid Gated Attention + DeltaNet) and is under active review. Does it supersede this Qwen3-30B PR, or do you see both families shipping? If 3.6 covers your use case I'd close this one; if you believe 3.0-30B has a distinct audience, say so and we'll keep it — but it needs a rebase either way (last update Jul 29).

JustVugg:
Closing after the direct question above went unanswered for 8 days and the branch has needed a rebase for a month.

For the record, precisely: this is **not** an architecture supersession — Qwen3-30B-A3B is `qwen3_moe` (standard-attention MoE), and the qwen36 engine that just landed (#712 CPU + #713 CUDA tier) exact-matches `qwen3_5_moe` only, deliberately. What has moved on is the *demand*: the newer 35B-A3B family of the same size class is now fully supported with pre-converted containers, GPU tier, and token-exact CI gates.

If you (or anyone) want to revive Qwen3-30B support: reopen or file fresh against current `dev` — the qwen36 review checklist (A/B/C lists in #712) is the template such an engine would be held to, and the serve codec + family registry that landed since would remove most of this PR's boilerplate. Thank you for the work that went into it.

PR #380, opened 2026-07-18
feat: add local-cluster MoE, dense sharding, and WebGPU workers

gauravsaini:
## Summary

This PR adds a progressive local-cluster execution path for consumer Macs and browser/WebGPU devices:

- coordinator/worker registration and distributed MoE expert routing
- native batch-union expert RPC for disk-backed workers
- contiguous dense MLP activation sharding across local nodes
- browser/mobile WebGPU expert workers over WebSocket
- native TCP-to-WebSocket activation proxy
- WebGPU expert weight exporter for readable floating-point checkpoints
- browser worker UI and protocol/dispatch tests

The coordinator keeps token generation, routing, attention, and KV state local. Remote workers execute assigned expert FFNs or dense MLP ranges.

## Validation

- native `make test`
- 73 Python tests
- 17 web tests
- TypeScript/Vite production build
- Python and JavaScript syntax checks

## Scope and follow-ups

The WebGPU path currently uses f32 buffers and is intended for a trusted LAN. TLS/authentication, WebRTC transport, quantized WebGPU weight formats, and broader dense/tensor-parallel sharding can follow as separate changes.

## Not yet tested (sorry for being gpu poor) :
Real C coordinator → dense worker COLIDN01 end-to-end
Dense shard output equivalence against local computation
Real C → WebGPU browser integration
WebGPU FFN numerical correctness against a CPU reference
Exporter tests
Actual iPhone/Android/WebGPU hardware smoke test

PR #550, opened 2026-07-23
feat: add dense MLP activation sharding

gauravsaini:
## Summary

- add a coordinator-to-worker TCP activation protocol for contiguous dense MLP layers
- add `coli cluster dense-worker` and `--dense-shards host:port:first:last`
- keep attention, KV cache, routing, and token generation on the coordinator
- load only the selected dense gate/up/down tensors on each worker
- add wire-level protocol/ownership tests and a token-exact CPU teacher-forcing gate

## Correctness gate

The optional `c/tests/test_dense_sharding.py` test starts a worker and compares the exact `TF=1` oracle mismatch signature between local CPU execution and delegated dense execution. It skips cleanly when the gitignored `glm_tiny` fixture is unavailable.

Native verification on macOS:

- `make colibri`
- `make test-c`
- `./tests/test_dense_protocol`
- dense sharding correctness test: skipped only because `glm_tiny` is not present locally

This is intentionally limited to dense MLP activation sharding. Expert-worker and WebGPU work remain separate PRs. The existing bundled PR #380 remains open while the work is split.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,32 @@ the right experts get. It works because routing has measurable structure (see
 the [expert atlas](https://github.com/JustVugg/colibri/issues/175)) — and
 structure is cacheable.
 
-The engine is a single C file (`c/glm.c`) plus small headers. No BLAS, no Python
-at runtime, no GPU required.
+The engine is a single C file (`c/colibri.c`) plus small headers. No BLAS, no
+Python at runtime, no GPU required.
+
+### Dense MLP activation sharding
+
+The first local-cluster seam can move contiguous dense MLP layers to another
+machine. Attention, KV state, routing, and token generation remain on the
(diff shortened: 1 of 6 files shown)


JustVugg:
This is the strongest of your three (it ships `test_dense_sharding.py`, a real token-exact baseline-vs-delegated gate — exactly the discipline the others need). But same honest note: dense MLP sharding across nodes is breadth, and the priority right now is depth on single-box models, so I'm **deferring** rather than merging today. It's env-gated OFF (default path safe) and it conflicts with #551/#552 anyway (all three edit the same `colibri.c` lines). Kept open — if any of the trio lands when breadth is the goal, this is the one with the correctness gate, so it'd be first. Thank you for the clean split and the test.

PR #551, opened 2026-07-23
feat: add distributed expert workers

gauravsaini:
## Summary

- add a coordinator-owned expert-worker data plane over persistent TCP
- keep routing, token generation, and KV cache on the coordinator
- batch each layer's routed expert union into one RPC to amortize network latency
- add disk-backed worker execution using the same converted model and native FFN kernels
- add a dependency-free registration/heartbeat/topology service via `coli cluster coordinator`
- add `coli cluster worker` and coordinator discovery for `coli serve`

## Correctness and scope

The single-machine path remains unchanged unless `CLUSTER_WORKERS` or `--cluster-coordinator` is configured. The control-plane tests cover registration, heartbeat, stale topology filtering, and expert discovery. Native C tests and `make colibri` pass on macOS.

This PR is intentionally the expert-worker slice only. Dense activation sharding is in #550; WebGPU/browser workers will be proposed separately. The original bundled PR #380 remains open as requested and is not this PR's merge unit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,40 @@ the right experts get. It works because routing has measurable structure (see
 the [expert atlas](https://github.com/JustVugg/colibri/issues/175)) — and
 structure is cacheable.
 
-The engine is a single C file (`c/glm.c`) plus small headers. No BLAS, no Python
-at runtime, no GPU required.
+The engine is a single C file (`c/colibri.c`) plus small headers. No BLAS, no
+Python at runtime, no GPU required.
+
+### Local cluster mode
+
+The coordinator keeps token generation, routing, and KV state local while
+disk-backed expert workers execute routed FFNs on other Macs. A layer's routed
+batch-union is sent as one persistent TCP request, so a token does not incur one
+round trip per expert.
+
+Start the optional registration service:
+
+```bash
+./coli cluster coordinator --host 0.0.0.0 --port 8765
+```
+
+On each worker, with the same converted model available locally:
+
+```bash
+./coli cluster worker --model /nvme/glm52_i4 --port 9100 \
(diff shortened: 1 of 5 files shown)


JustVugg:
Thanks for splitting this out. Two blockers, and I want to be straight about both. **(1) Priority:** distributed expert workers are breadth (multi-machine), and the focus right now is depth on single-consumer-box models — so this isn't where attention is going this week. **(2) Correctness, and this is the real one:** `cluster_worker_run` reimplements the expert FFN (`expert_gate_up` → `siluf` → `e8_rot_rows` → `matmul_qt`) as a parallel copy of `moe()`'s inner compute, but `test_cluster.py` only exercises the Python HTTP registry — nothing proves the worker's output matches local `moe()`. A parallel expert path with no token-exact worker-vs-local gate can silently drift from the real kernel. Compare your own #550, which *does* ship that gate. Before this could land, it needs a correctness test proving worker output == local output. It's env-gated OFF so the default path is safe, but I won't merge a re-implemented core kernel on trust. Deferring for now on both counts.

JustVugg:
Closing in favour of **#1036**, which is this work carried forward rather than replaced.

It keeps your `coli cluster coordinator/worker` split, the COLIEX01 framing and the `CLUSTER_WORKERS` gate, and fixes two things this branch could not have caught without a parity gate: the worker's parallel copy of `moe()` omitted the fmt=6 gate/up input pre-rotation (wrong numbers on E8/IQ3), and `cluster_init`'s inverted connect bookkeeping made the coordinator always report "no expert workers reachable".

Your #550 is where the token-exact discipline in this area came from, and #1036 mirrors it explicitly. Thanks for the original design — it's the one that shipped.

PR #567, opened 2026-07-23
Fix stateful KV tail at the NGEN limit

winklemad:
## Summary

When generation stopped exactly at the NGEN limit, `spec_decode` emitted the final token but returned before forwarding it into the stateful KV cache. As a result, `:more` resumed from the preceding token and `.coli_kv` omitted the visible tail of the answer.

Let stateful callers take the existing zero-draft forward path for the capped token before `kv_out` advances. One-shot generation keeps its existing saved terminal forward.

Add a dependency-free zero-layer regression covering NGEN 0/1, continued generation and `:more`, persisted history, accepted speculative tokens, EOS, and normal non-limit termination.

## Validation

No CUDA paths or performance claims are involved.

- [x] `make -C c check`

## Compatibility

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -392,6 +392,9 @@ tests/test_grammar$(EXE): tests/test_grammar.c grammar.h
 tests/test_schema_gbnf$(EXE): tests/test_schema_gbnf.c schema_gbnf.h grammar.h json.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
+tests/test_spec_decode_state$(EXE): tests/test_spec_decode_state.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h route_trace.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
 tests/test_decode_batch$(EXE): tests/test_decode_batch.c decode_batch.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -4555,7 +4555,12 @@ static int spec_decode(Model *m, int *all, int kv, int n_new, int eos, float *lo
         if((eos>=0 && next==eos) || is_stop(next)) break;
         emit(next,ud); all[kv]=next; emitted++; m->n_emit++;
         gr_feed(&g_grd,next);                           /* il walker segue l'output emesso */
-        if(emitted>=n_new) break;                       /* l'ultimo token non serve forwardarlo */
+        /* One-shot generation does not need logits or KV for the last token.
+         * Stateful callers do: their kv_out becomes chat history, feeds MORE,
+         * and is persisted to .coli_kv.  Let those callers take the normal
+         * g=0 step_all path below so the final token is really committed; just
+         * incrementing kv would serialize an uninitialized KV row. */
+        if(emitted>=n_new && !kv_out) break;
         int g = 0, gsrc = 0;                            /* sorgente: 1=grammatica 2=MTP/n-gram */
         if(g_grd.on){                                   /* metodo F: prima la grammatica — dove
                                                          * forza, l'acceptance e' ~1 (#48) */
--- a/c/tests/test_spec_decode_state.c
+++ b/c/tests/test_spec_decode_state.c
@@ -1,0 +1,224 @@
+/* Stateful spec_decode callers use kv_out as both the live history length and
+ * the .coli_kv record count.  At an NGEN cap the last emitted token therefore
+ * has to take the normal forward path before kv_out advances.  Otherwise MORE
+ * starts from the preceding token and persistence silently omits the visible
+ * tail of the answer.
+ *
+ * A zero-layer toy model keeps this test dependency-free.  Its embedding and
+ * lm_head implement an explicit token->next-token table, while m.n_fw proves
+ * the final state was actually forwarded rather than only counted. */
+#define main coli_glm_main_unused
+#include "../colibri.c"
+#undef main
+
+#define CHECK(condition) do { \
+    if (!(condition)) { \
+        fprintf(stderr, "%s:%d: check failed: %s\n", __FILE__, __LINE__, #condition); \
+        return 1; \
+    } \
+} while (0)
+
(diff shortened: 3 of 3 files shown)


JustVugg:
Thanks for this — the stateful-KV-tail bug is real (the last token dropped from persistence and re-forwarded on MORE), and shipping a dependency-free regression test is exactly right. But **the test itself is failing in CI** on both Linux and Windows, so I can't merge yet:

```
tests/test_spec_decode_state.c:99: check failed: out.n == 1 && emitted[0] == 2 && all[2] == 2
tests/test_spec_decode_state.c:216: check failed: test_ngen_zero_and_one_shot() == 0
```

It's the **one-shot** case (`kv_out == NULL`) that trips, which is the path the fix is supposed to leave byte-for-byte unchanged. So either the fix changes one-shot behavior after all, or the test's expectation for that case is off. Worth running `make -C c test-c` locally (or just `tests/test_spec_decode_state`) to see which — the CI is running the exact same thing.

Once the test is green I'll merge it — the fix direction is correct and the coverage is good; it's just that the gate has to pass before it lands on `dev` (a red test in the suite blocks everyone). Ping me when it's updated.

ZacharyZcR:
This is a real bug and `!kv_out` is the correct predicate — and I owe you a heads-up, because #581 (single-slot serve MTP, merged to dev a few hours ago) both **collides with this PR and adds a second caller your fix covers**.

**Why it's now CONFLICTING:** #581 gave `spec_decode` a 10th parameter — `float **logit_out` — so the mux scheduler can own the continuation logits. The signature and every call site changed:

```c
static int spec_decode(Model *m, int *all, int kv, int n_new, int eos, float *logit,
                       void (*emit)(int,void*), void *ud, int *kv_out, float **logit_out);
```

Your one-line change to line ~4721 is untouched by that (I didn't move the `break`), so it rebases cleanly. The conflict is your new `test_spec_decode_state.c`: its `spec_decode(...)` calls are 9-argument and won't compile against the new signature. Add a trailing `NULL` (the non-`logit_out` path, exactly like the four production callers) and they're correct again.

**And here's why your fix matters more after the rebase, not less:** #581 added a fifth `spec_decode` caller — the single-slot mux serve path — and it is a stateful caller with the exact exposure you're fixing:

```c
int prod = spec_decode(m, sc->hist, sc->len, r->maximum - r->emitted, eos, lg,
                       mux_spec_emit, &ud, &sc->len, NULL);   // kv_out = &sc->len (non-NULL)
```

At an NGEN cap this hits the same `if(emitted>=n_new) break` before the final token is forwarded, so `mux_done`'s `kv_disk_append` persists `sc->len` *without* the last visible token — the mux `.coli_kv` / prefix-match version of the `:more` truncation you documented. Your `&& !kv_out` guard fixes this caller for free, since it passes `kv_out` non-NULL.

So the rebase is worth it: (1) trailing `NULL` on the test's calls, and (2) if you want the regression to cover the caller #581 introduced, a case driving the mux single-slot spec path (or at minimum a comment noting `spec_decode`'s stateful callers now include serve, not just chat/run_serve). Happy to help wire that test case — it's my exposure your fix closes.

+1. Correct root-cause fix; the merge order just means it needs a touch-up against the new signature.

winklemad:
Thanks, Zachary — this is really helpful, especially the context around the new single-slot mux caller.

I also found the cause of the earlier Linux/Windows CI failure. It is in my test helper, not the stateful fix: choose() allocates the logits with falloc() (which uses malloc) but initializes only the selected slot. The remaining entries contain uninitialized values, so another token can accidentally win on some platforms. I will initialize the full logits buffer before setting the chosen token.

I will rebase onto the current dev branch, add the trailing NULL logit_out argument to every spec_decode call in the test, and rerun make -C c test-c and make -C c check before pushing the update.

The mux path is useful confirmation that this fix now protects more than the original :more case. I can also add a focused mux regression, or adjust the existing coverage, if you think that would be useful — just let me know what you would prefer.

JustVugg:
**Rebase request, with one extra thing you should know before you start.**

`dev` moved a lot in the last day (19 PRs, including the Vulkan backend and the shared routing telemetry). Your only *textual* conflict is `c/Makefile`, and that one is now trivial: the hand-written `TEST_BINS` line is gone, replaced by a list derived from the build rules (#733), so you can simply drop your entry — `test_spec_decode_state` is picked up by its own rule.

**But there is a second, non-textual problem, and it is the reason I did not resolve this one for you.** I tried: the Makefile merge is clean, and then the build fails.

```
tests/test_spec_decode_state.c:94:11: error: too few arguments to function ‘spec_decode’
tests/test_spec_decode_state.c:98:11: error: too few arguments to function ‘spec_decode’
tests/test_spec_decode_state.c:116:11: error: too few arguments to function ‘spec_decode’
```

`spec_decode` gained parameters on `dev` while this was open. Git merges the Makefile happily because nothing textually collides, and the mismatch only shows up at compile time — which is exactly the kind of thing a maintainer should not paper over on someone else's behalf. Your test knows what it is asserting about the KV tail; I would be guessing at what the new arguments should be for each of those three calls.

So: rebase onto current `dev`, drop the `TEST_BINS` line, and update the three `spec_decode` call sites to the current signature.

For context on the churn: `c/Makefile` used to be one shared line that every test-adding PR appended to, so any two of them conflicted by construction. That is fixed now, and merge order is oldest-clean-first from here on, so this should be the last rebase this PR needs.

The fix itself — the stateful KV tail at the `NGEN` limit — is one we want. If you would rather not carry it further, say so and we will close it with thanks; the branch stays yours.

winklemad:
Carrying it forward — thanks for the heads up on the Makefile change.

Dropped my `TEST_BINS` entry; the derived list picks the test up from its own rule now. I also moved my rule up next to `test_decode_batch` instead of `test_idot`, since `test_idot`'s dep line gained `route_trace.h` on dev and an insertion right above a changed line is what git was choking on. It's clean against current dev now.

One thing worth flagging: your paste showed three `spec_decode` call sites but there are eight in that file, all needing the trailing `NULL` for `logit_out` — fixing only the three still fails the build. Probably just truncated compiler output.

And I fixed the Linux/Windows flake I mentioned earlier: `choose()` was using `falloc` (plain malloc) and only setting the chosen slot, so garbage in the rest could out-rank it on some platforms. It zeroes the buffer now. `make -C c test-c` is green here; `make -C c check` only trips on `safetensors` not being installed locally.

PR #570, opened 2026-07-23
build: portable build_cuda.bat with MSVC/CUDA/arch auto-detection

ghostnumber7:
## Summary

Replace the hardcoded-path build script (a specific user's Desktop path, CUDA v12.8, fixed VS location) with one that discovers what it needs, so coli_cuda.dll can be built by double-click or from any shell.

- MSVC: find the host compiler with vswhere, preferring the 2022 toolset (range [17.0,18.0)) and falling back to -latest. vswhere is invoked by bare name from its own directory so its quoted path plus the quoted version arg don't confuse for /f's implicit cmd /c.
- CUDA: use nvcc from PATH, else %CUDA_PATH%; add -L<cuda>\lib\x64 only when the toolkit root is known.
- Arch: auto-detect compute capability from GPU 0 via nvidia-smi (12.0 -> sm_120, 8.9 -> sm_89), validated before use; default sm_120. CUDA_ARCH overrides for cross-compiling.
- Build in %~dp0 so backend_cuda.cu resolves and the DLL lands next to colibri.exe. All paths quoted, so spaces in any path are handled. CCBIN and ALLOW_UNSUPPORTED knobs for edge cases.
- 
## Validation

- [ ] `make -C c check`

Built `coli_cuda.dll` on Windows (RTX 5090, VS 2022 BuildTools, CUDA on PATH). Also verified it builds from a path containing a space. Build-only change — no effect on the engine at runtime.

## Compatibility

### DIFF
--- a/c/build_cuda.bat
+++ b/c/build_cuda.bat
@@ -1,5 +1,133 @@
 @echo off
-call "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvars64.bat"
-cd /d C:\Users\Mark\Desktop\Projects\colibri\c
-"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.8\bin\nvcc" -O3 -std=c++17 -arch=sm_120 -Xcompiler=-W3 -shared -DCOLI_CUDA_BUILDING_DLL -L"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.8\lib/x64" -lcudart backend_cuda.cu -o coli_cuda.dll
-echo EXITCODE=%ERRORLEVEL%
+rem ---------------------------------------------------------------------------
+rem build_cuda.bat - one-shot build of coli_cuda.dll (Windows, nvcc + MSVC).
+rem
+rem Mirrors the Makefile's `cuda-dll` recipe so it can be double-clicked or run
+rem from a plain shell: it locates vcvars64.bat and the CUDA toolkit itself
+rem rather than hardcoding install paths.
+rem
+rem Overrides (set before calling):
+rem   CUDA_ARCH   GPU compute capability (sm_NNN). Auto-detected from the
+rem               installed GPU via nvidia-smi; set to override or to
+rem               cross-compile. sm_120 = Blackwell / RTX 50xx, sm_89 = Ada /
+rem               RTX 40xx, sm_86 = Ampere / RTX 30xx.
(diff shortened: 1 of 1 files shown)


ZacharyZcR:
Reviewed the script end to end — it replaces a real footgun with a well-considered one, and the error handling is the standout. +1 to merge.

**The problem is as bad as the summary says.** The old script hardcoded `cd /d C:\Users\Mark\Desktop\Projects\colibri\c`, CUDA `v12.8`, the VS BuildTools path, and `sm_120` — it built for exactly one machine and one person. Anyone else got a wrong-directory failure on line 3.

**What I checked in the new one, all sound:**
- **vswhere range `[17.0,18.0)`** correctly pins the 2022 toolset, with a `-latest` fallback — the right call because nvcc rejects VS *newer* than it supports, so "grab the latest" would be the wrong default. Invoking vswhere by bare name from its own directory sidesteps the `for /f`-quoting trap. 
- **`nvidia-smi ... compute_cap` parsing**: `skip=1` drops the CSV header, and the `^[0-9][0-9]*\.[0-9][0-9]*$` validation before splitting means a stderr message swallowed by `2^>nul` can't turn into a garbage `-arch`. Clean.
- **Quoting** survives spaces in every path I traced (`CUDA_LIB`, `CCBIN_FLAG`, the `%NVCC%` call, `pushd %~dp0`).
- Mirrors the Makefile `cuda-dll` recipe (`-shared -DCOLI_CUDA_BUILDING_DLL -L…/lib/x64 -lcudart`), so the two build paths stay in step.
- The `:failed` / `:no_cuda` / `:no_msvc` messages with winget commands and the `CCBIN` / `CUDA_ARCH` overrides are genuinely better than most build scripts ship with.

**Two honest non-blocking notes:**
1. `--query-gpu=compute_cap` is a relatively recent nvidia-smi field (driver ~510+). A very old driver emits nothing, so detection falls back to `sm_120` — which then needs CUDA 12.8+ that such a host almost certainly doesn't have, so it fails at the nvcc step with the `:failed` guidance instead of silently mis-building. The chain is self-consistent; might be worth one comment line saying detection needs a recent driver, and `CUDA_ARCH` is the manual override.
2. The closing hint `make colibri.exe CUDA_DLL=1 ARCH=native` is right for building on the machine that runs it, but `native` bakes in that CPU's ISA — a DLL/host built on one box and copied to another will SIGILL. If the docs ever cover redistribution, `ARCH=x86-64-v3` is the portable host counterpart.

Build-only, no runtime effect, CI green, single file. Nice cleanup.

Issue #585, opened 2026-07-24
[Feature]: METAL Attention support for fmt:4


RDouglasSharp:
### Problem

Currently line 2366 in colibri.c gates METAL support to fmt:2 only:
&& l->kv_b.fmt==2.  The Metal mm_gemv kernel (line 16) only supports fmt=1 (int8), fmt=2 (int4 per-row), fmt=3 (int2), and f32 — no fmt=4 (grouped int4).

As a result, PROF=1 PROMPT="Compare the myths of Lucifer and Prometheus" NGEN=64 COLI_METAL=1 DIRECT=1 MTP=0 TEMP=0 COLI_METAL_RESSET=1 SNAP=~/glm52_i4 ./colibri 64 achieves 2.39 tok/s on fmt:2 but only  0.81 tok/s on fmt:4

### Proposed solution

Add fmt:4 support to the Metal attention path.  This involves

 - Adding gs to AttnW struct and bind_gemv
 - Adding grouped-int4 handling in the mm_gemv Metal kernel
 - Relaxing the l->kv_b.fmt==2 guard

### Alternatives considered

We could leave processing as is, but the existing code performs Apple Silicon Mac attention processing on the CPU.  This is ~3X slower than the fmt:2 model on my machine.

### Scope and compatibility

Limited to fmt:4 on Mac apple silicon

RDouglasSharp:
Profile output for the two formats:

PROF=1 PROMPT="Compare the myths of Lucifer and Prometheus" NGEN=64 COLI_METAL=1 DIRECT=1 MTP=0 TEMP=0 COLI_METAL_RESSET=1 SNAP=~/glm52_i4_g128 ./colibri 64
prefill 9 tokens in 10.64s | decode 64 tokens in 79.43s (0.81 tok/s) | expert hit rate 72.7% (pin 57.6% + lru 15.1%) | RSS 95.27 GB
experts loaded/token: 590.6 (per-layer 7.88 across 75; baseline topk=8) | TOPK=0 TOPP=0.00
speculation: 1.02 tokens/forward (63 forwards per 64 tokens) | MTP acceptance 0% (0/0)
PROFILE: expert-disk 58.157s service / 18.979s wait | expert-matmul 7.481s | attention 48.441s (including kvb 0.000s) | lm_head 0.000s | other 4.527s
ATTENTION: projection/RoPE 15.995s | score-softmax-value 4.806s | output projection 27.634s
[...]

Issue #583, opened 2026-07-24
[Performance]: GLM-5.2 (744B MoE) on a single AMD R9700 32 GB: 0.15 tok/s, 80% expert hit — capacity-bound benchmark


limcheekin:
### Commit

81f08a09e5651ce52616dc720f68810f9021c0be

### Hardware and storage

Hi there,

I use Opus 4.8 to drive the benchmark, it add the following benchmark results to the `docs/benchmarks.md`, appreciate if you have any idea to improve the performance:

## Community benchmarks (measured)

Real numbers from real machines, stock build (`setup.sh`, gcc 13), greedy decoding, `--ngen 32`, MTP active:

| machine | disk (iobench, 19 MB × 64, 8 threads) | config | measured |
|---|---|---|---|
| Intel i5-8600K (6C/6T, AVX2) · native Linux · 125 GB RAM · SATA SSD · **AMD Radeon AI PRO R9700 32 GB** (gfx1201/RDNA4, ROCm 7.2.2 **HIP backend**), measured 2026-07-24 | — (SATA SSD, not iobench'd) | HIP build · `CUDA_EXPERT_GB=30 CUDA_RELEASE_HOST=1 DIRECT=0` · warm cache · 1586 pinned experts (30 GB VRAM) | **0.15 tok/s** · hit 80.2% · RSS 97.7 GB (64-tok, MTP acc 0%) · at 26 GB budget → 0.12 tok/s / 77% |

### AMD single-card setup (ROCm/HIP, R9700) — measured 2026-07-24

First end-to-end AMD-discrete datapoint (the [GPU backends matrix](../GPU_BACKENDS.md)
had a `hip-test` pass on gfx1201 but no tok/s). Two setup notes not in the build docs:

```bash
# 1) Build the HIP backend. The documented command links as a PIE and fails
#    (backend_cuda.o is not -fPIC), so force position-independent code:
rm -f c/backend_cuda.o c/colibri c/glm
ROCM_HOME=/opt/rocm-7.2.2 make -C c glm HIP=1 HIP_ARCH=gfx1201 \
  HIPCCFLAGS="-O3 -std=c++17 -x hip --offload-arch=gfx1201 -fPIC -Wall -Wextra"

# 2) The `coli` wrapper is NVIDIA-only: cuda_binary() greps ldd for libcudart
#    (a HIP build links libamdhip64), and resource_plan.discover_gpus() only calls
[...]

PR #587, opened 2026-07-24
Metal fmt=4 grouped-int4 decode: attention + routed experts (#585)

RDouglasSharp:
# Metal fmt=4 (grouped-int4) decode support — Scope of Work, gating & tests

_GLM-5.2 (744B MoE, int4-g128) on the Apple-GPU (Metal) backend. Tracking #585 / #583._

## 1. Problem

On the `metal-attn-fmt4` branch the grouped-int4 (fmt=4, g128) checkpoint decoded at
**0.81 tok/s** against **2.39 tok/s** for the per-row int4 (fmt=2) checkpoint on the same
binary/prompt — a 2.95× regression the original fmt=4 PR (#457) had characterized as
"similar." The gap was entirely CPU fallback on the Metal path: two of the three decode
hot paths had no fmt=4 implementation and silently returned to the CPU.

1. **Fused attention** returned 0 (→ CPU) for any layer where `kv_b.fmt != 2`.
2. **Routed-expert GEMV** — the dominant cost in a 744B MoE decode — had *no* fmt=4 kernel
   branch and was hard-gated to CPU. Every routed expert of every layer ran on the CPU.

Attention is a thin slice of MoE decode; the experts are the bulk. Fixing (1) alone moved
decode 0.81 → 0.83. Fixing (2) moved it **0.83 → 2.13 tok/s** (vs 2.39 fmt=2).

### 2a. Fused-attention fmt=4 (`kv_b`)

`kv_b` is dequantized *inline* inside the `a_qabs` and `a_ctx` shader kernels (not via the
generic gemv), through the helper `a_deqrow`, which applied a single per-row scale
(`sc[row]`) — fmt=2 semantics. For fmt=4 the scale is per gs-wide group along the latent
dim (`[O][ng]`, `ng=ceil(A_KVL/gs)`).

- `a_deqrow` made format-aware: `sc[row*ng + i/gs]` for fmt==4, `sc[row]` otherwise.
- `a_qabs` / `a_ctx` gained `fmt`/`gs` constant buffers (index 4/5), bound from
  `W->kvb_fmt` / `W->kvb_gs` at their two dispatch sites.
- `AttnW` gained `kvb_gs`; both decode-function initializers updated. (The decode
  signatures already received `kvb_gs`; it was previously dropped.)
- Guard relaxed at the two colibri.c attention call sites:
  `l->kv_b.fmt==2` → `(l->kv_b.fmt==2 || l->kv_b.fmt==4)`.

### 2b. Routed-expert fmt=4 (`moe_gemv`)

- **`moe_gemv` kernel**: new `else if (fmt==4)` branch — grouped int4 with the per-group
  scale folded into the accumulator in-loop (group index `(8c+j)/qgs`), mirroring the
  already-validated `mm_gemv` fmt==4 branch. New `qgs` constant (buffer index 10). Final
  store made conditional: `yout = (fmt==4) ? acc : acc*sc[o]` (fmt=4 folds scale in-loop;
  every other format keeps the trailing per-row multiply). Per-expert scale slab read as
  `sc + o*ng`, matching the QT fmt=4 `[O][ng]` layout used by `mm_gemv` and `qt_bytes`.
- **`moe_submit`**: new `int qgs` parameter, bound at kernel index 10. Host gate relaxed
  `fmt!=1 && fmt!=2` → `fmt!=1 && fmt!=2 && fmt!=4`.
- **`coli_metal_moe_block` / `coli_metal_moe_block_begin`** (+ `backend_metal.h`): new
  `int qgs` parameter threaded to `moe_submit`.
- **`colibri.c`**: `MB_BUILD` now captures the quant group size `mgs = e->g.gs`
  (shared expert: `l->sh_gate.gs`) alongside `mfmt`, passed at both submit sites.

### Files touched

| File | Change |
|---|---|
| `c/backend_metal.mm` | `a_deqrow`/`a_qabs`/`a_ctx` fmt-aware; `AttnW.kvb_gs`; `moe_gemv` fmt=4 branch + `qgs`; `moe_submit`/`moe_block`/`moe_block_begin` `qgs` + gate |
| `c/backend_metal.h` | `moe_block` / `moe_block_begin` decls gain `int qgs` |
| `c/colibri.c` | attention guard relaxation (×2); `mgs` capture + pass (×2) |
| `c/tests/test_backend_metal.mm` | `run_moe` and `run_attn` gain grouped-int4 arms + oracle |

### Out of scope (deliberately)

The large-batch **prefill** paths `coli_metal_gemm` (`backend_metal.mm:885`) and
`coli_metal_matmul` (`:596`) still gate fmt=4 to CPU. These are `S ≥ 16` prefill kernels,
not exercised by decode (`S=1`), so they do not affect the tok/s numbers here. A separate
follow-up if prefill fmt=4 becomes relevant.

## 3. How the change is gated to affect **only** Metal

**Compile-time.** Every colibri.c edit is inside `#ifdef COLI_METAL` (the two expert-submit
blocks at `:3160–3215` and `:3253–3280`, including the `mgs` declaration, `MB_BUILD`
capture, and both call sites; the attention block at `:2355–2387`). `backend_metal.mm` and
`backend_metal.h` are compiled only into the Metal build. A CPU-only or CUDA build sees
none of this code.

**Runtime.** Within the `#ifdef`, all new paths are additionally guarded by
`g_metal_enabled` (set from `COLI_METAL=1` / `--metal`). With Metal disabled, the unchanged
CPU expert/attention paths run.

**Format-branch isolation.** The kernel change is a new `else if (fmt==4)` clause; the
`fmt==2` (int4 per-row) and int8 branches are byte-for-byte unchanged, and the conditional
final store leaves all non-fmt=4 output bit-identical. The host gate only *adds* fmt=4 to
the allow-list — fmt 3/5/6 still return nil/0 → CPU fallback, as before. Likewise
`a_deqrow` returns exactly `sc[row]` for fmt≠4, so fmt=2 attention is unchanged.

**Fallback safety.** If any expert weight/scale slab fails to resolve (e.g. not
page-registered), `moe_submit` still returns nil → CPU fallback, unchanged. fmt=4 experts
that aren't GPU-resident therefore still compute correctly on the CPU.

Net: fmt=2 decode, CPU decode, and CUDA decode are all provably unaffected; the only
behavioral change is that fmt=4 layers/experts that previously fell to CPU now run on the
GPU.

## 4. How correctness is established

**Kernel unit tests vs an independent CPU oracle** (`make metal-test && ./backend_metal_test`).
Each arm builds registered quantized weights, runs the Metal kernel, and compares against a
from-scratch CPU reference implementing the same math; pass = relative error below
threshold (`nerr < 1e-4` for expert/gemm, `< 2e-4` for attention, plus KV-cache write-back
`< 1e-4`).

The grouped oracle folds the per-group scale *per term*, which algebraically reduces to the
per-row `a·s[o]` when `qgs=0` — so a **single** reference validates both formats, and the
retained fmt=2 arms prove the shared code path is unchanged.

New fmt=4 arms:

- **Routed experts** (`run_moe`, grouped `[O][ng]` scale slab + grouped oracle):
  `moe decode nb=8 fmt4-g128`, `moe ragged nb=6 fmt4-g128`, `moe ragged nb=6 fmt4-g64`.
- **Fused attention** (`run_attn`, grouped `kv_b` + grouped oracle):
  `attn S=1 pos=0 kvb-fmt4-g128`, `attn S=1 pos=37 kvb-fmt4-g128`,
  `attn S=4 pos=12 kvb-fmt4-g128 (MTP)`, `attn S=3 pos=0 kvb-fmt4-g64`.

Both g128 (divides the dims) and g64 arms are included so a non-trivial number of groups
(`ng>1`) and the group-boundary index arithmetic are exercised. All prior fmt=2 arms and
the `gemm-test` grid-cliff regression remain as guards.

**End-to-end A/B on the real checkpoint** (`PROF=1 … COLI_METAL=1 … NGEN=64`):

| Checkpoint | tok/s | Metal attention | Routed experts on GPU |
|---|---|---|---|
| `~/glm52_i4` (fmt=2) | 2.39 | active | yes |
| `~/glm52_i4_g128` (fmt=4) — before | 0.81 | CPU fallback | **no (all CPU)** |
| `~/glm52_i4_g128` (fmt=4) — after | **2.13** | active | **yes** |

The decisive counter is the PROF line at `colibri.c:4918`
(`METAL: … | expert su GPU N | …`): ~0 for fmt=4 before, ≈ the fmt=2 level after. Decode
temp/draft pinned (`TEMP=0 DRAFT=0/MTP=0`) so the comparison is apples-to-apples; residual
gap to 2.39 is expert-cache warmth / I/O, not compute path.

## 5. Build & run

```bash
cd colibri-dev/c
make glm METAL=1                              # engine
make metal-test && ./backend_metal_test       # unit tests (new fmt4 arms should read "ok")
make gemm-test  && ./gemm_largebatch_test     # grid-cliff regression

# real-model A/B (decode)
PROF=1 COLI_METAL=1 TEMP=0 MTP=0 NGEN=64 ./colibri --metal --prompt "…" ~/glm52_i4       # fmt2
PROF=1 COLI_METAL=1 TEMP=0 MTP=0 NGEN=64 ./colibri --metal --prompt "…" ~/glm52_i4_g128  # fmt4
```

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -961,6 +961,8 @@ tests/test_pipe_block$(EXE): tests/test_pipe_block.c colibri.c st.h uring.h json
 
 tests/test_pilot_ring$(EXE): tests/test_pilot_ring.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+tests/test_moe_gs_guard$(EXE): tests/test_moe_gs_guard.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_omp_tune$(EXE): tests/test_omp_tune.c omp_tune.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
--- a/c/Makefile
+++ b/c/Makefile
@@ -967,6 +967,9 @@ tests/test_moe_gs_guard$(EXE): tests/test_moe_gs_guard.c colibri.c st.h uring.h 
 tests/test_omp_tune$(EXE): tests/test_omp_tune.c omp_tune.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
+tests/test_kvb_notice$(EXE): tests/test_kvb_notice.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
 test-c: $(TEST_BINS)
 	$(PYTHON) tools/run_tests.py $(TEST_BINS)
 
--- a/c/backend_metal.h
+++ b/c/backend_metal.h
@@ -80,7 +80,7 @@ int coli_metal_layer_decode(float *x,
     const void *qa_w, const float *qa_s, int qa_fmt, int qa_gs, const float *qa_ln,
     const void *qb_w, const float *qb_s, int qb_fmt, int qb_gs,
     const void *kva_w, const float *kva_s, int kva_fmt, int kva_gs, const float *kva_ln,
-    const void *kvb_w, const float *kvb_s, int kvb_fmt,
+    const void *kvb_w, const float *kvb_s, int kvb_fmt, int kvb_gs,
     const void *o_w, const float *o_s, int o_fmt, int o_gs,
     const void *shg_w, const float *shg_s, int shg_fmt, int shg_gs,
     const void *shu_w, const float *shu_s, int shu_fmt, int shu_gs,
@@ -116,7 +116,7 @@ int coli_metal_attn_decode(const float *x,
     const void *qa_w, const float *qa_s, int qa_fmt, int qa_gs, const float *qa_ln,
     const void *qb_w, const float *qb_s, int qb_fmt, int qb_gs,
     const void *kva_w, const float *kva_s, int kva_fmt, int kva_gs, const float *kva_ln,
-    const void *kvb_w, const float *kvb_s, int kvb_fmt,
+    const void *kvb_w, const float *kvb_s, int kvb_fmt, int kvb_gs,
     const void *o_w, const float *o_s, int o_fmt, int o_gs,
     float *Lc, float *Rc, int S, int pos_base, int st0, float eps, float theta, float ascale, float *out);
 
@@ -137,18 +137,16 @@ int coli_metal_resset_stats(double *flush_s);
  *
  *  D           = hidden size, Iinter = moe intermediate size
  *  g/u/d[e]    = pointers to expert e's gate/up/down quantized weights (in RAM slabs)
- *  gs/us/ds[e] = pointers to expert e's per-row scales
- *  fmt         = quant format (shared across experts). NOTE: fmt=4 (grouped int4) is
- *                NOT yet supported here -- gates to {1,2} and returns 0 (CPU fallback)
- *                for fmt=4 experts, same as before this stage. Grouped-int4 gained GPU
- *                support in mm_gemv (coli_metal_matmul/coli_metal_gemm/bind_gemv) only;
- *                extending the batched routed-expert path is future work (see PR_BODY.md).
+ *  gs/us/ds[e] = pointers to expert e's per-row (fmt=1/2) or per-group (fmt=4) scales
+ *  fmt         = quant format (shared across experts): 1=int8, 2=int4 per-row, 4=int4
+ *                grouped. qgs is the fmt=4 group size (ignored, pass 0, for fmt!=4).
+ *  qgs         = fmt=4 group size shared across experts in this block (0 for fmt!=4)
  *  xg          = packed activations [total_rows, D]; xoff[e] = row offset of expert e
  *  nr[e]       = rows for expert e; rows[]/rw[] map packed rows back to out positions
  *  out         = [S, D] accumulate target
  * Returns 1 on success, 0 to signal the caller to fall back to the CPU path.
  */
-int coli_metal_moe_block(int nb, int D, int Iinter, int fmt,
+int coli_metal_moe_block(int nb, int D, int Iinter, int fmt, int qgs,
                          const void *const *g, const void *const *u, const void *const *d,
                          const float *const *gs, const float *const *us, const float *const *ds,
                          const float *xg, const int *xoff, const int *nr,
@@ -163,7 +161,7 @@ int coli_metal_moe_block(int nb, int D, int Iinter, int fmt,
  * end returns 0 on GPU fault (caller redoes those experts on CPU).
  */
 typedef struct ColiMetalMoeHandle ColiMetalMoeHandle;
-ColiMetalMoeHandle* coli_metal_moe_block_begin(int nb, int D, int Iinter, int fmt,
+ColiMetalMoeHandle* coli_metal_moe_block_begin(int nb, int D, int Iinter, int fmt, int qgs,
                          const void *const *g, const void *const *u, const void *const *d,
                          const float *const *gs, const float *const *us, const float *const *ds,
                          const float *xg, const int *xoff, const int *nr,
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -1237,6 +1237,14 @@ static id<MTLCommandBuffer> moe_submit(int nb, int D, int Iinter, int fmt, int q
       off += n;
     }
   }
+  // COLI_METAL_MOE_EXACT=1: route fmt=4 routed experts to the CPU reference path (bit-exact,
+  // matches matmul_i4_grouped) instead of the fast batched float4 kernel. Opt-in; default is
+  // the fast GPU path. Returning nil makes moe() fall back to CPU for fmt=4 experts, exactly
+  // like an unresolved slab -- fmt=1/2 stay on the GPU, attention/dense are untouched.
+  // (PR #587 gate-2: token-exact mode for trajectories whose gap dips under the drift tail.)
+  { static int g_moe_exact = -1;
+    if (g_moe_exact < 0) { const char *e = getenv("COLI_METAL_MOE_EXACT"); g_moe_exact = (e && e[0] && e[0] != '0'); }
+    if (fmt == 4 && g_moe_exact) return nil; }
   if (g_resset_enabled) {   // E5: commit any pending slab adds before we may skip useResource:
     double t0 = mnow(); resset_flush(); g_t_resset_flush += mnow() - t0;   // METAL-RESSET line
   }
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -1244,7 +1244,11 @@ static id<MTLCommandBuffer> moe_submit(int nb, int D, int Iinter, int fmt, int q
   // (PR #587 gate-2: token-exact mode for trajectories whose gap dips under the drift tail.)
   { static int g_moe_exact = -1;
     if (g_moe_exact < 0) { const char *e = getenv("COLI_METAL_MOE_EXACT"); g_moe_exact = (e && e[0] && e[0] != '0'); }
-    if (fmt == 4 && g_moe_exact) return nil; }
+    if (g_moe_exact) return nil; }  /* exact mode is path-scoped, not fmt-scoped: the resident
+                                     * tier (fmt=1 on mixed containers) carries the same
+                                     * accumulation-order drift, so ALL routed experts fall to
+                                     * CPU under the flag (measured: 4/5 prompt flips -> 0/5,
+                                     * real g64 744B container, #587) */
   if (g_resset_enabled) {   // E5: commit any pending slab adds before we may skip useResource:
     double t0 = mnow(); resset_flush(); g_t_resset_flush += mnow() - t0;   // METAL-RESSET line
   }
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -154,7 +154,10 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
 }
 
 // Batched bindless expert GEMV: each row gr belongs to expert erow[gr], whose weight and
-// scale live at gpuAddresses waddr[e]/saddr[e] (zero-copy in the RAM slab). fmt 1=i8, 2=i4.
+// scale live at gpuAddresses waddr[e]/saddr[e] (zero-copy in the RAM slab). fmt 1=i8, 2=i4
+// per-row, 4=i4 grouped (scale layout [O][ng], ng=ceil(K/qgs) -- same convention as mm_gemv
+// fmt=4 above, but folded into the vectorized uchar4/float4 dot-product loop this kernel's
+// fmt=2 branch already uses, since moe_gemv has no scalar strided branch to reuse).
 // One SIMDGROUP per output row, 4 rows/threadgroup, 8-value loads: measured 1.5-2.1x over
 // one-threadgroup-per-row with uchar2 loads (358-389 GB/s on engine-like block shapes).
 kernel void moe_gemv(device const ulong* waddr [[buffer(0)]], device const ulong* saddr [[buffer(1)]],
@@ -162,7 +165,7 @@ kernel void moe_gemv(device const ulong* waddr [[buffer(0)]], device const ulong
                      device float* yout [[buffer(4)]],
                      constant int& O [[buffer(5)]], constant int& K [[buffer(6)]],
                      constant int& Kin [[buffer(7)]], constant int& fmt [[buffer(8)]],
-                     constant int& NT [[buffer(9)]],
+                     constant int& NT [[buffer(9)]], constant int& qgs [[buffer(10)]],
                      uint tg [[threadgroup_position_in_grid]],
                      uint slane [[thread_index_in_simdgroup]],
                      uint sgid [[simdgroup_index_in_threadgroup]]) {
@@ -210,13 +213,25 @@ kernel void moe_gemv(device const ulong* waddr [[buffer(0)]], device const ulong
                        as_type<float>(uint(b.z)<<16), as_type<float>(uint(b.w)<<16));
       acc+=dot(w0,x4[2*c])+dot(w1,x4[2*c+1]); }
     for(int i=K8*8+slane;i<K;i+=32) acc += as_type<float>(uint(w[i])<<16)*xr[i];
+  } else if (fmt == 4) {                            // grouped int4: per-expert scale [O][ng]
+    int rb=(K+1)/2, ng=(K+qgs-1)/qgs; device const uchar* w=(device const uchar*)(waddr[e])+(long)o*rb;
+    device const float* sr=sc+(long)o*ng;           // grouped scales for this output row
+    device const uchar4* w4=(device const uchar4*)w;
+    for(int c=slane;c<K8;c+=32){ uchar4 b=w4[c];
+      float4 w0=float4(float(int(b.x&0xF)-8),float(int(b.x>>4)-8),float(int(b.y&0xF)-8),float(int(b.y>>4)-8));
+      float4 w1=float4(float(int(b.z&0xF)-8),float(int(b.z>>4)-8),float(int(b.w&0xF)-8),float(int(b.w>>4)-8));
+      int g0=(8*c+0)/qgs,g1=(8*c+1)/qgs,g2=(8*c+2)/qgs,g3=(8*c+3)/qgs;
+      int g4=(8*c+4)/qgs,g5=(8*c+5)/qgs,g6=(8*c+6)/qgs,g7=(8*c+7)/qgs;
+      acc+=dot(w0*float4(sr[g0],sr[g1],sr[g2],sr[g3]),x4[2*c])
+          +dot(w1*float4(sr[g4],sr[g5],sr[g6],sr[g7]),x4[2*c+1]); }
+    for(int i=K8*8+slane;i<K;i+=32){ uchar b=w[i>>1]; int v=(i&1)?(b>>4):(b&0xF); acc+=float(v-8)*xr[i]*sr[i/qgs]; }
   } else { device const char* w=(device const char*)(waddr[e])+(long)o*K;
     device const char4* w4=(device const char4*)w;
     for(int c=slane;c<K8;c+=32) acc+=dot(float4(w4[2*c]),x4[2*c])+dot(float4(w4[2*c+1]),x4[2*c+1]);
     for(int i=K8*8+slane;i<K;i+=32) acc+=float(w[i])*xr[i];
   }
   acc=simd_sum(acc);
-  if(slane==0) yout[row] = (fmt==6||fmt==5) ? acc : acc*sc[o];   // fmt 6: in-block scales; fmt 5: none
+  if(slane==0) yout[row] = (fmt==4 || fmt==5 || fmt==6) ? acc : acc*sc[o];   // fmt4 grouped / fmt6 in-block scales folded; fmt5 none
 }
 
 // fmt=6 activation rotation for the GPU-resident down-projection input: one FWHT
@@ -277,14 +292,20 @@ kernel void a_copy(device const float* src [[buffer(0)]], constant int& off [[bu
 // ---- absorption core (S query rows, per-row causal). q:[S,H*QH]; qabs/clat:[S*H,KVL];
 //      sc:[S*H,T]; ctx:[S*H,VH]. Query row s (abs pos PB+s) attends keys [0, PB+s]. ----
 constant int A_QHH=A_H*A_QH;
-inline float a_deqrow(device const uchar* base, int row, int i, device const float* sc){
-  device const uchar* w=base+(long)row*((A_KVL+1)/2); uchar b=w[i>>1]; int val=(i&1)?(b>>4):(b&0xF); return float(val-8)*sc[row]; }
+// kv_b inline dequant of column i of output row `row`. fmt=2 -> one scale per row;
+// fmt=4 -> grouped int4, one scale per gs-wide group along the A_KVL input dim
+// (scale layout [O][ng], ng=ceil(A_KVL/gs)), matching QT fmt=4 / mm_gemv above.
+inline float a_deqrow(device const uchar* base, int row, int i, device const float* sc, int fmt, int gs){
+  device const uchar* w=base+(long)row*((A_KVL+1)/2); uchar b=w[i>>1]; int val=(i&1)?(b>>4):(b&0xF);
+  float s = (fmt==4) ? sc[(long)row*((A_KVL+gs-1)/gs) + i/gs] : sc[row];
+  return float(val-8)*s; }
 kernel void a_qabs(device const uchar* kvb [[buffer(0)]], device const float* sc [[buffer(1)]],
                    device const float* q [[buffer(2)]], device float* qabs [[buffer(3)]],
+                   constant int& fmt [[buffer(4)]], constant int& gs [[buffer(5)]],
                    uint gid [[thread_position_in_grid]]) {
   int s=gid/(A_H*A_KVL), r=gid%(A_H*A_KVL), h=r/A_KVL, i=r%A_KVL; int rbase=h*A_ROWSH;
   device const float* qp=q+(long)s*A_QHH+(long)h*A_QH;
-  float a=0; for(int d=0;d<A_NOPE;d++) a+=qp[d]*a_deqrow(kvb,rbase+d,i,sc); qabs[(long)(s*A_H+h)*A_KVL+i]=a;
+  float a=0; for(int d=0;d<A_NOPE;d++) a+=qp[d]*a_deqrow(kvb,rbase+d,i,sc,fmt,gs); qabs[(long)(s*A_H+h)*A_KVL+i]=a;
 }
 kernel void a_score(device const float* qabs [[buffer(0)]], device const float* Lc [[buffer(1)]],
                     device const float* Rc [[buffer(2)]], device const float* q [[buffer(3)]],
@@ -312,9 +333,11 @@ kernel void a_clat(device const float* sc [[buffer(0)]], device const float* Lc 
   for(int t=0;t<T;t++) a+=s[t]*Lc[(long)t*A_KVL+i]; clat[(long)sh*A_KVL+i]=a;
 }
 kernel void a_ctx(device const uchar* kvb [[buffer(0)]], device const float* sc [[buffer(1)]],
-                  device const float* clat [[buffer(2)]], device float* ctx [[buffer(3)]], uint gid [[thread_position_in_grid]]) {
+                  device const float* clat [[buffer(2)]], device float* ctx [[buffer(3)]],
+                  constant int& fmt [[buffer(4)]], constant int& gs [[buffer(5)]],
+                  uint gid [[thread_position_in_grid]]) {
   int sh=gid/A_VH, j=gid%A_VH, h=sh%A_H; int row=h*A_ROWSH+A_NOPE+j; device const float* cl=clat+(long)sh*A_KVL;
-  float a=0; for(int i=0;i<A_KVL;i++) a+=cl[i]*a_deqrow(kvb,row,i,sc); ctx[(long)sh*A_VH+j]=a;
+  float a=0; for(int i=0;i<A_KVL;i++) a+=cl[i]*a_deqrow(kvb,row,i,sc,fmt,gs); ctx[(long)sh*A_VH+j]=a;
 }
 
 // ===== full-layer tail kernels =====
@@ -853,14 +876,14 @@ static bool bind_gemv(id<MTLComputeCommandEncoder> e, const void* w, const float
 
 // Weight-pointer bundle for one layer's attention (+optional layer tail). All pointers
 // must be inside registered allocations. *_gs: fmt=4 group size for the corresponding
-// weight (0 if that weight isn't grouped). kv_b has none: it never flows through bind_gemv
-// -- a_qabs/a_ctx dequantize it inline with a PER-ROW-only helper (a_deqrow), so a fmt=4
-// kv_b is out of scope for this stage (see PR_BODY.md UNCERTAINTIES).
+// weight (0 if that weight isn't grouped). kv_b never flows through bind_gemv -- a_qabs/
+// a_ctx dequantize it inline via a_deqrow, which is fmt/gs-aware (fmt=2 per-row, fmt=4
+// grouped along A_KVL); kvb_gs is that group size (0 for fmt=2).
 typedef struct {
   const void *qa_w; const float *qa_s; int qa_fmt; int qa_gs; const float *qa_ln;
   const void *qb_w; const float *qb_s; int qb_fmt; int qb_gs;
   const void *kva_w; const float *kva_s; int kva_fmt; int kva_gs; const float *kva_ln;
-  const void *kvb_w; const float *kvb_s; int kvb_fmt;
+  const void *kvb_w; const float *kvb_s; int kvb_fmt; int kvb_gs;
   const void *o_w;  const float *o_s;  int o_fmt; int o_gs;
 } AttnW;
 
@@ -890,6 +913,7 @@ static bool encode_attn_projections(id<MTLComputeCommandEncoder> e, const AttnW 
     bind_gemv(e,W->qb_w,W->qb_s,W->qb_fmt,W->qb_gs,AQLORA,AHQH,aqr_,aqf_,S); rms(Lb,Loff,akvaln_,AKVL,S); rope(Rb,Roff,0,AROPE,0,1); BAR();
     rope(aqf_,0,ANOPE,AHQH,AQH,AHEADS); BAR();
     [e setComputePipelineState:g_a_qabs]; [e setBuffer:kvbW offset:kvbwoff atIndex:0]; [e setBuffer:kvbS offset:kvbsoff atIndex:1]; [e setBuffer:aqf_ offset:0 atIndex:2]; [e setBuffer:aqabs_ offset:0 atIndex:3];
+    [e setBytes:&W->kvb_fmt length:4 atIndex:4]; [e setBytes:&W->kvb_gs length:4 atIndex:5];
     [e dispatchThreads:MTLSizeMake((size_t)S*AHEADS*AKVL,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
     return true;
 }
@@ -902,7 +926,8 @@ static bool encode_attn_projections(id<MTLComputeCommandEncoder> e, const AttnW 
 static bool encode_attn_core_chunk(id<MTLComputeCommandEncoder> e,
                              id<MTLBuffer> Lb, size_t loff, id<MTLBuffer> Rb, size_t roff,
                              id<MTLBuffer> kvbW, size_t kvbwoff, id<MTLBuffer> kvbS, size_t kvbsoff,
-                             int r0, int ch, int T, int pos_base, float ascale) {
+                             int r0, int ch, int T, int pos_base, float ascale,
+                             int kvb_fmt, int kvb_gs) {
     size_t qabs_off=(size_t)r0*AHEADS*AKVL*4, qf_off=(size_t)r0*AHQH*4, ctx_off=(size_t)r0*AHVH*4;
     int PB=pos_base;
     auto BAR=[&]{ [e memoryBarrierWithScope:MTLBarrierScopeBuffers]; };
@@ -915,6 +940,7 @@ static bool encode_attn_core_chunk(id<MTLComputeCommandEncoder> e,
     [e setComputePipelineState:g_a_clat]; [e setBuffer:ascore_ offset:0 atIndex:0]; [e setBuffer:Lb offset:loff atIndex:1]; [e setBuffer:aclat_ offset:0 atIndex:2]; [e setBytes:&T length:4 atIndex:3];
     [e dispatchThreads:MTLSizeMake((size_t)ch*AHEADS*AKVL,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
     [e setComputePipelineState:g_a_ctx]; [e setBuffer:kvbW offset:kvbwoff atIndex:0]; [e setBuffer:kvbS offset:kvbsoff atIndex:1]; [e setBuffer:aclat_ offset:0 atIndex:2]; [e setBuffer:actx_ offset:ctx_off atIndex:3];
+    [e setBytes:&kvb_fmt length:4 atIndex:4]; [e setBytes:&kvb_gs length:4 atIndex:5];
     [e dispatchThreads:MTLSizeMake((size_t)ch*AHEADS*AVH,1,1) threadsPerThreadgroup:MTLSizeMake(256,1,1)]; BAR();
     return true;
 }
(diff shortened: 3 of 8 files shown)


RDouglasSharp:
Quick note on the overlap with #457: I put this together before I'd properly clocked that #457 already covers grouped-int4 (fmt=4) on the dense GEMV path — apologies for the duplicated effort there. Credit to @monotophic, and #457 should land first.

What's actually additive here are the two paths #457 doesn't touch: the fused-attention kv_b dequant (the inline a_qabs/a_ctx helper is per-row-only today, so grouped kv_b falls back to CPU) and the routed-expert moe_gemv path (still gated to CPU for fmt=4). On a g128 MoE those two are where the decode speedup comes from — the dense kernel alone leaves the experts, which dominate a 744B MoE decode, on the CPU.

Once #457 merges I'll rebase this onto it and drop the overlapping dense changes, so the diff shows only the additive attention + expert work. And if any of it is more useful folded back into #457, I'm glad to do that instead.

monotophic:
Hi! Thanks for the shout-out and the teamwork — much appreciated.

Since we're deconflicting, two pieces I've been holding until the open queue clears, both relevant here:

kv_b: I have a small PR ready that adds a one-time notice when kv_b's format forces fused attention back to the CPU — a diagnostic for exactly the gate your 2a fixes. So they're complementary rather than overlapping, and your change actually improves mine: today my notice text warns that --kvb-bits 4 with --group-size still misses the gate, and once your 2a lands that stops being true — fmt=4 becomes a valid configuration. I'll update and rebase the notice on whichever order things merge.

Routed experts (your 2b): I also have a Metal fmt=4 experts-path branch built and measured behind [#457](https://github.com/JustVugg/colibri/pull/457) — experts-on-GPU gave me +109% at cap-1 on my M5 Max (older dev base), so your 0.83→2.13 matches what I saw. Rather than dueling implementations, I'd rather put review weight behind #587 once it rebases. I've read the diff — it's solid work — and I have a couple of small findings to bring (a latent mixed-group-size edge in the shared-expert inclusion, and one hunk that's inert until [#457](https://github.com/JustVugg/colibri/pull/457) lands since coli_metal_gemm's internal guard still rejects fmt=4), plus a hunk-by-hunk rebase map: [#457](https://github.com/JustVugg/colibri/pull/457) authored the other side of most of your conflicts, including the same falloc→qalloc fix in qt_from_disk — nice independent convergence on that one.

Validation offer: my fmt=4 container is g64 (yours looks like g128) — happy to run the temp-0 CPU-vs-GPU token-identity check on real weights on my M5 Max against your branch, after rebase or on the current head if sooner is useful.

Agreed on [#457](https://github.com/JustVugg/colibri/pull/457)-first sequencing. Happy to see all of it merge regardless of whose diff carries it.

RDouglasSharp:
I merged the two as discussed.  The temporary merge branch is currently at https://github.com/RDouglasSharp/colibri/tree/merged-457-587  .

monotophic:

Thanks for standing the merge branch up so quickly — it builds clean and passes the Metal suite on my M5 Max, and the #457 commits came through with authorship intact, appreciated. One catch from validation, with a two-line fix: the colibri.c conflict resolution dropped the two `kv_b` guard relaxations (lines 2353/4078 still `fmt==2`-only), so 2a is unreachable on the branch as merged. Details, the full validation results (g64 attention path now token-identical CPU↔Metal with engagement counter-verified), and the hunk-disposition map are on #585.

RDouglasSharp:
My mistake.  The correct combined code is now on https://github.com/RDouglasSharp/colibri/tree/merged-457-587

JustVugg:
**Two open PRs now add fmt=4 (grouped-int4) to the Metal path — #457 (@monotophic, the GEMV kernel + two latent fmt=4 fixes) and #587 (@RDouglasSharp, attention + routed-expert decode).** They overlap on the same fmt=4 Metal surface, so we can't merge both blind, and there's a verification gap: **CI builds the Metal target but never executes it on a GPU**, so a green check proves it compiles, not that it's token-correct.

Two things unblock this, and then whichever clears them lands first (the other rebases on top):

1. **Reconcile the overlap.** Between you two: does one supersede the other, or are they complementary (GEMV kernel vs attention path) and mergeable in sequence? If complementary, say which goes first.
2. **Token-exactness from real hardware.** A short run on real Apple Silicon showing fmt=4 (a gs64 container) output is **bit/token-identical to the CPU path**, not just faster — the fmt=4 oracle pattern from #298. Post the diff (ideally the `make metal-test` correctness check on an fmt=4 fixture, or a greedy decode matching CPU).

Once (1) is agreed and (2) is posted for the leading PR, we merge it. This is the last gate — the work itself looks good, we just won't ship an unverified GPU decode kernel.

lBroth:
Same hardware run I posted on #457, for this one, plus the container half of the gate, since this is the PR whose new paths a container actually exercises.

I merged `2219144` into `dev` at `7b03ce0`, clean, no conflicts, and ran the gates on the merge rather than the branch since the branch is 21 commits behind. M5 Pro, 48 GB, macOS 26.5.2, Apple clang 21.

`make metal-test` passes: 38 cases, no failures, repeatable across runs. The fmt=4 part, which is the routed-expert and kv_b work this PR adds:

```
  moe decode nb=8  fmt4-g128     R=8 nerr=2.30e-06  ok
  moe ragged nb=6  fmt4-g128     R=16 nerr=2.07e-06  ok
  moe ragged nb=6  fmt4-g64      R=16 nerr=2.17e-06  ok
  attn S=1 pos=0   kvb-fmt4-g128 nerr=4.27e-06 cache=1.29e-05  ok
  attn S=1 pos=37  kvb-fmt4-g128 nerr=5.27e-06 cache=1.31e-05  ok
  attn S=4 pos=12  kvb-fmt4-g128 (MTP) nerr=5.71e-06 cache=1.29e-05  ok
  attn S=3 pos=0   kvb-fmt4-g64 nerr=5.46e-06 cache=1.05e-05  ok
```

`make check` on the same tree is 186 tests, OK with 13 skipped.

Then the gs64 container, built locally with `tools/convert_fp8_to_int4.py --indir glm_tiny --group-size 64`, the tiny oracle snapshot converts in seconds. The routed-expert path engages for real: over a 20-token greedy run the counters read 42 GPU blocks, 85 experts on GPU, 0 fallbacks. As far as I can tell that is the first time this path has run on hardware against a container. Three measurements, offered as data:

Teacher-forced, stock dev already disagrees CPU-versus-Metal at two positions out of 32 on this container, before any of this PR's paths exist. On this merge one of the two comes back into agreement with the CPU, and one remains, position 27, CPU 7 versus Metal 112.

Greedy, stock is token-identical CPU-versus-Metal, since fmt=4 all falls back there. On this merge greedy diverges from the fifth token onward.

All of it is deterministic. Three Metal runs produce the same sequence, two CPU runs the same.

A caveat that cuts both ways: this fixture container is pathological. Four-bit quantizing a toy model leaves near-tie logits everywhere, it matches its own fp16 oracle at only 7 of 32 teacher-forced positions, so numeric differences that would rarely flip a token on a real container flip several here, and teacher-forcing localizes the whole thing to one position in 32. I cannot say which way it goes on a real g64 container, none fits on 48 GB. But the gate as stated is a greedy decode matching the CPU, and on this fixture, with the new paths engaged, it does not yet, while the same fixture on stock and on the #457 merge does. Output diffs and exact commands are one comment away if useful.

monotophic:

@lBroth — thank you; the engagement counters (42 GPU blocks / 85 experts / 0 fallbacks) make this the first real hardware exercise of the routed-expert path, and the divergence you caught is exactly what that instrumentation exists to catch.

**A concrete hypothesis for the greedy divergence, offered for testing rather than asserted.** In our line-by-line review of this PR (notes on #585), we flagged one latent we called F2: `MB_BUILD` captures the batch's quantization *format* from its first member and now (correctly) captures the group size too — but it never checks that **group size is equal across all batch members, including the shared expert**. Format is checked; gs is not. A mismatched member silently dequantizes on the wrong group stride. That failure mode is invisible on uniform containers and fires precisely on small heterogeneous fixtures — which matches the signature here: the #457-only merge (no routed batching) stays token-identical on the same fixture, while this merge diverges once the routed path engages.

**We hold a tested guard for exactly this** — `mb_gs_compat`: capture the batch gs, reject any mismatched batch to the existing fallback path, with a unit test. It came out of the same review; it's a few lines and folds naturally into the post-#457 rebase along with the disposition map and the two `colibri.c` gate relaxations from our earlier note.

**The discriminating experiment is cheap, and you offered cheap re-runs:** same fixture, same merge, plus the guard patch. If greedy returns to CPU-identity, the diagnosis is confirmed and the fix simply rides the rebase; if it doesn't, we've eliminated the leading candidate and the map's kernel-variant notes (the `mm_gemv` restructuring differences) are the next place to look. We can hand the guard as a patch against any tree you name; alternatively @RDouglasSharp folds it into merged-457-587 directly.

One separable observation from your data, worth not conflating with this PR: the two teacher-forced disagreements on *stock* dev (one of which your merge heals, one persists at position 27) are a pre-existing fmt=4 prefill-path item — likely deserving its own small issue once this lands.

Sequencing per the maintainer's rule stands: #457 first (its gate evidence is complete), this PR rebases with map + fix + guard, then the hardware gate re-runs on the rebased head — your standing offer covers the fixture side, and (me/my) M5 Max covers a real g64 container run at full scale.

lBroth:
Ran the discriminating experiment, and it took the investigation somewhere neither of us expected.

First the direct answer. With the fused-attention call forced off and the routed-expert path left on, the divergent greedy sequence is bit-identical to the full-Metal run, 85 experts on GPU. Then a stronger version of the same fact: the fused attention never engages on this fixture at all, the METAL-ATTN counter stays at zero on every tree, presumably because the kernel is compiled for GLM-5.2 dims and the toy snapshot has different ones. So the divergence lives entirely in the routed-expert path.

On F2 specifically, the precondition is absent here: all batch members carry the same layout. Gate and up detect as gs=64; down has a 32-wide input, a single partial group where any gs of 32 or more aliases to the same scale index; the shared expert matches the routed ones. And the teacher-forced localization, one position in 32, looks like accumulation-order drift rather than a wrong stride, which would corrupt broadly. So a gs mismatch cannot explain this particular divergence, and your kernel-variant notes are the next place to look.

But your guard's shape is right, and testing it caught me. While checking whether a per-member skip would even work as a control, I found that the batch contract makes any per-member skip wrong: the row-copy pass walks the whole class, so a skipped member desyncs the offsets, and the CPU loop skips experts by class, not by batch membership, so the skipped contribution is silently lost. My int3 branch's format guard had exactly that bug. I spliced a mixed fixture (one expert per layer rewritten to per-row int4 inside an int3 container) and instrumented the guard: it fires exactly once, layer 3 miss class, and teacher-forced the per-member skip produces wrong tokens at several positions beyond the container's known prefill drift, deterministically, reproduced bit-for-bit by re-simulating the old behavior on the fixed tree. Rejecting the whole class to the CPU, your mb_gs_compat shape, restores CPU-identity modulo that same pre-existing drift. Fixed on my branch, credited to this review.

So: F2 does not explain the #587 fixture divergence, but the whole-batch-reject design it comes from is load-bearing, and I would fold mb_gs_compat into the rebase regardless. Standing offer holds for re-running the fixture side on the rebased head.

monotophic:

@lBroth — that's a beautifully run discrimination, and we'll score our own hypothesis first: **F2-as-the-cause is refuted for this divergence.** Fused attention never engaging (METAL-ATTN at zero — dim-gated against the toy snapshot) plus uniform layouts across batch members removes the precondition entirely; we offered F2 as the leading candidate and your experiment eliminated it cleanly. That's what the outcome pre-registration is for.

And the finding was more informative than if the hypothesis was confirmed: the mixed-fixture demonstration that **any per-member skip is structurally wrong under the batch contract** (offset desync + silently dropped contributions) is exactly the failure class the whole-batch-reject shape exists for — finding and fixing a live instance of it on your int3 branch is the best validation `mb_gs_compat` could have gotten. Agreed on folding it into the rebase regardless; @RDouglasSharp, it's ready to hand whenever you pick up the map.

On the actual divergence, one datapoint we hold narrows it usefully: our integration tree ran **this same routed-expert kernel** (the `moe_gemv` fmt=4 branch — H6 in the disposition map, applied verbatim from this PR) **token-identical CPU↔Metal over a real 744B g64 container** — the routed-experts half of our validation arc, the half that survived our own correction pass (the attention half was the one we retracted and later counter-proved separately). Same kernel, real container: identical; toy fixture: diverges from token 5. Combined with your one-position-in-32 teacher-forced localization, that pattern says *fixture-conditional near-tie sensitivity* — accumulation-order drift crossing a tie on a pathological distribution — rather than systematically wrong math.

The discriminating next step is cheap, and we're running our half of it now rather than proposing it: **underway on our side** — (1) we're refreshing our integration tree (this PR's hunks per the disposition map + the guard) onto current dev — i.e., the shape the post-#457 rebase will take — and running the full **real-container g64 temp-0 CPU-vs-Metal identity suite on it**, results posted here when done; (2) that tree carries top-2 logit-gap instrumentation at every decoded position, so the same run reports the real container's margin distribution — if the near-tie theory is right, real-weights margins should sit far above float-rounding territory while your pathological fixture's sit inside it. The other half only you can run (you hold the fixture): the same instrumented tree, your fixture, one command — we'll point you at the branch when the refresh is up, which turns your standing offer into a zero-setup run. A gap in rounding territory confirms the order-drift class and the fix is mechanical (match the fmt=4 branch's accumulation shape to the CPU expert loop — the discipline that got #457's dense path to bit-identity); a large gap means we're wrong twice and your fixture is the minimal reproducer of something real.

RDouglasSharp:
Agreed on the sequencing: #457 first, #587 rebases on top. They're complementary, not competing — #457 owns the dense GEMV kernel + the two latent fmt=4 fixes; #587 adds the fused-attention kv_b and routed-expert decode paths.

I've already stood up that shape: merged-457-587 rebased onto current dev → merged-457-587-ondev. It's monotophic's two #457 commits (author attribution intact) plus the #587 work, clean rebase, no conflicts. backend_metal.mm/.h come through byte-identical to the reviewed #587 kernel, and the colibri.c merge preserved every fmt=4 hunk (both guard relaxations, the kv_b.gs decode args, the mgrp expert threading). Secondary note, since it came up: that rebase also closes the small decode-throughput gap I'd been seeing between the two branches — it was base drift (~a dev cycle), not anything in the fmt=4 paths.

What it does not yet carry: mb_gs_compat and the disposition-map hunks. @monotophic — I'll fold those in directly; hand them against merged-457-587-ondev (or name the tree you'd prefer) and I'll apply and re-run metal-test. Agreed the whole-batch-reject shape is worth carrying regardless of whether F2 turns out to be the fixture's cause — and good catch on the per-member-skip desync, @lBroth.

On the divergence itself, the near-tie / accumulation-order read matches what I'd expect: attention never engages on the toy fixture (dim-gated, so it's purely the routed path), the same moe_gemv branch is token-identical on a real g64 container, and the localization is one position in 32. I'll run the fixture side of the instrumented experiment as soon as the branch is up — point me at it. And +1 on spinning the pre-existing stock-dev prefill disagreement (position 27) into its own issue rather than blocking here.

Once the guard + map are in and the real-container temp-0 CPU↔Metal identity run posts, the rebased head is the thing to gate on.

monotophic:

The probe we said we'd run has results — they start with a correction, and they ended up answering more than initially asked, because #622/#624 landed mid-analysis and gave us the missing control.

**Correction first.** The real-container datapoint we leaned on above — "same kernel, real container: identical" — was **prompt-specific**. On a different fixed prompt (temp-0 greedy, MTP off, 744B g64 container on my M5 Max), the integration tree diverges CPU↔Metal at generated position 9, and a control on the pre-refresh integ head diverges identically, so the refresh didn't introduce it. Each backend is individually deterministic (byte-identical across repeated runs and across a 16-commit rebase); this is systematic cross-backend drift, not flakiness. lBroth's fixture divergence deserved more credit as representative than we gave it.

**The full matrix.** After @JustVugg's #624 landed the `COLI_METAL_GEMM_MIN` isolation protocol, we ran both trees × three arms on the same prompt/container (all on current dev `6fcc555`; integ head = `4b6d56e` = #457 + this PR's hunks per the disposition map + `mb_gs_compat` + a top-2 logit-gap dump):

| tree | arm | greedy vs CPU |
|---|---|---|
| #457-only (`649b552`) | Metal, default | diverges @ pos 9 |
| #457-only | Metal, `GEMM_MIN=100000` | **token-identical, 64/64** |
| integ (+ this PR's paths) | Metal, default | diverges @ pos 9 |
| integ (+ this PR's paths) | Metal, `GEMM_MIN=100000` | **still diverges @ pos 9** |

Three attributions fall out:

1. **#457's kernels are token-exact on real 744B weights** once the pre-existing #622 prefill drift is excluded — including its decode GEMV and the serial one-expert-per-block GPU expert path active on that tree.
2. **The #622 prefill-GEMM drift flips real containers, not just toy fixtures**: it alone accounts for the #457-tree divergence (prefill-KV perturbation), on the first real prompt we tried. One datapoint for the crossing-rate question #624 deferred: real margins are not comfortably wide (below).
3. **This PR's added paths carry their own drift**: with prefill fully clean, the integ tree still flips the same pair at position 9. That is the divergence this PR's gate-2 is about, and it is real on real weights.

**The gap numbers.** No near-ties anywhere: CPU per-position top-2 gaps run min 0.375 / p10 1.07 / median 8.3, with nothing below 1e-2 in any arm — so the rounding-scale near-tie story is dead at real scale, including the version of it we proposed. We were wrong twice, per our own pre-registered scoring. What the instrument shows instead: pre-divergence cross-backend |Δgap| medians of 0.35–0.54 (max 1.4–2.2) — roughly five orders above the sub-3e-6 fixture-scale drift — and the flip lands exactly at the trajectory's first gap under ~2 (1.67 at pos 9). With p10 gaps ≈ 1.1, positions inside that drift band are prompt-routine. Mechanism, in one line: **flips occur wherever the gap dips below the drift tail, which is why identity is prompt-dependent on every tree.**

**The encouraging part — a proven-exact target exists on the same tree.** The serial GPU expert path is token-exact on the identical container/prompt where the batched `moe_gemv` path diverges. So exactness is demonstrably achievable at this scale on this hardware; the batched kernel's per-expert accumulation shape (and possibly the fused-attention kv_b path — our run can't separate the two; @lBroth's fixture, where fused attention never engages, can) just needs to match the serial path's or the CPU loop's, the same discipline that got #457's dense path there.

**@RDouglasSharp** — the branch is the handoff: `integ/post-wave-r2` @ `4b6d56e` on my fork carries the disposition-map hunks and `mb_gs_compat` already applied on current dev, plus the probe. Lift them from there onto `merged-457-587-ondev`, or say the word and we'll hand them as patches against your tree. **@lBroth** — one command stands: `COLI_LOGIT_GAP=1` on that branch dumps `pos/top1/top2/gap` per greedy step; on your fixture it prices the token-5 flip, and running it under both `COLI_METAL_GEMM_MIN` settings applies the same isolation split to the batched path in isolation.

Caveats on record: one prompt, one container; the integ tree's prefill also engages the batched path, so this PR's drift may partly enter at prefill; serial-path exactness is shown for prefill+decode jointly, not per-phase. Full arms/scripts/analyzer archived; margin distributions across a prompt set is the natural next instrument if #624's crossing-rate question needs more than one datapoint.

RDouglasSharp:
Gate-2 datapoint from a real g128 container (M-series Max, 128 GB, experts streamed from disk).

Ran greedy temp-0, MTP off, Metal vs CPU with the top-2 logit-gap probe (COLI_LOGIT_GAP=1) across 5 prompts / ~450 generated positions (64–96 tokens each). Zero top1 flips — token-identical CPU↔Metal on every prompt. The batched fmt=4 routed-expert path (this PR's moe_gemv) engages throughout; fused-attention kv_b also engages (real GLM-5.2 dims), so both fmt=4 decode paths are exercised.

The probe confirms the drift is real but sub-decision: per-position |Δlogit| is typically ~1e-3, with a single outlier ~0.29 that still sat well under that trajectory's gap. That's the same magnitude reported on g64 — the difference is only whether a trajectory's decision gap dips into the drift band, which is prompt/container-dependent. On g128, across these prompts, it does not.

So on real g128 weights the greedy-identity gate holds empirically. The open item is the g64 case reproduced earlier; I don't have a real g64 container to test locally (the pathological toy fixture is a separate, known-degenerate case). If g64 turns out to need CPU-order accumulation to stay identical at that grouping, the plan is to put that behind an opt-in flag (COLI_METAL_MOE_EXACT) and keep the fast float4 batched path as the default — i.e. demonstrate exactness with the flag on, ship speed with it off, so no one eats a slowdown by default.

Raw per-position gap dumps (both backends, all 5 prompts) available on request.

lBroth:
Ran the fixture side on integ/post-wave-r2 at 4b6d56e. The isolation you wanted falls out, with one structural surprise worth stating first.

On this harness the two COLI_METAL_GEMM_MIN arms are bit-identical, and they should be: the greedy prompt is 12 tokens, below the 16-row threshold, so matmul_qt never reaches the GPU GEMM in greedy on this fixture in either arm. The isolation is structural rather than flag-driven. With the fused attention dim-gated off (METAL-ATTN never prints) and the dense matmuls on the CPU by threshold, the only path on the GPU is the batched routed-expert one, 85 experts across the run, and stock dev at the same settings is token-identical with zero experts on the GPU. So the fixture's residual greedy divergence attributes to the batched moe path alone, and kv_b is exonerated here, which is the split my earlier runs could not make and yours could not either.

Numbers from the probe, pre-divergence only (positions 0 to 3, trajectories comparable): cross-backend |Δgap| median 1.1e-2, max 2.0e-2. The flip lands at position 4, the trajectory's first gap under that band, 9e-4. Same mechanism you measured at real scale, flip where the gap dips under the drift tail, at roughly 30 to 50x smaller drift than your g64 run's 0.35 to 0.54. So the toy fixture is degenerate in scale, not in kind, and it prices the batched path's own contribution with everything else off.

This is consistent with the serial-path-exact observation and with @RDouglasSharp's plan: the surface is the batched kernel's accumulation shape, and an opt-in exact mode plus the fast default seems like the right trade given g128 crossed nothing in 450 positions. Determinism held on every run here, and the standing offer stays for the rebased head.

RDouglasSharp:
Implemented the opt-in exact path — `COLI_METAL_MOE_EXACT=1`.

When set, fmt=4 routed experts fall back to the CPU reference (`matmul_i4_grouped`) via the
existing nil-return-from-`moe_submit` fallback; fmt=1/2 stay on the GPU, and the fused-attention
`kv_b` path is untouched (@lBroth's fixture already exonerated it). Unset, nothing changes — the
fast batched `float4` kernel is still the default. Fast by default, bit-exact on demand.

**Routing verified on a real g128 container** (temp-0 greedy, decode):

| mode | `expert su GPU` | `METAL-ATTN layer GPU` | expert-matmul |
| ---- | --------------- | ---------------------- | ------------- |
| default | 41454 | 4914 | 12.9s |
| `COLI_METAL_MOE_EXACT=1` | **0** (`blocchi GPU 0`) | 4914 | 57.4s (CPU-side) |

So experts drop cleanly off the GPU while attention stays on Metal. g128 itself can't demonstrate
the flip-fix — it crosses no gap in 450 positions in either mode — so this is the hunk for the
standing offer: **run `COLI_METAL_MOE_EXACT=1` on the fixture (pos-4 flip) and the g64 container
(pos-9)**; both should go token-identical, which closes gate-2 with the fast path preserved.

Branch (zero setup — it's `integ/post-wave-r2` @ `4b6d56e` + this one hunk):
**`RDouglasSharp/colibri:metal-moe-exact`**

Plan on confirmation: this folds onto the post-#457 rebased head alongside `mb_gs_compat` + the
disposition-map hunks — fast batched ships as the default, exact mode is the on-demand
token-exact / verification path.

monotophic:

As a potential alternate to the exact-mode flag being designed in: we think the drift is repairable at its source, and we're attempting the repair — declared now so the work doesn't collide with anyone else's.

**Why we think it's repairable without a speed tax.** Reading the batched fmt=4 branch of `moe_gemv` against the CPU loop, the drift has two structural sources, and the larger one looks like a free fix:

1. *Scale placement:* the batched branch folds the group scale into every weight before the dot (`dot(w0*float4(sr[g0..g3]), x4)`) — one extra float rounding per element, at scale values that get finer as the grouping gets finer. The CPU accumulates each group's partial first and scales once per group. This is a candidate explanation for why the measured drift is ~1e-3 on g128 but 0.35–0.54 on my g64 (more groups, finer scales, 64 roundings per group instead of one) — and repairing it *removes* float work rather than adding it.
2. *Lane-striding:* each SIMD lane's accumulator mixes elements across many groups before the `simd_sum`, a completely different association than the CPU's sequential-by-group order.

One more datapoint says the fix class works: the serial one-expert-per-block GPU path is empirically token-exact on the same g64 container where the batched path flips — same hardware, same weights, different accumulation shape. The batching speedup itself lives in dispatch structure, not in the accumulation order, so reordering the math should keep the throughput. (We measured the batching win at ~2.5× on this container; nobody wants to lose it, least of all us.)

**The plan, pre-registered:**
- *Stage 1 (building now):* group-local accumulation + one post-scale per group in the batched branch, Metal-side only, dispatch geometry untouched. Predictions on the record: on a 5-prompt real-g64 sweep (temp-0, decode-isolated per #624's protocol), greedy flips go to zero and per-position drift collapses to the g128 class or below; champion-class throughput on the same container stays within noise of the current batched path. If either fails, we'll say so with the numbers.
- *Stage 2 (design question, flagged for @RDouglasSharp / @JustVugg):* true exactness-by-construction needs a common documented reduction order on both sides — either a shared block-pairwise order (touches CPU kernels; shifts historical CPU outputs slightly) or Metal replicating the CPU's sequential-group order. That's an architecture conversation, not a patch we'd spring on anyone.
- If stage 1 misses its bars, the `COLI_METAL_MOE_EXACT` flag remains the sensible fallback and our own speed data supports the fast-default shape — this attempt is about making the flag unnecessary, not about opposing it.

Mechanics: the baseline sweep is running now on my g64 container (current kernel — the "before" distribution); the patch lands as a tested diff offered here for this PR to fold, same as the guard. Validation package will include the before/after sweep, an fmt=2 control (must be byte-unchanged), and interleaved old/new throughput pairs at champion-class.

monotophic:

Results from the declared repair attempt, plus the flag run @RDouglasSharp asked for — and the attribution moved somewhere we didn't predict, so a correction leads again.

**Correction first.** Our earlier claim that **#457's kernels are token-exact on real 744B g64 weights** was measured on one prompt, and one prompt was not enough. On the same 5-prompt sweep protocol used below, the pure-#457 tree (`649b552`, serial GPU experts + dense GEMV, prefill isolated per #624) flips **4 of 5 prompts**. The single surviving prompt was the one our original matrix used. We were reporting prompt-luck as exactness — the same failure mode we corrected once before on this thread, now caught in our own strongest claim. @JustVugg — this bears directly on gate-2 for #457's merge decision, so flagging it before anything lands on our stale datapoint.

**The full matrix** (real g64 container, 5 prompts, temp-0 ngen-64, `COLI_METAL_GEMM_MIN=100000` everywhere; cells = first flip position, — = token-identical):

| condition | p1 | p2 | p3 | p4 | p5 | flips |
|---|---|---|---|---|---|---|
| batched + fused-attn (this PR's paths) | 9 | 7 | 62 | — | 36 | 4/5 |
| our stage-1 "repair" kernel | 9 | 7 | 62 | 3 | 36 | 5/5 |
| `COLI_METAL_MOE_EXACT=1` | 55 | 7 | 62 | — | 6 | 4/5 |
| flag + fused-attn forced off | — | 7 | 62 | 3 | 35 | 4/5 |
| pure #457 tree | — | 7 | 62 | 3 | 35 | 4/5 |

Pre-divergence drift medians 0.15–1.07 logits in every condition. Three conclusions fall out:

1. **The g64 drift lives in the shared fmt=4 GPU GEMV family, present since #457** — not in this PR's additions. #587's batched path and fused attention shift *which* prompts and positions flip (each path contributes its own rounding pattern), but they add no measurable increment to the drift class. The convictions we and lBroth accumulated against the batched path were made against a baseline that wasn't clean.
2. **Our stage-1 repair is refuted** per its own pre-registration (5/5 flips; the scale-placement theory died exactly as our audit warned it might). One salvage: the restructured kernel incidentally ran its expert compute ~2× faster at equal (non-)exactness — relevant if a defined-order kernel design ever gets built.
3. **The flag as implemented doesn't close gate-2 on real g64** (4/5 flips) — through no fault of its design intent: on this tree the flag's early-return routes experts to the *serial GPU* path rather than CPU (`blocchi GPU 75`, not the `blocchi 0` your g128 table showed — worth reconciling; config-dependent path selection is our guess), and the fused-layer kernel (`coli_metal_layer_decode`), which carries most decode attention, sits outside the flag's gate entirely.

**What the state of gate-2 now is:** on real g64 containers, *no current fmt=4 GPU configuration is token-exact against CPU* — batched or serial, attention on or off, flag on or off — while @RDouglasSharp's real-g128 runs are flip-free with the same kernels. The gs-dependence (finer groups → more scale roundings per row?) or a container-layout difference is now the central open question, and it's answerable: we hold the g64 container and the instrument, and can mint intermediate-gs variants from the retained fp8 source if useful. Options as we see them, in the #624 spirit: scope the decode token-exact claim by grouping (what the docs did for prefill), or treat exactness as a kernel-family design goal (defined reduction order — a real conversation, not a patch), or extend the flag to genuinely cover all fmt=4 paths as the verification mode. That choice is yours and the maintainer's; our part was making sure it gets made on data rather than on our earlier one-prompt claim.

Raw sweep arms, per-position gap dumps, toggle-probe commits, and the scored repair spec are all archived and available on request.

monotophic:
Gate-2 update on #457: the allpaths re-sweep result is in — 4/5 with the flag
as shipped (mixed-tier container leaves fmt=1 residents on GPU), **0/5
token-exact with a one-line fmt-agnostic expert gate** (pre-registered,
counters-verified). Revises our earlier
fmt=4-family attribution: the residual drift class rides the expert tier
generally, not the fmt=4 kernels specifically.

RDouglasSharp:
This is great work — thank you for running it properly, and especially for pre-registering the fmt-agnostic patch before the re-run. The 0/5 token-exact result on a real g64 container is exactly the outcome I was after.

Happy to concede the scorecard: right on the mechanism, wrong on the tier my fixture couldn't represent. A uniform-format fixture was never going to surface the mixed-tier case where fmt=1 residents keep riding the batched kernel, so the fmt == 4-scoped gate sailing past them makes complete sense in hindsight. Your path-vs-fmt reframing is the correct one — the residual drift is the expert-tier accumulation class, not the fmt=4 kernels specifically.

I'm on board with the fix and the tradeoff: scoping COLI_METAL_MOE_EXACT by path (all routed experts to CPU under the flag) is the right call, and moving the resident tier to CPU in exact mode is consistent with the flag's opt-in-exactness contract. The fast default stays untouched, which is the part I care about.

Since you've already got the one-liner working and validated — with counters and provenance — it'd be best if you make the change and commit it directly; the credit for the diagnosis and the fix is yours. And yes, please do attach the results doc (L10_ALLPATHS_RESULTS_2026-07-28.md) so the positions and engine md5s are on the record alongside the commit.

Thanks again — this is the first GPU-enabled config that's token-exact on a real container, which is a genuinely nice result to land.

monotophic:

@RDouglasSharp — thanks for the generous read. Here's the change as a
`git am`-able patch so it lands with one command and zero transcription risk.

**Apply** (from your `metal-moe-exact-allpaths` checkout — the patch is cut
against d7ca0ed):

```bash
git am PATCH_587_path_scoped_gate.patch.txt
```

[PATCH_587_path_scoped_gate.patch.txt](https://github.com/user-attachments/files/30464944/PATCH_587_path_scoped_gate.patch.txt)

If it ever conflicts after your rebase, the change is one line:
`backend_metal.mm` expert-gate `if (fmt == 4 && g_moe_exact)` →
`if (g_moe_exact)` — plus the comment block.)

**Verify** (30 seconds, no sweep needed): any prompt, flag on, real mixed
container — the engagement line should read
`blocchi GPU 0 | fallback CPU 0 | expert su GPU 0`. With the fmt-scoped
gate it reads `75 | 0 | 75` on our container.

**Provenance for the record** (full detail in the attached results doc):
- Sweep protocol: identical to the #587 matrix (5 prompts, temp-0 ngen-64,
  `COLI_METAL_GEMM_MIN=100000`, real 744B g64 container + kv_b overlay).
- fmt-scoped gate @ d7ca0ed: 4/5 flips (positions —/7/62/3/35), engine md5
  35bd12d4.
- Path-scoped gate (this patch): **0/5 — token-exact**, engine md5 305b16fa.
  Pre-registered before the run.
- The shipped patch differs from the validated diagnostic build only in the
  comment text: `__TEXT` sections of the two binaries are md5-identical
  (3d1a7f02…), so the machine code you're applying is byte-for-byte the
  code that scored 0/5.

Fast default untouched, as you said — the flag just now means what it says
on the whole expert path.

*(Calibration coordinates: positions/md5s are for our container + d7ca0ed
lineage, macOS M5 Max, 2026-07-28; the mechanism and the gate semantics are
the durable part.)*

Requested results:
[L10_ALLPATHS_RESULTS_2026-07-28.md](https://github.com/user-attachments/files/30465010/L10_ALLPATHS_RESULTS_2026-07-28.md)

RDouglasSharp:
Landed and verified. Applied with git am (your authorship on the commit) and pushed to metal-moe-exact-allpaths.

Confirmed on a real container: flag off → METAL: blocchi GPU 166; flag on → blocchi GPU 0; output correct ("2 + 2 = 4") both ways. Experts fully off the GPU in exact mode, fast default untouched — matches your 0/5.

Thanks for the clean patch and the pre-registered validation; one-command land, no second-guessing. This is the one-liner that makes the flag mean what it says across the whole expert path.

JustVugg:
Heads-up, and I want to be transparent rather than pick a winner quietly: **#457 (@monotophic) and #587 (@RDouglasSharp) implement the same thing** — Metal grouped-int4 (`fmt=4`) — and they conflict with each other in `c/backend_metal.h`, `c/backend_metal.mm` and `c/colibri.c` (4 conflicts). Each is clean against `dev` on its own; merging either one forces the other into a rebase or makes it redundant.

Both of you have put real work in and I would rather you two decide how to combine it than have me choose based on merge order. It worked well last week: @steve-m and @ZacharyZcR hit the same overlap on Kimi and resolved it by one absorbing the other's commit with authorship preserved (#676/#684) — clean history, credit intact, no duplicated effort.

Whatever you agree on, I will merge. If it helps, tell me which one should be the base and I will hold the other until it rebases on top.

monotophic:
@JustVugg we worked together and agreed that #587 follows #457. The two PRs are complementary and have been tested together extensively. This thread contains a record of that coordination and testing. If there is anything you would like to see, say the word. I'd love to see these two PRs merge.

This one's yours @RDouglasSharp, anything to add?

RDouglasSharp:
Just as monotophic said, these are intended to stack.  Merge #457, then #587.  We have tested the combination extensively.

JustVugg:
**#457 is merged** — thank you both for sorting the order out between yourselves rather than leaving it to merge luck. That's the second time this week two contributors resolved an overlap by coordinating (steve-m/ZacharyZcR did it on Kimi), and it's a better outcome than anything I would have picked.

@RDouglasSharp — this now needs a rebase before it can land. Since you applied monotophic's patch with `git am` onto this branch, those commits are now in `dev` too, so the branch conflicts with itself: 4 conflicts in `c/backend_metal.h`, `c/backend_metal.mm` and `c/colibri.c`. After `git rebase origin/dev` the duplicated commits should drop out and this PR reduces to your own delta — the same shape #684 took once #676 landed.

CI is green as it stands, so this is purely mechanical. Ping me when it's clean and I'll merge it straight away — the stack is already agreed, I'm not re-litigating anything.

JustVugg:

RDouglasSharp:
Rebased onto current dev and green. The conflicts were the expected fmt=4 (this PR) vs fmt=6 (now in dev) overlap across moe_gemv, the fused-attention gates, and the moe_block signature — all resolved as unions (both formats coexist). make metal-test passes every arm including fmt4-g128/g64, e8, and kvb-fmt4; make check green. One small reconcile commit updates two of dev's newer test call sites for the qgs param. Ready to merge.

JustVugg:
@RDouglasSharp — you told me on 1 August this was rebased and green and ready, and it has sat since. That is on me, and there is now a second thing I owe you.

**#829 opened today and does the same work.** @aaristov, a first-time contributor, hit the same defect from #813 and wrote a fix touching the same four files this does. He had no way to know you were two weeks into it — nothing on #585 or #813 says so, and the only reason I spotted it is that I compared file lists.

I have not merged it, and I am not going to merge it over you. **I have proposed on #829 that this PR lands the decode path first**, and that his work rebases onto yours as a focused follow-up carrying two things I do not see here:

- `run_serve` calling `profile_print` under `PROF=1`. This is the actual reason #813 went unnoticed: the `METAL:` GPU-vs-fallback counters exist, but serve mode never printed them, so the banner said "GPU-enabled" while every routed block ran on CPU and nothing contradicted it.
- `MB_BUILD` poisoning the group size when experts disagree, so a mixed-format container falls back to CPU cleanly instead of computing with the wrong `gs`.

**But you get the call, not me.** If you read his diff and think it is the better base — or that merging the two is cleaner than stacking them — say so and I will follow that. You have 27 comments of context on this code and I have none. The last two overlaps here were settled better by the contributors talking to each other than by anything I decided.

**What I can do is make sure the rebase is the last one.** `dev` moved a lot today — 21 PRs, including DeepSeek V4 (#165), Windows HIP (#788), CUDA fmt=8 (#817) and the paged ragged KV runtime (#795). This conflicts again. I know how tired that sentence must be by now, so: tell me when you are rebasing and I will hold everything that touches `backend_metal.*`, `colibri.c` and the Metal tests until you land. Nothing goes in ahead of you.

For whatever it is worth, the fact that a stranger independently reproduced your diagnosis today is decent evidence the analysis in this PR was right from the start.

RDouglasSharp:
I'll do the rebase Saturday morning in one focused session and ping you the moment I start — no need to hold anything before then.

JustVugg:
Saturday morning works. Thank you for taking it that way — and for "no need to hold anything before then", which is more generous than I asked for.

I will still tell you what is coming, because a rebase you can predict is cheaper than one you discover.

**Open PRs that touch `backend_metal.*` or the Metal tests today:**

| PR | author | overlap with yours |
|---|---|---|
| #829 | @aaristov | the same four files — held, waiting on you |
| #790 | you | your own Kimi K3 backend |
| #798 | @ZacharyZcR | repo-wide clang-format — **held until the queue drains**, precisely so it does not land under work like this |
| #399 | @NeuralNotwerk | KV quantisation, conflicting already |

So nothing is queued to land on your files before Saturday. If that changes I will say so here rather than let you find it.

**One thing that did move under you**, and it is mine: `dev` took 24 PRs yesterday, including DeepSeek V4 (#165) and Windows HIP (#788). None of them touch `backend_metal.*`, but `colibri.c` did change — the no-OpenMP shim block gained `omp_set_num_threads` from #807, which is the METAL=1-on-stock-macOS build fix. If your branch touches that region you will see a small conflict there; it is additive, take dev's side, nothing of yours is dropped.

**And a measurement from yesterday that strengthens the case for #813 being real**, in case it is useful when you write the final description: I ran DeepSeek V4 on an i7-1355U and got 75 s/token, because the fp4 rows16 kernels were gated on `__AVX512F__ || __aarch64__` and consumer Intel has had no AVX-512 since Alder Lake. Same shape as your Metal finding: a fast path that announces itself and then silently is not taken. Different backend, same class of bug, and both invisible because the counters that would have shown it are not printed on the path users actually run.

Ping me Saturday and I will be here.

bradydibble:
Merge conflict with the base branch (dev). Please rebase onto current dev and re-push so CI re-runs. (stale >7d sweep)

RDouglasSharp:
@JustVugg — rebased onto current dev (8d06bd1) and green. Force-pushed, so the PR now reduces to my own 8-commit delta on top of dev; the duplicated #457 commits dropped out in the rebase, exactly as you called it.

The conflicts were the expected fmt=4 (this PR) vs fmt=5 / fmt=6 (both now in dev) overlap, plus dev's chunked-prefill attention refactor — all resolved as unions, nothing dropped:

moe_gemv: added the fmt=4 grouped-int4 branch alongside dev's new fmt=5 (bf16-raw) and fmt=6 (E8/IQ3) branches; the final-store and the moe_submit host gate unioned to cover 1/2/4/5/6.
fused attention: dev split the core into encode_attn_projections + a chunked encode_attn_core_chunk. Kept that refactor and threaded kvb_fmt/kvb_gs through to the a_ctx dispatch so grouped kv_b dequantizes on-GPU (a_qabs was already covered). Taking my old pre-refactor hunk verbatim would have reverted the chunking — avoided.
Makefile: my test_kvb_notice / test_moe_gs_guard now self-register via the #733 rule-derived gate; no TEST_BINS line left to touch.
colibri.c: took dev's #807 no-omp.h shim; my kv_b fmt-gate notice was unaffected.
test harness: took dev's ref_fp8_nblk (#838) over my local rename — same quant.h clash fix, independent convergence.

make metal-test passes every arm on an M-series Max, including the fmt=4 paths this PR adds:

moe decode nb=8  fmt4-g128     nerr=2.30e-06  ok
moe ragged nb=6  fmt4-g128     nerr=2.07e-06  ok
moe ragged nb=6  fmt4-g64      nerr=2.17e-06  ok
attn S=1 pos=0   kvb-fmt4-g128 nerr=3.99e-06  ok
attn S=1 pos=37  kvb-fmt4-g128 nerr=3.14e-05  ok
attn S=4 pos=12  kvb-fmt4-g128 (MTP) nerr=3.24e-05  ok
attn S=3 pos=0   kvb-fmt4-g64  nerr=4.92e-06  ok

fmt 1/2/3 negative control unaffected. Sequencing per our agreement (#457 first, this on top) is already in place — ready to merge whenever you are.

PR #592, opened 2026-07-24
metal: two-phase prefill attention for arbitrary sequence lengths

RDouglasSharp:
Split the fused Metal attention encoder into three phases:

1. **encode_attn_projections** — RMS norm, Qa/KVa/Qb projections, RoPE, Q→abs once for all S rows
2. **encode_attn_core_chunk** — score/softmax/clat/ctx over a row chunk, respecting Metal's ~1B thread limit
3. **Output projection GEMV** — once on the accumulated ctx buffer

Removes the AMAXS=4 cap from `coli_metal_attn_decode`. Scratch buffers are now reserved dynamically via `attn_scratch_reserve(S,T)`. The S≤4 decode path is unchanged (backward compatible — same monolithic command buffer). The S>4 path uses 1 projections buffer + N core chunks + 1 output buffer.

Also lifts the S≤4 gate in `colibri.c:attention_rows` — the absorb path now delegates to Metal for any S when `g_absorb!=0`.

### Key changes
- `c/backend_metal.mm`: +119 / -33 — split + reserve + two-phase dispatch
- `c/colibri.c`: +5 / -10 — gate lift + cumulative timer in prefill progress

Needs `JustVugg/colibri#587` (host-side _gs params) if routed-expert fmt=4 decode is also wanted, but this PR is independent.


RDouglasSharp:
Closing — opened in error without my approval; disregard.

PR #600, opened 2026-07-24
fix(metal): compile-time SDK floor for the residency-set path

### LINKED ISSUE #596: [Bug]: METAL=1 — backend_metal.mm fails to compile against a macOS SDK < 15 (MTLResidencySet not preprocessor-gated)
### Commit or version

`81f08a09e5651ce52616dc720f68810f9021c0be` (v1.1.1, current `main` HEAD).

### Environment

- OS/SDK: macOS with the **macOS 14.x SDK** (e.g. `apple-sdk 14.4`, current nixpkgs default)
- CPU: Apple Silicon
- Compiler: clang (Apple/LLVM), `-fobjc-arc`
- RAM / GPU / storage: N/A (compile failure)

### Reproduction steps

```shell
make -C c colibri METAL=1   # active SDK = macOS 14.x
```

(Reproduced under Nix with the darwin stdenv on `apple-sdk 14.4`; the same build succeeds once `apple-sdk_15` is used.)

### Expected behavior

Either `METAL=1` compiles, or it fails with a clear "requires the macOS 15 SDK" message, and `docs/metal.md` documents the minimum SDK/OS.

### Actual behavior and logs

```shell
clang++ -x objective-c++ -std=gnu++17 -fobjc-arc -O3 -c backend_metal.mm -o backend_metal.o
backend_metal.mm:359:41: error: no type or protocol named 'MTLResidencySet'
backend_metal.mm:359:71: error: no known instance method for selector 'addAllocation:'
backend_metal.mm:459:9:  error: unknown type name 'MTLResidencySetDescriptor'; did you mean 'MTLRenderPassDescriptor'?
backend_metal.mm:462:41: error: no known instance method for selector 'newResidencySetWithDescriptor:error:'
backend_metal.mm:464:20: error: no known instance method for selector 'addResidencySet:'
backend_metal.mm:547:47: error: no known instance method for selector 'removeResidencySet:'
14 errors generated.
make: *** [Makefile:308: backend_metal.o] Error 1
```

**Analysis (important nuance)**

The residency-set block (`COLI_METAL_RESSET=1`) is already correctly guarded **at runtime** by `@available(macOS 15.0, *)`, and `g_resset_obj` is a bare `id` to keep `-Wunguarded-availability` clean. But `@available` is a **runtime** guard, not a **compile-time** one: it does not stop the compiler from resolving the `MTLResidencySet` / `MTLResidencySetDescriptor` protocol names and the `newResidencySetWithDescriptor:` / `addResidencySet:` / `removeResidencySet:` selectors against the SDK headers. Compiled against an SDK older than macOS 15 — where those symbols are simply not declared — compilation fails (`no type or protocol named 'MTLResidencySet'` + unknown selectors under ARC).

`MTLResidencySet` is a macOS 15.0 API (consistent with the code's own `@available(macOS 15.0, *)` annotations and Apple's docs). `docs/metal.md` (`make colibri METAL=1 — no Xcode needed`) documents no minimum SDK/OS.

**Suggested fix** (PR against `dev`): gate the RESSET block with a **preprocessor** SDK-floor check (e.g. `#if defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 150000`, with a clean fallback that omits residency sets), and/or document the macOS 15 SDK minimum for `METAL=1` in `docs/metal.md`. The dependency-free default CPU path is unaffected.

lBroth:
Fixes #596. Written working with Claude, and I am not a Metal expert by training, so please read the diff with that in mind.

`@available(macOS 15.0, *)` is a runtime guard: the compiler still resolves `MTLResidencySet`, `MTLResidencySetDescriptor` and the residency-set selectors against the SDK headers, which do not declare them before macOS 15, so `METAL=1` fails to build on older SDKs exactly as reported. This adds `COLI_HAS_RESSET` (`__MAC_OS_X_VERSION_MAX_ALLOWED >= 150000`, overridable with `-DCOLI_HAS_RESSET=0`) around the five sites that touch the API, keeps the `@available` guards for the runtime side, has the init path print that the gate was compiled out instead of staying silent, and documents the SDK floor in `docs/metal.md`.

Verification, since I had no macOS 14 SDK on hand: I reconstructed one (Metal.framework from the 26.5 SDK with `MTLResidencySet.h` and the related declarations removed, sanity-checked with a plain `#import <Metal/Metal.h>` build). Stock `dev` fails against it with the 11 errors from the issue; this branch compiles clean with the gate at 0. The repro earned its keep, my first version missed the `removeResidencySet:` call in `coli_metal_shutdown`. With the gate on, the emitted object is byte identical to stock `dev`, so macOS 15+ SDK builds are unchanged.

On the exact tree of this PR (macOS 26.5, M5 Pro 48 GB): `make check` 167 tests pass, `make -C c colibri METAL=1` builds with no warnings, `make -C c metal-test` is green, oracle 32/32 teacher-forced and 20/20 greedy.

One oracle note that may be worth a line in CONTRIBUTING: `tools/make_glm_oracle.py` needs `transformers >= 5.11`, and on Python 3.9 pip silently caps at 4.57.6 via `requires-python`, which looks like the version does not exist. Python 3.12 resolves it fine.

It would be good if @jerome-benoit could confirm on the real `apple-sdk 14.4` Nix build, since my SDK 14 is a reconstruction.

### DIFF
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -333,8 +333,20 @@ static std::mutex g_slab_mtx;   // expert_load registers slabs from parallel Ope
 // g_resset_obj is a bare `id` (holds id<MTLResidencySet>) so the global's declared type
 // carries no availability annotation -- the protocol name only appears inside
 // @available(macOS 15.0, *) guards below, keeping -Wunguarded-availability clean.
+// @available is a runtime guard only: the compiler still resolves MTLResidencySet and its
+// selectors against the SDK headers, which don't declare them before macOS 15 (#596).
+// COLI_HAS_RESSET is the compile-time floor (helpers become no-ops below it); the @available
+// guards stay on top so a macOS 15 SDK build still runs right on macOS 14. -DCOLI_HAS_RESSET=0
+// forces the fallback branch on a modern SDK.
+#ifndef COLI_HAS_RESSET
+#if defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 150000
+#define COLI_HAS_RESSET 1
+#else
+#define COLI_HAS_RESSET 0
+#endif
+#endif
 static id g_resset_obj;
-static bool g_resset_enabled;   // COLI_METAL_RESSET=1, macOS 15+, and creation succeeded
+static bool g_resset_enabled;   // COLI_METAL_RESSET=1, macOS 15 SDK+OS, and creation succeeded
 static bool g_resset_dirty;     // addAllocation: calls pending commit; g_resset_mtx-guarded
 // Set mutations + dirty flag get their OWN mutex, never held together with g_slab_mtx: no
 // live Metal call may run under the slab lock the parallel OMP loader threads contend on
@@ -355,8 +367,12 @@ static double g_t_resset_flush;   // sec committing pending adds in moe_submit (
 // no separate counter for the add/remove side.
 static void resset_add(id<MTLBuffer> b) {
   if (!g_resset_enabled) return;
+#if COLI_HAS_RESSET
   std::lock_guard<std::mutex> lk(g_resset_mtx);
   if (@available(macOS 15.0, *)) { [(id<MTLResidencySet>)g_resset_obj addAllocation:b]; g_resset_dirty = true; }
+#else
+  (void)b;
+#endif
 }
 // Remove + commit immediately, NOT deferred: the caller frees the underlying host memory
 // right after coli_metal_unregister returns, so the removal must be applied before that --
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
Static review only — I have no macOS box, so I did not build this. Reading the diff against `dev`, the gate looks complete and behaviour-preserving:

**Coverage.** All six sites that name a macOS 15 symbol are inside `#if COLI_HAS_RESSET`: `addAllocation:` (`resset_add`), `removeAllocation:`/`commit` (`resset_remove`), `commit` (`resset_flush`), `MTLResidencySetDescriptor`/`newResidencySetWithDescriptor:`/`addResidencySet:` (init), and `removeResidencySet:` (`coli_metal_shutdown`). Nothing outside them mentions `MTLResidencySet`, and `g_resset_obj` stays a bare `id`, so the fallback build has no unresolved names left.

**The state writes the gate excludes are unreachable, not skipped.** Worth stating explicitly since it's the one thing that looks asymmetric: `resset_remove`'s `g_resset_dirty = false` and `resset_flush`'s dirty handling now sit inside the `#if`, so under `COLI_HAS_RESSET=0` they don't run. That's fine because the only writer of `g_resset_enabled = true` is the init block, which is itself gated — so `g_resset_enabled` is permanently false and all three helpers return at their first line. No path can leave a stale `g_resset_dirty` behind. Same reasoning covers `shutdown`'s gated `if (g_resset_enabled)`, and the unconditional reset of the three globals after it keeps the teardown state identical either way.

**The `-DCOLI_HAS_RESSET=0` escape hatch is the right call** — it's what makes the fallback branch testable on a modern SDK, which is exactly how you caught the missing `removeResidencySet:`. Reconstructing an SDK 14 to test against rather than asserting it compiles is a notably higher bar than this class of fix usually clears.

One nit, take it or leave it: the init-path `fprintf` for the compiled-out case says the binary was built against an SDK < 15, which is the common cause but not the only one — `-DCOLI_HAS_RESSET=0` on a modern SDK lands there too. "compiled without residency-set support (macOS 15 SDK required)" would cover both. Not worth a round trip on its own.

Agreed that `@available` alone was never going to be enough here, and that documenting the SDK floor in `docs/metal.md` belongs in the same change.

Issue #523, opened 2026-07-22
[Performance]: Vulkan (#418) vs ROCm/HIP on RDNA4 (RX 9070 XT) — CORRECTED: Vulkan is 19–24% faster (original comparison was confounded)


noobdev-ph:
> [!IMPORTANT]
> **CORRECTED — the conclusion below is wrong.** My original comparison was confounded: `COLI_CUDA=1` tripped the `g_draft` guard and disabled MTP on the HIP arms while the Vulkan arms drafted at 2.00 tokens/forward, so the two backends were not running the same decode configuration.
>
> With speculation controlled on both, **Vulkan is 19–24% faster than ROCm/HIP**, not 32–35% slower. @steve-m's `~35% faster than ROCm on RDNA4` note is directionally correct and my challenge to it is withdrawn.
>
> **See the [corrected 80-run results](https://github.com/JustVugg/colibri/issues/523#issuecomment-5056835970).** The ReBAR finding and the correctness results below still stand; the performance comparison does not.

---

Independent measurements of the Vulkan backend from #418 on RDNA4, since @JustVugg mentioned not having a Vulkan GPU to verify against. Filing as an issue rather than a PR comment so the numbers are citable on their own and the review thread stays focused.

Summary up front: **the backend works correctly on RDNA4 with no ROCm installed, and produces exact output on a real fmt=4/gs=64 container.** That's the headline and it's a good one — this is the only path that gets the engine onto a lot of hardware.

Two things I ran into that I can't explain on my own, and would value your read on: it needs Resizable BAR (and fails silently without it), and on this particular box it comes out 23–35% slower than ROCm/HIP at matched VRAM — which runs opposite to the `~35% faster than ROCm on RDNA4` note in `c/Makefile:260`. I'm assuming I've got something configured differently rather than that the note is wrong, so I've documented my full setup below and would happily re-run anything.

## Commit

`550df5c` (branch of #418, rebased onto `dce7012` / v1.1.0). One binary built with **both** backends — they are independent in the Makefile — so every arm below is the *same executable* and only environment variables differ.

## Hardware and storage

- **CPU:** AMD Ryzen 7 5700G (8C/16T, Zen 3)
- **RAM:** 64 GB DDR4
- **GPU:** AMD Radeon RX 9070 XT (gfx1201, RDNA4), 16 GB VRAM, **Resizable BAR enabled** (16 GB host-visible)
[...]

PR #602, opened 2026-07-25
Add Qwen3.6-35B-A3B engine: Vulkan MoE backend, resident expert pinning, serve GPU support + fixes

minne100:
## Summary
This PR adds a complete, self-contained engine for **Qwen3.6-35B-A3B** (35B params / 3B active, Apache-2.0) to colibri, plus a set of fixes and a benchmark. The goal is to run this large MoE model on a 16 GB machine by streaming experts from disk on demand.

### What's included
- **Phase-2 engine (`c/qwen36.c`)** — Gated Attention (GQA + partial RoPE, `rope_dim=64`) blended with a pure-C **Gated DeltaNet** recurrent linear-attention path (per-layer `DN_rec`/`DN_conv` state), followed by a streaming MoE (shared expert + group-limited top-k). Hybrid layers use Gated Attention at `i%4==3` (10/40 layers), the rest are DeltaNet.
- **Vulkan MoE backend (`c/vulkan_gemv.c` + headers)** — int4 weights are unpacked **inside the shader** and multiplied as float GEMV. This is a deliberate workaround for the **AMD 780M driver `OpSDotKHR`/int8 compiler segfault** (0x800184): we only use `OpCapability Shader` and never touch int8 types. The backend auto-probes a Vulkan compute device and silently falls back to CPU when none is present (dynamic `vulkan-1` load, zero-dependency binary).
- **Bug fix: GPU produced all-zero output on the real 35B int4 model.** Two root causes, both in `vulkan_gemv.c`:
  1. `vg_expert_ensure → vg_expert_loaded` had its `layer`/`expert-id` arguments **swapped** (`loaded(layer, eid, li)` was called as `(layer, li, eid)`), so weights were uploaded to the wrong slot and the shader read zeros. The self-test missed this because `eid==li` by coincidence.
  2. `load_expert_merged`'s int4 slice was gated behind `vg_use_int4()`, so the int4 container (`g4/u4/d4`) was never filled when an expert loaded before `vg_init`. Now the int4 container is always populated.
  Verified: GPU vs CPU MoE cosine = **1.00000** at layer 0 / token 0, and **byte-identical** generated text vs CPU-only.
- **Resident expert pinning (`COLIBRI_RESIDENT=1|2`)** — pin the experts a prompt actually routes to into RAM (and, via the GPU `g_w` buffer, into VRAM) so they are not repeatedly evicted and re-streamed from disk. `=1` pins once after prefill; `=2` keeps collecting and pinning incrementally across the decode phase. Per-layer pin budget = `cap` to avoid LRU deadlock.
- **Serve mode (`c/qwen36_serve.c`)** — integrates the Vulkan backend (enabled automatically per `g_gpu_backend`) and fixes two HTTP POST parsing bugs:
  1. `Expect: 100-continue` was never answered, so clients that wait for `100` never sent the body.
  2. `handle_conn` NUL-truncated the request buffer in place at the first space (`*sp1=0`), which destroyed the body and made `strstr(req,"\r\n\r\n")` fail → `400`. Now method/path are copied into local arrays and the full buffer is searched.
  Verified end-to-end: `/v1/chat/completions` returns valid OpenAI-format JSON ("The capital of France is Paris.").
- **Conversion toolkit** (`c/tools/convert_qwen36.py` default true-int4, `make_qwen36_oracle.py` attention-only, `make_qwen36_tiny.py` weight-free smoke) and **design docs** (`docs/qwen36-*.md`).
- **Benchmark** (`c/int4_vs_int8_cold.md`): cold-start comparison on an **integrated-GPU 16 GB laptop** (AMD 780M, no dedicated VRAM), long prompt, `N_NEW=256`.

### Benchmark (cold start, long prompt, default GPU unless noted)
| Metric | int4 +GPU | int8 +GPU | **int8 CPU-only** |
|---|---|---|---|
| First-token latency (TTFT) | 50.18 s | 28.68 s | **13.83 s** |
| Throughput | 0.29 tok/s | 0.48 tok/s | **1.08 tok/s** |
| Peak RSS | 11.30 GB | 11.27 GB | **10.25 GB** |
| 256-token wall time | 873 s | 532 s | **238 s** |

**Takeaway:** on this iGPU machine, **int8 CPU-only is the fastest and smallest** — the integrated GPU shares system memory and gains no bandwidth, while int4 pays a CPU-side unpack-to-int8 penalty before upload. int4's only hard advantage is **half the on-disk size** (19.7 GB vs 34.6 GB). A dedicated GPU with the experts resident in VRAM is expected to show the opposite result; that path is implemented (`COLIBRI_RESIDENT`) but not yet measured on such hardware.

### Test plan
- [x] int4 self-test (`convert_qwen36.py --selftest`) passes
- [x] GPU vs CPU MoE cosine = 1.00000; byte-identical generation
- [x] `COLIBRI_RESIDENT=2` prefill pin hit-rate ~39%, text correct, <16 GB
- [x] serve POST returns valid OpenAI JSON (curl round-trip)
- [ ] measure on a dedicated-GPU machine (not available here)

### Notes / limitations
- NFC normalization is not implemented (the tokenizer assumes NFC input).
- The AMD integrated-GPU path uses float GEMV (not int8 dot) by design.
- Model weights (`modles/`) and the Vulkan SDK headers (`c/vulkan/`) are intentionally git-ignored; the generated `*.spv.h` headers are committed.


JustVugg:
Thanks for this — it's a substantial, well-built contribution, and we'd like to bring the Qwen3.6-35B engine in. Two things before it can land, both about **shape**, not quality:

1. **Retarget to `dev`, not `main`.** All contributions merge into `dev` (main is release-only and protected). Please change the base branch to `dev` and rebase onto it.

2. **Split it — this is +15,515 lines in one PR.** It bundles at least three separable things: (a) the **Qwen3.6 engine** (`qwen36.c` / `qwen36_serve.c` / converter / oracle tooling), (b) a **Vulkan MoE backend**, and (c) **resident-expert pinning + serve GPU** changes. We can't responsibly review or merge that as a single unit, and if any one piece needs iteration it blocks the whole thing. Please break it into an incremental series we can land one at a time — ideally starting with the **self-contained engine** (new `qwen36.c` that doesn't touch the shared GLM/Inkling path), which is the lowest-risk piece and can go in first. The Vulkan backend is its own big review and also overlaps #418 / #601 — worth coordinating so we don't end up with two Vulkan backends.

Land the engine piece on `dev` first, and we'll take the GPU/Vulkan parts as follow-ups. Happy to review each increment quickly.

JustVugg:

JustVugg:
Context that affects this PR: **#418 is now merged to `dev`** — @steve-m's Vulkan backend (expert tier + dense + MLA attention, any Vulkan 1.2 driver, opt-in via `VK=1`/`COLI_VULKAN=1`).

That makes it the project's Vulkan backend, and this PR carries a second one (`vulkan_gemv.c`, `vulkan_core.h`). Two Vulkan implementations in one repo is the kind of duplication we'd rather not carry — and you'd inherit steve-m's shader work and the field reports coming from #523/#510 for free.

Concretely, the ask when you rebase: **split this into the Qwen3.6 engine alone**, dropping the Vulkan backend (use #418) and the standalone HTTP serve (`qwen36_serve.c`, `test_http_parse.c` — `openai_server.py` already serves every engine, and Kimi/Inkling both route through it). That takes the PR from +15,515 to something reviewable, and the engine is the part only you have written.

There's also #544 (@opxyc) adding a Qwen3-30B-A3B engine — same family, different file. Worth the two of you agreeing on one Qwen engine before either lands; @steve-m and @ZacharyZcR did exactly that on Kimi last week and it worked well.

kreuzzelg:
Hi @minne100,

first of all: thank you for #602 — the qwen36 engine, the converter and the
tiny/oracle validation setup are excellent groundwork. We had a machine with
dedicated GPUs (RTX 3070 8 GB + Quadro RTX 4000 8 GB) and picked up exactly
the item you left open in your test plan: *"measure on a dedicated-GPU
machine"*.

Your expectation was right — dedicated GPUs do show the opposite result, but
only once the experts are truly **resident in VRAM**. Following the
maintainers' guidance (split, target `dev`, engine first), we submitted:

- **#712** — the engine slice of your work, retargeted to `dev`, with your
  co-authorship, plus CPU performance work (dense int8 + AVX2 GEMV, DeltaNet
  recurrence parallelized) and two converter fixes (bundled `tokenizer.json`,
  flat `config.json` for the VL checkpoint).
- **#713 (draft)** — a CUDA VRAM expert tier on top: heat-based placement via
  the existing shared `backend_cuda.cu` (no new backend), LFRU from `tier.h`,
  warmstart + persisted heat. Measured: **11.3 tok/s with 2×8 GB, 9.9 tok/s
  with a single 8 GB card** (35B int4, 200-token decode) — above
  llama.cpp/Ollama on the same box.

A few findings from our runs that may be useful for your Vulkan path:
1. The `gpu_probe` constants in `qwen36.c` (`VK_QUEUE_COMPUTE_BIT` as 0x20,
   stride 32) prevent the probe from ever enabling the GPU on NVIDIA/Linux
   (0x20 is VIDEO_DECODE; the struct is 24 bytes).
2. With large caches (>4 GB weight pool) the 32-bit slot arithmetic in
   `vulkan_gemv.c` wraps and aliases slots (layers 32–39 ⇄ 0–7), and the
   precompiled GEMV shader itself cannot address past 4 GB — we hit both at
   `cache=256`. Fixes are on our branch if useful.
3. Your published i4/i8 containers just need two small files to run out of
   the box: `tokenizer.json` and a flattened `config.json` — the updated
   converter in #712 bundles both; a self-contained container is here:
   https://huggingface.co/Kreuzzelg/qwen36-35b-a3b-colibri-i4

To be clear, we see the two GPU paths as complementary: your Vulkan/iGPU
approach is the right answer for shared-memory laptops and AMD (no CUDA),
the CUDA tier for dedicated NVIDIA cards. Happy to coordinate however works
best for you — and thanks again for the foundation.

minne100:
Closing in favor of #712 (engine slice, with co-authorship) — thanks @kreuzzelg for picking up the dedicated-GPU benchmarking I left open in the original test plan.

Local follow-ups I've applied (happy to fold into #712 or a separate Vulkan PR):
1. gpu_probe bug: VK_QUEUE_COMPUTE_BIT was 0x20 (VIDEO_DECODE_BIT), so the GPU was never probed on NVIDIA/Linux. Fixed to 0x00000002u in qwen36.c.
2. vulkan_gemv.c 32-bit slot wrap: slot byte offsets are uint32_t, so any weights pool >= 4GB wraps/aliases (layers 32-39 -> 0-7 at cache>=256). Added a >4GB guard that disables the Vulkan backend and falls back to the CPU MoE path instead of silently corrupting output. 64-bit shader offsets (buffer_device_address) are a follow-up.
3. Published i4/i8 containers now include tokenizer.json (was missing) — uploaded to both HF repos, so they run out of the box.

A Vulkan MoE GEMV backend for integrated/AMD GPUs (this implementation) is complementary to #418 (steve-m, CUDA/NVIDIA-oriented). I'm opening a separate draft PR on dev for it. Happy to coordinate so we don't end up maintaining two backends if you'd rather unify.

PR #620, opened 2026-07-25
docs(api): OpenAI server supports tool-calling

jeswr:
`docs/api.md` (the OpenAI `coli serve` section) currently says Tools "return an explicit error rather than being silently ignored." That's stale for **Tools**: `c/openai_server.py` at HEAD ships full tool-calling — `parse_tool_calls()` (renders GLM tool markers back into OpenAI `tool_calls`), `render_chat(..., tools, tool_choice)` (injects the `<tools>` XML and honours `tool_choice` including `none`/`required`/forced-function), plus a dedicated `test_openai_tools_e2e.py`. Tool schemas only 400 when *malformed* (landed via #401/#408). The Anthropic section of the same doc already documents full tools support, so the OpenAI paragraph contradicted both the code and its sibling section.

This corrects only the Tools sentence; the image/audio input, custom stop sequences, log-probabilities, and token-penalty items are left unchanged.

_Prepared for review — happy to adjust wording. (Heads-up: there may be a delay before I can respond to review comments.)_


JustVugg:
Thanks — and this docs fix is correct: tool-calling **is** supported on the OpenAI `/v1/chat/completions` path (the previous wording was wrong). I retargeted it from `main` to `dev` (all contributions land on dev), which surfaced a conflict in `docs/api.md` since that file moved on dev. Could you rebase onto current `dev` and resolve it? Then it merges.

JustVugg:

Issue #343, opened 2026-07-17
[Feature]: Support Anthropic protocol endpoint


yurivict:
### Problem

'coli serve' serves the OpenAI protocol.
Would it be too difficult to add the Anthropic protocol endpoint?

I know that many people love the Claude CLI software. But Claude connects to only Anthropic endpoints.

### Proposed solution

Choices:
* run these 2 endpoints at the same time
* make this choice in the global config file
* make it based on an environment variable

JustVugg:
The OpenAI-compatible endpoint (`/v1/chat/completions`) already covers most clients. An Anthropic-protocol endpoint (`/v1/messages`) is a thin adapter on top of the same generation path — it's mostly request/response shape translation (system prompt handling, content blocks, the `stop_reason`/`usage` mapping) in the Python gateway (`openai_server.py`), no engine change. It'd make the engine drop-in for Anthropic-SDK clients and tools. Reasonable feature; it lives entirely in the gateway. Would you use it against a specific client (Claude Code, an SDK app)? That'd help scope which parts of the messages API to cover first (streaming + tool_use being the interesting ones).

yurivict:
I would use it with Claude Code.

So far I wasn't able to get any CLI program that supports the OpenAI protocol to work with Colibri. I also didn't manage to compile some CLI programs. I am working on this in the scope of FreeBSD ports. But Claude Code is the best at this time, IMO.

JustVugg:
i'm working on!

JustVugg:
Built it — **#525**, `/v1/messages` served alongside the OpenAI endpoint on the same port, no second process and nothing to enable.

For Claude Code:
```bash
export ANTHROPIC_BASE_URL=http://localhost:8000
export ANTHROPIC_API_KEY=local        # only enforced if you set COLI_API_KEY
export ANTHROPIC_MODEL=glm-5.2-colibri
claude
```

It is a translation layer over the same generation path, so tools, streaming and the KV cache behave exactly as they do on `/v1/chat/completions`. `x-api-key` authenticates (Bearer still works). Streaming emits the full named-event sequence, and long cold prefills are kept alive with the protocol's own `ping` event. `stop_sequences`, `top_k` and non-text blocks are refused explicitly rather than ignored.

14 tests cover the client-visible contract, but **what I can't test here is Claude Code itself against a real model** — that's your ask and your setup. Would you try the branch and report back, especially on a turn that uses tools? Either result is useful.

[...]

PR #625, opened 2026-07-26
Apply the implicit GLM stop sequences on /v1/messages too

bharathgnana:
`/v1/messages` never went through `StopFilter`. `generation_options`'s stop sequences are
discarded into `_stop_sequences`, and both Anthropic paths call `engine.generate` with a raw
sink (`output.append` / `on_text`), so the stop policy `/v1/chat/completions` already has
never reaches the Anthropic endpoint.

In serve mode the engine arms only `<|endoftext|>` and filters the other stop ids (#401, and
#549 for containers where a different end-of-turn special wins the final-token margin), which
makes role markers the gateway's boundary to enforce. On `/v1/messages` nothing enforced it, so
`<|user|>` was detokenized into the reply and generation ran past the end of the turn until it
hit `max_tokens`.

Claude Code is the reference client from #343 and speaks only `/v1/messages`, so it saw this on
every reply while OpenAI-shaped clients were already clean.

### Reproduction

GLM-5.2 int4 container (the `-with-int8-mtp` HF build), `coli serve`, DGX Spark GB10, `dev` at
6fcc555. Prompt: *Reply with the single word: ready*

```
/v1/chat/completions -> "ready"
/v1/messages         -> "ready<|user|>the user asks about cake"   (streaming leaks it too)
```

Engine log for the same run:

```
[stop] serve mode: filtered 17 non-EOS stop tokens (tool-call safety, #401)
[stop] 1 stop tokens: 154820
```

### Fix

Reuse `stop_policy()` / `StopFilter` rather than adding a second mechanism. `body` in
`anthropic_generation` is the translated OpenAI-shaped request, and `anthropic_messages()` has
already refused a client `stop_sequences`, so it resolves to the same implicit GLM role
boundaries — `StopFilter`'s prefix-aware holdback keeps prose streaming token by token.

Passing `stopped=stop_filter.stopped` also ends generation at the marker instead of decoding a
tail that is discarded. On a streaming 744B model that is real time: a 200-token budget now
returns in 34 s instead of roughly 4 minutes.

### Tests

Three tests in `test_anthropic_messages.py` covering the non-streaming shape, the SSE stream,
and a marker split across engine chunks. All three fail on `dev` without the change:

```
FAIL: test_marker_split_across_chunks_is_still_caught
FAIL: test_role_marker_ends_a_streamed_turn
FAIL: test_role_marker_ends_the_turn
```

`FakeEngine.generate` gained the `stopped`/`on_accept` kwargs the real engine already receives,
and honours `stopped()` so the tests can assert the tail is cancelled rather than merely hidden.

`make check`: 213 passed, 24 skipped.

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -1989,6 +1989,10 @@ class APIHandler(BaseHTTPRequestHandler):
     def anthropic_generation(self, body, prompt, request_id, tools, enable_thinking):
         maximum, temperature, top_p, grammar, _stop_sequences = generation_options(
             body, self.server.max_tokens)
+        # Same policy as /v1/chat/completions: `body` is the translated OpenAI-shaped
+        # request, and anthropic_messages() has already refused a client `stop_sequences`,
+        # so this resolves to the implicit GLM role boundaries.
+        stop_sequences, ignore_leading_stop = stop_policy(body, True)
         cache_slot = body.get("cache_slot")
         if (cache_slot is not None and
                 (isinstance(cache_slot, bool) or not isinstance(cache_slot, int) or
@@ -2028,9 +2032,11 @@ class APIHandler(BaseHTTPRequestHandler):
             queue_headers = {"x-colibri-queue-wait-ms": str(round(queue_wait * 1000))}
             if not stream:
                 output = []
+                stop_filter = StopFilter(stop_sequences, output.append, ignore_leading_stop)
                 stats = self.server.engine.generate(
-                    prompt, maximum, temperature, top_p, output.append, cache_slot,
-                    self.client_disconnected, grammar=grammar)
+                    prompt, maximum, temperature, top_p, stop_filter.feed, cache_slot,
+                    self.client_disconnected, grammar=grammar, stopped=stop_filter.stopped)
+                stop_filter.finish()
                 content, stop_reason = blocks_and_stop("".join(output), stats)
                 self.send_json(200, {
                     "id": message_id, "type": "message", "role": "assistant",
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
Reviewed against `dev` at 6fcc555 — static only, no GLM container here, so I verified the reasoning rather than the reproduction. Every premise the fix rests on checks out:

- **`body` really is the OpenAI-shaped request.** `anthropic_messages` builds `translated` and hands that to `anthropic_generation`, and it carries no `stop` key — so `parse_stop_sequences` returns `()` and `stop_policy(body, True)` necessarily lands on the implicit GLM branch. Not an accident of the current call site: `translated` is constructed explicitly a few lines above.
- **A client `stop_sequences` cannot slip through.** `anthropic_messages` rejects it with a 400 as its very first action, before translation, so `stop_policy` can't resolve to a client-supplied set here.
- **`StopFilter.stopped` exists** (`:940`, `self.matched is not None`) and **the real `Engine.generate` already accepts `stopped=`** (`:1227`) — worth stating explicitly, since `FakeEngine` gaining the kwarg is exactly the kind of change that can paper over a signature mismatch in the real engine. It doesn't here.
- **`stop_reason` stays correct.** A `stopped()` cutoff is a soft stop (`STOP <id>` to the native process), not a truncation, so `stats["length_limited"]` stays false and `blocks_and_stop` yields `end_turn` rather than `max_tokens`.

The point I'd add to the description: **this isn't a new mechanism, it's the fourth call site of an existing one.** `stopped=stop_filter.stopped` is already what `/v1/chat/completions` does in all three of its generation paths (`:1693`, `:1861`, `:1890`). That's the strongest argument for this shape over a bespoke Anthropic-side guard — the semantics are already carried in production, and the diff is +12/-4 because it reuses rather than reimplements.

Two things I'd call out as done right:

- `stop_filter.finish()` before `split.finish()` in the streaming path. The held-back tail has to flush through `on_text` into the thinking splitter *before* that splitter closes, and the ordering here does that.
- Making `FakeEngine` honour `stopped()` instead of just accepting it. That's what lets the tests assert the tail is **cancelled** rather than merely filtered out of the output — the 34s-vs-4min difference is real behaviour, and without that change the tests would pass on a version that still decodes the whole discarded tail.

For what it's worth, I looked at this exact area yesterday from the C side (the `stops_arm_tok` filter and its "boundaries the Python server owns" comment) and concluded the layering was sound — because I checked `/v1/chat/completions` and stopped there. The gateway does own the boundary; it just only owned it on one of the two endpoints. Nice catch, and the Claude Code framing explains why it stayed invisible for so long.

PR #650, opened 2026-07-27
tools: deprecate convert_olmoe.py

CodePrometheus:
## Summary

convert_olmoe.py still writes the old per-expert layout (separate gate_proj/up_proj/down_proj), but olmoe.c only loads the merged merged_weight tensor that convert_olmoe_merged.py produces. The two drifted apart when the engine moved to single-read-per-expert loading.

## Validation

- [x] `make -C c check`

## Compatibility


CodePrometheus:
Conflict is resolved, kindly ping @JustVugg

PR #659, opened 2026-07-27
fix(cuda): correct RAM budget on integrated/unified-memory GPUs (#653)

### LINKED ISSUE #653: GB10/unified memory: auto RAM_GB ignores the CUDA expert tier (pre-allocation snapshot) -> OOM-kill
## Summary

On unified-memory hosts (GB10 / DGX Spark), the automatic `RAM_GB` budget does not account for the CUDA expert tier, because both the budget and the overcommit guard are computed from a **memory snapshot taken before the tier is allocated**. On a discrete GPU this is correct (VRAM is a separate pool); on Grace-Blackwell the GPU tier and the RAM cache draw from the *same* physical LPDDR5X, so the budget is over-committed by exactly `CUDA_EXPERT_GB` and the process is OOM-killed during prefill.

## Environment

- Dell/NVIDIA **GB10 (DGX Spark)**, Grace `Cortex-X925` + `Cortex-A725`, aarch64
- 121 GB unified LPDDR5X, `sm_121`, CUDA 13.0, driver 580.159.03
- gcc 13.3.0, Ubuntu 24.04, kernel 6.17
- colibri `81f08a0`, built `make -C c glm CUDA=1 CUDA_ARCH=native ARCH=armv9-a+i8mm+bf16`
- Model: `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp`

This is the same class of machine as the GB10 row in `docs/benchmarks.md` (#136, #161).

## Reproduce

```sh
COLI_CUDA=1 COLI_GPU=0 CUDA_EXPERT_GB=24 SERVE_BATCH=1 DIRECT=1 PIPE=1 \
  coli serve --model /path/to/glm52 --host 127.0.0.1 --port 8123
# then send any chat completion
```

Startup log — note the tier is allocated first, then the budget is sized as if that memory were still free:

```
[CUDA] hot expert tier: 1268/1268 experts, VRAM 23.98 GB (budget 24.0 GB, reserve 2.0 GB)
[PIN] placement: 1268 VRAM + 0 RAM expert (0.0 GB warm)
[RAM_GB=57.7 auto] cap raised 8->10: budget allows it (projected peak 55.1 GB)
```

`MemAvailable` was ~65 GB at process start, so `65 * 0.88 = 57.2 GB` ≈ the 57.7 GB budget. The tier had already taken 24 GB of that same pool, leaving ~41 GB actually available. The engine dies mid-prefill:

```
[prefill] layer 37/78 - 13 token
[api] request failed: colibri engine exited unexpectedly
```

```
oom-kill: task=colibri,pid=3368105
Out of memory: Killed process 3368105 (colibri)
  total-vm:81543976kB, anon-rss:42231328kB
```

RSS at kill (42.2 GB) matches the ~41 GB that was genuinely left, confirming the budget overshot by roughly the tier size.

## Cause

- `colibri.c:6535` — `g_mem_avail_boot = mem_available_gb();` captured once, at startup
- `colibri.c:6131` — `ram_gb = g_mem_avail_boot*0.88;` (auto budget)
- `colibri.c:6175` — the `projected peak > g_mem_avail_boot` overcommit guard uses the same stale value, so it cannot catch this

The CUDA expert tier is allocated *after* that snapshot. I verified that `cudaMalloc` does correctly decrement `MemAvailable` on this platform (8 GB alloc moved it 61016 MB -> 52426 MB), so re-reading `/proc/meminfo` after the tier is placed would observe the real figure — the snapshot is simply taken too early for this case.

## Suggested fix

When the device is integrated / unified-memory (`cudaDeviceProp::integrated`, or `unifiedAddressing` with the host), subtract the bytes actually placed in the expert tier from the auto budget and from the guard at `colibri.c:6175` — or re-read `mem_available_gb()` after tier placement and use the lower of the two. Discrete-GPU behaviour would be unchanged.

A warning when `RAM_GB(auto) + CUDA_EXPERT_GB` exceeds physical RAM on an integrated device would also have made this immediately obvious.

## Workaround

Setting both budgets explicitly so their sum leaves headroom works fine:

```sh
CUDA_EXPERT_GB=16 RAM_GB=38   # on a 121 GB box with ~27 GB held by another process
```

## Note

`CUDA_DENSE=1` also reproducibly hangs during prefill on this hardware (stalls around layer 65/78, no output, no error). Happy to open that separately if useful.

JustVugg:
## What

On **integrated / unified-memory GPUs** (Grace-Blackwell **GB10 / DGX Spark**, Jetson) the hot expert tier and the host RAM cache draw from **one physical memory pool**. The auto RAM budget reads `g_mem_avail_boot` — a `MemAvailable` snapshot taken at startup **before** the expert tier is placed — so on unified memory it over-counts free RAM by exactly the tier size. The budget then over-commits and the process is **OOM-killed during prefill** (reported on GB10, `RAM_GB` auto).

## Fix

- Add `coli_cuda_device_integrated(device)` → `cudaDeviceProp::integrated` (also valid under HIP).
- After the expert tier is placed, if the primary device is integrated, subtract the placed bytes from `g_mem_avail_boot`.
- Both consumers — `cap_for_ram` (auto budget) and the overcommit guard — read the corrected value, so the budget shrinks by the tier size and stops over-committing.

**Discrete GPUs are unaffected**: `integrated` is `0`, so the block is a no-op and their behaviour is byte-for-byte unchanged.

## Testing

- CPU build (`make glm`) clean, no warnings.
- `-DCOLI_CUDA` syntax check clean (validates the new branch).
- ⚠️ **The OOM only reproduces on unified-memory hardware, which I can't test here** — asking the reporter to verify on GB10.

Closes #653 (pending reporter confirmation on GB10).

### DIFF
--- a/c/backend_cuda.cu
+++ b/c/backend_cuda.cu
@@ -710,6 +710,16 @@ extern "C" int coli_cuda_mem_info(int device, size_t *free_bytes, size_t *total_
     return cuda_ok(cudaMemGetInfo(free_bytes, total_bytes), "memory info");
 }
 
+/* #653: 1 when the device shares physical memory with the host (Grace-Blackwell /
+ * GB10, Jetson, integrated GPUs). On these the expert tier and the RAM cache draw
+ * from the same pool, so the RAM budget must account for the tier; on a discrete GPU
+ * VRAM is a separate pool and this returns 0. */
+extern "C" int coli_cuda_device_integrated(int device) {
+    cudaDeviceProp prop{};
+    if (!cuda_ok(cudaGetDeviceProperties(&prop, device), "device properties")) return 0;
+    return prop.integrated ? 1 : 0;
+}
+
 extern "C" void coli_cuda_stats(int device, size_t *tensor_count, size_t *tensor_bytes) {
     size_t count = 0, bytes = 0;
     for (int i = 0; i < g_nctx; i++) if (device < 0 || g_ctx[i].device == device) {
--- a/c/backend_cuda.h
+++ b/c/backend_cuda.h
@@ -30,6 +30,7 @@ COLI_CUDA_DLLEXPORT void coli_cuda_shutdown(void);
 COLI_CUDA_DLLEXPORT int coli_cuda_device_count(void);
 COLI_CUDA_DLLEXPORT int coli_cuda_device_at(int index);
 COLI_CUDA_DLLEXPORT int coli_cuda_mem_info(int device, size_t *free_bytes, size_t *total_bytes);
+COLI_CUDA_DLLEXPORT int coli_cuda_device_integrated(int device);
 /* device < 0 returns aggregate statistics for all configured devices. */
 COLI_CUDA_DLLEXPORT void coli_cuda_stats(int device, size_t *tensor_count, size_t *tensor_bytes);
 COLI_CUDA_DLLEXPORT void coli_cuda_group_stats(uint64_t *calls, uint64_t *experts, uint64_t *rows,
--- a/c/backend_loader.c
+++ b/c/backend_loader.c
@@ -33,6 +33,7 @@ typedef void           (*fn_shutdown)(void);
 typedef int            (*fn_device_count)(void);
 typedef int            (*fn_device_at)(int index);
 typedef int            (*fn_mem_info)(int device, size_t *free_bytes, size_t *total_bytes);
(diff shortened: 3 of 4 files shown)


JustVugg:
Merged to dev. Verification requested on GB10 hardware (see #653) — the OOM only reproduces on unified-memory GPUs, so CI green + reporter confirmation is the gate here.

PR #667, opened 2026-07-28
feat(core): model-architecture seam, chat templates, text antiprompt

khalilswdp:
## Summary

The engine is hardcoded to GLM-5.2: chat template, eos token, and forward-pass dispatch are all baked into colibri.c, so a second model family currently means a second binary (like olmoe.c). The smallest change that fixes this is a load-time seam:

- arch.h + arch_glm.c — a ModelArch descriptor registry selected from config.json "architectures"[0] (falling back to "model_type"). Unknown architectures are refused up front with the supported list, instead of silently mis-reading another model's weights. A config with neither field predates the registry and keeps loading as GLM with a notice, so existing snapshots and oracle fixtures are unaffected.

- engine.h — the shared engine types (Cfg, QT, Layer, Model, …) move verbatim out of colibri.c so a future arch module can compile as a standalone .c. The forward pass dispatches attention_rows/moe through a two-pointer vtable bound at load; GLM binds the existing built-ins. No speculative fields — a new family adds its fields in the PR that consumes them.

- Chat template/eos from the descriptor — serve/run build turns from chat_prefix/chat_turn/chat_think/chat_eos instead of hardcoded GLM strings; the GLM descriptor is byte-identical to the previous literals. router_bias gains NULL guards for families without e_score_correction_bias.

- antiprompt.h — a quantized model can emit a role delimiter like <|user|> as ordinary text tokens, so the id stop never fires (and serve drops role-marker ids from the stop set, #401) and the model hallucinates the next turn. This matches the decoded text, streamed with a (longest-needle − 1) hold-back so delimiters split across tokens are caught and never printed. COLI_ANTIPROMPT overrides or disables it.

Also: COLI_TPS_EVERY makes the tok/s heartbeat cadence configurable (default 8; the hardcoded 16 meant short replies never produced a reading; 0 silences).

## Validation

- [x] `make -C c check` — clean build, 0 new warnings, C unit tests + Python
      stdlib tests pass (the two `-Wformat-truncation` notes gcc 15 emits are
      pre-existing on `dev`; a separate small PR fixes them)
- [x] CUDA changes — n/a (no kernel changes; `arch_glm.o` is pure C linked into
      every build)
- [x] No performance claims in this PR

## Compatibility

- [x] The default CPU build remains dependency-free — the registry is pure C,
      two new headers + one `.c`, no new deps
- [x] No model files, generated binaries, or benchmark artifacts included
- Defaults changed: none. All new behavior is selected by the model's own
  `config.json` or opt-in env vars (`COLI_ANTIPROMPT`, `COLI_TPS_EVERY`).
- `.gitignore` gains pattern rules for build outputs (`*.o`, `*.obj`, …) so new
  sources' objects don't leak into `git status`.

### DIFF
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -26,3 +26,11 @@ tests/test_int3
 tests/test_int3_load
 tests/test_tok_o200k
 tests/test_pipe_block
+
+# build outputs, by pattern (per-file lists leak new sources' objects)
+*.o
+*.obj
+*.lib
+*.exp
+*.pdb
+*.bak
--- a/c/Makefile
+++ b/c/Makefile
@@ -392,8 +392,13 @@ $(file >.build-config,$(BUILD_CONFIG))
 endif
 .build-config: ;
 
-colibri$(EXE): colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h quant.h sample.h kv_persist.h telemetry.h route_trace.h omp_tune.h $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) $(VK_SPV) .build-config
-	$(CC) $(CFLAGS) colibri.c $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) -o colibri$(EXE) $(LDFLAGS)
+colibri$(EXE): colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h quant.h sample.h kv_persist.h antiprompt.h telemetry.h route_trace.h omp_tune.h arch.h engine.h arch_glm.o $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) $(VK_SPV) .build-config
+	$(CC) $(CFLAGS) colibri.c arch_glm.o $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) -o colibri$(EXE) $(LDFLAGS)
+
+# Model-architecture registry: model-agnostic engine + per-model descriptors.
+# Pure C, no deps — compiles on every platform/backend unchanged.
+arch_glm.o: arch_glm.c arch.h .build-config
+	$(CC) $(CFLAGS) -c arch_glm.c -o $@
 
 # Vulkan backend object (plain C + vulkan headers) and its SPIR-V shaders.
 backend_vulkan.o: backend_vulkan.c backend_vulkan.h .build-config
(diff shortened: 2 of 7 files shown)


JustVugg:
Thanks for this — the architecture seam is a nice direction. It now conflicts on `dev` (colibri.c / Makefile moved with recent merges). Could you rebase onto latest `dev`? CI is green; once it's clean we can land it, then #670 on top of it.

khalilswdp:
Rebased onto latest dev (a3a5a75) — applied cleanly, no conflicts. The conflicts you saw were from an accidental fork-sync merge that briefly polluted the branch; it's been removed and the branch is back to the single seam commit, now based on current dev. Ready to land — and the windows suite is already re-stacked on top for the follow-up.

khalilswdp:
Rebased onto latest dev (c67dbab). One real conflict with [#679](https://github.com/JustVugg/colibri/pull/679): both changes guard the spec-decode emit loop — your g_mux_stop/g_mux_cancel (client STOP/CANCEL mid-turn) and this PR's g_astop (text antiprompt) — resolved as the union of the stop conditions so both features coexist.

Why the antiprompt guard exists at all: a heavily-quantized GLM can emit a role delimiter like <|user|> as ordinary text tokens (<,|,user,|,>) instead of the atomic control-token id — the id stop never fires and the model runs on, hallucinating the next turn. And in serve mode the role-marker ids are deliberately dropped from the stop set anyway (#401, protecting <tool_call> blocks from int4 argmax noise), so even the atomic token wouldn't stop there. The antiprompt matches the decoded text instead, streaming with a short hold-back (longest needle − 1) so a delimiter split across tokens is caught and never leaked to the client.

Full local gate green at the rebased tip: build + C/Python suites, token-exact oracle 32/32 TF / 20/20 greedy, no new warnings. Ready to land — the windows suite is re-stacked on top and ready to follow.

JustVugg:

terrizoaguimor:
Read this properly because it overlaps my #716, and it turns out the overlap is almost nothing
textually and something real semantically. Both are worth knowing before either lands.

The design reads well to me. Refusing an unknown `architectures[0]` up front with the supported
list, rather than letting it mis-read another model's weights, is the right default — it is the
same reasoning behind the identity check in #700, and for the same reason: a wrong answer that
looks healthy is worse than a refusal. Keeping a config with neither field loading as GLM with a
notice is a good call too; that is exactly the tiny oracle fixture, and breaking it would have cost
someone an afternoon.

## Textual overlap with #716: only the Makefile

I diffed it rather than guessing. Of the functions #716 touches:

| | overlap |
|---|---|
| `model_init`, `pin_load`, `stats_dump_q`, `usage_load`, the `ROUTE_TRACE` block | none |
| `moe` | none — you change the **call site** to `g_ops.moe(...)`, #716 changes the **inside** |
| `eusage` | moved verbatim into `engine.h`; #716 aliases it in `model_init`, which stays |
| **Makefile dependency lines** | **conflicts** — you add `arch.h engine.h arch_glm.o`, #716 adds `route_trace.h`, to the same ~17 lines |

So the only conflict is mechanical and both sides of it are wanted. Whichever lands second takes
five minutes. No reason to sequence these on my account.

## The interaction that does matter

Today `arch_glm.c` registers one architecture with `.model_type = "glm_moe_dsa"`, and #716
hardcodes that same string as the identity written into `.coli_usage`:

```c
rt_init("glm_moe_dsa", c->n_layers, c->n_experts);
```

They agree by coincidence, and the coincidence ends the moment this seam does its job. With a
second architecture registered — the Qwen3.6 work in #712/#713 is the obvious next one — one
binary loads two model families and **stamps both histories with the same identity**. Two
consequences, both bad:

- A Qwen history is labelled `glm_moe_dsa`.
- Worse, the cross-model refusal in #700 does **not** fire, because the identities match. A
  placement history from one model would silently seed the other's expert pinning — which is the
  exact silent-misattribution failure that issue exists to prevent, arriving through a door I left
  open.

This is a problem with my PR, not yours; your seam is what makes it visible. The fix is small and
belongs on my side: derive the identity from the descriptor rather than hardcoding it, so it tracks
whatever `model_arch_select` returned:

```c
rt_init(arch->model_type, c->n_layers, c->n_experts);
```

`route_trace.h` only needs a string it can hash, so it does not care where it comes from. I will
make that change in #716 if this lands first, or note it here if mine does — either order works,
it just must not be forgotten. If you would rather I hold #716 until this is in so the seam is
available to use, say so and I will.

## One question about the descriptor

`chat_turn` is documented as "per-turn printf fmt, exactly two %s: (user, think)". Since the
strings are compile-time constants in a static registry that is fine today, but it is a format
string driving `printf` from a table — if an architecture is ever registered from anything but a
literal in this file, that becomes a formatted-output injection point. Worth a comment on the
struct field saying it must remain a literal, so the constraint outlives the current registry.

terrizoaguimor:
Correcting myself before this misleads anyone, including me.

I said above that the history identity in #716 would become wrong once this seam lands, and
committed to changing `rt_init("glm_moe_dsa", …)` to derive from the descriptor. I checked properly
before writing that patch, and **it would have been a no-op**: `REGISTRY[]` holds exactly one entry
today, whose `.model_type` is the same `"glm_moe_dsa"` string I hardcoded. Same output, plus a new
dependency from merged code onto an unmerged PR.

More to the point, I assumed the direction was one binary serving several families. The evidence
says otherwise — #712/#713 add `c/qwen36.c` as its own binary, the way `kimi_k3.c` and `olmoe.c`
already are. With one engine per family, an identity per binary is correct rather than fragile,
which is what #716 does.

So the hazard is real but conditional, and the condition is not this PR: it arrives the day a
**second** `ModelArch` is registered here and one binary can load two families. At that moment
`rt_init` has to take its identity from `g_arch->model_type` instead of a literal, or the two
families write histories claiming to be each other and #700's cross-engine refusal silently stops
firing. `load_cfg` already runs before `rt_init`, so the descriptor is available when that day
comes — it is a one-line change, just not one worth making before it means anything.

Sorry for the noise. The rest of the review stands, and #716/#719 have since merged, so the
Makefile conflict I flagged is gone too — @JustVugg's `TEST_BINS` rework in #731/#733 removed the
shared line entirely.

JustVugg:

khalilswdp:
Back now — I had some personal affairs that needed my attention for a stretch, and this
ended up waiting on them. Sorry for the silence, and thanks for the patience (and for the
queue/TEST_BINS changes, which made this rebase far less painful than the previous ones).

Everything is updated. What changed since you last looked:

- **Rebased onto current dev**, single pass. The engine-side conflicts, since you asked
  rather than guessed — the right call, two were real:
  - The fmt=4 grouped-scale fix this branch carried in `expert_group_issue`'s fallback is
    **dropped**: #762's descriptor path + refusal guard supersedes it, and depot-computed
    fmt=4 experts now ride that path instead.
  - New upstream signatures threaded through the depot code: `pin_load`'s trust flag
    (`0`, auto-discovered) and `qt_resolve_fmt`'s `stamped_name` (`NULL` — routed experts
    are never stamped), matching the existing call sites.
  - #716/#719's `rt_trace_open()` kept as-is; #766's expert-width term folded into the
    reworded cap=1 warning; the #686/#687 release-host logic intact alongside it.
  - The fmt=8 / private-ordinal documentation blocks moved with the QT typedef into
    `engine.h`, unchanged.
- **Review feedback addressed**: `chat_turn` now carries the must-remain-a-literal
  constraint on the struct field itself, as suggested. On the `rt_init` identity — agreed
  with the (self-corrected) framing: it's a one-liner (`g_arch->model_type`) that belongs
  to whichever PR registers a second descriptor in this binary; it's noted in the
  description so it isn't forgotten when that day comes.
- **Structure**: the DirectStorage transport is split out of the Windows-suite PR into
  its own PR, stacked between the suite and the VRAM depot — one concern per PR.
- **Validation** re-run at every commit of the stack: `make -C c check`, token-exact
  oracle (32/32 TF, 20/20 greedy), no warnings beyond dev's own baseline on the same
  toolchain. One unrelated observation from those runs: the new `test_fp8_repack` /
  e2e tests fail on a plain Windows box at current dev — the e2e harness compiles with a
  bare `gcc` invocation and `compat.h` `#error`s without `-D_FILE_OFFSET_BITS=64`.
  Happy to file a separate issue with details.

PR #670, opened 2026-07-28
feat(win): fix silent CPU fallback, launcher suite, DirectStorage expert loads

khalilswdp:
> **Stacked on https://github.com/JustVugg/colibri/pull/667
#667 ** (model-architecture seam) — its commit shows here until
> it merges. This PR's own work is three commits, reviewable independently:
> 1. `fix(win)`: static cudart + launcher/cap-warning fixes
> 2. `feat(win)`: PowerShell launcher suite
> 3. `feat(win)`: DirectStorage expert-load transport (opt-in)

## Summary

On Windows every run silently fell back to CPU: `coli_cuda.dll` linked dynamic
cudart, but CUDA 13.x keeps `cudart64_*.dll` in `bin\x64` (not on PATH), so
`LoadLibrary` failed quietly. Fix: `-cudart static` (the DLL then imports only
KERNEL32). The rest makes the fix visible and usable:

- **`coli` launcher**: banner prints the real backend instead of a hardcoded
  "streaming CPU"; numeric env vars parse tolerantly (`CUDA_EXPERT_GB=auto` no
  longer crashes); `--ram` accepts fractional budgets.
- **cap=1 warning**: quotes the actual RAM shortfall and what `--ram` would
  buy a second cache slot, instead of unconditionally blaming `PIN_GB`.
- **`windows/*.ps1` suite**: Build (toolchain detect + doctor), Run (wizard,
  attaches to a running serve instead of starting a second engine), Chat, Web,
  Warmup. Saved with a UTF-8 BOM — PowerShell 5.1 reads BOM-less files as
  ANSI, which breaks the multilingual warmup prompts.
- **`warmup.ps1`**: `-Backend auto|gpu|cpu` (the `.coli_usage` pin is
  backend-flavoured), `-PromptFile`, +40 multilingual/code prompts.
- **DirectStorage expert loads** (opt-in `COLI_DSTORAGE=1`, default off): the
  miss path's coalesced ~19 MB slab pread can be served by a DirectStorage
  host-destination DMA instead. Pure transport swap — bytes land in the slab
  the expert cache already owns; cache/eviction/layout untouched; primary
  replica only; any failure (no SDK at build, no `dstorage.dll`, no adapter,
  request error) falls back to pread. The depot PR reuses this transport for
  disk→VRAM. No performance claims yet — correctness-verified only.

## Validation

- [x] `make -C c check` full pass at this tip; zero new warnings
- [x] `make CUDA_DLL=1 colibri.exe` clean; DLL builds with and without
      `DSTORAGE_HOME`
- [x] DirectStorage A/B on a real int4 container: `COLI_DSTORAGE=0` vs `1`
      byte-identical output; served loads reported at exit
- [x] Token-exact oracle with the flag off (default): 32/32 TF, 20/20 greedy
- [x] All six `.ps1` scripts parse under Windows PowerShell 5.1 and PS 7
- [x] Static-cudart DLL: dumpbin shows KERNEL32 imports only; GPU engages on a
      box where the dynamic link failed silently

## Compatibility

- [x] Default CPU build unchanged and dependency-free — DirectStorage is
      opt-in at build time (`DSTORAGE_HOME`), optional in the loader;
      Linux/macOS unchanged
- [x] No model files or binaries included
- Defaults changed: none in the engine.


JustVugg:
Holding this behind #667 (its arch-seam base). Please rebase #667 first onto latest `dev`, then this one on top — CI green on both, just needs the merges resolved.

JustVugg:

JustVugg:

JustVugg:
Closing this one because #825 already carries all of it, and asking you to rebase two stacked branches for one body of work would be a waste of your time.

Verified by file set, not by title: all 15 files here are also in #825, which adds `backend_dstorage.cpp`, `backend_loader.c` and `backend_cuda.h` on top. There is nothing in this branch that #825 does not have.

#825 stays open. It is waiting on one decision that is mine, not yours, and it is the same decision that #667 is waiting on: whether the model-architecture seam is still wanted now that eight engines have shipped as separate binaries. I owe you that answer and I am sorry it has taken this long. Once it is settled, the DirectStorage transport is the part with no equivalent anywhere in the tree and I would like it either way.

PR #671, opened 2026-07-28
entropy: in-tree rANS codec, int4-rans256-g0 container format, offline tools

monotophic:

## What this is

The first deliverable of the transport-layer build declared on #651
(follow-through on #594): a dependency-free C99 rANS codec (`c/rans.h`,
header-only), the **`int4-rans256-g0`** container record format, and the
offline tools to mint, verify, and document it. **No engine inference path
is touched** — `colibri.o` compiles byte-identical to base; everything here
is codec + tools + tests + docs, useful standalone and consumable by any
tier (the CUDA capacity tier included — one codec, every consumer, no
external dependency).

- **Codec**: scalar + vectorized batch decode (AVX-512 F+BW and NEON arms),
  each carrying the i4-precedent runtime envelope — first-use round-trip
  selftest (with machine-checked renorm-band coverage) and env
  kill-switches. Vector arms are byte-identical to scalar by contract and
  by test — this is a codec: no tolerance, only identity.
- **Format**: per-tensor records under original safetensors names (256
  round-robin-interleaved rANS streams, GPU/threadgroup-friendly), the
  shared table embedded per-shard in `__metadata__`, and a **mandatory
  format stamp** — entropy-coded sizes are data-dependent, so no
  byte-arithmetic inference exists; unstamped tensors can never be
  misread (refuse-rather-than-guess, same philosophy as #529).
- **Tools**: `repack_rans.py` (deterministic: two runs byte-identical;
  positive per-row geometry verification; crash-evident multi-shard
  manifest) and `rans_verify.py` (TRUST-VERIFY-REFUSE: a named refusal
  class for every corruption class, never a crash — fuzz-hardened).
- **Docs**: `docs/int4-rans256-g0.md` is complete enough that a third-party
  reader AND writer can be built from the doc alone (independently
  demonstrated during review).

## Measured

| | ratio / throughput |
|---|---|
| Real-corpus repack (2 shards, 5.4 GB, GLM-5.2 int4) | **0.7305** (27.0% smaller incl. framing); full-corpus census anchor 0.763 |
| Decode, Zen 5 AVX-512 (32T, pure C) | **20.2 GB/s** nibble-out (10.1 GB/s packed), 3.0× scalar |
| Decode, M-series NEON (12T, pure C) | 5.0 GB/s nibble-out (2.5 GB/s packed), 1.5× scalar |
| Decode, Metal (prototype, prior work) | 24.3 GB/s single-dispatch warm median |
| Encoder reproduces the proven prototype bitstream | **byte-for-byte** (offsets + payload) |

## Capstone matrix

| requirement | decisive evidence |
|---|---|
| engine untouched | `colibri.o` byte-identical base-vs-branch; diff = new files + Makefile/test wiring only |
| lossless, always | round-trip property suite (random/adversarial/real tensors) + 4,907-mutant structured fuzz under ASan+UBSan: zero memory errors, zero arm divergence |
| vector = scalar, bit-exact | byte-identity across arms on every fuzz-accepted record + real chunks, on both silicons |
| refuse, never guess | named-refusal battery (25+ classes incl. u64-wraparound, decompression-bomb bounds, hostile headers) with **C/Python refusal parity as a tested property** |
| deterministic containers | two-run byte-identity + stable-sort table construction + cross-implementation (C vs pure-Python) byte-identity |
| third-party consumable | reference parser + decoder built from the docs alone during review, decoding real records byte-exact |

<details>
<summary>Review + hardening detail</summary>

Built under a two-reviewer gate (blind validator + deep audit, independent
harnesses). The hardening round closed everything found: non-wrapping
header validation + amplification bounds (`E_OVERSIZE`), corrected encoder
scratch bound (+`E_SCRATCH`), `E_SYMBOL_UNCODABLE` pre-scan, never-crash
verifier (hostile-header suite: zero tracebacks), positive 2-D geometry
verification (g64/g128 refused by name — this PR scopes v1 to per-row int4
deliberately), `E_NSTREAMS`/`E_UNALIGNED` guards, big-endian `#error`,
crash-evident repack manifest. The fuzzer ships as `make fuzz-rans`
(seeded, sanitizer-built, non-gating). Suite: `make check` +227-test Python
battery; gcc-13 `-Wall -Wextra -Wpedantic -Wshadow` clean; the AVX-512
arm's first silicon run passed the full suite with the envelope exercised
both ways.

</details>

Scope notes: v1 targets per-row int4 expert tensors (the census's 0.763
container class); g64-class and scale-array coding are documented future
rounds with their own (weaker) payoff cases. Engine integration (loader +
CPU streaming decode, then Metal batched decode) follows as separate PRs —
this one is deliberately inert: format + codec + tools, reviewable and
usable on their own.

*Durable vs current-state: format, codec, envelope, and refusal semantics
are durable; throughput figures are calibrations at (Zen 5 Strix Halo /
M5 Max, 2026-07-28, dev fa599e0 base) and the ratio at (GLM-5.2 int4
per-row container, 2-shard sample vs 372 GB census).*

### DIFF
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -8,3 +8,6 @@ tests/test_st_pread
 tools/libiq3.so
 tools/libiq3.dylib
 tools/iq3.dll
+tools/librans_c.so
+tools/librans_c.dylib
+tools/rans_c.dll
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -11,3 +11,5 @@ tools/iq3.dll
 tools/librans_c.so
 tools/librans_c.dylib
 tools/rans_c.dll
+tests/fuzz_rans
+tests/*.dSYM/
--- a/c/Makefile
+++ b/c/Makefile
@@ -460,6 +460,18 @@ tests/test_e8_kernel$(EXE): tests/test_e8_kernel.c quant.h
 tests/test_rans$(EXE): tests/test_rans.c rans.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
+# Structured-mutation fuzz for the rans record parser/decoder under
+# ASan+UBSan — seeded and bounded (a few thousand mutants, well under a
+# minute), asserting no sanitizer report and byte identity across every
+# compiled decode arm on every accepted record. NOT a test gate (sanitizer
+# flags differ from the normal build); build+run on demand:
+#   make fuzz-rans
+fuzz-rans: tests/fuzz_rans.c rans.h
+	$(CC) -O1 -g -fsanitize=address,undefined -fno-sanitize-recover=all \
+	  -Wall -Wextra -Wno-unused-parameter -Wno-unused-function \
+	  tests/fuzz_rans.c -o tests/fuzz_rans -lm
+	./tests/fuzz_rans
+
 tests/test_int3$(EXE): tests/test_int3.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
@@ -505,7 +517,12 @@ tests/test_pipe_block$(EXE): tests/test_pipe_block.c colibri.c st.h uring.h json
 test-c: $(TEST_BINS)
 	$(PYTHON) tools/run_tests.py $(TEST_BINS)
 
-test-python:
+# test-python builds the rans ctypes bridge first: without it the repack e2e
+# test's C-vs-Python byte-identity pin would silently compare Python against
+# itself (the test ALSO skips loudly if the library is somehow still absent,
+# e.g. when invoked via unittest directly — but a green `make test` must mean
+# the pin was actually checked).
+test-python: $(RANSLIB)
 	$(PYTHON) -m unittest discover -s tests -p 'test_*.py'
 
 test: test-c test-python
@@ -567,4 +584,4 @@ clean:
 
 bench: iobench$(EXE)
 	@if [ -n "$(ARGS)" ]; then ./iobench$(EXE) $(ARGS); else echo "built iobench$(EXE) — run: ./iobench$(EXE) <file> <MB> <iters> <threads> <direct 0|1>"; fi
-.PHONY: all colibri glm iq3 rans cuda-test hip-test gpu-compile cuda-bench cuda-dll portable test-c test-python test check clean install uninstall bench
+.PHONY: all colibri glm iq3 rans fuzz-rans cuda-test hip-test gpu-compile cuda-bench cuda-dll portable test-c test-python test check clean install uninstall bench
--- a/c/Makefile
+++ b/c/Makefile
@@ -337,6 +337,22 @@ iq3: $(IQ3LIB)
 $(IQ3LIB): tools/iq3_encode.c
 	$(CC) $(CFLAGS) -fPIC -shared $< -o $@ $(LDFLAGS)
 
+# int4-rans256-g0 codec bridge for tools/repack_rans.py + tools/rans_verify.py.
+# OPTIONAL: both tools fall back to a pure-Python codec producing the same
+# bytes far more slowly when the library is absent (fine for tests, not for a
+# real repack). Same ARCH story as the engine: an AVX-512/NEON host gets the
+# batched vector decoder, anything else the scalar paths.
+ifneq (,$(DARWIN))
+RANSLIB = tools/librans_c.dylib
+else ifneq ($(IS_WIN),)
+RANSLIB = tools/rans_c.dll
+else
+RANSLIB = tools/librans_c.so
+endif
+rans: $(RANSLIB)
+$(RANSLIB): tools/rans_ctypes.c rans.h
+	$(CC) $(CFLAGS) -fPIC -shared $< -o $@ $(LDFLAGS)
+
 cuda-test: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/test_backend_cuda.cu
 	@command -v "$(GPUCC)" >/dev/null 2>&1 || { echo "$(GPUCC_NAME) not found" >&2; exit 1; }
 	"$(GPUCC)" $(GPUFLAGS) backend_cuda.cu tests/test_backend_cuda.cu -o backend_cuda_test$(EXE)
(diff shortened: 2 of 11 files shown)


monotophic:

@ZacharyZcR — the in-tree record schema is concrete: this PR. Your
requirements list from the 6×5090 work, mapped against it as shipped:

| your requirement | status in this PR |
|---|---|
| independently decodable tensor/expert records (aggregate archives hit a CUDA boundary) | met by design — per-tensor records under original safetensors names, each self-contained (own stream offsets; shard-local table). No aggregate archive exists in this shape. |
| per-record original size, encoded size, codec bound, shape, truncation validation | met — the verifier enforces all of it as named refusal classes (non-wrapping count checks, an amplification bound, per-stream truncation, table CRC), fuzz-hardened, with C/Python refusal parity as a tested property. |
| CPU decoder for the streamed transport case | met — in-tree, dependency-free; scalar + NEON + AVX-512 (10.1 GB/s packed measured on Zen 5, 32T). |
| optional pointer-batch GPU decoder for compressed-VRAM | format-ready by design — the 256-stream round-robin interleave exists precisely for descriptor-batch GPU dispatch (24.3 GB/s measured on the Metal prototype of this exact construction). A CUDA port is close to line-for-line and needs no external codec. |
| exact model/traversal identity incl. placement order | order-identity dissolves under name-addressed random access — no consumer needs traversal order anymore, which we'd argue is the point of the portable shape. Model identity: names + shapes + the mandatory per-tensor format stamp. **Container-level** identity (digest/stamp of the whole artifact) is deliberately NOT duplicated here — that is #529's machinery, and when it lands it applies to these shards like any others. |
| alignment permitting direct reads without a second staging copy | workable today with the same aligned-window read your sidecar loader uses (one read, offset delta, no second copy); records are 16-byte aligned internally. **Pre-registered trade, decided by your data:** if your A/B shows the aligned-window pattern costs materially versus true 4 KiB tensor alignment, we add an `--align` option to the repack tool — a shard-layout change only, the record wire format does not move. |
| fail-fast, never silently substituting a missing tier | shared philosophy — the stamp is the only dispatch signal (entropy sizes admit no byte-arithmetic inference), so unstamped or malformed tensors refuse rather than guess. Tier-substitution policy lives in the consumers; ours will match yours. |

Your validation offer is exactly what we'd like this PR to carry into
review: packing, complete-record verification, direct loading, and the
6×5090 A/B against the 110 GB sidecar — run against v1 as it stands. If it
surfaces a real defect, that's what review is for; the alignment question
above is the one open trade and your A/B is its designated decider.

*(Throughput figures are calibrations at their stated hosts/dates —
E1_SILICON_RESULTS in the PR evidence; ratio 0.7305 on a 2-shard real-corpus
sample vs the 0.763 full-corpus census anchor.)*

ZacharyZcR:
@monotophic Thanks for mapping the requirements to concrete code. I reviewed the actual v1 scope and want to separate what the 6x5090 host can validate now from what needs the consumer PR.

What is testable in this PR:

- deterministic repack and complete-record verification on the real GLM shards;
- corruption/truncation refusal and C/Python parity;
- record-size distribution and the real cost of aligned-window direct reads;
- whether 16-byte record alignment is sufficient in practice or a 4 KiB `--align` layout materially improves direct-read throughput.

What is not yet testable as a capacity-tier A/B:

- this PR has no CUDA pointer-batch decoder or Colibri loader/dispatch path, so there is no rANS-backed GPU tier to compare with #651's 110 GB DietGPU sidecar;
- v1 deliberately supports per-row int4 and refuses g64, while current Colibri conversion defaults and the newer public GLM container are g64. The old per-row checkpoint can exercise v1, but that result would not establish coverage for the current default container;
- fail-fast tier substitution remains a consumer property, not something this inert format PR can demonstrate;
- until #529's container identity lands, names + shapes prevent traversal coupling but do not by themselves prevent attaching same-shaped records from the wrong checkpoint. The first consumer still needs an external whole-model digest/manifest check.

So my review plan is:

1. validate the phase-1 format/tooling claims independently on the per-row GLM source;
2. use the resulting record offsets/sizes to settle 16-byte vs 4 KiB alignment;
3. reserve the 6x5090 throughput/capacity comparison for the CUDA consumer PR, where identical placement, routed replay, decode-time CPU/GPU windows, and fail-fast behavior can actually be measured.

That does not block a deliberately inert codec/format PR from standing on its own. It just keeps “format-ready for GPU dispatch” distinct from “GPU capacity tier validated.” I will post the packing/verification/alignment results here when that host run is complete.

monotophic:

@ZacharyZcR — this is the review split we wanted; the
testable-now list matches what the PR claims. Three
responses on the three substantive points:

**g64 coverage — deliberate scope, chosen on census numbers.** The
744G-nibble census measured per-row int4 at 0.763 corpus-wide but
g64-class at ~0.890 — better quantizers make less-compressible codes, and
on g64 the scale arrays are 12.3% of the stream (vs 0.13% per-row), so a
worthwhile g64 round is really a different design that adds scale-array
coding. v1 targets the container class where the payoff is proven; the
PR's scope notes acknowledge that as does your framing. The g64 extension 
is deferred work, to be built when its (weaker) case is worth 
its complexity.

**The wrong-checkpoint identity gap — agreed, and here's where each layer
lands.** Container identity proper is #529's machinery (open, CI-green;
when it lands these shards get stamped like any others). Interim and
consumer-side: the repack tool already emits a per-run manifest
(crash-evidence, from the hardening round) — we're growing it with
optional per-tensor content digests as a tool-layer commit **on this PR**
(record wire format untouched; includes a retro-manifest mode so
already-minted shards can be covered by hashing in place), and the first
consumer PR ships the whole-model manifest check at load. So: #529 =
identity of the artifact; manifest digest = the external byte-level
verification you're pointing at, each with a concrete home.

**Alignment — your experiment remains the designated decider**, as
pre-registered above: if the record offsets/sizes show aligned-window
reads paying materially vs a true 4 KiB layout, `--align` lands in the
repack tool, wire format unmoved either way.

Looking forward to the per-row host run — the phase-1 validation you've
scoped is the strongest external evidence this PR could carry, and the
capacity-tier A/B waiting for the consumer PR is the right sequencing on
our side of the ladder too (the loader/dispatch consumer is in build now).

monotophic:

Pushed `cc4a589` — the manifest-digest commit promised above, landing the
interim whole-artifact verification @ZacharyZcR's review pointed at. Record
wire format untouched; the commit is tools/tests/docs only (engine object
still byte-identical to base).

What it adds:

- **`repack-manifest.json` now carries two digest layers**: a whole-file
  sha256 per shard (covers `.qs` sidecars, headers, padding — the parts no
  per-record check can see) and a per-tensor sha256 of each record's bytes
  (localizes damage to a tensor). Hashed while streaming the write — no
  second read.
- **`rans_verify.py` checks both layers when present**: file gate first
  (`E_SHARD_DIGEST_MISMATCH`), then per-record
  (`E_DIGEST_MISMATCH`/`E_DIGEST_MISSING`). Manifest absent or digest-less
  → note-and-proceed (older mints stay loadable); manifest present but
  damaged → refuse (`E_MANIFEST_MALFORMED`) — a damaged integrity artifact
  is evidence, not noise. Manifest *deletion* remains the inherent sidecar
  gap; that's the consumer/identity layer's job (#529 lineage + the
  loader check arriving with the consumer PR).
- **`--manifest-only <dir>`**: retro-manifests already-minted output by
  hashing in place — no re-encode. Retro over a minted dir reproduces the
  mint-time manifest byte-for-byte (tested). If your host run has already
  repacked, one command covers those shards with digests after the fact.

Verification: tamper bites demonstrated for both layers (including a
`.qs`-region flip that trips *only* the file gate — the scale-swap case,
asserted with zero record-level noise); two-run mint determinism includes
the manifest; retro-vs-mint byte identity; full suite green. Independent
blind re-validation covered the same claims with its own hashing and
header parser before this push.

*(Digest fields are sidecar-file additions — nothing about record layout,
table serialization, or stamps moved; v1 as frozen.)*

Issue #455, opened 2026-07-20
GLM-5.2 int4 think mode: stochastic reasoning loops / EOS starvation on stock v1.0.0; #307's TEMP=0.9 NUCLEUS=0.95 stopgap makes it severe


lukascechovic:
## Environment

- colibrì v1.0.0, commit 72874f3, fully stock (`coli chat` — no HTTP layer; also confirmed via stock `coli serve`)
- GLM-5.2 744B MoE, int4 checkpoint in the **pre-fix per-row-scale format** (the format #225 identified as an incoherence root cause)
- Linux, CPU-only (no ROCm/CUDA), 122 GB RAM, `--ram 80 --topp 0.7`, `DIRECT=1 PIPE_WORKERS=2 PILOT_REAL=1`
- `.coli_kv` deleted before each run, `KVSAVE=0` (except the KV probe below)
- Prompt: `2+2 = ?`, `THINK=1`

## Repro

```bash
rm -f "$MODEL/.coli_kv"
printf '2+2 = ?\n' | env DIRECT=1 PIPE_WORKERS=2 PILOT_REAL=1 THINK=1 KVSAVE=0 \
  TEMP=0.9 NUCLEUS=0.95 \
  coli chat --no-attach --model "$MODEL" --ram 80 --topp 0.7
```

## Observed

**With #307's suggested `TEMP=0.9 NUCLEUS=0.95`** (severe, reproduced 1/1): the model hallucinated a different problem and never terminated — killed after 30 min / ~1050 tokens. Excerpt:

```
2 × 1 = 2
Simplify 4(???). What is the value of ?+?+?* ?
...
Okay, I think we're going to to the theater. (Ha ha!) But I'm going to the movie. (Ha, ha!)
...
So, what is the denominator? Is there a second "denominator"? ...
```

Stray `</think>` tokens appear mid-text; no EOS ever sampled.

**With defaults (0.7/0.90)** (stochastic): one `coli serve` sample was clean; the identical repeat produced a 10× "Wait, …" self-correction spiral before finally answering. A stock `coli chat` sample never closed its `<think>` block (answered inside reasoning, then EOS).

**KV resume amplifies it:** with KV saving enabled, two fresh-server requests were clean, but after a server restart resumed the saved `.coli_kv` (29 MB), the same request ran into a "Wait" loop to the `max_tokens` cap with no post-think answer — and the cap-hit was reported as `finish_reason:"stop"` instead of `"length"`.

## Notes / suspected causes

[...]

Issue #537, opened 2026-07-22
[Analysis] The DRAM bandwidth wall: decode performance ceiling for streaming MoE, measured paths forward


ZacharyZcR:
## Summary

Streaming MoE decode (S=1) on GLM-5.2 744B is **DRAM-bandwidth-bound**, not PCIe-bound and not compute-bound. This analysis quantifies the physical ceiling, shows we are already at 79–96% of it on a 6x RTX 5090 rig, identifies where the remaining 4–21% is lost, and proposes research directions — both software and hardware — for the community to discuss.

All numbers below are from controlled measurements on a 6x RTX 5090 + dual Xeon Silver 4510 (24 physical cores) + 251 GB DDR5 system, with GLM-5.2 int4 (gs=64), `DRAFT=0`, usage-snapshot-controlled placement, and `PROF=1` instrumentation. PRs #429 (NUMA arena), #432 (device router), #433 (grouped kernels), #434 (resident accumulation), and #445 (pin budget fix) are all applied.

---

## 1. The bandwidth hierarchy — where the wall actually sits

For decode S=1, every expert evaluation is a GEMV: pure memory-bound, zero compute reuse. The only thing that matters is how fast weight bytes reach the arithmetic units.

| Tier | Theoretical peak | Measured | Time to read one expert (~20 MB) |
|---|---|---|---|
| GPU VRAM (GDDR7, per card) | ~1,792 GB/s | — | **0.011 ms** |
| PCIe 5.0 x16 (per card) | ~63 GB/s | ~50 GB/s | **0.4 ms** |
| CPU DRAM (DDR5, dual-socket) | ~120 GB/s | 62–70 GB/s | **0.3 ms** |
| NVMe SSD (PCIe 4.0 x4) | ~7 GB/s | host-dependent | **~3 ms** |

The critical ratio: **VRAM bandwidth is 25–28x DRAM bandwidth.** Once an expert lives in VRAM, it is effectively free. Once it lives in RAM, it costs 0.3 ms regardless of who computes it.

## 2. PCIe is not the bottleneck (and NVLink would not fix it)

A common intuition is that PCIe is the chokepoint between GPU and CPU memory. The data says otherwise:

```
PCIe 5.0 x16 throughput:   ~63 GB/s (measured ~50 GB/s)
DDR5 DRAM throughput:      ~120 GB/s theoretical, 62-70 GB/s measured
```

[...]

Issue #387, opened 2026-07-18
Datapoint: M5 Max 128 GB — decode-tuning A/Bs (MTP × cache 2×2): 0.4 → 2.0 tok/s at identical output


monotophic:
*Authored by Fable 5 in Claude Code, analysis in partnership with @Monotophic.*

**Coordinates (per-measurement, so future readers can judge staleness):** measured 2026-07-15..17
on a MacBook Pro M5 Max (`Mac17,7`, 128 GB unified, 6P+12E, internal Apple NVMe), macOS 26.5,
engine `main` @ `72d3d37`, `make glm METAL=1`, `COLI_METAL=1`, **zero local patches**. Model:
`mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp` (144 shards / 384 GB). Everything below is
**identical-output tuning** — no routing substitution, no expert pruning, greedy temp-0 decode.

## TL;DR

Two independent, additive levers take this box from ~0.4 tok/s (defaults/auto-RAM) to ~2.0 tok/s:

**Champion config:** `MTP=0 CAP_RAISE=0 AUTOPIN=0 coli run --ram 90 --cap 1`

| Lever | Direction | Effect (this box) |
|---|---|---|
| **`MTP=0`** | speculative decode off on Metal | **+79% at a cached config, +10–22% at cap1**; −20–45% bytes moved |
| **`--cap 1`** (no LRU cache) | smaller cache = faster, always | expert-matmul 20.6→35.4→78.4s as the cache grows |
| `--ram` | flat on speed across 20–40 | pick on the heat/energy axis, not the speed axis |
| `COLI_MMAP=1` | do not use on Metal | 39× slower (GPU demand-faults a 358 GB mapping) |

The counterintuitive headline: **the engine's own expert cache hurts on a big-RAM Mac.** The
mechanism (Metal per-buffer residency cost on LRU slot churn) is diagnosed in #379, and PR #386
turns the `--cap 1` half of this into a platform-aware default. This post is the data behind both.

## Method (what it took to make these numbers trustworthy)

Our first two measurement rounds were invalid and we threw them away. What survived, applied to
every run below:

- `--temp 0` (greedy): identical token stream, so any text diff is a real change (md5-checked).
- `.coli_usage` restored to a pristine snapshot before every run (`usage_save()` rewrites it each
  run and silently poisons AUTOPIN comparisons otherwise).
[...]

PR #674, opened 2026-07-28
docs: define colibri as an experimental inference engine

ZacharyZcR:
## Summary

- define colibri as an experimental inference engine and research platform
- state the objective: optimize the complete inference path to reduce hardware dependency and inference cost
- distinguish measured findings, conditional wins, correctness properties, and open hypotheses instead of presenting every technique as guaranteed acceleration
- document the core techniques: hierarchical placement, learned weight staging, I/O overlap, heterogeneous execution, compressed state, and measured speculation
- publish an open research matrix covering routing history, multi-SSD I/O, automatic hardware planning, representation/compression, speculation, and CPU/GPU overlap
- give contributors a reproducible experiment protocol and explicitly welcome controlled negative results
- update the roadmap and correct the absolute #455 wording
- align the English, Simplified Chinese, Traditional Chinese, and Italian READMEs

## Scope

README positioning and research documentation only. No runtime, build, API, or configuration behavior changes.

## Validation

- `git diff --check`
- verified the positioning, measured-findings, and open-experiments sections in all four language files
- reviewed modified Markdown links, tables, and structure


ZacharyZcR:
Added the Discord community link to the top navigation and contribution section in all four README languages: https://discord.gg/fpQxKnRb

monotophic:
*Authored by Fable 5 in Claude Code, analysis in partnership with @Monotophic (We/our = the joint analysis; me/my = @Monotophic's own hardware and commitments).*

Glad to see this merged — the measured / conditional / correctness / hypothesis taxonomy is the repo formalizing what its best threads already do, and we want to put two concrete datapoints behind it plus a flag of interest on one research-matrix row.

## The taxonomy does real work — two cases from our own record

**A conditional win, correctly filed:** `COLI_METAL_RESSET=1` on my M5 Max took decode from 0.89 to a flat 2.28–2.39 tok/s (measured both ways, #379) — a >2.5× win that is *platform-conditional* by mechanism, not a guaranteed acceleration. Under the old framing that number invites over-claiming; under this taxonomy it files cleanly as a conditional win with a stated condition. That's the framework earning its keep.

**A controlled negative result, welcomed:** we built and instrumented three generations of prefetch on the storage tier and refuted all three with the instrument (full record on #387; occupancy-vs-saturation trap documented on #537). Under "every technique is an acceleration," that work looks like failure; under this document it's exactly the contribution the protocol asks for. A repo that welcomes scored refutations gets honest numbers.

## The "automatic hardware planning" row: we're active in this lane

The research matrix names automatic hardware planning as an open direction. We've been building the measurement-first end of it: a **digital twin** of Colibri decode — a serial-sum model (disk + expert-matmul + attention + lm_head + other) whose every parameter carries measured provenance and an uncertainty class, validated by hindcast against a 34-run corpus on my M5 Max. Components are atomic and substitutable, so the same model prices a candidate config — or candidate *hardware* — before anything runs.

What makes planning tractable is that some parameters turn out to be class-constants rather than machine-specifics. Two measured examples from my fleet: sustained STREAM lands at **44% of the spec-sheet bandwidth on two different SoC families** (114 of 256 GB/s theoretical on Strix Halo, 120 of 273 on GB10 — best-of-10 triad, working set ≫ LLC), and the ~74% expert cache-hit ceiling on my M5 Max behaves as a routing-concentration constant of model+workload rather than a cache-size effect (#537). Constants like these are what let a planner generalize beyond the box it was calibrated on.

This lane also composes with #673: `coli tune` searches the config space empirically on one machine; a twin predicts across machines; profiles and predictions cross-validate each other. We've offered a twin-pair profile-transfer experiment on that thread that tests exactly this seam.

Flagging interest, not claiming the row: our methodology and fleet dataset (three machine classes measured across five boxes, more standing up) are available to whoever works this direction, and we expect to bring a concrete proposal here once our current transport-layer work (#671's follow-on rungs) clears the queue.

Issue #510, opened 2026-07-22
RDNA2 (gfx1030) field report: AMD backend works; greedy decode not token-stable across mixed CPU/GPU expert tiers


ab2m:
## Summary

Field report: the v1.1.0 AMD backend **works on RDNA2** (RX 6900 XT, gfx1030) — we believe this is the first validation outside the RDNA4 card it was developed on. It runs and the VRAM expert tier behaves as designed, but **greedy decode is not token-stable when experts are split across the CPU and GPU tiers**, which differs from the token-exact result reported for RDNA4 in #339/#499.

## Environment

- GPU: AMD Radeon RX 6900 XT 16 GB (RDNA2, gfx1030 — no WMMA), also driving the desktop
- ROCm 7.1.1 / hipcc 7.1.1 (Ubuntu distro packages, `ROCM_HOME=/usr`), kernel 7.0.0-27-generic
- CPU: Ryzen 9 5900X, 60 GB RAM, model on a Corsair MP600 PRO XT NVMe
- Build: dev branch (v1.1.0), `make HIP=1 HIP_ARCH=gfx1030 ROCM_HOME=/usr` — compiles clean (85 host warnings), engine detects the card as `sm_103`
- Model: `mateogrgic/GLM-5.2-colibri-int4-with-int8-mtp` (int4 + int8 MTP heads)
- Requires the `COLI_TEMP` patch from #509 (comment) for any `--temp` run

## What works

- `COLI_CUDA=1 COLI_GPU=0 CUDA_EXPERT_GB=12`, "routed experts only" mode: hot tier pins learned experts (166 → 1,335 tensors across runs as `.coli_usage` grows), thousands of calls served from VRAM.
- Numbers on this disk-bound box (24-token runs, `--topp 0.85`): cold CPU 0.10 tok/s (hit 46%) → GPU 0.38–0.39 tok/s (hit 63–64%, pin up to 36%). Warm CPU alone also reaches ~0.37 tok/s, so on this machine the GPU tier's main effect is the higher hit rate; disk streaming remains the bottleneck, as expected.

## The finding: greedy decode is placement-dependent on the mixed tier

Same binary, same flags (`--temp 0`, `DRAFT=0`, `--topp 0.85`), same prompt, 3-way matrix:

| Comparison | Result |
|---|---|
| CPU-only run 1 vs run 2 | **token-identical** (deterministic) |
[...]

PR #681, opened 2026-07-29
fix(hip): support GPUs without matrix cores (gfx9xx, gfx101x/103x)

MrMetric:
> **Authorship**: this patch and write-up were authored by Claude Fable 5 (Anthropic's AI) working interactively on MrMetric's machine; MrMetric reviewed the diff, ran the builds on his hardware, and submits it. Questions in review will be answered by both.

## Problem

`make glm HIP=1` fails on any AMD GPU without matrix cores — MI50/Vega (gfx9xx without MFMA), RDNA1 (gfx101x), RDNA2 (gfx103x) — because `backend_gpu_compat.h` unconditionally includes rocWMMA, whose headers `static_assert("Unsupported architecture")` for those targets (tracked upstream as ROCm/TheRock#1944; these archs will never be supported).

The engine itself doesn't need WMMA to run: both WMMA host-dispatch sites already have fallbacks (`compute_major<7` pre-Volta path), and the grouped expert kernels (`grouped_hidden_g4_dual` etc.) are plain HIP.

## Change

- **`c/backend_gpu_compat.h`** — rocWMMA is included only when the target arch can use it. Otherwise `COLI_GPU_HAS_WMMA=0` and `__CUDA_ARCH__` stays undefined, so the WMMA kernel bodies compile out and host dispatch takes the existing fallbacks. Missing `rocwmma-dev` on a WMMA-*capable* arch remains a hard `#error` (a silent tensor-core downgrade would be worse). Non-WMMA archs no longer need `rocwmma-dev` installed at all.
- **`c/Makefile`** — the decision is a compile define (`-DCOLI_HIP_NO_WMMA`) derived from `HIP_ARCH`, because it must be identical in both hipcc passes: an arch-macro check inside the header fires only in the device pass, and the host side would then dispatch kernels whose bodies were compiled out (empty launch, no error, stale output buffer). To make the arch filter reliable, `HIP_ARCH` is normalized first: comma lists accepted, and `native` is resolved to real gfx names via `rocm_agent_enumerator` at make time. `HIP_ARCH` may now be a list, producing a fat binary (`--offload-arch` per entry); a mixed capable/incapable list warns loudly that WMMA is off for all targets (shared host dispatch requires one consistent setting).
- **`c/backend_cuda.cu`** — gate the opt-in `COLI_CUDA_TC_INT4` branch on `COLI_GPU_HAS_WMMA`, exactly as its `TC_W4A16` sibling already is. `grouped_s4_wmma`'s body is `__CUDA_ARCH__>=750`-guarded, so on a no-WMMA build the env var would launch an empty kernel: no error, output buffer never written.

## Verified

- ROCm 7.2.4, Ryzen 9 5950X, Radeon Instinct MI50 32 GB (gfx906) + RX 6900 XT (gfx1030), Arch Linux.
- `make glm HIP=1` (native → resolves to `gfx906 gfx1030`, fat binary), `HIP_ARCH=gfx906`, and `HIP_ARCH="gfx906,gfx1030"` all build clean.
- Engine serves GLM-5.2 int4-gs64 on the MI50 with the VRAM expert tier active (`[CUDA] hot expert tier: 727/727 experts` on first bring-up; pinned tier grows with usage history as designed). Related field report for this hardware class: #510.
- `make check`: 229 tests pass.

## Noticed but not touched (flagging for a maintainer)

- The same empty-kernel hazard exists on **CUDA sm_70/Volta**: `grouped_s4_wmma` needs `>=750` but the host never checks the minor version. This PR's `COLI_GPU_HAS_WMMA` gate doesn't cover that case (compile-time 1 on CUDA).
- `backend_gpu_compat.h` references `HIP-KNOWN-BUGS.md` (BUG-002), which doesn't exist in the repo.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -183,7 +183,22 @@ HIP       ?= 0
 ROCM_HOME ?= /opt/rocm
 HIPCC     ?= $(ROCM_HOME)/bin/hipcc
 HIP_ARCH  ?= native
-HIPCCFLAGS ?= -O3 -std=c++17 -x hip --offload-arch=$(HIP_ARCH) -Wall -Wextra -fPIE
+# HIP_ARCH accepts a list (space- or comma-separated, e.g. "gfx906 gfx1030")
+# to fat-bundle several targets into one binary; COLI_GPUS picks at runtime.
+# `native` is resolved to the local GPUs' arch names HERE (not left to hipcc):
+# the NO_WMMA_ARCHS filter below must see real gfx names to know whether the
+# rocWMMA paths can compile at all. `override` beats command-line assignment.
+comma := ,
+override HIP_ARCH := $(subst $(comma), ,$(HIP_ARCH))
+ifeq ($(strip $(HIP_ARCH)),native)
+ifeq ($(HIP),1)
+override HIP_ARCH := $(shell $(ROCM_HOME)/bin/rocm_agent_enumerator 2>/dev/null | grep -v gfx000 | sort -u)
+ifeq ($(strip $(HIP_ARCH)),)
+$(error HIP_ARCH=native but rocm_agent_enumerator found no GPUs; set HIP_ARCH=gfxNNNN explicitly)
+endif
+endif
+endif
+HIPCCFLAGS ?= -O3 -std=c++17 -x hip $(foreach a,$(HIP_ARCH),--offload-arch=$(a)) -Wall -Wextra -fPIE
 GPUCC      = $(NVCC)
 GPUFLAGS   = $(NVCCFLAGS)
 GPUCC_NAME = nvcc (set CUDA_HOME or NVCC)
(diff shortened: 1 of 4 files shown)


MrMetric:
Follow-up finding from the same hardware (write-up again by Claude Fable 5, tested on MrMetric's box): after this PR made mixed CPU+GPU runs *possible* on these cards, they measured **slower than CPU-only** — and the cause is general, not AMD-specific.

## Symptom

Short decode requests (GLM-5.2 int4-gs64, `DRAFT=0`, warm page cache, same 24-token prompt, fresh serve per config):

| Config | tok/s | routed-CPU rate | tier straggler |
|---|---|---|---|
| CPU only | 0.35–0.39 | 21.5 GB/s | — |
| CPU+MI50 | 0.18 | 3.0 GB/s | 25.2x |
| CPU+MI50+6900XT | 0.24 | 3.3 GB/s | 14.7x |

Enabling `COLI_CUDA=1` collapses the **CPU** expert tier ~7x, which swamps the GPUs' contribution (they finish their share in ~2.5 s).

## Root cause

The OMP self-tune re-exec block at the top of `main()` is skipped whenever `COLI_CUDA`/`COLI_METAL` is set. The load-bearing variable turns out to be **`OMP_PROC_BIND=close`**: this host is a Ryzen 9 5950X (two CCDs, two 32 MB L3 complexes), and unbound OMP threads migrating across CCXs destroy the tiny ~200 KB-per-worker expert regions. Thread-state sampling during the slow runs shows ~25 cores busy in userspace while delivering 3 GB/s — cycles burned on cross-CCX cache misses, not serialization and not I/O.

Eliminated along the way, in case it saves someone the same detours: `OMP_WAIT_POLICY=active` + `GOMP_SPINCOUNT` alone recover nothing (3.0→3.5 GB/s); XEXP is not a factor (fmt=4 container never takes it); thread count is identical (32) in all configs.

## Verification

Exporting the full tune set manually with CUDA enabled (works without the re-exec because a spawned process's libgomp sees them at constructor time):

```
OMP_WAIT_POLICY=active GOMP_SPINCOUNT=200000 OMP_PROC_BIND=close OMP_DYNAMIC=FALSE KMP_BLOCKTIME=200
```

| Config | tok/s | routed-CPU rate | tier straggler |
|---|---|---|---|
| CPU+MI50, tuned | 0.42 | 21.4 GB/s | 1.8x |
| CPU+MI50+6900XT, tuned | **0.48** | 21.7 GB/s | 1.8x |

Both GPU configs now beat CPU-only; the workload becomes expert-I/O-bound (~51%), which is the correct limiter for this machine.

## Suggested fix

Let the tune run under CUDA/Metal too — `GOMP_SPINCOUNT` and `KMP_BLOCKTIME=200` already bound the idle-spin concern (#341) that presumably motivated the skip, and `overwrite=0` keeps user env authoritative. If you'd rather be conservative, seeding only `OMP_PROC_BIND=close` + `OMP_DYNAMIC=FALSE` captures the win we measured. Happy to push a commit to this branch either way.

Caveat: numbers are single short requests (coarse), but the A/B is stable across repeated fresh serves, and the effect needs a multi-CCX CPU — single-CCX hosts may not reproduce it.

PR #684, opened 2026-07-29
Kimi K3: integrate coli chat, API, and Web

ZacharyZcR:
## Summary

- add the shared `SERVE=1` protocol to `kimi_k3`
- route Kimi models through `coli chat`, `coli serve`, and `coli web`
- add validated multi-turn K3 message framing and reasoning/content streaming
- add top-p sampling, cancellation, state reset, and profiling output
- document the supported workflow and current limitations

Depends on #676. This is intentionally stacked on its exact commits; once #676 lands, this PR should reduce to the integration commit.

## Validation

- `make -C c kimi_k3`
- `make -C c check` — 224 Python tests passed, 13 skipped, plus all C tests
- `PYTHONPATH=c python3 -m unittest c.tests.test_openai_server` — 104 passed
- K3 wire fixture compared against Moonshot's official `encoding_k3.py`: 77 token IDs, exact match
- `git diff --check`

## Full-model test boundary

The local validation host has 93 GiB RAM and a 12 GiB RTX 4070, so it cannot load the full Kimi K3 snapshot. Tokenization, protocol, gateway, TUI/Web paths, compilation, and regression tests are covered here; a complete model generation run remains the release gate on a host that can load #676's snapshot.

## Current limitations

- text chat only; tools and image input return explicit unsupported errors
- one KV slot only; sibling state is not isolated for concurrent decode yet
- CPU engine path from #676; no new GPU claim

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@ c/glm
 c/glm.exe
 c/olmoe
 c/inkling
+c/kimi_k3
+c/kimi_k3.exe
 c/olmoe.exe
 c/iobench
 c/iobench.exe
--- a/c/Makefile
+++ b/c/Makefile
@@ -410,6 +410,9 @@ olmoe$(EXE): olmoe.c st.h json.h compat.h
 inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ)
 	$(CC) $(CFLAGS) inkling.c $(INK_CUDA_OBJ) -o inkling$(EXE) $(LDFLAGS)
 
+kimi_k3$(EXE): kimi_k3.c st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h quant.h
+	$(CC) $(CFLAGS) kimi_k3.c -o kimi_k3$(EXE) $(LDFLAGS)
+
 # Use a baseline that matches the compiler target. macOS already targets a
 # portable baseline when ARCH is empty; forcing the x86 value there breaks
 # Apple Silicon. Unknown targets use native rather than an invalid x86 flag.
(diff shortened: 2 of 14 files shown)


steve-m:
Thanks @ZacharyZcR — this was exactly the right direction, and #676 has now absorbed it: the branch is rebased onto current dev and carries your integration commit as-is (`ca3dcc6`, authorship preserved via cherry-pick `-x`).

Two things we added on top, from the box that holds the full snapshot:

- the wire fixture re-validated independently against Moonshot's official `encoding_k3.py` + `tiktoken.model` on the real container tokenizer: **77/77 token ids exact**;
- the full-model release gate you named is closed — the `SERVE=1` protocol and multi-turn `coli chat` / `coli web` sessions were run against the complete 2.8T snapshot on the 62 GiB test box (int4 dense, ~5 s/token decode; details in the #676 validation table).

So once #676 lands, this PR is fully covered by it.

JustVugg:
Closing as **absorbed** — not rejected.

@steve-m cherry-picked your integration commit into #676 as-is with authorship preserved (`5450c27e` — "feat: integrate Kimi chat with TUI and Web", authored by @ZacharyZcR), rebased the whole branch onto current `dev`, and verified the ten original commits unchanged by range-diff. #676 is now merged, so **your work is in `dev`** under your name.

@ZacharyZcR — thank you: routing Kimi through the gateway instead of templating in `coli` was the right call, and it turns out to fix the same gap for **Inkling** too (`coli chat` was applying the GLM template to non-GLM models). That's a second engine fixed for free.

@steve-m — thanks for handling the overlap cleanly rather than duplicating it.

Issue #689, opened 2026-07-29
[Bug]: deep speculative verify batches are not token-exact vs CPU on CUDA (near-tie flips + lower draft acceptance)


terrizoaguimor:
## Summary

On the CUDA path, a decode run whose drafts are accepted in **long streaks** can diverge
from the unbatched path by a near-tie token. The same run on CPU is byte-exact. It also
shows up as a lower acceptance rate for an identical draft set: **90% on GPU vs 100% on
CPU**, same model, same prompt, same corpus.

This is the CUDA/MoE analogue of #622 (Metal prefill GEMM not token-exact on near ties).
It matters because it is invisible unless drafts are accepted deeply: MTP at auto depth 1,
and even `DRAFT=8` with low acceptance, never surface it.

## Repro and isolation

GLM-5.2 int4, commit 2d62381, 96-token greedy (`COLI_TEMP=0`), single H200 fully resident
(`CUDA_EXPERT_GB=120 CUDA_RELEASE_HOST=1 COLI_CUDA_MTP=1 CUDA_DENSE=1 COLI_CUDA_ATTN=1
COLI_CUDA_PIPE=2 COLI_CUDA_TC_W4A16=1 PIN=auto PIN_GB=all RAM_GB=auto`), token ids captured
with `TOKENS=1`. The deep drafts came from an external draft source, but the source is not
the variable — four controls pin that down:

| Run | Config | vs baseline |
|---|---|---|
| A | baseline, MTP depth 1 (S=2 batches) | — |
| D/G | baseline, `DRAFT=8` (S=9 batches, 336 proposals, 16% acceptance) | **identical** |
| F | deep source at depth 1 (S=2 batches, 100% acceptance) | **identical** |
| B | deep source at depth 8 (S=9 batches, **90% acceptance**) | **diverges at token 85/96** (first 84 identical) |
| E | B repeated | identical to B (deterministic, not flaky) |
| H/I | the same A/B pair, **CPU only** | **byte-identical to each other** |

So: batch *shape* alone is fine (D/G match A). The draft *source* is fine (F matches A, and
H/I are byte-exact on CPU with the identical source at depth 8). What differs in B is that
emitted tokens come from **deep rows** of the verify batch — rows 2..8 rather than 0..1 —
because acceptance is high.

## Why it looks like near ties rather than a logic error

The first 84 tokens are identical and the divergence is a single token; a verification or
[...]

Issue #688, opened 2026-07-29
[Performance]: GLM-5.2 on a single H200 (141 GB) + 235 GB RAM — 4.75 tok/s median, matching the 6x RTX 5090 reference class


terrizoaguimor:
Per CONTRIBUTING's benchmark-report format.

**Commit:** 2d62381 (main). **Model:** GLM-5.2 int4 g64 + int8 MTP container (372 GB).

**Hardware:** DigitalOcean GPU droplet (nyc2): 1× NVIDIA H200 (143,153 MiB, driver
580.173.02, CUDA 13.1), Intel Xeon Platinum 8592+ (24C), 240 GB RAM, model on a 5 TB
NVMe volume (`/scratch`, ~1.9 GB/s buffered scattered reads measured during runs),
Ubuntu 22.04.5.

**Commands (exact):**
```
COLI_CUDA=1 COLI_GPU=0 CUDA_EXPERT_GB=120 CUDA_RELEASE_HOST=1 COLI_CUDA_MTP=1 \
CUDA_DENSE=1 COLI_CUDA_ATTN=1 COLI_CUDA_PIPE=2 COLI_CUDA_TC_W4A16=1 PIPE=1 \
PIN=auto PIN_GB=all RAM_GB=auto COLI_TEMP=0 PROF=1 COLI_MODEL=/scratch/glm52_i4 \
./coli run --ngen 96 "<fixed prompt>"
```

**Warm-up policy:** one prior CPU turn (33-token smoke) to seed `.coli_usage` for
`PIN=auto`; page cache warm from the container download. Placement per run: 5,640
experts → VRAM (119.99 GB) + 9,311 → RAM (197.7 GB warm) in 62 s; resident set
130.4 GB VRAM incl. 10.15 GB dense.

**Runs (96-token greedy decode):** 4.18 / 4.75 / 5.37 tok/s — **median 4.75**, best
5.37, first-16-token peak 7.26. Expert hit 99.1→99.7% (pin), residual disk 2.9–4.8 s
per run. Prefill 32 tokens in 5.1–6.0 s. MTP (`COLI_CUDA_MTP=1`): 1.72–1.88 tok/fw.
CPU-only baseline, same box: 0.85 tok/s (hit 82%, MTP 1.79 tok/fw).

**Notes**
- Single-GPU large-RAM needed two departures from defaults, filed separately:
  `CUDA_RELEASE_HOST=1` (defaults off on ndev==1; costs 3.8×) and explicit
[...]

Issue #687, opened 2026-07-29
[Bug]: CUDA_EXPERT_GB=auto fills all VRAM on single-GPU; lazy dense uploads then fail 60x and fall back to CPU silently


terrizoaguimor:
## What happened

Single H200 (143,153 MiB), GLM-5.2 int4, commit 2d62381, first GPU run with the
documented large-RAM recipe (`CUDA_EXPERT_GB=auto CUDA_DENSE=1 COLI_CUDA_ATTN=1
COLI_CUDA_PIPE=2 COLI_CUDA_TC_W4A16=1 PIN=auto PIN_GB=all RAM_GB=auto`):

1. The auto expert tier claimed the GPU to within 4 MiB:
   `nvidia-smi`: **143,149/143,153 MiB used** by the engine process.
2. The resident dense tensors upload lazily on first use (per docs) — every upload then
   failed: 60× `[CUDA] tensor [..,..] on device 0 disabled after an error; falling back
   to CPU` + repeated `[CUDA] tensor allocation: out of memory`.
3. Because failed tensors are disabled permanently, the run continued as a de-facto CPU
   run while holding 143 GB of VRAM: decode 0.82 tok/s — CPU speed (CPU-only baseline
   on this box: 0.85), with the GPU 100% allocated and ~0% useful.

The docs say auto is "clamped against free VRAM after reserving the projected dense
resident set and 2 GB of runtime headroom", but with `CUDA_DENSE=1 + COLI_CUDA_ATTN=1 +
PIPE=2` on this topology the reservation under-counts (dense+attn measured 10.15 GB once
uploads succeed, plus workspace). Same auto-sizing family as #653 (GB10), different knob.

Workaround that produced clean runs: explicit `CUDA_EXPERT_GB=100` (later 120) — dense
uploaded 10.15 GB, resident set 130.4 GB, zero allocation errors, 4.18–5.37 tok/s.

## Suggested fix

- Include the ATTN/PIPE/TC workspace in auto's dense projection, or
- treat a dense-upload OOM as a signal to evict N experts and retry once before
  permanently disabling the tensor (the tier adapts every 16 tokens anyway), and
- log a single prominent line when >K resident tensors get disabled ("GPU effectively
  idle — reduce CUDA_EXPERT_GB"), since the current failure mode reads as healthy.

Full logs (60 failure lines + nvidia-smi snapshots) available.
Hardware: as in the companion CUDA_RELEASE_HOST report.

[...]

Issue #686, opened 2026-07-29
[Bug]: CUDA_RELEASE_HOST defaults off on single-GPU, wasting ~100 GB of RAM on large-RAM hosts (worth 3.8x decode)


terrizoaguimor:
## What happened

On a single H200 (141 GB) + 235 GB RAM host running GLM-5.2 int4 with the documented
"large-RAM host" recipe (`CUDA_EXPERT_GB=auto/120 CUDA_DENSE=1 COLI_CUDA_ATTN=1
COLI_CUDA_PIPE=2 PIN=auto PIN_GB=all RAM_GB=auto`, commit 2d62381):

- `CUDA_RELEASE_HOST` defaults to `ndev>1` (colibri.c:6867), so on a single GPU the
  host copies of the entire VRAM expert tier (~100-120 GB) stay resident in RAM.
- `RAM_GB=auto` then budgets around them: the RAM expert tier starved at
  **9,297/19,456 experts resident** and decode ran disk-bound.

Measured on the same box, same model, same prompt (96-token greedy, 3 runs, median):

| Config | Resident experts | Expert hit | Decode |
|---|---|---|---|
| default (`RELEASE_HOST` off) | 9,297 (4,690 VRAM + 4,607 RAM) | 86.2% | **1.10 tok/s** |
| `CUDA_RELEASE_HOST=1` | 14,951 (5,640 VRAM + 9,311 RAM) | 99.1→99.7% | **4.18–5.37 tok/s** |

`[PROF] expert I/O` for the default config: 166.9 GB fetched, 87.7 s read service in an
87.4 s decode — the whole token budget went to disk misses that the wasted RAM could
have absorbed.

## Suggested fix

Either default `CUDA_RELEASE_HOST=1` whenever `PIN_GB=all`/large `PIN_GB` is requested
(the host copy is provably redundant there — reload-on-CUDA-failure still works from
disk), or emit a startup warning when single-GPU + `RELEASE_HOST=0` leaves the RAM
budget projecting over (the `[RAM_GB auto] WARNING ... PIN_GB is inflating` line
already fires in exactly this situation and could name the knob).

Happy to PR either variant. Full logs available.

Hardware: DigitalOcean H200 droplet (nyc2), Xeon Platinum 8592+ 24c, 240 GB RAM,
NVMe scratch, driver 580.173.02, CUDA 13.1, Ubuntu 22.04.5.

JustVugg:
Confirmed in code — `colibri.c:6947`:

[...]

PR #690, opened 2026-07-29
feat(spec): frozen-corpus draft source (COLI_DRAFT_CORPUS)

terrizoaguimor:
## Summary

`ngram_draft()` already proposes a continuation by matching the tail of the current
sequence against earlier positions in that same sequence. This adds the same lookup over
a **persistent file of previously generated token ids**: at each decode step the engine
proposes what followed the longest suffix of the live context found in the corpus (suffix
lengths 8 down to 3, most recent occurrence first). Proposals enter the existing
batch-union verify forward alongside grammar/MTP/n-gram, so the target model's argmax
still decides what is accepted — the source only ever proposes.

Unlike MTP (one token per forward) a corpus hit proposes a whole span.

## Measurements

GLM-5.2 int4, 96-token greedy (`COLI_TEMP=0`), replaying a prompt whose generation is in
the corpus:

| Host | Baseline (MTP) | With corpus drafts |
|---|---|---|
| H200, fully resident | 1.78 tok/forward, 3.69 tok/s | 6.00 tok/forward, **4.52 tok/s** (+22%), 90% acceptance |
| CPU (24C Xeon 8592+, 120 GB pin) | 1.76 tok/forward, 0.82 tok/s | 7.92 tok/forward, **1.00 tok/s** (+22%), 100% acceptance |

**Leading with the ceiling rather than the multiplier:** 3–4× fewer forwards buys ~22% of
wall clock, not 3–4×. In an MoE every row of a verify batch activates its own experts, so
speculation amortizes the dense path, attention and per-forward overhead but never the
expert work itself. The same ~20% ceiling reproduced across four regimes here (resident
GPU, CPU, and two disk-streaming configurations of another MoE). `docs/corpus-draft.md`
documents this explicitly so nobody sizes an expectation off the forward count.

Rejected drafts cost real time, so the source pauses below `COLI_CORPUS_MINACC`
acceptance (default 50% — the measured break-even: 90% acceptance gave +22%, 19% gave
−25%). A prompt unrelated to the corpus costs nothing: the suffix match finds nothing and
the source stays inert (measured — identical forward count to a run with no corpus at
all). The risk lives in the *neighbourhood* of the corpus, where near-miss prefixes
produce spurious matches; that is what the guard is for.

## Compatibility

- **Off by default.** Without `COLI_DRAFT_CORPUS` the corpus pointer is `NULL` and
  behaviour is bit-identical to today.
- No new dependencies; the default CPU path stays dependency-free.
- Corpora are built from any run with `TOKENS=1`, which already dumps the generated ids —
  no new tooling needed.
- New env vars documented in `docs/ENVIRONMENT.md`; full write-up in
  `docs/corpus-draft.md` (structured after `docs/grammar-draft.md`).

## Validation

- [x] `make check` on this branch — **229 tests OK**, clean build, **0 warnings**
- [x] Oracle **32/32 TF + 20/20 greedy**
- [x] Oracle **20/20 greedy with the source armed** — not required by CONTRIBUTING, added
      because a new draft source should be token-exact when ON, not only when OFF
- [x] `tests/test_corpus_draft.c` covers the lookup contract: longest suffix wins,
      recency breaks ties, the `-1` span separator terminates a proposal, caps hold,
      degenerate inputs propose nothing
- [x] No performance claims beyond the table above, all from the hardware described

## Known caveat, filed separately

On CUDA, a run whose drafts are accepted in long streaks can diverge from the unbatched
path by a single near-tie token (1 in 96 measured, at position 85, first 84 identical),
and the identical corpus accepts at 90% on GPU vs 100% on CPU. Four controls show this is
a property of deep verify batches on that path rather than of this source (`DRAFT=8` with
MTP batches the same shape without diverging; this source at depth 1 reproduces the
baseline exactly; CPU is byte-exact at depth 8). Filed as #689 with the full isolation
table. The CPU path is byte-exact.

Related reports from the same hardware: #686, #687, #688.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -214,7 +214,7 @@ else
 PYTHON    ?= python3
 endif
 CUDA_OBJ  =
-TEST_BINS = tests/test_json$(EXE) tests/test_st$(EXE) tests/test_st_mirror$(EXE) tests/test_st_pread$(EXE) tests/test_tier$(EXE) tests/test_grammar$(EXE) tests/test_schema_gbnf$(EXE) tests/test_decode_batch$(EXE) tests/test_idot$(EXE) tests/test_i4_grouped$(EXE) tests/test_stops$(EXE) tests/test_topp$(EXE) tests/test_sample_nan$(EXE) tests/test_temp_env$(EXE) tests/test_tok_o200k$(EXE) tests/test_kv_alloc$(EXE) tests/test_int3$(EXE) tests/test_int3_load$(EXE) tests/test_i4_acc512$(EXE) tests/test_compat_direct$(EXE) tests/test_dsa_select$(EXE) tests/test_logit_nan$(EXE) tests/test_router_nan$(EXE) tests/test_pipe_block$(EXE) tests/test_e8_kernel$(EXE) tests/test_rans$(EXE)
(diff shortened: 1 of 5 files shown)


JustVugg:
Merged to `dev` — thanks, this is a nicely-scoped piece of work, and I appreciate that you led with the ceiling ("3–4× fewer forwards buys ~22% of wall clock, not 3–4×") instead of the multiplier. That kind of honesty in a performance PR is rare and it's why this was easy to review.

I pushed one follow-up docs commit (68082da) adding a **"When this helps, and when it does not"** section: the +22% is a *replay* best case, and on genuinely novel text the corpus has nothing to propose so the gain approaches zero. Your doc already said "replaying a prompt whose generation is in the corpus" — I just made it impossible to skim past, so nobody reads the table as a blanket speed-up. Also cross-linked #689 there, since deep drafts from this source are the easiest way to reach that open verify-batch divergence.

No change to your code. Shout if you'd word the scope section differently.

PR #696, opened 2026-07-29
fix(cuda): evict cold experts and retry instead of losing a resident tensor (#687)

terrizoaguimor:
Item 2 of #687 — the evict-N-and-retry-once you described. #694 (the loud warning) is item 1 and is independent; this one can land after it.

## The change

On a resident-tensor upload failure, free the coldest experts on that device and retry once, instead of disabling the tensor for the rest of the run.

## Measured on the failure it fixes

1× H100 80GB, GLM-5.2 int4, `CUDA_EXPERT_GB=auto` with `CUDA_DENSE=1 COLI_CUDA_ATTN=1 COLI_CUDA_PIPE=2 COLI_CUDA_TC_W4A16=1`. `auto` does **not** over-commit on this card (it leaves ~3.6 GB and runs clean), so I reproduced the #687 condition with `CUDA_RESERVE_GB=0.5`, which moves the tier 71.06 → 72.56 GB and is enough to break it:

| | Tensors disabled | Allocation failures | Output |
|---|---|---|---|
| `dev` | **11** | 107 | correct, at CPU speed |
| this patch | **0** | 8 | correct, **byte-identical** |

## Three details that turned out to matter

- **Eviction is sized to the failing tensor** (`qt_bytes(w)`), not a fixed expert count. My first version evicted 8 experts per attempt and rescued 10 of 11 — a fixed count rescues the small tensors and leaves the large ones behind.
- **`qt_cuda_reset(w)` before the retry.** A failed attempt can leave a partially uploaded tensor with a non-NULL handle; the retry then reuses it instead of re-uploading, and computes on garbage. This is the interesting one: the rescued run was *deterministic* and its output looked perfectly plausible, but its first token differed from a run that never hit the OOM. I only caught it by diffing token ids against a clean run — fault injection had reported success, because injecting a failure *before* the allocation does not reproduce the state an OOM leaves *during* it.
- **`expert_host_ensure()` before freeing the device copy.** With `CUDA_RELEASE_HOST=1` the host copy is already gone, so the expert has to exist somewhere before its VRAM is taken away.

Accounting (`gpu_expert_count`/`gpu_expert_bytes`) follows the REPIN downgrade path so the conventions match, and a hard cap of 1024 experts stops a pathological run from trading the whole tier away.

## Validation

- [x] `make check` — OK
- [x] Clean build, **0 warnings**, `CUDA=1` and `portable`
- [x] Oracle **32/32 TF + 20/20 greedy**
- [x] A/B above, **byte-identical token ids** vs a run that never hit the OOM
- [x] Deterministic across repeated runs

## Note for item 3

While reproducing this I swept `CUDA_RESERVE_GB` on the H100: 0.5 → 11 disabled, 2.0 (default) → 0, 4.0 → 0. On the H200 in #687, the same default 2.0 **did** fail. The difference tracks residency — 6,345 experts in VRAM there vs 3,411 here — which fits the grouped-expert staging buffers scaling with the tier while the reserve stays fixed. Two topologies is too thin for me to propose a coefficient, but if that shape looks right to you I'm happy to gather more points.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -377,6 +377,45 @@ static void cuda_disabled_note(void){
         "[CUDA] per device for dense + attention + workspace). See issue #687.\n",
         CUDA_DISABLED_LOUD_AT);
 }
+/* #687: a resident-tensor upload that OOMs disables that tensor for the WHOLE run and
+ * silently drops it to CPU, while the expert tier keeps the VRAM it took. The tier is
+ * the side that can give memory back: it re-adapts every 16 tokens under REPIN, and an
+ * evicted expert simply reloads from disk (expert_host_ensure). So trade the coldest few
+ * experts for the dense path and retry once, instead of losing the GPU for the run. */
+static Model *g_model;                 /* set after model_init; mirrors pilot_m */
+static void expert_host_ensure(Model *m, int layer, ESlot *s);
+static long g_cuda_evicted;            /* total experts traded away, for the cap + report */
+static int cuda_evict_coldest(int device, int64_t need_bytes){
+    Model *m=g_model;
+    enum { EVICT_CAP = 1024 };         /* never thrash the tier away wholesale */
+    if(!m||need_bytes<1||g_cuda_evicted>=EVICT_CAP) return 0;
+    int freed=0; int64_t got=0;
+    /* free at least what this tensor needs, with headroom for the allocator: a fixed
+     * expert count rescues the small tensors and leaves the big ones behind. */
+    while(got<need_bytes*2 && g_cuda_evicted<EVICT_CAP){
+        int bl=-1,bz=-1,have=0; uint32_t bh=0;
+        for(int l=0;l<=m->c.n_layers;l++){
+            ESlot *P=m->pin[l]; if(!P) continue;
+            for(int z=0;z<m->npin[l];z++){
+                ESlot *s=&P[z];
+                if(!s->g.cuda||s->g.cuda_device!=device) continue;
+                uint32_t h=(s->eid>=0&&s->eid<m->c.n_experts)?m->eheat[l][s->eid]:0;
+                if(!have||h<bh){ have=1; bh=h; bl=l; bz=z; }
+            }
+        }
+        if(!have) break;                                   /* nothing left on this device */
+        ESlot *s=&m->pin[bl][bz];
+        int64_t was=(int64_t)coli_cuda_tensor_bytes(s->g.cuda)
+                   +(int64_t)coli_cuda_tensor_bytes(s->u.cuda)
+                   +(int64_t)coli_cuda_tensor_bytes(s->d.cuda);
+        expert_host_ensure(m,bl,s);        /* CUDA_RELEASE_HOST may have freed the host copy */
+        qt_cuda_reset(&s->g); qt_cuda_reset(&s->u); qt_cuda_reset(&s->d);
+        s->g.cuda_eligible=s->u.cuda_eligible=s->d.cuda_eligible=0;
+        m->gpu_expert_count--; m->gpu_expert_bytes-=was;
+        freed++; g_cuda_evicted++; got+=was;
+    }
+    return freed;
+}
 static int g_cuda_e8_ready;   /* codebook published to the devices (see cuda_boot) */
 #endif
 #ifdef COLI_VULKAN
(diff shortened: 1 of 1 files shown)


JustVugg:
Merged #694 first (the loud warning — smaller and independently useful, the order suggested in #687), so this one now needs a rebase: both PRs touch the same region of `colibri.c` and they conflict with each other, not with `dev`.

Could you rebase onto latest `dev`? The warning is already in, so this PR reduces to the evict-and-retry logic itself — which is the substantive half. CI was green on both.

Thanks for splitting it in two; landing the diagnostic first means anyone hitting #687 today already gets a message that names the knob, even before the retry logic ships.

JustVugg:

terrizoaguimor:
Rebased onto `dev` (`292ed4c`) — but **please hold this one, do not merge it yet.** I have a
result that disqualifies it in its current form, and I would rather say so before it looks
ready than after.

## What the rebase found first

`dev` already fixes #687 from the other side: `cuda_disabled_note()` counts the disabled
tensors and says it once, loudly, naming `CUDA_EXPERT_GB`. That is a *diagnostic*; this PR
*recovers*. I resolved the conflict keeping both, so the retry runs first and your notice only
fires when recovery genuinely failed — which also makes it more precise, since it then means
"we tried to hand memory back and still could not."

## What the measurement found second

Forcing the condition on GLM-5.2 with `CUDA_EXPERT_GB=79 CUDA_RESERVE_GB=0` (80 GB H100):

| | `dev` | this PR |
|---|---|---|
| resident tensors disabled | **8** | **0** |
| rescued by eviction | 0 | 1 |
| decode | 1.46 tok/s | **1.15 tok/s** |
| expert hit rate | 95% | 93% |
| **generated text vs an unpressured run** | **identical** | **differs** |

The mechanism works — it prevented all eight disables with one eviction round. But it is
**slower** here, because the experts it trades away cost more in disk service than the dense
path gains from staying on the GPU, and more importantly **it changes the output** in a
situation where the behaviour it replaces does not.

That last row is the one that matters. `dev` falling back to CPU for eight dense tensors
produced token-identical output to a run with no pressure at all. My rescue did not. Whatever
the cause, "makes the model say something different" is not an acceptable price for "keeps the
GPU busy", and it is the same class of failure I hit in an earlier revision of this patch.

## Being investigated

Whether the divergence is deterministic. If it reproduces exactly, it is a numeric shift —
evicting a VRAM-resident expert moves it to CPU arithmetic, which is inherent to the approach
and would mean the premise needs rethinking rather than the code. If it varies between runs, it
is state corruption and a real bug in the eviction path. Two identical runs are in flight; I
will post the answer either way.

Either way this should not merge as-is, so I am moving it to draft. Sorry for the noise on a PR
you asked me to rebase — the rebase is sound, the idea underneath it may not be.

terrizoaguimor:
Answering my own question from above, and correcting it: **the divergence is my bug, not an
inherent property of moving experts between tiers.** I had offered that as the benign
possibility. The control says otherwise.

## The determinism test

Two identical pressured runs of this branch: same output, same single rescue. So it is not
state corruption in the racing sense — it is deterministic.

## The control that mattered

On **pristine `dev`, with none of my code**, two runs differing only in how many experts live
in VRAM:

| | experts in VRAM | disabled | generated text |
|---|---|---|---|
| `CUDA_EXPERT_GB=40` | 1883 / 6555 | 0 | — |
| `CUDA_EXPERT_GB=70` | 3293 / 7969 | 0 | **identical** |

So residency does **not** change the output. Your GPU and CPU expert paths agree, tier
placement is genuinely lossless, and my "it is just numerics" hypothesis is dead. That control
is the one I should have run before writing the previous comment.

## Three-way, which localises it

| run | first generated token |
|---|---|
| unpressured (all on GPU) | `NVMe (Non-Volatile Memory Express) is a high-speed protocol…` |
| `dev` under pressure, 8 tensors disabled to CPU | **identical to unpressured** |
| this branch, tensor rescued | `NVM Express (NVMe) is a high-speed storage protocol…` |

Your CPU fallback agrees with the healthy GPU run to the token. Only the rescued path differs,
and it differs from the **first generated token**, not after drift. A tensor that computes a
different answer than both the GPU original and the CPU fallback is not producing a valid
alternative — it is producing a wrong one.

So `qt_cuda_reset(w)` before the retry, which I added for exactly this failure in an earlier
revision, is necessary but **not sufficient**. The two candidates I can see are that the retry's
re-upload still inherits something stale beyond `w->cuda`, or that `cuda_evict_coldest` frees
expert tensors the in-flight forward pass is still using — `expert_host_ensure` restores the
host copy, but the GPU copy goes away underneath a computation that may still reference it.
I have not distinguished those yet.

## Where that leaves it

Staying in draft. I am not going to defend a patch that changes what the model says, and I
would rather have found this here than have you find it after a merge.

Worth saying plainly: **`dev` does not need this to be correct.** `cuda_disabled_note()` already
turns #687 from a silent degradation into a legible one, which was the actual complaint. This
patch was trying to convert a legible failure into no failure, and in its present form it
converts it into a wrong answer instead — strictly worse. If you would rather close it than
carry a draft, that is a reasonable call and I will not be precious about it. If you are happy
to leave it open I will keep digging, at a lower priority than phase B.

terrizoaguimor:
Closing this.

Not only because of the correctness bug — though that alone would be enough, and it is still
unexplained. The measurements in #687 today changed what I think this PR *is*.

The shortfall that makes dense uploads fail is **predictable before a single expert is placed**:
it is 0.741 MB per resident expert against a flat 2 GB reserve, and the sizing at `colibri.c:7262`
already knows how many experts it is about to commit to. Reactive eviction is the right tool when
you cannot see the wall coming. Here you can, and fixing the sizing removes the need for the retry
instead of leaving it armed for when the sizing gets it wrong.

So even with the correctness bug solved, I would not argue for merging this ahead of a corrected
projection. It would be a second mechanism guarding a hole that should not exist.

For the record, so it is not lost if someone wants to revisit it:

- The eviction works mechanically — on the forced-OOM case it rescued all 8 tensors that `dev`
  disables.
- It **changes model output**, deterministically, from the first generated token. `dev`'s CPU
  fallback is token-identical to an unpressured GPU run; the rescued path is not. A rescued tensor
  computing a different answer than both the GPU original *and* the CPU fallback is producing a
  wrong one.
- `qt_cuda_reset(w)` before the retry, which I added for exactly this failure in an earlier
  revision, is necessary but not sufficient. I did not find out why.

That last point is a real bug in the retry path and it would need solving before this idea comes
back in any form. If the sizing fix lands and you still want eviction as a backstop for the cases
sizing cannot predict — a second process taking VRAM mid-run, say — this is the right place to
start from, but starting from a version that changes what the model says would be a mistake.

Thanks for the review time on it, @JustVugg. The measurement that killed it came out of trying to
defend it.

PR #701, opened 2026-07-29
fix(inkling): expert cache served the wrong experts; int4 dense container; --chat

JustVugg:
Three changes that together make Inkling runnable — and correct — on a host that cannot hold its bf16 dense set. Validated end-to-end against the transformers oracle **and** against the real 975B snapshot.

## 1. Correctness: the expert cache silently used the wrong experts

`moe()` acquired a slot for every `(token, expert)` pair up front and kept the pointers until the compute pass. But `slot_acquire` evicts the LRU when the cache is full — **including slots already handed out in the same call**. The pointer stays valid; the slot now holds a *different* expert. No crash, no warning, just incoherent output.

So correctness silently required a cache large enough for **every distinct expert in the batch**: a prefill of 18 tokens at `topk=6` needs up to **108 slots per layer**. Below that, the model computed with whatever landed in the recycled slots.

**Fix:** process the routed experts in **rounds of `cap` pairs** — acquire, fill, compute, accumulate. Nothing can be evicted while in use. The MoE output is a weighted sum, so accumulating in rounds is identical to a single pass, and the cache can go down to *one* slot per layer (more disk reads; memory proportional to `cap` instead of to the batch).

**Proof** — `tools/make_tiny_inkling.py`'s transformers oracle:

| Cache | Before | After |
|---|---|---|
| `cap=4` (> topk=2) | 36/36, 24/24 | 36/36, 24/24 |
| `cap=1` (< topk=2) | would reuse one slot for both experts | **36/36, 24/24** |

## 2. Memory: optional int4-gs64 dense container

The pre-converted container ships int4 experts but **bf16 dense weights** — 49.4 GB measured, resident, needed for every token — and `load_w` expands bf16 to f32 while loading (~99 GB peak). Hosts below ~64 GB die before generating anything.

`c/tools/convert_inkling_dense_int4.py` quantizes **only the dense set** into a separate container (**15.3 GB**, one pass, originals read-only). The engine auto-detects `<snap>/dense-int4g64/`; `INK_DENSE_Q4=0` ignores it. **Without the container nothing changes** — the bf16 path is untouched.

Adds `matmul_i4g` (group-scaled int4) and `matmul_i8r` (per-row int8), plus a bounds guard in `matmul_w`: the container is a file, not an invariant, so a geometry mismatch exits instead of reading past the buffer.

Quantization error vs the real bf16 weights:

| Tensors | Format | Rel. L2 |
|---|---|---|
| attention, shared experts, dense MLP | int4-gs64 | ~11% |
| `embed_tokens`, `lm_head` | int8 per-row | ~0.9% |
| norms, biases, router, conv1d | untouched | 0 |

11% is what 4 bits costs on this distribution (noise ≈ 0.135σ for 16 levels over ±amax in groups of 64), not a defect of the conversion — and it is enough:

```
== Inkling C engine, 66 layers, experts @ container, cache 2/layer ==
<|message_model|><|content_text|>My name is Inkling.<|end_message|>
[eos after 8 tokens]  RSS 15.1 GB
```

A 975B model answering correctly, with a clean EOS, on a 25 GB box.

## 3. Usability: `--chat`

Renders Inkling's chat template in the engine, so the direct CLI stops feeding an instruct model raw out-of-distribution text. `coli chat/serve/web` already render it through the gateway.

## Testing

- Oracle: **36/36 teacher-forced, 24/24 tokens** — at `cap=4` and at `cap=1`
- All four engines build clean: `glm`, `inkling`, `olmoe`, `kimi_k3`
- Real 975B snapshot: coherent answer, correct EOS, 15.1 GB RSS
- Converter validated numerically against the source weights (per-class rel-L2 reported by the tool itself)

## Honest limits

**Speed is unchanged and still disk-bound.** ~8 GB of expert cache against a 464 GB expert bank is ~1.7% residency: tens of seconds per token on one NVMe (measured: 279 GB read for 8 tokens, 1.2% cache hit). This makes the model *runnable and testable* on a small host, not fast — `docs/inkling.md` states that rather than quoting a best case. `EXPERT_BUDGET`-style trimming was deliberately **not** ported from `colibri.c`: it is quarantined there (measured 0.13 vs 0.30 tok/s, *more* I/O).

Also documents the models table in the README: GLM-5.2, Inkling, Kimi K3, OLMoE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,27 @@ the full 756 GB on disk at once:
 ./coli convert --model /nvme/glm52_i4     # download+convert shard by shard (python, one-time)
 ```
 
+#### Other supported models
+
+GLM-5.2 is the reference model, but the same streaming approach runs three more
+families. Each is a **sibling engine** — one C file, its own architecture, the same
+`coli chat` / `coli serve` / `coli web` front end (the launcher picks the binary from
+the model's `config.json`):
+
+| Family | Total / active | Weights | Build | Docs |
+|---|---|---|---|---|
+| **GLM-5.2** | 744B / 40B | [`mastouri/…-int4-g64-with-int8-mtp`](https://huggingface.co/mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp) (372 GB) | `make -C c glm` | this page |
+| **Inkling** (Thinking Machines) | 975B / 41B | [`nbeerbower/Inkling-colibri-int4`](https://huggingface.co/nbeerbower/Inkling-colibri-int4) (469 GB) | `make -C c inkling` | [inkling.md](docs/inkling.md) |
+| **Kimi K3** (Moonshot) | 2.8T / 104B | [`moonshotai/Kimi-K3`](https://huggingface.co/moonshotai/Kimi-K3) — original checkpoint, routed experts stay **native MXFP4** | `make -C c kimi_k3` | [kimi_k3.md](docs/kimi_k3.md) |
+| **OLMoE** (AI2) | 7B / 1B | converted with `c/tools/convert_olmoe_merged.py` | `make -C c olmoe` | — |
+
+Kimi K3 needs no conversion: its QAT-trained MXFP4 experts are streamed straight from
+the original Hugging Face shards, and the bf16 dense set is quantized at load time.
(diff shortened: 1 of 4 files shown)


kbsoso9999-maker:
- [ ]

PR #703, opened 2026-07-29
release: merge dev → main (v1.3.0 — Kimi K3, Inkling on small RAM)

### LINKED ISSUE #685: [Bug]: 1.2.0 gibberish responses with Cuda A100
### Commit or version

2d623816eecbf740b1fc06eb422974bda0af8931

### Environment

Ubuntu 24.04 AMD EPYC 7413 512G 4xA100

cuda 12.9.2

COLI_CUDA=1
COLI_GPUS="0,1,2,3"
CUDA_EXPERT_GB=auto
CUDA_DENSE=1
COLI_CUDA_ATTN=1
PIN_GB=all
RAM_GB=auto
COLI_CUDA_PIPE=2 
COLI_CUDA_TC_W4A16=1 
CTX=1024 

./coli web --host 0.0.0.0 --port 8000 --model-id glm-5.2-colibri

[CUDA] hot expert tier: 6967/17363 experts, VRAM 147.93 GB (budget 148.0 GB, auto, reserve 2.0 GB)
[CUDA]   device 0: 1824 experts, 38.73 GB
[CUDA]   device 1: 1408 experts, 29.90 GB
[CUDA]   device 2: 1867 experts, 39.64 GB
[CUDA]   device 3: 1868 experts, 39.66 GB
[PIN] placement: 6967 VRAM + 10396 RAM expert (220.7 GB warm) in 252s da /root/.cache/huggingface/hub/models--mastouri--GLM-5.2-colibri-int4-g64-with-int8-mtp/snapshots/5276684ba30ac0026c07220d3f389171a84eb074/.coli_usage
[RAM_GB=458.1 auto] cap raised 8->134: budget allows it (projected peak 456.6 GB; set CAP_RAISE=0 to disable)

### Reproduction steps

Hi, I can't get colibri answering correctly with cuda enabled and 4 H100 40G cards

I only get random tokens as response

Am I missing something in the configuration ? (tested dev branch, with the same result)

Best regards

### Expected behavior

English correct answer

### Actual behavior and logs

Image

```shell

What is your name?
My is not you. I Name{{1name.name 

 

 .

 

 



　   

 are點 
```

JustVugg:
Promotes **51 commits** from `dev`. The headline: colibrì goes from one model family to **three MoE families, 744B to 2.8T parameters** — and a 975B answers on a 25 GB box.

## New model families

**Kimi K3** (Moonshot, 2.8T total / 104B active, 93 layers) — #676, by **@steve-m**, with the chat/API/Web integration by **@ZacharyZcR**. A sibling engine (`c/kimi_k3.c`) that streams the QAT-trained **MXFP4 routed experts straight from the original Hugging Face shards** — never re-encoded — and quantizes the bf16 dense set at load time. Includes the KDA + gated-NoPE-MLA + AttnRes + LatentMoE architecture, an MXFP4 matmul kernel (scalar + AVX2 + int8-activation), chunked prefill (bit-identical, 2.6×), parallel `O_DIRECT` expert reads, and `--chat` for K3's XTML format.
*Status: engine and tokenizer validated — the multi-turn wire matches Moonshot's official `encoding_k3.py` **77/77 token ids** — and the C/Python regression suite passes. Full-model generation is still gated on a host with ~1.6 TB of storage for the checkpoint; nobody has run it end to end yet. Said plainly so the people with that hardware know it's theirs to try.*

**Inkling** (Thinking Machines, 975B / 41B active) now runs on a small host — #701. Two problems were in the way:
- The pre-converted container ships int4 experts but **bf16 dense weights**: 49.4 GB resident, expanded to f32 while loading (~99 GB peak). A new one-pass tool quantizes just the dense set to **int4-gs64 (15.3 GB)**; the engine auto-detects it, and without it the bf16 path is untouched.
- **The expert cache silently computed with the wrong experts.** `moe()` acquired a slot for every `(token, expert)` pair up front and kept the pointers until compute, but a full cache evicts slots *already handed out in the same call*. Correctness therefore required a cache big enough for every distinct expert in the batch — up to 108 slots per layer for an 18-token prefill at `topk=6`. Below that: incoherent output, no error. Now the experts are processed in **rounds of `cap`** and accumulated, so nothing is evicted while in use, and the cache can go down to one slot per layer.

Result: **`My name is Inkling.` — a 975B model, correct answer, clean EOS, 15.1 GB RSS on a 25 GB machine.** Verified against the transformers oracle at 36/36 teacher-forced and 24/24 tokens, at both `cap=4` and `cap=1` (where the old code corrupted).

## Correctness fixes

- **#692 (@ZacharyZcR)** — `attention_absorb_ragged_kernel` applied per-row scales to **group-scaled (`fmt=4`)** tensors, so the recommended g64 GLM container produced **gibberish** through `coli serve`/`coli web` while `coli run` was fine. Located to a single line, fixed with the shared `absorb_scale()` helper, plus a two-group regression fixture and the previously-orphaned ragged test wired into CI. **Field-confirmed fixed by @terrizoaguimor** on his exact reproduction.
- **#679** — serve honours `STOP`/`CANCEL` **mid-turn** in single-slot speculative decode; before, a stop matched by the gateway sat unread until `max_tokens` (minutes of apparent hang on CPU hosts).
- **#681 (@MrMetric)** — HIP builds work on AMD GPUs **without matrix cores** (MI50/Vega, RDNA1/2), and a latent bug where the tensor-core INT4 branch could launch an empty kernel and keep stale output.
- **#659** — OOM on integrated/unified-memory GPUs (GB10 / DGX Spark): the RAM budget was computed before the VRAM expert tier was placed, and on unified memory both draw from the same pool.
- **#665** — `doctor` and `resource_plan` recognize **AMD/ROCm**; a working HIP build is no longer reported CPU-only.

## Performance

- **#691 (@terrizoaguimor)** — `CUDA_RELEASE_HOST` now defaults on for single-GPU hosts with a large `PIN_GB`, where the duplicated host copy was starving the RAM expert tier. Measured on a single H200: **1.10 → 4.75 tok/s (3.8×)**, 9,297 → 14,951 resident experts.
- **#694 (@terrizoaguimor)** — one loud line when resident tensors fall back to CPU, so `CUDA_EXPERT_GB=auto` filling the card no longer degrades 60× in silence.
- **#698 (@ariannamethod)** — `olmoe` stopped calling `fadvise(DONTNEED)` after every expert read, which was dropping pages the LRU would immediately want again (`EXPERT_DROP=1` restores the old behaviour for RAM-tight boxes).
- **#690 (@terrizoaguimor)** — `COLI_DRAFT_CORPUS`: a frozen-corpus draft source proposing whole spans. +22% wall clock when replaying a prompt already in the corpus; the docs state the scope rather than the headline.

## Thanks

This release is mostly other people's work, and the hard parts were found by people with hardware we don't have:

- **@steve-m** — the Kimi K3 engine, and for absorbing #684 cleanly instead of duplicating it
- **@ZacharyZcR** — the Kimi chat/API/Web integration, and for locating #685 to one line while three of us were chasing the wrong theory
- **@terrizoaguimor** — four measured reports (#686-#689) with A/Bs and medians, two PRs, and the field confirmation that closed #685
- **@MrMetric**, **@ariannamethod**, **@monotophic**, **@tasoo-oos**, **@RDouglasSharp**, **@khalilswdp**
- **@remip2**, **@kvbkvbkvb**, **@mrchevyceleb**, **@snailwei**, **@AlphaMo99**, **@sepotvin** — for reports precise enough to debug from

Version bumped to **1.3.0** (`c/version.py`). Full suite green: 28 C tests + 233 Python tests. Tag `v1.3.0` after merge to publish the release artifacts.


JustVugg:
Two additions since the description above — both already on `dev`, so this PR carries them:

### `TOPP` for Inkling (#704)

`colibri.c` had `TOPP` and `kimi_k3.c` had `K3_TOPP`; `inkling.c` was the only engine without it, and it is the one where it matters most — Inkling on a small host spends **92% of decode in expert I/O**, so every expert the router's tail contributes is ~28 MB read for little gain. Same semantics as the other two engines (sort routed pairs descending, cut at cumulative `p`, no renormalisation), opt-in, off by default.

New here and not in the other engines: **it reports what it trimmed**, so the trade is measurable rather than assumed.

```
[topp] 0.30: 36/72 routed used (50.0% trimmed, 1.00 experts/token avg)
```

Measured on the tiny fixture: 50% of routed experts trimmed, expert loads 47 → 27 (**−43%**). `TOPP=0` (the default) leaves the computation bit-identical — oracle still 36/36 and 24/24, at `cap=4` and `cap=1`.

Honest limit: no quality A/B on the real 975B yet. The fixture is random-init so its argmax happens not to flip when the tail is dropped — that proves the plumbing, not the quality. Documented as a quality lever accordingly.

### README: one command line for all three models (3937c00)

The models table listed the engines but the *Run it* section only ever showed GLM, so nothing told a reader that `coli` picks the engine from the model's `config.json` and renders that family's chat template — i.e. that **the command line does not change between models**:

```bash
COLI_MODEL=/nvme/glm52_i4   ./coli chat
COLI_MODEL=/nvme/inkling_i4 ./coli chat
COLI_MODEL=/nvme/kimi_k3    ./coli chat
./coli web --model /nvme/inkling_i4
```

Plus the two caveats that actually bite: Inkling needs `--cap 2` on a RAM-tight host, Kimi needs ~1.6 TB for the snapshot.

PR #710, opened 2026-07-30
docs: README header covers all four model families (main)

JustVugg:
Same header change as #709, cherry-picked onto `main` so the landing page matches what `main` actually ships.

`main` already contains all four engines (`colibri.c`, `inkling.c`, `kimi_k3.c`, `olmoe.c`) and the roster table at line 333 — it was only the header that still said **"Explore GLM-5.2 (744B-parameter MoE)"**, contradicting the repo description and hiding three of the four families 300 lines down the page.

## What is in this PR

- scale claim is a range, `744B to 2.8T`, instead of one model name
- the four families announced under the tagline, linking to the roster table that is already on `main`
- blockquote reworked: *"an inference engine you can run today, and an open research platform"*, with the product/research tension resolved by a distinction rather than a hedge — **no SLA on speed, and a hard guarantee on semantics**
- sample banner `v1.1.0` → `v1.3.0` (v1.3.0 is the current release)
- all four files: `README.md`, `README.it.md`, `README.zh-CN.md`, `README.zh-TW.md`

## What is deliberately NOT in this PR

`dev` also has README changes documenting the **Vulkan backend** (#418) and the OLMoE chat mode. Those describe code that is on `dev` and **not on `main`**, so copying the README files wholesale would have given `main` a page promising a backend it does not ship. Only the header commit was cherry-picked; the Vulkan documentation travels with the Vulkan code whenever `dev` next merges.

Every screenshot caption is untouched — they describe real measured runs, and inflating them to sound bigger would be falsifying a measurement.

Docs only. No code, no build, no behaviour change.


JustVugg:
Superseded by #711, which merges all of \`dev\` into \`main\` and already contains this exact commit (\`2dbfdc1\`, "docs: the README is about four model families, not one") along with the four updated README files.

This PR existed to get the header onto \`main\` on its own, in case the rest of \`dev\` was not ready to ship. It is, so the cherry-pick is redundant — and merging both would apply the same change twice. Closing rather than merging.

The Vulkan documentation that was deliberately held out of this PR travels with the Vulkan code in #711, which is where it belongs.

PR #714, opened 2026-07-30
Async transient CUDA experts for faster multi-GPU prefill

ZacharyZcR:
## Summary

Add an opt-in multi-GPU RAM spill tier for cold INT4 experts during prefill.

- reuse one gate/up/down CUDA staging set per selected GPU
- update staging weights from RAM and execute cold experts in parallel waves
- keep routing, weights, and quantization unchanged
- restrict the path to `S > 1`; decode keeps the existing resident-GPU/CPU path
- fall back to CPU per expert if upload or execution fails
- expose `COLI_CUDA_TRANSIENT_PREFILL` and `COLI_CUDA_TRANSIENT_MIN_ROWS`

The feature remains experimental and off by default.

## Real hardware measurement

GLM-5.2 744B INT4 on 6x RTX 5090, 2x Xeon Silver 4510, 251 GiB RAM. Same 703-token README prompt, greedy generation, 32 output tokens, 9,335 experts resident in VRAM and 10,121 in RAM.

| Mode | Prefill | Prefill throughput | Decode |
|---|---:|---:|---:|
| baseline | 74.00 s | 9.50 tok/s | 2.05 tok/s |
| transient prefill | 47.89 s | 14.68 tok/s | 2.28 tok/s |

Prefill throughput improves 54.5% (1.545x); latency falls 35.3%. The decode difference is normal short-run variance, not attributed to this feature because the transient path is disabled for `S == 1`.

Transient profile: 8,572 cold expert calls / 141,194 routed rows; summed worker upload time 64.60 s, compute time 6.95 s, overlapped across six devices. Routed CPU prefill rows fall from 143,357 to zero.

Test configuration includes `CUDA_EXPERT_GB=auto`, `CUDA_DENSE=1`, `COLI_CUDA_ATTN=1`, `COLI_CUDA_PIPE=2`, `COLI_CUDA_TC_W4A16=1`, `PIN_GB=all`, `DRAFT=0`, and `COLI_CUDA_TRANSIENT_MIN_ROWS=1`.

## Validation

- default CPU build
- 233 C/Python tests passed (13 skipped)
- CUDA build on sm_120
- CUDA backend correctness test
- ragged attention test (`relative_rms=0` for plain and grouped INT4)
- end-to-end baseline/feature A/B on the 744B model

PR #715, opened 2026-07-30
Overlap NVMe expert loads with transient CUDA

ZacharyZcR:
Depends on #714. The visible stack will shrink to this PR's single NVMe commit after #714 merges.

## What changes

- adds opt-in `COLI_CUDA_TRANSIENT_NVME=1` ready-first scheduling
- starts a transient CUDA wave as soon as at least one expert per GPU has finished loading, while the remaining NVMe reads continue
- supports both pthread `PIPE` loaders and non-blocking `io_uring` completion probing
- preserves the existing path byte-for-byte when the flag is off
- reports disk-expert and ready-first wave counts

Data path: NVMe -> RAM expert slab -> existing pinned transient staging -> GPU. No weights are approximated or skipped.

## 6x RTX 5090 / GLM-5.2 744B cold-disk A/B

403-token prefill, `CUDA_EXPERT_GB=auto PIN_GB=0 RAM_GB=24 PIPE=1 PIPE_WORKERS=16 DIRECT=1`, frozen usage placement.

- control: 48.84, 48.10 s; mean 48.47 s
- ready-first: 41.88, 46.03, 42.83 s; mean 43.58 s
- latency: -10.1%
- throughput: +11.2%
- main-thread expert wait fell from about 19.1-19.7 s to 10.7-12.0 s
- about 7,700 disk experts were consumed in about 1,000 ready-first waves per run

A threshold sweep also rejected over-batching: two experts/GPU raised the run to 45.48 s by waiting too long for I/O.

## Validation

- `make test`: all C tests pass
- Python: 233 passed, 13 skipped
- CUDA build passes with CUDA 13.3 / sm_120
- repeated real GLM-5.2 744B runs completed on 6x RTX 5090


ZacharyZcR:
Rebased onto the #714 decode-affinity fix. New tip: `d18b732` (base includes `35858fa`).

The NVMe ready-first waves now use the same pthread-isolated per-GPU transient workers, so prefill no longer changes the OpenMP team used by subsequent S=1 CPU decode. The rebased branch passes the full C test suite and an sm_120 CUDA build on the 6x RTX 5090 host.

ZacharyZcR:
Rebased again to inherit the multi-session decode guard from #714 (`a3cd561`). New tip: `68450ed`. NVMe ready-first remains prefill-only and no longer mistakes ragged mux decode (`S > 1`) for contiguous prefill.

ZacharyZcR:
#714's fresh current-dev A/B invalidated the warm-resident transient path it
depends on: Decode is now parity, but Prefill regresses 20.1%. I am therefore
leaving this PR as Draft.

The cold-NVMe ready-first result is a distinct regime and may still be useful,
but this stack can no longer merge as-is. The next honest step is to rebase it
into a standalone NVMe-only change (without exposing #714's warm-resident
feature), then repeat the cold-disk A/B against current `dev`. Until that is
done, the earlier +11.2% figure is historical evidence, not a current merge
claim.

ZacharyZcR:
Closing this stacked implementation now that its dependency #714 has been
rejected on the correctly configured NUMA baseline.

This does not invalidate the cold-NVMe ready-first idea or erase its historical
result. It means the current PR is no longer an honest merge unit: it carries
the warm-resident transient machinery from #714, while the only remaining
potential value is the narrower NVMe streaming case.

If that case is resumed, it should start as a new PR from current `dev` with:

1. an NVMe-only ready-first path, without #714's general transient feature;
2. serialized cold-cache A/B runs against the current baseline;
3. Decode and multi-session regression checks;
4. a benefit demonstrated specifically on a RAM-constrained configuration.

The branch and benchmark history remain available for reference.

PR #726, opened 2026-07-31
cuda: add native grouped E8 expert kernels

ZacharyZcR:
## Root cause

The generic CUDA expert-group kernels support formats 0 through 4. fmt=6
previously fell through to the int2 decoder, which read the 98-byte E8 blocks
with the wrong row layout and caused illegal memory accesses on GLM-sized
tensors.

## Fix

- add native homogeneous grouped E8 gate/up and down CUDA kernels
- expand each 32-weight E8 sub-block once per projection
- fuse gate/up accumulation and SiLU
- rotate the grouped intermediate buffer once before down projection
- support both synchronous and asynchronous grouped issue paths
- retain a safe per-expert fallback for mixed-format groups
- add a two-expert fmt=6 group oracle to the CUDA backend test

## Validation

- `make cuda-test CUDA=1 CUDA_ARCH=sm_120`: pass on RTX 5090
- E8 grouped fixture, including ragged plain/g64 cases: RMS 0
- full 270 GB GLM-5.2 E8 container, 6x RTX 5090, 125 GB GPU expert tier,
  NUMA enabled, all 19,200 routed experts resident:
  - no illegal access, fallback, or disk I/O
  - 703-token Prefill: 119.83 s / 5.87 input tok/s
  - 56-token Decode: 14.11 s / 3.97 tok/s
  - GPU critical expert time: 1.580 s, down 6.3% from 1.686 s
  - CPU routed time varied from 5.917 s to 6.533 s, masking the GPU gain in
    end-to-end Decode

### DIFF
--- a/c/backend_cuda.cu
+++ b/c/backend_cuda.cu
@@ -939,7 +939,7 @@ extern "C" int coli_cuda_expert_group(ColiCudaTensor *const *gates,
     if (!first) return 0;
     int device=first->device,D=first->I,I=first->O,total=0,max_rows=0;
     GroupDesc host[64]; if(count>64) return 0;
-    int all_s4=1,all_q4=1,any_g4=0;
+    int all_s4=1,all_q4=1,any_g4=0,any_e8=0;
     for(int c=0;c<count;c++){
         ColiCudaTensor *g=gates[c],*u=ups[c],*d=downs[c];
         if(!g||!u||!d||rows[c]<1||g->device!=device||u->device!=device||d->device!=device||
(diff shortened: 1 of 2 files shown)


ZacharyZcR:
Full-container validation after the fix, current `dev`, 6x RTX 5090, explicit
125 GB E8 expert tier (leaves resident tensors/workspace headroom), NUMA enabled,
all 19,200 routed experts resident across VRAM+RAM:

- no CUDA illegal access
- no resident-tensor fallback
- 100% expert hit, zero disk I/O
- 703-token Prefill: **123.54 s / 5.69 input tok/s**
- 56-token greedy Decode: **13.40 s / 4.18 tok/s**
- CPU E8 expert bandwidth: **33.94 GB/s**
- GPU critical expert time: 1.686 s vs CPU 5.917 s

The correctness blocker is closed. Performance is intentionally conservative:
fmt=6 groups now call the safe per-expert kernel until a native grouped E8
kernel exists. This result should not be presented as the format's performance
ceiling.

ZacharyZcR:
Native grouped E8 is now implemented in `a842d50`.

Exact A/B on the full GLM-5.2 E8 container:

| path | Prefill (703) | Decode (56) | GPU critical | CPU routed |
|---|---:|---:|---:|---:|
| safe per-expert fallback | 123.54 s | 4.18 tok/s | 1.686 s | 5.917 s |
| native grouped E8 | 119.83 s | 3.97 tok/s | 1.580 s | 6.533 s |

The native kernel improves the GPU-critical expert path by **6.3%** and Prefill
by **3.0%**. The lower end-to-end Decode number is CPU-side run variance: the
CPU routed half became 0.616 s slower while the GPU half became 0.106 s faster.
No CUDA error, resident fallback, or disk I/O occurred; expert hit rate stayed
100%.

So this closes the missing native fmt=6 grouped path, but it also confirms that
this mixed CPU/GPU placement is presently limited by the CPU/RAM tier rather
than the grouped GPU kernel.

PR #728, opened 2026-07-31
fix(cli): no invented model path — say what is missing instead (#724)

### LINKED ISSUE #724: [Bug]: Hardcoded paths
### Commit or version

1.3.0

### Environment

Windoze, Lenovo P16s gen 3

### Reproduction steps

Run `python coli plan` without any arguments

### Expected behavior

An error message, a specific one informing me that it can't find a model to load

### Actual behavior and logs

```shell
`cannot create resource plan: missing config.json: C:\home\vincenzo\glm52_i4`
```

terrizoaguimor:
Fixes #724.

`DEF_MODEL` defaulted to `/home/vincenzo/glm52_i4`, so every error named a directory from someone
else's machine. There is no default now — `plan`, `info` and `need_model` say
`no model directory given` and how to fix it, and `doctor` reports it as a failed check instead of
exiting, since reporting is what doctor is for.

Same class, fixed with it: `tools/download_glm52.py` defaulted a **400 GB** download to that path
(`GLM_DIR` is now required), the two `scripts/*.sh` use `${COLI_MODEL:?…}`, and four tools printed
`/home/vincenzo/…` in their usage examples. `grep -rn vincenzo c/` is now empty.

Tested with and without a model directory; `doctor --json` against a real model is byte-identical
to `dev` apart from its live `available_bytes`.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -1034,6 +1034,11 @@ def cmd_bench(a):
 
 def cmd_convert(a):
     banner("convert")
+    # here --model is the DESTINATION the converted weights are written to, not an
+    # existing snapshot, so it needs its own wording rather than need_model's.
+    if not a.model:
+        sys.exit(f"{C.yel}no output directory given.{C.r}\n"
+                 f"  --model <dir> is where the converted model is written")
     # python con torch/safetensors: l'ambiente del progetto se c'e', altrimenti quello corrente
     venv_py = os.path.join(HERE, "mio_env", "Scripts" if sys.platform == "win32" else "bin", "python3")
     py = venv_py if os.path.exists(venv_py) else sys.executable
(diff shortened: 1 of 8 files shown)


nonbasketless:
Why is everything here so verbose

I feel like this deserved three or four sentences

Thanks much for the fix though!

terrizoaguimor:
Fair, and you are right — that was a one-line default removal wearing a report. I have cut the description down. Thanks for saying so.

terrizoaguimor:
Found a regression in my own patch while sweeping all eleven subcommands rather than the three I had checked: `coli convert` died on a TypeError, because there `--model` is the *destination* the converted weights are written to, not a snapshot to read. Fixed in d93be65 with its own wording. `build` and `stop` also fail here, but identically on `dev` — a missing `make` and a missing pidfile dir, nothing to do with the model path.

PR #729, opened 2026-07-31
Vulkan MoE GEMV backend for integrated/AMD GPUs (draft, complements #418)

minne100:
## Vulkan MoE GEMV backend for integrated / AMD GPUs (DRAFT)

Isolates the Vulkan compute path for the qwen36 engine routed-expert GEMVs so it can be reviewed independently of the engine PR (#712).

### What
- vulkan_gemv.c (+ headers): Vulkan storage-buffer GEMV for the MoE expert forward.
- int4 weights unpacked in-shader + plain float GEMV (avoids the AMD OpSDotKHR segfault on Radeon 780M); int8 float path + optional IDP shader (blacklisted on known-broken drivers).
- Transparent startup probe; falls back to CPU MoE if vg_init() fails or COLIBRI_GPU=0.

### Relation to #418
#418 (steve-m) is the project Vulkan backend, CUDA/NVIDIA-oriented (expert tier + dense + MLA). This one targets integrated/AMD GPUs (shared memory, no CUDA) — complementary, not competing. Happy to coordinate with steve-m / maintainers toward a single backend if preferred.

### Status: DRAFT
- Depends on #712 (qwen36 engine) as the caller; not yet wired into dev build. Integration point: vg_init() / vg_expert_ensure().
- The engine-side gpu_probe fix (VK_QUEUE_COMPUTE_BIT was 0x20 = VIDEO_DECODE, should be 0x2) lives in #712.

### Fix included here
- 32-bit slot offsets wrap at >=4GB weights pool (layers 32-39 alias 0-7 at cache>=256). Now guarded: backend auto-disables + CPU fallback instead of silent corruption. 64-bit shader offsets (buffer_device_address) are a follow-up.

This supersedes the Vulkan portion of the now-closed #602 (engine taken over by #712).


Neppord:
multiple files are autogenerated from scripts that are not part of the pr, is that intentional?

> /* AUTO-GENERATED from gemv_int4.spv by tools/gen_spv_header.py. Do not edit. */

JustVugg:
@minne100 status check: this draft predates several rounds of `backend_vulkan.c` evolution (fmt=5 decode in all three shaders landed since, plus the Lavapipe CI job). Is the GEMV approach here still distinct from what the backend now does, or has the gap closed? If it's still complementary, a rebase would tell us; if not, closing is fine and the design notes remain linked.

JustVugg:
Thanks for isolating the Vulkan path so it can be read on its own, and for documenting the driver findings (the `OpSDotKHR` crash on the 780M is worth knowing).

As it stands this is a draft by its own description: about 11,800 lines that nothing in the engine calls, so merging it today would add code that is compiled by no target and exercised by no test. I am keeping it open rather than merging it. What would make it reviewable for `dev`:

- a wiring point in `qwen36` (or whichever engine you target first) behind an opt-in flag, so the path runs end to end;
- a token-exact comparison against the CPU path on the tiny fixture used by the CI, or a documented reason it cannot be exact and the measured drift;
- the SPIR-V blobs generated from checked-in shader source with the compiler and flags recorded, so they can be rebuilt.

With those three it becomes a backend PR we can test; without them it stays a draft, which is fine, but it cannot go in yet.

PR #741, opened 2026-08-01
fix(cuda): stage the VRAM tier in rounds so host RSS is bounded (closes #730)

### LINKED ISSUE #730: [Bug]: OOM trying to use more VRAM than system RAM
### Commit or version

292ed4c6164886adfa1b5ede54ecf15c917e1c0b

### Environment

- Ubuntu 26.04
- AMD Ryzen 9 5900X
- MemTotal:       65745408 kB
- NVIDIA RTX PRO 6000 Blackwell
- CUDA 13.3, driver 610.43.02
- gcc 15.2.0, nvcc 13.3.73
- Storage: WD_BLACK SN850X 4 TB NVMe (Gen4 x4)

### Reproduction steps

I've been running colibri (a great project!) on a system that has a RTX PRO 6000 Blackwell Max-Q (96GiB VRAM) but only 64 GiB system RAM, thanks to making kind of piecemeal upgrades.

I'd been trying to get it to use the full VRAM, under the assumption that's probably the tier where I can best store the most experts. I ran into an OOM while trying to configure `CUDA_RELEASE_HOST` for this system (which will free system RAM after the expert is loaded into GPU).

For easy reproduction on other HW (assuming you have a systemd-based distro), the following example sets a cgroup so that it has 24G system RAM & tries to use 40GB GPU experts.

```sh
make -C c/ colibri CUDA=1

systemd-run --user --wait --pipe -p MemoryMax=24G -p MemorySwapMax=0 \
  env SNAP=/mnt/hermes/models/glm52_i4_g64 COLI_CUDA=1 CUDA_RELEASE_HOST=1 \
      CUDA_EXPERT_GB=40 PIN=/dev/null PIN_FILL=1 PIN_GB=1 \
  "${PWD}/c/colibri" 64 4 4
```

(should also be able to exclude `PIN`, `PIN_FILL`, and `PIN_GB` if you already have an existing profile that hits this)

### Expected behavior

Goal is to use the full budget of CUDA_EXPERT_GB on less RAM. Or to cap CUDA_EXPERT_GB if it's not feasible (I think it might be, but unsure of performance cost so far).

### Actual behavior and logs

At commit https://github.com/JustVugg/colibri/commit/292ed4c6164886adfa1b5ede54ecf15c917e1c0b I get:

```
Running as unit: run-p493616-i513565.service; invocation ID: 2265d46862d84d64a1ab0e9ec7059538
[CUDA] device 0: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, 102.0 GB VRAM, sm_120
[CUDA] mode: routed experts only (resident dense on CPU); VRAM experts without host backing
== GLM C engine (glm_moe_dsa), cache=64 experts/layer | experts@4-bit dense@4-bit | idot: avx2 ==
loaded in 6.57s | resident dense: 10869.90 MB | layers=78 experts=256 | MTP ACTIVE (draft=0)
[MTP] active: native speculative decoding (draft=0)
          Finished with result: oom-kill                  
Main processes terminated with: code=killed, status=9/KILL
               Service runtime: 11.415s
             CPU time consumed: 58.739s
                   Memory peak: 24G (swap: 0B)
```

Setting other RAM budget variables like `RAM_GB=24` and `RSS_GUARD_GB=8` doesn't change the OOM.

Hypothesis is that this loop tries to load all of the GPU experts into host RAM:

https://github.com/JustVugg/colibri/blob/292ed4c6164886adfa1b5ede54ecf15c917e1c0b/c/colibri.c#L7303

and nothing gets released (as specified by `CUDA_RELEASE_HOST=1`) until it's done and in the next loop:

https://github.com/JustVugg/colibri/blob/292ed4c6164886adfa1b5ede54ecf15c917e1c0b/c/colibri.c#L7331

More details in Claude's doc (preceding is my summary & verified by me):
https://gist.github.com/mcollinswisc/b9baeb61dd4be170b56a698862743f55

JustVugg:
Fixes #730, reported and correctly diagnosed by @mcollinswisc.

## The bug

`pin_fill` loaded the **entire** VRAM-ranked prefix into host RAM in one parallel pass, and only then entered the upload loop that releases each slab:

```c
#pragma omp parallel for schedule(dynamic,1)
for(int a=0;a<(gpu_prefix?gpu_prefix:npin);a++)
    expert_load(...);                        /* ALL of them, into host RAM */

for(int a=0;a<gpu_limit && m->gpu_expert_bytes<budget;a++){
    ...
    if(uploaded && g_cuda_release_host) expert_host_release(m,s);   /* too late */
}
```

So `CUDA_RELEASE_HOST=1` worked exactly as documented and **could not help** — by the time the first release ran, every slab was already resident. **Peak host RSS was the whole `CUDA_EXPERT_GB` budget**, whatever that budget was.

The code's own comments state the intended contract and show the gap:

> *"Load the VRAM-ranked prefix first. Once uploaded its host backing is released before the disjoint RAM-ranked suffix is allocated."*
>
> `npin+=prefix_est;` — *additive: prefix RAM is returned after upload*

Both are true of the prefix **as a whole**. The design treats prefix RAM as transient, and it is — transient in aggregate rather than per expert, and aggregate is the number that has to fit.

It also explains why `RAM_GB=24` and `RSS_GUARD_GB=8` changed nothing for the reporter: those bound the *steady state*, and this peak is a startup transient that no steady-state budget models.

## Why it surfaced now

@mcollinswisc has **more VRAM than system RAM** — 96 GB on an RTX PRO 6000 Blackwell against 64 GB of RAM. Every assumption that host RAM comfortably exceeds the GPU tier quietly breaks in that regime, and it is going to become more common rather than less.

They also reproduced it for everyone else with a cgroup, which is what made this checkable in minutes instead of requiring their hardware:

```sh
systemd-run --user --wait --pipe -p MemoryMax=24G -p MemorySwapMax=0 \
  env SNAP=... COLI_CUDA=1 CUDA_RELEASE_HOST=1 CUDA_EXPERT_GB=40 \
      PIN=/dev/null PIN_FILL=1 PIN_GB=1 "${PWD}/c/colibri" 64 4 4
```

## The fix

Stage in rounds: load a round in parallel, upload it, release it, next. Peak host staging becomes `stage × expert_bytes` instead of the full budget, and the parallel load keeps its bandwidth.

Round size is `min(4 GB, budget/8)`, at least one expert and never more than the prefix. `budget/8` keeps rounds large enough to saturate the loader threads; the absolute cap protects the little-RAM, big-budget case that is the report. When it engages, it says so:

```
[CUDA] tier staging: N experts per round (X GB host peak) instead of M at once
       (Y GB) — CUDA_RELEASE_HOST frees each round before the next (#730)
```

Same bug family as the Inkling expert-cache defect fixed recently: acquiring the whole working set up front when the working set is larger than what fits, where processing in rounds is both correct and bounded.

## Behaviour is unchanged for everyone else

Without `CUDA_RELEASE_HOST`, without CUDA, or with no VRAM prefix, `stage == prefix` and **the loop runs exactly once** — which is the previous code path, statement for statement. The accumulators (`remaining`, `placed_b`, `placed_n`, `gpu_expert_bytes`) already lived outside the loop, so they carry across rounds untouched.

## Verification, and its limit

- Builds clean; **syntax-checks with `-DCOLI_CUDA`**, which the plain build does not compile.
- `make test-c` passes.
- The non-CUDA path is structurally identical with `stage == prefix`.

**Not verified functionally: I have no CUDA hardware.** Compile-level and non-CUDA evidence is what I can offer. @mcollinswisc has the reproduction and can confirm whether the OOM is gone and what the staging line reports; @4ny3l said they were investigating this and may want to review the approach.

One thing I did not change and want to flag: once the VRAM budget fills mid-prefix, later rounds still load experts that will not be uploaded, and they stay as RAM pins. That is the pre-existing behaviour — the old code loaded them too — so it is preserved deliberately rather than fixed silently here.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -8076,17 +8076,43 @@ static void pin_load(Model *m, const char *statspath, double gb, int trusted){
 #ifdef __linux__
     if(gpu_prefix>0) g_numa_skip_bind=1;   /* prefix slabs = transient upload staging: don't bind (#419) */
 #endif
-    #pragma omp parallel for schedule(dynamic,1)
-    for(int a=0;a<(gpu_prefix?gpu_prefix:npin);a++)
-        expert_load(m,r[a].l,r[a].e,&m->pin[r[a].l][slot_of[a]],1,0);   /* startup pin load; demand=0, never classified */
-#ifdef __linux__
-    g_numa_skip_bind=0;
+    int pre_n = gpu_prefix?gpu_prefix:npin;   /* quanti esperti nella fase "prefisso" */
+    /* #730: quanti caricarne per volta. Il default e' "tutti", che riproduce esattamente
+     * il comportamento precedente. Sotto CUDA_RELEASE_HOST i slab host del prefisso sono
+     * staging TRANSITORIO -- il commento di budget qui sopra lo dice: "prefix RAM is
+     * returned after upload" -- ma erano transitori in AGGREGATO: si caricava tutto il
+     * prefisso in RAM host e solo dopo si liberava un esperto alla volta. Il picco di RSS
+     * host era quindi l'intero CUDA_EXPERT_GB, e su un host con piu' VRAM che RAM (96 GB
+     * di VRAM contro 64 di RAM, #730) l'OOM arriva prima che il primo release parta.
+     * A round, il picco diventa stage*eb e il carico parallelo mantiene la sua banda. */
+    int stage = pre_n;
+#ifdef COLI_CUDA
+    if(g_cuda_enabled && g_cuda_release_host && gpu_prefix>0 && budget>0){
+        /* Tetto di staging: il piu' piccolo tra 4 GB e un ottavo del budget del tier,
+         * mai meno di un esperto (altrimenti il ciclo non avanza) e mai piu' del
+         * prefisso. Un ottavo mantiene i round abbastanza grandi da tenere occupati
+         * i thread del carico parallelo; il tetto assoluto protegge chi ha poca RAM
+         * e un budget enorme, che e' esattamente il caso segnalato. */
+        double cap = 4e9; if(budget/8.0 < cap) cap = budget/8.0;
+        int st = (int)(cap/eb);
+        if(st < 1) st = 1;
+        if(st < stage) stage = st;
+        if(stage < pre_n)
+            fprintf(stderr,"[CUDA] tier staging: %d experts per round (%.1f GB host peak) "
+                           "instead of %d at once (%.1f GB) — CUDA_RELEASE_HOST frees each "
+                           "round before the next (#730)\n",
+                    stage, stage*eb/1e9, pre_n, pre_n*eb/1e9);
+    }
 #endif
-    m->resident_bytes+=(int64_t)(gpu_prefix?gpu_prefix:npin)*eb;
+    for(int base=0; base<pre_n; base+=stage){
+    int hi = base+stage; if(hi>pre_n) hi=pre_n;
+    #pragma omp parallel for schedule(dynamic,1)
+    for(int a=base;a<hi;a++)
+        expert_load(m,r[a].l,r[a].e,&m->pin[r[a].l][slot_of[a]],1,0);   /* startup pin load; demand=0, never classified */
+    m->resident_bytes+=(int64_t)(hi-base)*eb;
 #ifdef COLI_CUDA
     if(g_cuda_enabled && budget>0){
-        int gpu_limit=gpu_prefix?gpu_prefix:npin;
-        for(int a=0;a<gpu_limit && m->gpu_expert_bytes<budget;a++){
+        for(int a=base;a<hi && m->gpu_expert_bytes<budget;a++){
             int li=r[a].l;
             { ESlot *s=&m->pin[li][slot_of[a]];
                 int64_t need=qt_bytes(&s->g)+qt_bytes(&s->u)+qt_bytes(&s->d);
@@ -8133,6 +8159,14 @@ static void pin_load(Model *m, const char *statspath, double gb, int trusted){
                 }
             }
         }
+    }
+#endif
+    }   /* fine del ciclo a round (#730) */
+#ifdef __linux__
+    g_numa_skip_bind=0;
+#endif
+#ifdef COLI_CUDA
+    if(g_cuda_enabled && budget>0){
         fprintf(stderr,"[CUDA] hot expert tier: %d/%d experts, VRAM %.2f GB (budget %.1f GB%s, reserve %.1f GB)\n",
             m->gpu_expert_count,npin,m->gpu_expert_bytes/1e9,
             g_cuda_expert_auto?safe_total/1e9:budget/1e9,


mcollinswisc:
> Not verified functionally: I have no CUDA hardware. Compile-level and non-CUDA evidence is what I can offer. @mcollinswisc has the reproduction and can confirm whether the OOM is gone and what the staging line reports; 

I can confirm it's successfully able to use the full VRAM now, after the fix from f6ae858, tested at 06f31b2:

```
$ COLI_CUDA=1 COLI_GPU=0 CUDA_RELEASE_HOST=1 CUDA_EXPERT_GB=96 RAM_GB=54 \
DIRECT=1 URING=1 THINK=1 CTX=4096 NGEN=64 COLI_TEMP=0 \
SNAP=/mnt/hermes/models/glm52_i4_g64/ COLI_PROMPT="How is it going?" ./c/colibri 64 4 4
[URING] queued expert I/O active (depth=512, workers=8, DIRECT=1)
[CUDA] device 0: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition, 102.0 GB VRAM, sm_120
[CUDA] mode: routed experts only (resident dense on CPU); VRAM experts without host backing
== GLM C engine (glm_moe_dsa), cache=64 experts/layer | experts@4-bit dense@4-bit | idot: avx2 ==
loaded in 6.45s | resident dense: 10869.90 MB | layers=78 experts=256 | MTP ACTIVE (draft=0)
[MTP] active: native speculative decoding (draft=0)
[USAGE] expert history: 199216 selections (/mnt/hermes/models/glm52_i4_g64//.coli_usage)
[CUDA] tier staging: 188 experts per round (4.0 GB host peak) instead of 4522 at once (96.0 GB) — CUDA_RELEASE_HOST frees each round before the next (#730)
[CUDA] hot expert tier: 4521/5374 experts, VRAM 96.00 GB (budget 96.0 GB, reserve 2.0 GB)
[CUDA]   device 0: 4521 experts, 96.00 GB
[PIN] placement: 4521 VRAM + 853 RAM expert (18.1 GB warm) in 32s da /mnt/hermes/models/glm52_i4_g64//.coli_usage
[RAM_GB=54.0] resident 29.5 GB + reserve 6.3 GB (ws 1.4, KV 1x4096 0.7, kvb 0.5), experts 21.2 MB x 77 layers -> cap lowered 64->11 (projected peak 53.8 GB)
[stop] 18 stop tokens: 154820 154827 154829 154821 154822 154823 154824 154825 154826 154828 154830 154831 154832 154833 154834 154835 154836 154837 (15 from the tokenizer's special set)
prompt: 7 tokens | generating up to 64 (EOS stop=154820) | n-gram draft=0
How is it going?PROFILO PREFILL (9.61s):
PROFILE: expert-disk 398.372s service / 0.513s wait | expert-matmul 7.618s | attention 1.017s (including kvb 0.087s) | lm_head 0.033s | other 0.434s
ATTENTION: projection/RoPE 0.322s | score-softmax-value 0.024s | output projection 0.583s
 I hope you are doing well. I am writing to ask you for a favor
[t=16  RSS 45.31 GB  hit 69%  0.42 tok/s  1.00 tok/fw]
. I am planning to travel to your city next month and I would love to
[t=32  RSS 45.31 GB  hit 68%  0.42 tok/s  1.00 tok/fw]
 stay with you for a few days. I know it is a lot to ask
[t=48  RSS 45.31 GB  hit 67%  0.41 tok/s  1.00 tok/fw]
, but I would really appreciate it if you could let me stay at your place
[t=64  RSS 45.31 GB  hit 65%  0.41 tok/s  1.00 tok/fw]

---
prefill 7 tokens in 9.61s | decode 64 tokens in 157.42s (0.41 tok/s) | expert hit rate 65.5% (pin 37.0% + lru 28.5%) | RSS 45.31 GB
experts loaded/token: 590.6 (per-layer 7.88 across 75; baseline topk=8) | TOPK=0 TOPP=0.00
speculation: 1.02 tokens/forward (63 forwards per 64 tokens) | MTP acceptance 0% (0/0)
CUDA expert tier: 4521 resident experts (96.00 GB) | 12189 calls served from VRAM
[CUDA] resident set: 13563 tensors, 96.00 GB VRAM
[CUDA] expert groups: 4563 call, 13811 expert, 14669 righe (3.03 expert/call)
PROFILE: expert-disk 0.000s service / 26.219s wait | expert-matmul 86.731s | attention 27.995s (including kvb 0.000s) | lm_head 0.000s | other 16.471s
ATTENTION: projection/RoPE 9.959s | score-softmax-value 7.661s | output projection 10.371s
```

During generation:

```
$ nvidia-smi
Sun Aug  2 13:09:20 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 610.43.02              KMD Version: 610.43.02     CUDA UMD Version: 13.3     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA RTX PRO 6000 Blac...    Off |   00000000:07:00.0  On |                  Off |
| 30%   50C    P1             68W /  300W |   97025MiB /  97887MiB |      1%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
...
|    0   N/A  N/A          712886      C   ./c/colibri                           95496MiB |
+-----------------------------------------------------------------------------------------+
```

Thanks!

Issue #707, opened 2026-07-30
[Bug]: OpenMP spin-wait tuning is inert on macOS — and applying it costs 2.2x decode on a 32 GB host


fredchu:
## Summary

The OpenMP hot-thread tuning block in `main()` never reaches the OpenMP runtime on macOS, because the self-re-exec that makes it effective is compiled only for Linux and FreeBSD. On this host that turns out to be **load-bearing in the good direction**: applying the same settings externally makes decode **2.2x slower**. So the report is less "please fix the missing branch" and more "please don't fix it without gating it, and here is the measurement".

Two separate items, both about how the OpenMP team is configured on macOS.

## Commit or version

`2d623816eecbf740b1fc06eb422974bda0af8931` (main, 2026-07-28)

## Environment

macOS 26.5.2 (Darwin 25.5.0), Apple M1 Max (8 performance + 2 efficiency cores), 32 GB unified memory, internal Apple NVMe / APFS, Apple clang 17.0.0, Homebrew `libomp`, `make colibri METAL=1`. Model: `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp` (429.3 GB, int8 MTP head). Full benchmark context in the companion performance report.

## Reproduction steps

```shell
cd colibri/c && make colibri METAL=1

P="Explain why a Mixture-of-Experts model needs less RAM than its parameter count."

# A — as shipped
RAM_GB=24 COLI_METAL=1 ./coli run --model <model> --temp 0 --ngen 32 "$P"

# B — the same four variables main() tries to set, exported before launch
RAM_GB=24 COLI_METAL=1 \
OMP_WAIT_POLICY=active KMP_BLOCKTIME=200 OMP_DYNAMIC=FALSE GOMP_SPINCOUNT=200000 \
  ./coli run --model <model> --temp 0 --ngen 32 "$P"

# then one variable at a time
```

## Expected behavior

Either the tuning applies on macOS the way it does on Linux, or it is explicitly not applied because it was measured not to help there.

## Actual behavior and logs

```text
=== Item 1: the tuning is inert on macOS, and that is currently a good thing ===

decode 32 tokens, prompt 24 tokens, greedy, RAM_GB=24, COLI_METAL=1
expert hit rate held at 9.9-10.1% across every run below, so cache state is not the variable

  A  as shipped                          218.55s / 219.39s   (mean 218.97s)
  B  four tuning vars exported           481.20s / 490.72s   (mean 485.96s)   +122%
[...]

PR #750, opened 2026-08-01
inkling: audio input, Metal expert MoE, and tuned Apple defaults

rgbkrk:
Adds audio input to Inkling and speeds it up on Apple Silicon: DMel frames in
at `<|audio|>` placeholders, routed-expert matmul on the Metal GPU
(`COLI_METAL=1`, opt-in), and three default changes on `__APPLE__` builds that
come out of a measured knob sweep rather than a guess. Text-only Inkling and
non-Apple builds are unchanged, everything here is `__APPLE__`/`COLI_METAL`
guarded or opt-in.

## Audio input (DMel)

Inkling's audio tower is small on purpose: a `[mel_bins*mel_vocab, D]`
embedding table plus one RMSNorm. A frame's embedding is
`rmsnorm(sum of 80 rows, eps 1e-6)`, and it *replaces* the placeholder's text
embedding rather than adding to it (`embed_norm` doesn't apply), matching
`masked_scatter` in HF's `modeling_inkling.py`. That's the whole audio path,
there's no separate encoder stack to load or run, which is why this lands as
a small patch to `inkling.c` instead of a new subsystem.

The tensors (`model.audio.*`, ~10 MB bf16 on Inkling-Small) load from either a
`--keep-audio` container or an `audio.safetensors` sidecar dropped next to the
snapshot. The sidecar matters for anyone running off a remote/lazy mount:
fetch ~10 MB via HTTP range requests instead of pulling the 532 GB bf16
checkpoint just to get the audio tower. Absent tensors, text-only engine,
nothing changes.

Three entry points land the DMel frames:

- `--audio file.dmel` on the CLI (implies `--chat`, emits TMLv0 audio framing)
- an optional 7th field on the serve wire protocol, carrying raw DMel bytes
- a `"dmel"` array in `ref_inkling.json` for the oracle harness

The 7th SUBMIT field was already reserved (glm uses it for grammar bytes), and
the two engines never see each other's extension, so there was no new framing
to design, just a slot to fill. Frame counts are checked against placeholder
counts up front, a mismatch is a hard error, not a silent skip.

On the gateway side, `input_audio` content parts on `/v1/chat/completions`
now work for Inkling: base64 WAV in, spoken answer out. The WAV decode (PCM16
or float32, mono after mixdown, 16 kHz enforced) and the DMel DSP are a numpy
port, imported lazily so text-only serving stays stdlib-only. The DSP is
validated byte-identical against `tinkernel-audio` (itself byte-golden
against the official `tml-renderers` 0.1.0 wheel) on 7 clips, including a
quiet-audio case that exercises the turn-level RMS boost. The HF processor
skips that boost; `tml-renderers` is what the model actually saw in training,
so that's what this matches.

`tools/make_tiny_inkling_audio.py` adds the multimodal oracle fixture: a
random-init `InklingForConditionalGeneration` with an audio tower, saved in
the native TML layout. The engine reproduces it token-for-token: 38/38
teacher-forced positions, 24/24 greedy.

## Metal expert MoE (`COLI_METAL=1`, opt-in)

Reuses colibri's existing `coli_metal_moe_block` rather than adding a second
GPU kernel. The container's int4 format (nibble-packed, -8 offset, per-row
scales) is already bit-identical to the kernel's `fmt=2` input, so this is
wiring, not new math.

Expert slots move from per-slot mallocs to one page-aligned slab per sparse
layer, registered with the Metal backend once at startup. Eviction reuses the
region instead of unregistering and re-registering, so there's no
`useResource` churn on cache turnover and unified memory stays zero-copy.
That single change is most of the win below, see the defaults section for
why.

Each cache round groups its (token, expert) pairs by expert and submits one
command buffer for the round; a failed submit falls back to the CPU loop for
that round only, so a bad dispatch degrades instead of crashing the run.

## Apple performance defaults

Three defaults change on `__APPLE__` builds, all three came out of the same
10-config sweep (2 reps per config, frozen probe: fixed prompt, cap, and
pins, `USAGE_SAVE=0`) rather than being turned on individually and hoped for:

- **`COLI_METAL_RESSET=1` default.** Without the residency set, each decode
  MoE block pays ~135 ms of per-buffer `useResource` churn to do ~3 ms of
  actual math. With it: decode goes 0.17 -> 1.76 tok/s, a 10x change from one
  registration pattern.
- **OMP active-spin tuning off on Apple.** Measured strictly worse even for
  pure-CPU decode (0.50 -> 0.84 tok/s without the spin) and prefill
  (49.7s -> 24.4s without the spin). Same SoC power-budget mechanism as
  `docs/METAL-M5MAX-PERF-REPORT.md` documented for GLM on the M5 Max: an
  active-spinning CPU team steals clock headroom the GPU needs, and Apple
  Silicon shares that budget across CPU and GPU.
- **GPU decode on by default** (`INK_METAL_MIN_S` 2 -> 1). With the residency
  set in place, GPU decode beats CPU decode 2x, so the old prefill-only gate
  was leaving performance on the table rather than protecting it.

Net effect on the probe: 0.50 tok/s decode / 49.7 s prefill at the old
defaults, 1.75 tok/s / 10.3 s at the new ones. Output tokens are identical
across every config in the sweep, this is a placement change, not a
precision change.

The prefill number matters more than usual for audio: a 5-second clip is
~100 DMel positions, so for audio turns prefill speed counts double compared
to a text-only prompt of the same token length.

Also: `mem_avail_bytes` gets a macOS branch (free + inactive + purgeable
pages), so expert-cache auto-sizing works on macOS instead of falling back to
a fixed 16 experts/layer regardless of how much RAM is actually free.

## Behavioral coverage

| Audio tensors | `COLI_METAL` | Platform | What happens |
|---|---|---|---|
| absent | any | any | text-only engine, byte-identical to before this PR |
| present | off | any | DMel frames replace placeholder embeddings, CPU expert matmul |
| present | on | Apple | same audio path, routed experts run on the Metal GPU via the residency-set slab, per-block CPU fallback on a failed dispatch |
| present | n/a | non-Apple | `COLI_METAL`/Metal code paths aren't compiled in; behaves like the CPU row |

## Validation

`make check` is green (275 Python tests, C unit tests, 0 build warnings). The
tiny multimodal oracle is exact on CPU-f32 and on the container-int4 path
both CPU and Metal. Gateway DSP is byte-identical to `tml-renderers` across
all 7 test clips. Beyond the synthetic fixtures, this also ran live against
the real Inkling-Small 131 GiB int4 container: real speech in, correct
transcripts out, including spoken-number normalization ("seven forty five
PM ... gate B twelve" comes back as "7:45 PM from gate B12").

## Benchmark report

- Commit: `e57083f4ae0b6b8c0fbbd977bee888ce6a3b76c5`
- Hardware: Apple M-series Mac, 128 GB unified memory
- Model/storage: Inkling-Small, 131 GiB int4 container, internal NVMe, warm
  page cache
- Workload: frozen probe (fixed prompt, cap, and pins), `USAGE_SAVE=0` so
  cache-warming doesn't drift the comparison between configs
- Run count: 2 reps per configuration across the 10-config sweep

| | Decode | Prefill |
|---|---|---|
| Old defaults | 0.50 tok/s | 49.7 s |
| New defaults | 1.75 tok/s | 10.3 s |

## How to try it

```sh
# converting yourself: keep the audio tower (adds ~10 MB to the container)
python3 c/tools/convert_inkling_int4.py --indir <bf16> --outdir <snap> --keep-audio

make -C c inkling METAL=1
COLI_METAL=1 SNAP=~/Models/inkling_i4 ./c/inkling --audio turn.dmel -n 128

# gateway: base64 WAV in, spoken answer out
./coli serve --model ~/Models/inkling_i4
curl localhost:8000/v1/chat/completions -d '{
  "model": "inkling-colibri",
  "messages": [{"role": "user", "content": [
    {"type": "input_audio", "input_audio": {"data": "<base64 wav>", "format": "wav"}}
  ]}]
}'
```

Already sitting on a text-only container? The two audio tensors
(`model.audio.encoder.weight`, `model.audio.final_norm.weight`) can be pulled
out of the original checkpoint's shards with HTTP range requests (~10 MB
instead of the 532 GB download) and written as an `audio.safetensors` sidecar
next to the snapshot; `st_init` already indexes every `*.safetensors` in the
directory, so the engine just picks it up.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -461,8 +461,8 @@ cuda-bench: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/bench_tens
 olmoe$(EXE): olmoe.c st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h omp_tune.h
 	$(CC) $(CFLAGS) olmoe.c -o olmoe$(EXE) $(LDFLAGS)
 
-inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ)
-	$(CC) $(CFLAGS) inkling.c $(INK_CUDA_OBJ) -o inkling$(EXE) $(LDFLAGS)
+inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ) $(METAL_OBJ)
+	$(CC) $(CFLAGS) inkling.c $(INK_CUDA_OBJ) $(METAL_OBJ) -o inkling$(EXE) $(LDFLAGS)
 
 kimi_k3$(EXE): kimi_k3.c st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h quant.h omp_tune.h route_trace.h $(VK_OBJ) $(VK_SPV)
 	$(CC) $(CFLAGS) kimi_k3.c $(VK_OBJ) -o kimi_k3$(EXE) $(LDFLAGS)
--- a/c/inkling.c
+++ b/c/inkling.c
@@ -38,6 +38,20 @@
 #include "backend_cuda_ink.h"
 static int g_cuda = 0;
 #endif
+#ifdef COLI_METAL
+/* Apple-GPU expert MoE (opt-in, COLI_METAL=1). Reuses colibri's batched
+ * coli_metal_moe_block: inkling's container int4 (nibble-packed, -8 offset,
+ * per-row scales) is bit-identical to the Metal fmt=2 kernel, and int8 to
+ * fmt=1. Expert slots live in page-aligned per-layer slabs registered once
+ * for zero-copy resolve — unified memory, no upload. Attention and the dense
+ * path stay on the CPU (at high hit rates ~90% of decode is expert matmul). */
+#include "backend_metal.h"
+static int g_metal = 0;
+#endif
+#if defined(__APPLE__)
+#include <mach/mach.h>
+#include <sys/sysctl.h>
+#endif
 
 #define MAXL 256
 
@@ -52,6 +66,7 @@ typedef struct {
     int n_experts, topk, n_shared, moe_inter, dense_inter;
     int eos;
     float eps, route_scale, mup;
+    int audio_tok, mel_bins, mel_vocab;   /* DMel audio input (TMLv0 <|audio|> placeholder) */
     unsigned char local[MAXL];            /* 1 = sliding-window layer */
     unsigned char sparse[MAXL];           /* 1 = MoE layer, 0 = dense MLP */
 } Cfg;
@@ -115,6 +130,8 @@ typedef struct {
     int xq;                               /* experts on disk are a colibri container (U8 + .qs) */
     Wt embed, lm_head;
     float *embed_norm, *final_norm;
+    Wt audio_enc;                         /* [mel_bins*mel_vocab, D] embedding table */
+    float *audio_norm;                    /* audio tower RMSNorm [D]; NULL = no audio */
     Layer *L;
     LCache *cache;
     int64_t rb13, rb2;                    /* container row-bytes (0 = not container) */
@@ -488,6 +505,12 @@ static void load_cfg(Cfg *c, const char *snap) {
     jval *eo = json_get(root,"eos_token_id");
     if (!eo || eo->t != J_NUM) eo = json_get(r,"eos_token_id");
     c->eos = (eo && eo->t == J_NUM) ? (int)eo->num : -1;
+    /* DMel audio: placeholder id at the top level (absent in the shipped
+     * config -> the TMLv0 constant), frame geometry under audio_config */
+    c->audio_tok = (int)jnum(root,"audio_token_id",200023);
+    jval *ac = json_get(root,"audio_config");
+    c->mel_bins  = ac ? (int)jnum(ac,"n_mel_bins",80)     : 80;
+    c->mel_vocab = ac ? (int)jnum(ac,"mel_vocab_size",16) : 16;
     /* real config.json: intermediate_size = MoE, dense_intermediate_size = dense.
      * HF-saved config (post_init applied): intermediate_size = dense, moe_intermediate_size = MoE. */
     jval *dis = json_get(r,"dense_intermediate_size");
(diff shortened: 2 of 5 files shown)


JustVugg:
Reviewed properly rather than on the strength of the description, and it holds up on every point I checked. Merging.

## What I verified, and how

**The text-only path really is unchanged.** Everything audio is behind `st_has(&m->S, "model.audio.encoder.weight")` and an `if (!m->audio_norm) return 0;` early-out, so a container without the two audio tensors never enters any of it.

**The wire protocol is byte-identical when there is no audio and no grammar.** This was the part I most wanted to check, because `openai_server.py` is shared by every engine. The header is built as

```python
+ (f" {len(xpayload)}" if xpayload else "")
```

so with both empty it produces exactly the previous bytes. GLM and Kimi see no change at all.

**The overloaded 7th field is safe, and I confirmed the assumption it rests on** rather than taking it on trust. Your comment says *"inkling rejects grammars upstream"* — that is `openai_server.py:1965`:

```python
if grammar is not None and ARCH in ("inkling", "kimi"):
    raise APIError(400, "`response_format` grammars are not supported by the {ARCH} ...")
```

Pre-existing code, and it fires before `generate()` is reached. Combined with your explicit `Grammar and audio cannot be combined` guard, there is no path where an engine reads one as the other. Worth noting `grep -i grammar c/inkling.c` returns nothing at all — the engine has never had grammar support, so the field was genuinely free.

**numpy is not a new dependency.** Lazily imported inside `_np()` and turned into a `400` with a clear message if absent. A text-only user never touches it.

**Built and tested here:** all four engines compile, `make check` is green (288 tests, `OK (skipped=13)`), the new `make_tiny_inkling_audio.py` parses, and `--keep-audio` is wired through the converter. CI is 13/13 including all three `All engines` platforms.

## The Apple OpenMP change is the most valuable thing in this PR, and you may not know why

```c
-#ifndef COLI_CUDA
+#if !defined(COLI_CUDA) && !defined(__APPLE__)
```

> *the active-spin team steals the shared SoC power budget … measured strictly worse on Inkling-Small even for pure-CPU decode — 0.50 vs 0.84 tok/s, and 2x the prefill*

**@fredchu measured the same effect independently last week in #707**, on different hardware (M1 Max, 32 GB), a different engine (GLM-5.2), and by a different route — he found the Darwin re-exec branch was missing and then measured what applying the settings externally costs: **2.2× slower decode**. He could not explain why beyond the I/O-pool contention argument.

Two independent measurements, different machines, different engines, same direction and similar magnitude. That moves this from "a tuning preference on one Mac" to a property of the platform, and it is the strongest evidence in the tracker for what to do with that block. I will link the two.

It also means `colibri.c` still has the problem your patch fixes in `inkling.c` — worth a follow-up, though `colibri.c` currently has 14 open PRs against it so it wants doing carefully rather than quickly.

## The one design choice I would flag, without objecting to it

Overloading the seventh `SUBMIT` field with two different meanings works, is guarded, and is documented in the comment — but it is subtle, and the safety depends on a check 400 lines away in a different file. If a third engine ever wants a per-request payload, please give the field a tag rather than adding a third meaning. Not a change request for this PR.

## What I appreciated

The behavioural coverage table, four rows including the "absent/any/any → byte-identical" one, is the review I would otherwise have had to do by hand.

And the note at the end — that the two audio tensors can be pulled out of the original shards with **HTTP range requests, ~10 MB instead of re-downloading 532 GB** — is the kind of detail that decides whether people actually try a feature. That belongs in `docs/inkling.md`, and I will put it there if you do not get to it first.

Also: running it against the real 131 GiB container and reporting that spoken numbers come back normalised ("*seven forty five PM … gate B twelve*" → "*7:45 PM from gate B12*") is worth more than any synthetic fixture. Thank you for doing the expensive version of the test.

Issue #700, opened 2026-07-29
The learning cache is engine-specific: .coli_usage has two incompatible writers and two engines that cannot produce it at all


terrizoaguimor:
## Summary

"A cache that learns" is one of the first things the README promises, and routing history is what feeds it (`PIN=auto`, `AUTOPIN`, `REPIN`, the `.coli_pairs` coupling prefetch, the expert atlas in the dashboard). That machinery is per-engine today, and the four engines disagree:

| Engine | `ROUTE_TRACE` | writes `.coli_usage` | reads it for `PIN` |
|---|---|---|---|
| `colibri.c` (GLM-5.2) | ✅ | ✅ sparse **text** triples | ✅ |
| `inkling.c` | ❌ | ✅ dense **binary** `IKU1` | ✅ (own format only) |
| `olmoe.c` | ❌ | ❌ | ❌ |
| `kimi_k3.c` (#676) | ❌ | ❌ | ❌ |

So: routing traces exist on exactly one engine, and the usage file is written in **two mutually unreadable formats under the same filename**.

I hit this while instrumenting K3 to collect routing data, and rather than send a K3-only patch I would rather propose the shared facility — if that shape appeals, I am happy to implement it.

## The format collision, precisely

`colibri.c` writes sparse text through `stats_dump_q()` (`telemetry.h`):

```c
FILE *f=fopen(tmp,"w");                       /* text */
... fprintf(f,"%d %d %u\n", i, e, m->eusage[i][e]);   /* layer expert count, non-zero only */
```

`inkling.c:798` writes a dense binary block with its own magic:

```c
FILE *f = fopen(tp, "wb");                    /* binary */
uint32_t hdr[3] = { 0x31554B49u /* "IKU1" */, (uint32_t)c->n_layers, (uint32_t)E };
fwrite(hdr, 4, 3, f);
for (int i = 0; i < c->n_layers; i++) fwrite(m->eusage[i] ? ... , 4, E, f);
```

Both default to `<snap>/.coli_usage` and both honour `PIN=<path>`. The comment above `inkling.c`'s writer says *"same contract as glm's `.coli_usage`"* — it is a different contract.

What that costs today:

1. Point `inkling` at a snapshot where GLM has run: the magic check at `inkling.c:750` fails and it declines cleanly — *"[pin] …: not an inkling usage file, ignoring"* — so the user does get told, but pinning is off and the history they accumulated is unusable.
[...]

PR #754, opened 2026-08-01
olmoe: adopt route_trace.h — the last engine on the shared history

terrizoaguimor:
The fourth and last engine for #700, after `colibri.c` (#716), `kimi_k3.c` (#719) and
`inkling.c` (#742). With this, every engine that routes experts writes the same history.

`m->freq` moves from one flat `[n_layers * n_experts]` block to `route_trace.h`'s row per
layer, so each read site — the hot-pin ranking, the bump in `moe()`, and the LFRU eviction
guard — changes shape but not meaning. `last_access` stays flat; it is a separate array and
`route_trace.h` does not own it.

The history follows `kimi_k3`'s convention exactly: opt-in behind `COLI_USAGE`. With the
variable unset nothing is loaded and nothing is written, so the default path is unchanged.
`PPL=1` deliberately does not save — a loss sweep should not fold its own tokens into the
persisted ranking.

Verified against the real OLMoE-1B-7B-0125-Instruct, converted with
`tools/convert_olmoe_merged.py`. The patched engine is byte-identical to `dev` on the same
run — same tokens and the same `hit=1216 miss=832`, so the eviction path decides exactly as
it did. The `COLI_USAGE` round trip writes `-1 16 64` / `-2 1 3841822045`, has no row for the
dropped layer 16, and accumulates 2048 → 4096 across two runs. `make check` green.

`route_trace.h` itself needs no change — `"olmoe"` was already in `rt_engine_names[]` from
#716, so the refusal message can name it on both sides.

Two things I found on the way and deliberately did not touch — `hot_pinned.bin` loads with no
header or identity check, and `ref_olmoe_real.json` no longer matches a fresh conversion — are
written up in #700, since both are behaviour questions rather than part of this move.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -458,7 +458,7 @@ cuda-bench: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/bench_tens
 	"$(GPUCC)" $(GPUFLAGS) backend_cuda.cu tests/bench_tensor_core.cu -o backend_cuda_bench$(EXE) $(ANS_NVCC_LIBS)
 	./backend_cuda_bench$(EXE)
 
-olmoe$(EXE): olmoe.c st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h omp_tune.h
+olmoe$(EXE): olmoe.c st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h omp_tune.h route_trace.h
 	$(CC) $(CFLAGS) olmoe.c -o olmoe$(EXE) $(LDFLAGS)
 
 inkling$(EXE): inkling.c st.h json.h compat.h $(INK_CUDA_OBJ) $(METAL_OBJ)
--- a/c/olmoe.c
+++ b/c/olmoe.c
@@ -36,6 +36,7 @@
 #include <omp.h>   /* omp_set_num_threads/omp_get_max_threads per omp_tune.h */
 #endif
 #include "omp_tune.h"
+#include "route_trace.h"                    /* shared routing telemetry (#700) */
 
 #ifdef _WIN32
 #include <windows.h>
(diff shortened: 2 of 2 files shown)


JustVugg:
Merging — **that closes the loop from #700**. Four engines, one history format, six weeks after you filed the issue that pointed out there were two mutually unreadable writers under the same filename and two engines producing nothing at all.

Verified before merging: 13/13 CI including all three `All engines` platforms, and the change is genuinely small at +24/-10 — which is the signature of the facility being right rather than the port being shallow. `m->freq` going from a flat `[n_layers * n_experts]` block to a row per layer touches three read sites and changes shape without changing meaning; that is exactly the kind of thing that gets subtly wrong, and your note that `last_access` stays flat because `route_trace.h` does not own it is the sentence that tells me you looked.

`COLI_USAGE`-gated with the default path untouched, matching `kimi_k3`, and `PPL=1` deliberately not saving — a loss sweep should not fold its own tokens into a persisted ranking. Agreed, and worth having written down.

The round-trip evidence is the useful part: `-1 16 64` / `-2 1 3841822045`, no row for the dropped layer, and 2048 → 4096 accumulating across two runs. That is the header, the drop, and the accumulation all shown at once.

**The two things you found and deliberately left alone** are the right call, and both deserve to be picked up rather than forgotten:

- `hot_pinned.bin` loading with no header or identity check is the same class of problem this whole issue was about — a file trusted because it is where it is expected to be. It should probably get the same refuse-by-name treatment.
- `ref_olmoe_real.json` no longer matching a fresh conversion is the more worrying of the two, because an oracle that has quietly drifted from what the converter produces is an oracle that stops catching things.

Separating them from the move was correct: they are behaviour questions, and bundling them would have made a +24/-10 refactor into an argument.

Thank you for seeing this through all four engines. #700 started as "I hit this while instrumenting K3 and would rather propose the shared facility than send a K3-only patch", and that is exactly what it turned into.

PR #755, opened 2026-08-01
build: Linux/aarch64 branch in the Makefile — probe DOTPROD/I8MM instead of trusting native (#631)

### LINKED ISSUE #631: build: Linux/aarch64 falls through to the x86-64 branch — i8mm SMMLA kernels silently compile out
## Summary

On Linux/aarch64 the default `make` produces a binary with **zero** `smmla`
instructions. The i8mm SMMLA kernels added for ARM are guarded by
`#if defined(__ARM_NEON) && defined(__ARM_FEATURE_MATMUL_INT8)`
(`c/quant.h:617`, `:730`, `:740`), and on this toolchain that macro is never
defined — so the kernels are compiled out with no error and no warning.

This is likely affecting the two published GB10 rows in `docs/benchmarks.md`,
both of which report a **matmul-bound** warm profile.

## Root cause

`c/Makefile` computes an `AARCH64` variable at line 37:

```make
AARCH64 := $(filter aarch64 arm64,$(TARGET_CPU))
```

but only ever consumes it inside the **Darwin** branch, where it correctly
appends `-mcpu=$(ARCH)` and the comment explicitly says this is what "unlocks the
i8mm SMMLA int8/int4 dot kernels in colibri.c" (lines 58–68).

There is **no Linux/aarch64 branch**. Linux ARM therefore falls through to the
final `else` at line 100, which is commented:

```make
# --- Linux / *BSD x86-64 (percorso originale, invariato) ---
CC      = gcc
ARCH   ?= native
CFLAGS  = -O3 -march=$(ARCH) ...
```

`-march=native` is correct on x86. On aarch64 it depends entirely on whether GCC
recognises the core.

## Evidence

Host: NVIDIA DGX Spark (GB10, Grace: 10× Cortex-X925 + 10× Cortex-A725),
Ubuntu 24.04, Linux 6.17, **gcc 13.3.0**, colibrì `81f08a0` (v1.1.1).

`/proc/cpuinfo` advertises the features:

```
asimddp sve sve2 svei8mm svebf16 i8mm bf16
```

But gcc 13.3 does not know Cortex-X925, so:

| flags | `__ARM_FEATURE_MATMUL_INT8` / `DOTPROD` / `SVE2` / `BF16` |
|---|---|
| `-march=native` | **none** (only 5 generic `__ARM_FEATURE_*` macros) |
| `-mcpu=native` | **none** |
| `-mcpu=cortex-x925` | **none** |
| `-march=armv9-a+i8mm+bf16` | **all four** |

Note that `-mcpu=native` — the fix the Darwin branch uses — **also fails here**,
so simply extending the Darwin branch to Linux would not be enough on gcc 13.

Resulting binaries, same source, same compiler:

```console
$ make clean && make glm
$ objdump -d colibri | grep -c '\bsmmla\b'
0

$ make clean && make glm ARCH="armv9-a+i8mm+bf16"
$ objdump -d colibri | grep -c '\bsmmla\b'
22
```

The startup banner confirms which kernel is selected — with the ARCH build it
reads:

```
== GLM C engine (glm_moe_dsa), cache=%d experts/layer | experts@%d-bit dense@%d-bit | idot: neon-i8mm ==
```

## Why this probably matters for the published GB10 numbers

From #136 (Dell Pro Max GB10), the warm profile per 128 tokens was:

> expert-disk 57.9 s / **expert-matmul 129.8 s** / attention 49.7 s — once the
> learning cache warms, this box is **matmul-bound, not disk-bound**.

and that report's own "where the headroom is" section says:

> Grace has `i8mm` (SMMLA ≈ 2× SDOT int8 peak) and **nothing in the tree uses it
> yet** … Back-of-envelope, halving the matmul bucket would be worth ~+30%
> end-to-end here.

The kernels have since landed. But if those runs used the default Linux/aarch64
`make`, they emitted 0 `smmla` and the flagged headroom is still unrealised — the
2.4 / 3.33 ladder in `docs/benchmarks.md` may be a **non-i8mm** measurement.

### Measured delta: ~2×, not ~30%

Interleaved A/B (arms alternated A B A B so cache drift loads onto both equally;
`.coli_usage` lives in the model dir and persists across runs, so a sequential
A-then-B would hand the second arm a cache the first one warmed). Both binaries
built up front from the same source. `./glm 64 4 8`, `NGEN=48`, `TEMP=0`,
`DIRECT=1 PIPE=1`, greedy, same prompt:

| precision | no-i8mm | i8mm | gain |
|---|---:|---:|---:|
| int8 experts, r1 / r2 | 0.55 / 0.56 | 1.06 / 1.11 | 1.93× / 1.98× |
| int4 experts, r1 / r2 | 0.57 / 0.57 | 1.13 / 1.11 | 1.98× / 1.95× |

p50 decode-forward latency 5202 ms → 2425 ms. Time shares move
`expert-matmul 67% → 59%`, and both builds still report
`verdict: compute-bound in expert matmuls`.

This is ~2× end-to-end, against the ~+30% estimated in #136. That estimate
assumed SMMLA halving *only* the matmul bucket (129.8 s of ~237 s → ~+27%); the
measured whole-forward speedup is larger than that model predicts. I am not
going to claim a mechanism for the excess — absolute read-service time also fell
(65.2 s → 52.9 s), but the i8mm arm happened to sit at a slightly higher hit rate
(70.3% vs 66.7%), so part of that is cache drift rather than a causal effect of
the kernel. The A/B ratio is the solid number; the decomposition is not.

**Consistency check with the published ladder:** #136 reports 0.50 tok/s warm,
pure CPU, on a GB10; we measure 1.10–1.13 stock CPU with i8mm on. That is ~2.2×,
in line with what this A/B predicts — but the two hosts are **not** a controlled
pair: #136's Dell OEM SSD measured 5.58 GB/s O_DIRECT against 9.99 GB/s here, so
disk is confounded with the kernel across that comparison. The within-host A/B
above is the evidence; the cross-host ratio is only suggestive. What would settle
it is someone re-running #136's box with `ARCH="armv9-a+i8mm+bf16"`.

## Suggested fix

Add an explicit Linux/aarch64 branch. Because `-mcpu=native` is unreliable on
older GCC for newer cores, prefer probing for a flag that actually defines the
macro rather than trusting `native`:

```make
else ifneq (,$(AARCH64))
# --- Linux / *BSD aarch64 (Grace, Graviton, Jetson, Ampere, RPi5...) ---
# -march=native is NOT safe here: GCC silently falls back to a bare baseline on
# cores it does not know (e.g. gcc 13.3 vs Cortex-X925), which compiles the i8mm
# SMMLA kernels in quant.h out with no diagnostic. Probe instead.
CC      = gcc
ARM_TRY := armv9-a+i8mm+bf16 armv8.6-a+i8mm+bf16 armv8.2-a+dotprod
# Probe must check the MACRO, not just that the flag is accepted: -march=native
# is accepted here and still defines nothing, so an accept-only probe would
# happily select it and reintroduce the bug.
ARCH    ?= $(firstword $(foreach a,$(ARM_TRY),\
             $(if $(shell $(CC) -march=$(a) -dM -E - </dev/null 2>/dev/null \
                   | grep -q __ARM_FEATURE_MATMUL_INT8 && echo $(a)),$(a))))
CFLAGS   = -O3 $(if $(ARCH),-march=$(ARCH)) -fopenmp -pthread -Wall -Wextra \
           -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function
LDFLAGS  = -lm -fopenmp -pthread
EXE      =
```

Verified on this host: the probe selects `armv9-a+i8mm+bf16`. (The last entry in
`ARM_TRY` is a dotprod-only fallback that will not satisfy the MATMUL_INT8 grep —
either drop it or widen the probe to accept `__ARM_FEATURE_DOTPROD` as a second
tier, depending on whether you want a partial-acceleration rung.)

A cheaper alternative, if you'd rather not add flag probing: keep the fallthrough
but emit a build-time warning when `__ARM_FEATURE_MATMUL_INT8` is undefined on an
ARM host, so the slow path is at least **loud**:

```c
#if defined(__ARM_NEON) && !defined(__ARM_FEATURE_MATMUL_INT8)
#warning "aarch64 build without i8mm: SMMLA kernels disabled. Rebuild with ARCH=armv9-a+i8mm+bf16"
#endif
```

Either way, a one-line note in the README's build section would help GB10 /
Graviton / Jetson owners, since the failure is completely silent today.

## Reproduce

```bash
gcc -march=native -dM -E - </dev/null | grep -c __ARM_FEATURE_MATMUL_INT8   # 0 here
make clean && make glm                    && objdump -d colibri | grep -c '\bsmmla\b'
make clean && make glm ARCH="armv9-a+i8mm+bf16" && objdump -d colibri | grep -c '\bsmmla\b'
```

## Separately — a note on the 30 tok/s target in #161

Not a bug, but possibly useful for that thread's roadmap. GB10's Grace LPDDR5X
measures **273 GB/s** on this host. At full top-8 int4, per decoded token
colibrì touches roughly 9.9 GB of resident dense weights plus 75 layers × 8
experts × ~19 MB ≈ 11.4 GB of experts — call it **~21 GB/token**, so a
**~13 tok/s memory roofline** even at 100% residency.

The 11.19 tok/s reported in #161 is therefore already ~87% of roofline. That
reframes ">11 → 30 needs 2–3× faster MLA": 30 tok/s at full top-8 int4 would
need ~630 GB/s of bandwidth, which this silicon does not have. Reaching 30 on a
GB10 looks like it requires lower K, lower precision, or more tokens per forward
(higher accept length) — not a faster attention kernel. Happy to be corrected if
the residency assumptions are off.

anrasi:
## Problem

#631 reported it precisely: on Linux/aarch64 there is no dedicated branch in
`c/Makefile`, so ARM falls through to the x86-64 `else` and gets
`-march=native`. On a gcc that does not recognise the core — gcc 13.3 vs
Cortex-X925 on a DGX Spark GB10, but any gcc older than its silicon qualifies —
`native` silently degrades to plain `armv8-a`: neither `__ARM_FEATURE_DOTPROD`
nor `__ARM_FEATURE_MATMUL_INT8` is defined, the SDOT/SMMLA int8-int4 kernels
compile out with no warning, and the banner reads `idot: neon`.

The nastiest part is quiet: without `DOTPROD`, `g_i4s` initialises to 2
(`c/quant.h:531-541`), so the `S>=g_i4s` gate at `c/colibri.c:883` sends
**int4 S=1 decode — the token-generation hot path — to the f32 fallback**.

`-mcpu=native` would not help: as #631's table shows, it degrades identically
on this toolchain, so extending the Darwin branch's approach to Linux is not
enough.

## Fix

A Linux/*BSD aarch64 branch next to the PowerPC one. For the default
`ARCH=native` it probes the compiler once:

1. If `-mcpu=native` already defines `__ARM_FEATURE_MATMUL_INT8`, use it
   unchanged (a gcc that knows the core keeps full native tuning).
2. Otherwise, re-add the features the silicon itself reports in
   `/proc/cpuinfo` (`asimddp` → `+dotprod`, `i8mm` → `+i8mm`) as
   `-march=armv8-a+…` — kept **only if this compiler accepts the modifiers**
   (gcc < 10 has no `+i8mm`; composing blindly would turn a slow build into a
   broken one). If it doesn't, fall back to `-mcpu=native` and say so with a
   `$(warning)` — the current behaviour, minus the silence.

An explicit `ARCH=` is respected verbatim: `armv*` values go to `-march`,
core names to `-mcpu` (same convention as the Darwin branch). On *BSD there is
no `/proc/cpuinfo`, so the probe degrades gracefully to `-mcpu=native`.
Cross-compilers keep passing an explicit `ARCH=` exactly as before — the probe
only fires on the `native` default, where host = target is already assumed.

x86-64, Darwin, Windows and PowerPC branches are untouched.

## Validation (DGX Spark GB10, Ubuntu 24.04, gcc 13.3.0, on top of current `dev`)

Flag selection, all paths:

| invocation | flags chosen |
|---|---|
| `make` (default) | `-march=armv8-a+dotprod+i8mm` (probe) |
| `make ARCH=armv8.6-a+i8mm` | `-march=armv8.6-a+i8mm` |
| `make ARCH=cortex-a76` | `-mcpu=cortex-a76` |
| `make` with a gcc that rejects `+i8mm` (simulated gcc 9) | `-mcpu=native` + explicit warning |

Resulting default binary, same source, same compiler as #631's evidence:

```console
$ make clean && make
$ objdump -d colibri | grep -c smmla
22        # was 0 before this patch
$ strings colibri | grep -o 'idot: .*'
idot: neon-i8mm ==
```

`make check` green on this branch on top of current `dev`: all C test
binaries + 273 Python tests OK (34 skipped) — every test binary visibly
compiled with the probed `-march=armv8-a+dotprod+i8mm`.

Kernel-level effect of the recovered instructions at the GLM-5.2 744B expert
shapes (`matmul_qt_ex`, 2048×6144 / 6144×2048, best-of-3 windows, idle
machine, 2 interleaved rounds — generic-armv8-a build vs this patch's default):

| shape | before (GF/s) | after (GF/s) | Δ |
|---|---|---|---|
| gate/up int8 S=8 | ~243 | ~274 | 1.13× |
| gate/up int4 S=8 | ~178 | ~223 | 1.26× |
| down int8 S=8 | ~269 | ~296 | 1.10× |
| down int4 S=8 | ~186 | ~241 | 1.30× |
| gate/up int8 S=1 (decode) | ~95 | ~126 | ~1.3× |
| gate/up int4 S=1 (decode) | ~36 | ~91 | **~2.5×** (leaves the f32 fallback) |

As #631 notes, the published GB10 rows in `docs/benchmarks.md` were measured
matmul-bound and likely predate (or missed) the i8mm kernels; happy to
contribute a re-measured row from this box in a follow-up if useful.

Credit: the root-cause analysis is @james-fwai's in #631 — this PR adds the
fix that issue was closed without.

Closes #631.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -97,6 +97,40 @@ ARCH   ?= native
 CFLAGS  = -O3 -mcpu=$(ARCH) -fopenmp -pthread -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function
 LDFLAGS = -lm -fopenmp -pthread
 EXE     =
+else ifneq (,$(AARCH64))
+# --- Linux / *BSD aarch64 (#631) ---
+# NEON e' baseline, ma i kernel SDOT/SMMLA int8-int4 vivono dietro
+# __ARM_FEATURE_DOTPROD / __ARM_FEATURE_MATMUL_INT8. A gcc that does not know
+# this core (gcc 13 vs Cortex-X925 on GB10: #631) degrades native to plain
+# armv8-a SILENTLY: the kernels compile out, no warning, the banner says
+# "idot: neon" and int4 S=1 decode falls back to f32. So for the default
+# ARCH=native, probe the macro; if it is missing, re-add the features the
+# silicon itself reports in /proc/cpuinfo -- but keep them only if this
+# compiler accepts the modifiers (gcc < 10 has no +i8mm: composing blindly
+# would turn a slow build into a broken one).
+CC      = gcc
+ARCH   ?= native
+ifeq ($(ARCH),native)
+ARCHFLAG := -mcpu=native
+ifeq ($(shell $(CC) -mcpu=native -dM -E - </dev/null 2>/dev/null | grep -c __ARM_FEATURE_MATMUL_INT8),0)
+CPUFEAT  := $(shell grep -m1 '^Features' /proc/cpuinfo 2>/dev/null)
+ARCH_EXT := $(if $(filter asimddp,$(CPUFEAT)),+dotprod)$(if $(filter i8mm,$(CPUFEAT)),+i8mm)
+ifneq ($(ARCH_EXT),)
+ifneq ($(shell $(CC) -march=armv8-a$(ARCH_EXT) -dM -E - </dev/null 2>/dev/null | grep -cE '__ARM_FEATURE_DOTPROD|__ARM_FEATURE_MATMUL_INT8'),0)
+ARCHFLAG := -march=armv8-a$(ARCH_EXT)
+else
+$(warning $(CC) does not accept$(subst +, +,$(ARCH_EXT)): int8/int4 idot kernels compile out (#631))
+endif
+endif
+endif
+else ifneq (,$(filter armv%,$(ARCH)))
+ARCHFLAG := -march=$(ARCH)
+else
+ARCHFLAG := -mcpu=$(ARCH)
+endif
+CFLAGS  = -O3 $(ARCHFLAG) -fopenmp -pthread -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function
+LDFLAGS = -lm -fopenmp -pthread
+EXE     =
 else
 # --- Linux / *BSD x86-64 (percorso originale, invariato) ---
 CC      = gcc


JustVugg:
Merging. This is a good catch on a failure mode that is worse than being slow: it is being slow **silently**.

`-mcpu=native` degrading to plain armv8-a when the compiler does not recognise the core means the SDOT/SMMLA kernels compile out, the banner prints `idot: neon`, and int4 S=1 decode quietly falls back to f32. Nothing errors. Nobody knows. That is the same class as the two bugs we fixed this week — engines that stopped compiling on Windows for months, and a `READY` sentinel mangled by the CRT — where the absence of a complaint was mistaken for the absence of a problem.

**What I checked:**

- The whole branch is inside `ifneq (,$(AARCH64))`, and `AARCH64 := $(filter aarch64 arm64,$(TARGET_CPU))`, so x86 and Windows paths are untouched. Verified by building here on x86-64: `colibri` and `kimi_k3` both clean.
- You do not compose flags blindly — `-march=armv8-a$(ARCH_EXT)` is only adopted if the compiler actually defines the feature macros afterwards. That matters: gcc < 10 has no `+i8mm`, and blindly appending it would turn a slow build into a broken one.
- The fallback **warns** rather than silently accepting less. That is the whole point of the PR and it would have been easy to leave out.

Reading `/proc/cpuinfo` `Features` for `asimddp`/`i8mm` is the right source — it is what the silicon reports, not what the compiler guessed.

Nice first contribution, and thank you for the datapoints on #165 as well. If you have the GB10 in front of you, a before/after `idot:` line from the banner plus decode tok/s would be worth adding to #631 — it turns "this should help" into a number, and yours is the hardware where the bug bites.

PR #756, opened 2026-08-01
docs: inkling audio input and the 10 MB sidecar fetch

rgbkrk:
Follow-up from the #750 review: the range-request trick moves into
`docs/inkling.md` where people can find it.

The new section covers what the audio tower actually is (an embedding table
plus a norm, which is why audio costs nothing when absent), the two ways to
get the tensors (`--keep-audio` at conversion, or the ~10 MB sidecar fetch
for anyone already holding a text-only container), the input contract, and a
gateway example.

The fetch script is not illustrative pseudocode: it was extracted from the
page as written and executed, and its output is byte-identical to the
sidecar used for the live validation runs in #750. The Modes table gains the
`--audio` row and the intro no longer claims the engine is text-only.

### DIFF
--- a/docs/inkling.md
+++ b/docs/inkling.md
@@ -3,7 +3,9 @@
 `c/inkling.c` runs [Thinking Machines Inkling](https://huggingface.co/thinkingmachines/Inkling)
 (975B total / 41B active, Apache 2.0) with colibri's expert-streaming approach:
 dense weights resident (RAM or VRAM), routed experts streamed from disk with an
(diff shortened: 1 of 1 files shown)


JustVugg:
Merging — and this is the doc I offered to write on #750 if you did not get to it first. You got there first, which is better.

The sidecar note is the part that matters most and it is easy to underrate: **~10 MB of HTTP range requests instead of a 532 GB re-download** is the difference between "I will try the audio path" and "I will not". Someone sitting on a text-only container was previously looking at a re-fetch they would never do.

Merged as-is.

Issue #622, opened 2026-07-25
[Bug]: Metal prefill GEMM is not token-exact vs CPU on near-tie logits (S >= GEMM_MIN); decode unaffected


lBroth:
On tiny fixture containers, teacher-forced prefill disagrees CPU vs Metal at 2 of 32 positions, deterministically, on stock `dev` at 7b03ce0 (M5 Pro, macOS 26.5.2). Greedy decode is token-identical on the same containers. This came out of the #587 hardware runs; filing it separately as agreed there, since it reproduces on stock dev where neither PR's code exists.

Reproduce, from `c/` with a tiny oracle snapshot in place (`tools/make_glm_oracle.py`, needs the transformers>=5.11 / Python 3.12 venv; the converter additionally needs torch and safetensors). One trap to know: if `--indir` points at a missing or empty snapshot the converter prints `converted 0 main shard(s)` and exits 0, and every command below then passes vacuously on an empty container, so check that line says 1.

```
python3 tools/convert_fp8_to_int4.py --indir glm_tiny --outdir /tmp/t_i4g64 --ebits 4 --io-bits 8 --group-size 64 --n-layers 5
python3 tools/convert_fp8_to_int4.py --indir glm_tiny --outdir /tmp/t_i3 --ebits 4 --io-bits 8 --xbits 3 --group-size 0 --n-layers 5

SNAP=/tmp/t_i4g64 TF=1 ./colibri 64 4 16                 # CPU reference
SNAP=/tmp/t_i4g64 COLI_METAL=1 TF=1 ./colibri 64 4 16    # Metal
# same CPU/Metal pair with SNAP=/tmp/t_i3 for the second container
```

Observed, identical across repeated runs:

| container | positions that differ (CPU token vs Metal token) |
|---|---|
| grouped int4 g64 | pos 14: 34 vs 29, pos 27: 7 vs 112 |
| int3 experts, per-row int4 dense | pos 19: 91 vs 103, pos 22 vs 23 shift |

[...]

PR #763, opened 2026-08-01
metal: GPU attention for S>4 prefill (single command buffer)

RDouglasSharp:
## Summary

The Metal fused-attention path (`coli_metal_attn_decode`) was capped at `S<=4`, so it only ran for decode / MTP. During **prefill** (S>4) every layer's attention fell back to the CPU even with `COLI_METAL=1`. This lifts the cap and runs large-S prefill attention on the GPU, in a **single command buffer**, cutting prefill attention time ~4× while producing output consistent with the CPU reference. It is **off by default** (`COLI_METAL_PREFILL=1` to opt in), since GPU prefill can pick a different token on near-tie logits (the [#622](https://github.com/JustVugg/colibri/issues/622) family) and a greedy stream is not guaranteed bit-identical to the CPU. CPU and CUDA paths are unchanged.

## Motivation

On Apple Silicon, `COLI_METAL=1` accelerates decode attention and the routed experts, but prefill attention was left on the CPU (the gate required `S<=4`). For any real prompt that means a large slice of the cold-prefill wall runs on the CPU while the experts run on the GPU.

## What changed

- **`c/colibri.c`** — the fused-attention gate no longer hard-caps at `S<=4`. It can engage for any `S` when Metal absorption is active, controlled by a new `COLI_METAL_PREFILL` env — **default `0`**, which keeps the old S≤4 cap and leaves S>4 on the CPU; `=1` opts in. Also adds a cumulative timer to the `[prefill]` progress line.
- **`docs/metal.md`** — documents the flag: what it buys (~4× prefill attention) and what it costs (near-tie divergence, #622).
- **`c/backend_metal.mm`** — `coli_metal_attn_decode` now handles arbitrary `S`. Projections → attention core → output GEMV run in **one command buffer**, ordered by memory barriers — the same structure the S≤4 path already used and which is token-exact vs the CPU. A single `a_score` dispatch is `S*AHEADS*T` threads; it is guarded under ~2³⁰ and falls back to the CPU for very large prompts (in-encoder row-chunking to keep those on the GPU is a follow-up).

The CPU and CUDA code paths are untouched.

## Performance

GLM-5.2 int4, Apple Silicon (M-series), 544-token prompt, greedy, default settings. CPU attention is timed per stage; on the GPU the three stages run fused in one command buffer and are reported as a single total.

| | CPU | GPU |
|---|---|---|
| **prefill attention** | **35.9 s** _(projection/RoPE 16.1 s · score·softmax·value 12.0 s · output 6.3 s)_ | **9.0 s** _(8.6 s GPU kernel)_ |
| **prefill wall** | 96.4 s | 67.3 s |

Prefill attention is ~4× faster on the GPU, trimming the prefill wall ~30% at this length. Prefill is dominated by streaming experts from disk, so the attention share — and the win — grows with S. Decode is unaffected.

## Correctness

Compared GPU prefill (`COLI_METAL_PREFILL=1`) against the pure-CPU reference (`COLI_METAL=0`), both at the model's default settings, greedy decoding:

- **Natural prompts** (e.g. a ~540-token README excerpt): the GPU output stays close to and semantically consistent with the CPU — same opening, same reading of the input. Where individual tokens differ they are argmax near-ties (equally-valid tokens), not errors.
- **Highly repetitive prompts** (e.g. a pangram repeated 40×): an early token can flip and greedy cascades it. Same near-ties, amplified — GPU/CPU int4 accumulation rounds differently and GLM-int4 sits on argmax ties.

This is the documented [#100](https://github.com/JustVugg/colibri/issues/100) / [#622](https://github.com/JustVugg/colibri/issues/622) sensitivity, identical in kind to the prefill GEMM and the existing S≤4 Metal and MTP paths: every emitted token remains a valid argmax and the continuation stays correct, it just isn't guaranteed to be the same stream. Because of exactly this, the flag ships **off by default**.

## Flags / behavior

- `COLI_METAL_PREFILL=0` (**default**) — S>4 prefill attention stays on the CPU; bit-exact vs the CPU, no behavior change.
- `COLI_METAL_PREFILL=1` — opt in to S>4 prefill attention on the GPU (the ~4× above; may diverge on near-tie logits, #622).
- Prompts above the single-dispatch thread cap fall back to the CPU automatically.
- Documented in `docs/metal.md`.

## Note (separate, pre-existing)

While validating this, we found that forcing `ABSORB=1` routes S>4 prefill through the CPU **absorbed** path, which produces incorrect output on long prompts. That path predates this PR and is off by default, so this change doesn't touch it — but as a side effect, S>4 absorbed attention now runs on the correct GPU kernel instead. Flagging in case it's worth a separate look.

### DIFF
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -799,15 +799,27 @@ extern "C" int coli_metal_matmul(ColiMetalTensor **tp, float *y, const float *x,
 // ---- fused decode attention scratch (GLM-5.2 dims) ----
 enum { AH=6144, AHEADS=64, AQLORA=2048, AKVL=512, AROPE=64, AVH=256, AQH=256, ANOPE=192, AROWSH=448, AHQH=AHEADS*AQH, AHVH=AHEADS*AVH, AMAXS=4 };
 static id<MTLBuffer> ax_,aqr_,aqf_,acomp_,aqabs_,ascore_,aclat_,actx_,aout_,aqaln_,akvaln_; static size_t ascore_cap;
-static id<MTLBuffer> axr_,anrm_,ash1_,ash2_,ashout_,asig_,aidx_,aw_,akeff_;   // full-layer tail
+static size_t ax_cap,aqr_cap,aqf_cap,acomp_cap,aqabs_cap,aclat_cap,actx_cap,aout_cap;
+static id<MTLBuffer> axr_,anrm_,ash1_,ash2_,ashout_,asig_,aidx_,aw_,akeff_;   // full-layer tail (AMAXS-sized)
 static void attn_scratch_init(){
   if(ax_) return;
   auto L=[&](size_t n){ return [g_dev newBufferWithLength:n*AMAXS options:g_res_opts]; };
-  ax_=L(AH*4); aqr_=L(AQLORA*4); aqf_=L(AHQH*4); acomp_=L((AKVL+AROPE)*4);
-  aqabs_=L((size_t)AHEADS*AKVL*4); aclat_=L((size_t)AHEADS*AKVL*4); actx_=L(AHVH*4); aout_=L(AH*4);
-  aqaln_=L(AQLORA*4/AMAXS); akvaln_=L(AKVL*4/AMAXS);   // norm weights are per-tensor, not per-row
   axr_=L(AH*4); anrm_=L(AH*4); ash1_=L(2048*4); ash2_=L(2048*4); ashout_=L(AH*4);
   asig_=L(256*4); aidx_=L(8*4); aw_=L(8*4); akeff_=L(4);
+  aqaln_=[g_dev newBufferWithLength:AQLORA*4 options:g_res_opts];
+  akvaln_=[g_dev newBufferWithLength:AKVL*4 options:g_res_opts];
+}
+static void attn_scratch_reserve(int S, int T){
+  attn_scratch_init();
+  ax_=ensure(ax_,&ax_cap,(size_t)S*AH*4);
+  aqr_=ensure(aqr_,&aqr_cap,(size_t)S*AQLORA*4);
+  aqf_=ensure(aqf_,&aqf_cap,(size_t)S*AHQH*4);
+  acomp_=ensure(acomp_,&acomp_cap,(size_t)S*(AKVL+AROPE)*4);
+  aqabs_=ensure(aqabs_,&aqabs_cap,(size_t)S*AHEADS*AKVL*4);
+  ascore_=ensure(ascore_,&ascore_cap,(size_t)S*AHEADS*T*4);
+  aclat_=ensure(aclat_,&aclat_cap,(size_t)S*AHEADS*AKVL*4);
+  actx_=ensure(actx_,&actx_cap,(size_t)S*AHVH*4);
+  aout_=ensure(aout_,&aout_cap,(size_t)S*AH*4);
 }
 // y[S,O] = quantized-weight(w) applied to xin[S,I]. Weights are registered (page-aligned,
 // zero-copy) at model load; resolve to (buffer,offset). Returns false to fall back to CPU.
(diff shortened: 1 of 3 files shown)


JustVugg:
The measurement is good and the finding behind it is real: the `S<=4` gate meant prefill attention sat on the CPU while the experts ran on the GPU, which is a strange split for anyone running `COLI_METAL=1`. 35.9 s → 9.0 s on the attention, 30% off the prefill wall at 544 tokens, single command buffer with the same structure the `S<=4` path already uses. All of that I want.

**One change requested: make `COLI_METAL_PREFILL` default to 0.**

## Why, in your own words

> *an early token can flip and greedy cascades it*
> *the GPU output stays **close to and semantically consistent** with the CPU*

That is a change to the emitted token stream, shipped on by default. And the front page of this project says:

> **no SLA on speed, and a hard guarantee on semantics**

with the accompanying line that placement changes speed, never the model's answers. A default that can alter which tokens come out contradicts the one promise we make unconditionally — the guarantee is not "the continuation stays sensible", it is "you get the same model".

## Your #100 argument is fair, and it is also why this needs the opposite default

You are right that this is the same *kind* of sensitivity as the existing `S<=4` Metal path and MTP, and right that every emitted token remains a valid argmax. Nobody is calling this a correctness bug.

But look at how the project treats the other members of that family:

- **#622 is open right now**: *"Metal prefill GEMM is not token-exact vs CPU on near-tie logits (S >= GEMM_MIN); decode unaffected"*. This PR expands exactly that surface, so turning it on by default would enlarge a bug we have not closed.
- **MTP** is disabled when acceptance does not repay verification, and #100 — the issue you cite — was closed by making non-reproducibility *visible and controllable*, not by accepting it.
- **`EXPERT_BUDGET`** is quarantined in `colibri.c` for being faster at the cost of semantics, with the reasoning left in the source as a reminder.
- The two most recent Metal PRs, #457 and #757, both led with **"token-identical through eos"** as the headline claim rather than the throughput.

Every one of those points the same way: speed that changes the output is opt-in here.

## What I am asking for

1. **`COLI_METAL_PREFILL` defaults to 0.** Same patch, same measurements, one character.
2. **A line in `docs/metal.md`** saying what the flag buys and what it costs — "~4× prefill attention; output may differ from CPU on argmax near-ties (#622)". People with a 544-token prompt and a Mac will turn it on; they should know what they are turning on.
3. Optionally, and only if it is cheap: have the engine **say** when it engages, the way `[OMP]` and `[CUDA]` lines do. A run whose output is not CPU-reproducible should carry that fact in its own log, so a benchmark posted here is self-describing.

If you would rather argue for default-on, the argument that would move me is a measurement rather than a principle: run the same prompt set both ways and report how often the streams actually diverge. If it is one prompt in fifty on natural text, that is a different conversation from a coin flip — and nobody has that number today, including me. Your repetitive-pangram case suggests the rate is input-dependent, which is exactly the sort of thing worth quantifying once rather than each of us guessing.

Two other things worth saying since they were easy to miss in a PR this well-written:

The **2³⁰ thread guard with a CPU fallback for very large prompts** is the kind of thing that gets left out and then found by a user with a 4,000-token prompt. Thank you for putting it in.

And separating the CPU attention timings by stage (projection/RoPE 16.1 s · score·softmax·value 12.0 s · output 6.3 s) is what makes the 4× claim checkable rather than assertable.

JustVugg:
Status note so this does not read as stalled: **held on one character**, and nothing else.

`COLI_METAL_PREFILL` defaulting to `1` ships a path that, by your own description, can change which tokens come out:

> an early token can flip and greedy cascades it

The front page promises **no SLA on speed, and a hard guarantee on semantics**, and #622 — *"Metal prefill GEMM is not token-exact vs CPU on near-tie logits"* — is open right now. Default-on would enlarge an unclosed bug.

Everything else in the PR I want: 35.9 s → 9.0 s on prefill attention, single command buffer, the 2^30 thread guard with its CPU fallback, and the per-stage CPU timings that make the 4× checkable rather than assertable. None of that is in question.

Flip the default to `0`, add a line to `docs/metal.md` saying what the flag buys and what it costs, and it merges.

And the standing offer: if you would rather argue for default-on, bring the **divergence rate** on natural prompts rather than a principle. One in fifty is a different conversation from a coin flip, your repetitive-pangram case suggests it is input-dependent, and **nobody has that number — including me**. A measurement would settle it either way, and we would know something today we do not.

RDouglasSharp:
Understood and done.  The default is now off.

PR #764, opened 2026-08-02
docs(experiments): GLM-5.2 on 4x RTX A6000 (Ampere) + EPYC 7402P

ThefloorMiner:
Adds `docs/experiments/glm52-4xa6000-2026-08-02.md`, following the convention of the existing 6x5090 report.

Engine v1.3.0 at commit `ecade075cfc2eae684097ea7de5570c3786ce199`, CUDA build `sm_86`, model `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`.

**Why this host is worth documenting:** the existing experiment set covers 6x5090 Blackwell, Metal M5 Max, Strix Halo and a 9950X3D+5090 box. This adds **Ampere multi-GPU on a pre-AVX-512 CPU** (EPYC 7402P, Zen 2, DDR4-2400), which behaves quite differently.

### Findings

- **`CUDA_DENSE=1` is worth x2.8** on a CPU-bound host (1.53 -> 4.26 tok/s). With the dense path on CPU, 192 GB of VRAM serve experts only while the CPU is the limiter. The flag does not appear in the public README.
- **Removing `URING`/`PILOT*` gains +26%.** Once experts are resident the disk reads 0 MB/s during decode, so the prefetch/overlap machinery only consumes the scarce resource.
- **Prefill is a different regime.** A 512-token slice already touches every expert of every layer, since `P(miss) = (1-8/256)^512 ~ 1e-7`. Chunked prefill therefore multiplies full-model sweeps instead of reducing work. Only one GPU is active during prefill.
- **KV cost.** Measured 0.316 MB/token vs 0.098 for llama.cpp `-ctk f16` on the same checkpoint (x3.2). Detailed in a separate issue.

### The part I would most like reviewed

Section 6 documents a **methodological problem affecting any A/B benchmark on this engine**: `.coli_usage` persists across restarts and grows monotonically. A configuration *identical in memory placement* measured **5.46 then 2.56 tok/s** after the learned profile re-tuned onto a different workload (240k -> 17.7M recorded selections).

`CONTRIBUTING.md` asks for a warm-up policy, but warm-up alone is insufficient: the profile needs to be snapshotted and restored per configuration. The report is explicit that its own numbers are **within-session rankings, not portable throughput**, and two hypotheses formed during the session are documented as refuted.

Suggested follow-up: expose per-tier access counts (VRAM/RAM/disk hits) via `/profile` in a scriptable form, so contributors can report the metric that actually decides placement trade-offs.

Docs-only change: no build, no test impact.

### DIFF
--- a/docs/experiments/glm52-4xa6000-2026-08-02.md
+++ b/docs/experiments/glm52-4xa6000-2026-08-02.md
@@ -10,6 +10,13 @@ experiment set does not cover (6×5090 Blackwell, Metal M5 Max, Strix Halo, 9950
 Its main contribution is not a throughput number, however: it is a **methodological finding
 that invalidates naive A/B benchmarking on this engine** (§6).
 
+> **Version and revision notice.** The §4 configuration matrix was measured on
+> `ecade075` (2026-07-30); `dev` was already 185 commits ahead. Two of its conclusions have
+> since been **retracted and replaced with controlled measurements** — the prefill I/O claim
+> in §5 and the CPU-fallback claim of issue #776. Both retractions are stated inline where
+> the original claim stood, rather than removed, because the way each was reached is the
+> useful part. §5–§8 were re-measured under the snapshot/restore protocol of §3.
+
 ---
 
 ## 1. Hardware
@@ -47,6 +54,26 @@ Median reported. RSS and VRAM sampled after load.
 > **Caveat (important):** two measurement clients running concurrently against the same
 > engine produce erratic results (0.43–1.04 tok/s) because the engine serialises requests.
 > Any concurrent measurement must be discarded.
+
+**Amended protocol, used for §5–§8.** The above is not sufficient on this engine, because
+`.coli_usage` carries state across restarts (§6). Every measurement from §5 onward adds:
+
+```sh
+cp -f "$MODEL/.coli_usage" snapshot          # once, before the campaign
+# ... then, before every single configuration:
+systemctl stop colibri && sleep 35           # VRAM is not released immediately
+cp -f snapshot "$MODEL/.coli_usage"          # restore the profile byte-for-byte
+systemctl start colibri
+```
+
+Two further rules learned the hard way, both of which produced a wrong published claim:
+
+1. **Never read cumulative log counters as current state.** Count matching lines *before* and
+   *after* the window under test and report the difference — a long-lived log file will
+   otherwise attribute months of history to the current configuration.
+2. **Never sample during a cold start.** Loading 429 GB takes minutes, during which GPU
+   utilisation is legitimately 0 % and one layer may take 45 s. A sample taken there
+   describes the loader, not the engine.
 
 ## 4. Configuration matrix (9 configs, same session, ~1 h)
 
@@ -100,9 +127,9 @@ Sampled during decode, after tuning:
 CPU↔GPU round trips, tiny matmuls at S=1 — not any single bandwidth. This explains both the
 failure of batching and the uselessness of maximising memory.
 
-### Prefill is a different regime entirely
+### Prefill is a different regime — but not for the reason stated in the first draft
 
-Decode activates 8 of 256 experts per layer; consecutive tokens reuse them, so the learned
+Decode activates 8 of 256 experts per layer and consecutive tokens reuse them, so the learned
 pinned set covers them. Prefill does not:
 
 ```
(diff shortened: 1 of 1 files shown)


ThefloorMiner:
**Version disclosure — please read before reviewing.**

The measurements in this report were taken on `ecade075cfc2eae684097ea7de5570c3786ce199` (2026-07-30). I have since checked that `dev` was **185 commits ahead** at the time of submission, which I should have done first.

I re-verified each claim against current `dev` (`fe308c9`):

| Claim in the report | Status on `dev` |
|---|---|
| `--cap` default of 8 starves the RAM expert cache | **fixed** — now `default=None` / auto (issue #765 closed by me) |
| `CUDA_EXPERT_GB=auto` leaves no prefill scratch | still reproducible (#767) |
| KV in fp32 + `kvb_all` materialisation, x3.2 vs llama.cpp | **unchanged**, verified line-by-line (#768) |
| `coli plan` computes against free VRAM | still applies (#769) |
| `COLI_PREFILL_CHUNK` absent from `ENVIRONMENT.md` | still 0 occurrences |

**What this means for the numbers.** The C0 baseline (1.53 tok/s) was partly caused by the `--cap 8` default, which no longer exists — so the headline x3.6 improvement would not reproduce as such on `dev`; part of it is now the default behaviour. The relative ranking of `CUDA_DENSE` and of removing `URING`/`PILOT*` should be unaffected, since neither touches the cap path, but I have not re-measured.

I am happy to either (a) re-run the full matrix on current `dev` before you consider merging, or (b) have you merge it with an explicit "measured at `ecade075`" banner at the top. Your call — I did not want to leave the stale-version problem undisclosed while you reviewed.

The methodological finding in §6 (`.coli_usage` persisting across restarts and invalidating cross-session A/B) is independent of the version and is the part I would most like reviewed.

ThefloorMiner:
**Second correction, on the prefill analysis in §5 of the report.**

The report states that a 512-token slice "re-reads the whole non-resident expert set" once per slice. Issue #153 measures otherwise on a Zen5 + 5090 host: expert-disk time is flat (19.4s → 21.2s) as the prompt doubles, while attention roughly doubles (49.2s → 118.0s) and is 41-51% of prefill.

So: **prefill is attention-bound, not disk-bound**, and my host was disk-visible only because its expert residency was incomplete. The probability argument I gave — `P(miss) = (1-8/256)^512 ≈ 1e-7`, so a slice touches every expert — is correct and I would keep it, but the conclusion I drew from it is not.

That makes two corrections on this PR (the other being the 185-commit version gap). I would rather flag both than have a reviewer find them. If you would prefer I rewrite §5 against current `dev` with the engine's own `PROFILE` breakdown instead of inferring from log-line rates, say so and I will.

The section I still stand behind is §6 on `.coli_usage` invalidating cross-session A/B — since writing it I re-ran three configurations with the profile snapshotted and restored, and an identical memory placement that had measured 5.46 tok/s measured 2.78 under control.

ThefloorMiner:
**Revised — the two corrections above are now in the document itself, along with the measurements that replace them.**

I had left both retractions as comments hanging off the PR. That was the wrong shape: a reviewer would have had to read the thread to know which parts of the file are still true. They are now stated inline, at the exact place the original claim stood, and the file carries a revision notice at the top.

What the revision adds, all of it under snapshot/restore of `.coli_usage`:

**§3 — the protocol that was missing.** Restore the profile byte-for-byte before every configuration. Plus the two rules that each of my wrong claims came from: never read cumulative log counters as current state (count before/after and diff), and never sample during a cold start (429 GB takes minutes, during which 0 % GPU is the loader, not the engine).

**§5 — the context curve**, `.coli_usage` restored at each level, ~1190-token prompt:

| `CTX` | prefill | decode | new OOM messages |
|---|---|---|---|
| 32,768 | 198 tok/s | 1.56 tok/s | 0 |
| 65,536 | 170 | 1.42 | 0 |
| 98,304 | 198 | 1.33 | 0 |
| 131,072 | 170 | 1.20 | 0 |
| 196,608 | 148 | 1.03 | 0 |

No cliff, and prefill never leaves the 148–198 band. `CTX=131072` buys 4× the context for −23 % decode.

**§6 — three placements under a controlled profile:** VRAM-heavy 2.81, balanced 2.78, RAM-heavy **3.51**. Moving 12 GB between the tiers changes nothing; the RAM budget is what moves the number. And configuration B is byte-for-byte the one that once measured 5.46 — under a controlled profile it gives 2.78. The difference is *specialisation*: 240 k selections from one repeated prompt then, 18.2 M from mixed use now. That is the same ×2 that motivated #780.

**§7 — I was attributing the 262144 collapse to the context.** It is a budget failure: `RAM_GB=215` drives the expert cache to its `cap=1` floor, so every one of the 8 experts a token needs becomes a disk read. At `RAM_GB=235` the sweep reaches 196,608 with no collapse.

**§8 — new: what the learned profile actually looks like**, 18,189,640 selections over 19,324 experts. It is **not a power law** — the top 1 % of experts carries 6.8 % of selections, not the 30–50 % a power law would give, the max/median ratio is 18.7, and no expert is ever unused. `noaux_tc` load-balancing is working exactly as designed, and that is precisely what limits hot/cold tiering: pinning the hottest 42.9 % of the model captures 72.1 % of accesses (×1.68), with each additional gigabyte capturing less than the last. It also says the engine's uniform per-layer cap is fine: 71.5 % against 72.1 % for a global top-k, so **0.6 points** — not worth the complexity.

Happy to split this into a separate file if you would rather keep the original as the record of what was measured on `ecade075`; my judgement was that one document that says where it was wrong is more useful than two, but it is your tree.

ThefloorMiner:
**Two more sections, and one of them makes me want to re-rank everything else in this report.**

## §5 — the engine's own per-phase counters

I had been sampling from outside (`top`, `nvidia-smi`, `iostat`) and could only conclude "nothing is saturated". The engine's own profiler attributes it. `CTX=8192`, 48 tokens, greedy:

```
expert-disk    2.345 s   ( 4.8 %)
expert-matmul  9.074 s   (18.6 %)
attention     31.264 s   (64.2 %)   including kvb 0.000 s
other          5.987 s   (12.3 %)

  of which  score-softmax-value  20.398 s   (41.9 % of total)
```

**Attention is 64 % of decode.** Expert compute is 19 %, disk is 5 %.

That reorders this report against itself. §6, §7 and §8 are all about where experts live — residency, tiering, pinning, the flatness of the routing distribution. Together those govern **under a fifth** of the time. A perfect expert placement, costing nothing, buys at most ~19 %.

It also explains §4's biggest result after the fact. `CUDA_DENSE=1` was worth ×2.8 and I could not say why beyond "it moves dense tensors to the GPU". It was the decisive lever because it was the only one aimed at the phase that actually dominates.

And it sharpens #768 in a way that supports your prioritisation there: **`kvb 0.000 s`**. During decode the absorbed path runs and `kvb_all` is never built. The 30.1 GB is a *prefill-only transient* that `cap_for_ram()` nevertheless reserves permanently at every context length.

One number in there does not make arithmetic sense and I have deliberately **not** written it up as a finding: at `CTX=8192` with a ~40-position KV cache, `score-softmax-value` moves ~7 MB per token and spends 425 ms doing it — about 17 MB/s, three orders of magnitude below this host. Either the counter absorbs waiting time, or there is a degenerate path at S=1. The section says exactly that and stops. I would rather hand you an anomaly labelled as one than a fourth claim I have to retract.

## §9 — the `DRAFT` re-test, and a correction to how I first read the result

`--auto-tier` forces `DRAFT=0` here. Re-tested under the §3 protocol:

| | mean |
|---|---|
| `DRAFT=0` | **1.70 tok/s** |
| `DRAFT=3` | 0.85 (−50 %) |
| `DRAFT=3` + `COLI_CUDA_MTP=1` | 0.96 (−44 %) |

`--auto-tier` is right. My first reading was "speculative decoding hurts, close the path" — and that was too broad, because I had not checked what `DRAFT` actually does. `colibri.c:944` calls it *"n token auto-speculati per forward via **n-gram lookup**"*, `colibri.c:6000` describes lossless self-speculation whose candidates come **from the context**, and the MTP head is only the fallback when the lookup returns nothing (`colibri.c:6093`). It is speculation by *copying*.

My prompt asks for three advantages of expert routing and generates 64 tokens of entirely novel prose. There is nothing to copy. So this measures the **structural worst case** for the technique, and the honest conclusion is the narrow one: on novel generation `DRAFT>0` costs 2× and returns nothing, disabling it by default is correct, and it remains untested on the workloads it targets — code editing, summarisation, structured output.

The harness also failed to capture the acceptance rate (`not reported` in all three conditions). That is my defect, not the engine's, and it is noted as such.

## Status of this PR

The 4× A6000 host is powered down, so the llama.cpp comparison you would want alongside §7 — and the `-c 32768` ratio question from #768 — are queued rather than done. The GGUF and the build are in place; it is a relaunch, not a rebuild.

Everything currently in the file is measured. Where it is not, it says so.

PR #770, opened 2026-08-02
fix(plan): warn when VRAM is already in use

### LINKED ISSUE #769: coli plan silently reports against free VRAM while an engine instance is running
`plan` is read-only and safe, but it computes against *currently free* VRAM. With a production instance holding 178 GB, it reports a 2.6 GB hot tier and ~29-46% projected residency — figures unrelated to the configuration being planned.

**Observed** (engine running, 178 GB VRAM in use):

```
VRAM   2.6 GB hot tier · ~122 experts · 0:RTX A6000, 1:..., 2:..., 3:...
limit  disk expert misses
hit    46% projected expert residency
```

The same command with the engine stopped yields the intended figures. Note also that VRAM is not released immediately on `systemctl stop` — a plan (or a restart) within ~30 s still sees the old allocation.

**Impact.** `plan` is the natural tool to reach for *before* changing a running deployment, and that is exactly when it misleads. Its output is the basis for capacity decisions, and nothing in the output signals that it is degraded.

**Suggested fix:** detect a running instance or non-trivial VRAM usage, and either warn loudly or plan against *total* rather than *free* VRAM.

---
*Observed on v1.3.0, commit `ecade075cfc2eae684097ea7de5570c3786ce199`, Linux, CUDA build `sm_86`, 4x RTX A6000, 251 GiB RAM, model `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`. Full context in #764.*

ThefloorMiner:
Closes #769.

`coli plan` sizes the hot tier from **free** VRAM. Run on a host where an engine instance already holds the GPUs, it reports a tiny tier and a pessimistic hit rate that describe nothing — with no indication that the output is degraded.

That is precisely when the command gets reached for: before changing a live deployment.

**Measured on a 4× A6000 host with the production engine running** (178 GB VRAM held, 91 % actual expert residency):

```
VRAM   2.6 GB hot tier · ~122 experts · 0:RTX A6000, 1:..., 2:..., 3:...
limit  disk expert misses
hit    46% projected expert residency
```

Those numbers were used, in good faith, as the basis for a capacity decision. They are simply the plan for the 2.3 GB that happened to be free.

**Change:** one warning when less than 75 % of aggregate VRAM is free, naming the figures. Verified against the real `nvidia-smi` state of that host:

```
203.8 GB of VRAM is already in use (only 2.3 GB of 206.1 GB free): this plan
plans against the remainder. Stop the running engine for a representative plan.
```

Uses the `total_bytes`/`free_bytes` already carried by `discover_gpus()`; no new probe. Threshold chosen so a few hundred MB of desktop compositor does not trip it.

Python-only, no build or engine impact.

### DIFF
--- a/c/resource_plan.py
+++ b/c/resource_plan.py
@@ -567,6 +567,20 @@ def build_plan(model, ram_gb=0, context=4096, gpu_indices=None, vram_gb=0,
         warnings.append("one or more requested GPUs were not detected")
     if gpus and vram_budget < requested_vram:
         warnings.append("VRAM tier was clamped by free VRAM or model expert size")
+    # The plan sizes the hot tier from *free* VRAM, so running it while an engine
+    # instance already holds the GPUs silently produces a tiny tier and a
+    # pessimistic hit rate that describe nothing. That is exactly when a user
+    # reaches for `coli plan` -- before changing a live deployment -- so say so
+    # rather than let the numbers be read as a capacity answer.
+    if gpus:
+        gpu_total = sum(g["total_bytes"] for g in gpus)
+        gpu_free = sum(g["free_bytes"] for g in gpus)
+        if gpu_total and gpu_free < 0.75 * gpu_total:
+            warnings.append(
+                f"{format_bytes(gpu_total - gpu_free)} of VRAM is already in use "
+                f"(only {format_bytes(gpu_free)} of {format_bytes(gpu_total)} free): "
+                "this plan plans against the remainder. Stop the running engine "
+                "for a representative plan.")
     if cold_bytes:
         warnings.append("cold expert misses may reach disk; normal decode speed depends on hit rate")
 


JustVugg:
Merging — and this is the right fix for the problem as stated.

#769 says `coli plan` reports against free VRAM **silently**. This removes the *silently* without touching the plan itself, which is correct: a read-only command given degraded input should say the input is degraded, not quietly invent a better one.

The framing in the comment is what sold it:

> That is exactly when a user reaches for `coli plan` — before changing a live deployment

A capacity answer that is wrong precisely when it is most consulted is worse than no answer, and nothing in the output distinguished the two. The 75% threshold is a judgement call but a reasonable one, and the message names both numbers so the reader can decide for themselves.

Note the CI here had never actually run — GitHub holds workflows for first-time contributors until a maintainer approves them, and nobody had. Not your doing, and now approved.

Closes #769.

Issue #767, opened 2026-08-02
CUDA_EXPERT_GB=auto leaves no prefill scratch; a real prompt permanently disables a device


ThefloorMiner:
`auto` clamps the expert tier against free VRAM minus the dense resident set and 2 GB of runtime headroom per device. That headroom is enough for short prompts but not for a realistic one: the prefill scratch allocation fails and the engine **permanently disables a tensor on that device, falling back to CPU** for the rest of the session.

**Reproduce:** `CUDA_EXPERT_GB=auto` on 4 devices, then send a ~7,000-token prompt.

**Observed**

```
[prefill] layer 33/78 · 7065 token
[CUDA] scratch allocation: out of memory
[CUDA] tensor [6144,2048] on device 3 disabled after an error; falling back to CPU
```

Throughput collapses and does not recover until restart.

**Why it is easy to miss:** benchmarking with short prompts hides this completely. Our 9-config matrix used a 21-token prompt and never triggered it; the first real conversation did.

**Workarounds found:** set `CUDA_EXPERT_GB` explicitly below `auto`, and/or set `COLI_PREFILL_CHUNK` (which is not in `ENVIRONMENT.md` — it only appears in the comment at `colibri.c:4464`) to bound the scratch.

**Suggested fix:** scale the per-device reserve with `CTX` and the expected prefill batch, or make the fallback non-permanent.

---
*Observed on v1.3.0, commit `ecade075cfc2eae684097ea7de5570c3786ce199`, Linux, CUDA build `sm_86`, 4x RTX A6000, 251 GiB RAM, model `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`. Full context in #764.*

ThefloorMiner:
**Version note.** Reported against `ecade075`; re-checked against current `dev` (`fe308c9`) — **still reproducible**: `CUDA_RESERVE_GB` (`colibri.c:503`) already existed in the version I tested and its default is still 2 GB per device, which is what runs out on a realistic prompt.

Raising `CUDA_RESERVE_GB` is a clean workaround I had missed — worth documenting alongside the failure mode, since the symptom (a device silently disabled for the rest of the session) gives no hint about which knob to reach for.

Also still true on `dev`: `COLI_PREFILL_CHUNK` has **0 occurrences** in `docs/ENVIRONMENT.md`, though it is the other remedy here.

JustVugg:
[...]

PR #771, opened 2026-08-02
docs(env): document COLI_PREFILL_CHUNK, and what CUDA_DENSE is worth

ThefloorMiner:
Two documentation gaps found while bringing GLM-5.2 up on a 4× A6000 host (#764).

**`COLI_PREFILL_CHUNK` has no entry in `ENVIRONMENT.md`.** It exists only in the comment at `colibri.c:4464`, yet it is the remedy when a long prompt exhausts CUDA scratch under `CUDA_EXPERT_GB=auto` (#767) — a failure whose symptom (`tensor [6144,2048] on device 3 disabled after an error; falling back to CPU`) gives no hint which knob to reach for.

The entry also states its cost, which is not obvious: a 512-token slice already routes to essentially every expert of every layer, since `P(miss) = (1-8/256)^512 ≈ 1e-7`. Each slice therefore re-reads the whole non-resident expert set, so smaller chunks multiply disk traffic rather than reduce work. Measured 2.6–4.3 tok/s prefill at `chunk=512` on an 8,595-token prompt.

**`CUDA_DENSE` has a one-line entry that does not say when it matters.** Off by default the engine reports `routed experts only (resident dense on CPU)`. On a CPU-bound host that leaves the dense path of every layer on the CPU while the VRAM tier serves experts only — measured **×2.8** (1.53 → 4.26 tok/s) on 4× A6000 + a 24-core EPYC. It was by a wide margin the most profitable single setting found on that host, and nothing in the docs suggested to try it.

Docs-only.

### DIFF
--- a/docs/ENVIRONMENT.md
+++ b/docs/ENVIRONMENT.md
@@ -18,6 +18,7 @@ Format: `VAR` — default — effect.
 |---|---|---|
 | `RAM_GB` | `0` (auto ≈ 88% of free RAM) | RAM budget in GB for the resident/streamed expert working set. Higher → more experts stay hot → higher cache hit rate. |
 | `CTX` | `4096` | Maximum context length (tokens) the KV cache is sized for. |
+| `COLI_PREFILL_CHUNK` | `0` (off) | Run a long prompt through the layers in N-token slices instead of one pass. Every S-scaled activation buffer shrinks from prompt-sized to chunk-sized, which is the remedy when a long prompt exhausts CUDA scratch. Byte-identical output (verified at N=256). Skipped under an active MTP draft. **Cost:** a slice of 512 tokens already routes to essentially every expert of every layer (`P(miss) = (1-topk/n_experts)^N`), so each slice re-reads the whole non-resident expert set -- prefer the largest N that still fits your scratch. |
 | `NGEN` | `256` (engine) | Max tokens to generate before stopping (stop tokens can end sooner). `coli --ngen` defaults to `1024`. |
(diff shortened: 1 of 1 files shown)


ThefloorMiner:
**Correction — I got the prefill claim wrong, and I would like to fix this PR before it is reviewed.**

The `COLI_PREFILL_CHUNK` entry I proposed states that each slice "re-reads the whole non-resident expert set". The probability part is right — a 512-token slice does route to essentially every expert of every layer — but the *re-read* conclusion is mine and it is contradicted by the measurements in #153:

| prompt | total | expert-disk | expert-matmul | attention |
|---|---|---|---|---|
| 256 | 120.9s | 19.4s | 34.6s | **49.2s** |
| 512 | 231.4s | **21.2s** | 58.3s | **118.0s** |

expert-disk is essentially flat while attention doubles: the model streams off disk **once per prefill**, and the cache retains experts across slices. My host was disk-visible only because its residency was incomplete, which is a property of that host and not of chunking.

I will amend the entry to describe what the knob actually does (bounding the S-scaled activation buffers) and drop the causal claim about disk traffic. Happy to push the amended wording, or to close this and resubmit — whichever you prefer.

Thanks to @NeuralNotwerk / the author of #153 for the profile data; it corrected me.

JustVugg:
Merging. Two lines, both filling a gap that was costing people real time.

`COLI_PREFILL_CHUNK` existed only in a comment at `colibri.c:4464` and is the remedy for #767 — an environment variable that fixes a real failure and is documented nowhere is effectively not there. And expanding `CUDA_DENSE` to say what the engine reports when it is off (`routed experts only (resident dense on CPU)`) connects a flag to the line people actually see in their logs.

**#767 stays open.** This documents the workaround, and the bug underneath — a failed prefill scratch allocation permanently disabling a tensor on that device for the rest of the session — is untouched. Written up there, along with why it is the same shape as your #776.

Thank you for writing these up while you had the context rather than moving on.

PR #773, opened 2026-08-02
feat: add DeepSeek V4 long-context inference

ZacharyZcR:
## Summary

Stacked on #772. This PR contains only the long-context allocation foundation:

- store the 128-token FlashInfer SWA cache as a ring instead of allocating by maximum context
- cap sparse-attention scratch space at the model top-k width
- accept up to 32,768 prompt + output tokens
- avoid allocating the duplicate CPU compressed cache in the all-GPU path

The branch was restacked after #772 was rebased; its delta is now limited to `c/backend_cuda_dsv4.cu` and `c/deepseek_v4.c`.

## Correctness boundary

- the exact sparse path is validated only through 2K tokens, where all compressed candidates fit in `index_topk=512`
- 32K allocation support is **not** a claim of correct 32K inference
- contexts beyond 2K remain incomplete until the checkpoint Lightning Indexer is connected

## Verification

- targeted long-context tests passed
- local CPU build passed
- the base #772 backend passed a full CUDA 13 / SM120 build on the 6x RTX 5090 host

## Status

Draft and dependent on #772. Keep it stacked until the base backend merges; then rebase and implement the FP8 Lightning Indexer query/key cache, MQA logits, top-512 selection, and real long-context A/B validation.


ZacharyZcR:
Closing this in favour of #772, as @JustVugg asked in [#772](https://github.com/JustVugg/colibri/pull/772#issuecomment-5185504066).

He is right that this is the same PR. #773 carries #772's 25 commits verbatim plus one — `ed563b4` *Add DeepSeek V4 long-context cache foundation* — over the same 46 files. Two reviewers could have read the same 6,700 lines without either noticing, and that is my fault for opening both.

#772 is the one with the review on it, so that is the one that stays. The long-context commit is not lost: it goes back up on its own once #772's shape is settled, which is not a small caveat right now — the blocking request there is an offload path, and the file layout is likely to move so the engine file belongs to #165's streaming implementation. A long-context cache built on top of the current layout would need redoing anyway.

Better to rebuild it on whatever #772 becomes than to keep a stale superset open.

Issue #748, opened 2026-08-01
[Bug]: Loading Kimi K3 layers is slow, and then Colibri stops doing anything.


brad-evony:
### Commit or version

release 1.4.0

### Environment

system: core i9 10940x 256gb ram 4tb nvme ssd, windows 10 pro 
running colibri in command prompt.

### Reproduction steps

set OMP_NUM_THREADS=14
set COLI_MODEL=E:\AI-MODELS\KIMI-K3-MXFP4
python coli web --ram 242 --gpu none --ngen 512

### Expected behavior

It is supposed to load the 93 layers within a reasonable time, then load the tokenizer, then open the browser page to begin chat functionality.

### Actual behavior and logs

It is slow to load taking 40 minutes because it only uses one CPU thread out of 14. Then after it loads the tokenizer, it stops using the CPU and SSD, and just doesn't do anything.  It never begins the chat or opens the browser window.  It doesn't give any errors it just stops doing anything.

```shell

dashboard: http://127.0.0.1:8000/  (opens automatically when the engine is ready)
[K3] loaded layer 1/93 (57.0s, RSS 1.4 GB)
[K3] loaded layer 9/93 (270.8s, RSS 4.1 GB)
[K3] loaded layer 17/93 (483.2s, RSS 6.9 GB)
[K3] loaded layer 25/93 (694.8s, RSS 9.8 GB)
[K3] loaded layer 33/93 (906.6s, RSS 12.6 GB)
[K3] loaded layer 41/93 (1117.8s, RSS 15.5 GB)
[K3] loaded layer 49/93 (1329.3s, RSS 18.4 GB)
[K3] loaded layer 57/93 (1540.7s, RSS 21.2 GB)
[K3] loaded layer 65/93 (1752.1s, RSS 24.1 GB)
[K3] loaded layer 73/93 (1963.6s, RSS 26.9 GB)
[K3] loaded layer 81/93 (2175.4s, RSS 29.8 GB)
[K3] loaded layer 89/93 (2386.9s, RSS 32.6 GB)
[K3] init done in 2538.7s | 93 layers | expert cache 4/layer (17.5 MB/slot) | RSS 35.1 GB
[K3] tokenizer.json loaded (family=kimi)
```
Pressing CTRL-C gives the following message:

Traceback (most recent call last):
  File "E:\AIPROGS2\colibri-v1.4.0-windows-x86_64\coli", line 1249, in <module>
    main()
[...]

Issue #195, opened 2026-07-14
run_serve_mux on native Windows (#177): health passes but generations never dispatch — request hangs, TTL unloads mid-request


juicyroots:
### What happened

Follow-up to #189 (thanks for the fast #177!): on native Windows, the #177 mux path gets further than before — the engine loads, the gateway binds, `/health` passes — but **generation requests never dispatch**. A `POST /v1/chat/completions` hangs indefinitely (killed it at 700 s); from llama-swap's perspective the model is idle, so its TTL fired and unloaded the model **mid-request**, leaving an orphaned `glm.exe`.

Setup: Windows 11, MinGW GCC 16.1, `main` @ 37ffb61 (+ the #158 aligned-free fix applied locally, needed for `CUDA_RELEASE_HOST` on Windows), GLM-5.2 744B int4, dual-GPU expert tier (this same rig/config as #162/#163/#189).

### Evidence it's the mux dispatch, not the environment

Control experiment through **identical plumbing** (same llama-swap entry, same job-object wrapper, same env, same engine binary): pinning the pre-#68 gateway (`c/coli` + `c/openai_server.py` @ 1bdaeee, which drives the legacy single-sequence `run_serve` protocol) works end-to-end — 99.7 s cold spin-up to a completed 48-token answer. Only the gateway/protocol differs.

So the engine-side stdin polling may be fine — the hang could equally be in the gateway↔mux handshake on Windows (request framing written but never read, or the response reader blocking). I didn't bisect further yet.

### Repro

```
# stock main on Windows, any model:
python coli serve --model <snap> --port 9070    # health passes
curl -d '{"model":"...","messages":[{"role":"user","content":"hi"}],"max_tokens":4}' \
     http://127.0.0.1:9070/v1/chat/completions   # hangs forever
```

Happy to run instrumented builds or bisects on this rig — turnaround is fast (warm cache, ~60 s per run).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

JustVugg:
Fixed in `main` (`2ead86a`) — and your control experiment pointed exactly at the right layer. Two Windows-only bugs were stacked in the #177 mux:

1. **`run_serve_mux` never set `_setmode(_O_BINARY)`** (legacy `run_serve` does, from #131 — the mux didn't inherit it). In TEXT mode the CRT collapses CRLF inside the `fread()` payload — the engine waits forever for bytes that will never arrive — and expands LF in the READY/STAT sentinels on the way out.
[...]

PR #775, opened 2026-08-02
feature: Hy3 support 

ErikTromp:
## Summary

I had the Hy3 support for Colibri already working for quite a while but now there is more focus on supporting different models from the main repo as well, figure a pull request would make sense. I made sure to rewrite anything off that I did before generalization was in place in Colibri to work in a way that consolidates with the current codebase

## Validation

- [X] `make -C c check`
- [X] CUDA changes were tested with `make -C c cuda-test` 
- [X] Performance claims include hardware, commands, and repeatable measurements

## Compatibility


JustVugg:
Reviewing properly takes longer than this note, but one thing is worth telling you now rather than after a user reports it.

**`hy3.c` emits the `READY` sentinel but never calls `coli_serve_binary_mode()`.**

On Windows the CRT opens stdout in text mode and rewrites the trailing `\n` as `\r\n`, so the engine sends `\x01\x01READY\x01\x01\r\n`, `coli` — which matches the bytes exactly — never recognises it, and the session hangs forever with **no error at all**. That is #748: a user loaded 93 Kimi layers over 42 minutes and then sat watching nothing happen.

It is one line, from `compat.h`, before the sentinel:

```c
static void serve_loop(...){
    coli_serve_binary_mode();   /* #748: TEXT-mode stdout mangles the READY sentinel */
    ...
    fputs("\x01\x01READY\x01\x01\n", stdout);
```

`colibri.c` has had it since #195; `inkling.c` and `kimi_k3.c` were written without it and it went unnoticed for months, which is exactly why it now lives in `compat.h` instead of being pasted per engine. Yours is the fourth engine to be written without it, so this is on our onboarding rather than on you — but you get to have it for free instead of finding out from a bug report.

**Credit where it is due:** your `coli` wiring is the most complete of any new engine we have received. `model_arch` recognises `hy_v3`, the banner is arch-aware, and the serve protocol is there. #712 (Qwen3.6) has none of that and is blocked on it, so you have already done the part that usually gets missed.

The larger question — whether we take a fifth engine right now — is the maintainer’s and is being discussed separately. Nothing about that reflects on the work; four engines already multiply every maintenance problem across four files, and this week produced two Windows bugs that had to be fixed in three of them at once.

ErikTromp:
Good catch on the ready emit, will fix it

ErikTromp:
Fyi sentinel issue is fixed

PR #777, opened 2026-08-02
mgua@tomware.it added a small clarification for windows invocation fr…

mgua:
A small documentation change to clarify invocation line on windows, from powershell

## Summary

the coli invocation on windows may not be automatic thru the python interpreter. (at least it is not in my case), probably because coli filename is without extension. 
I added a couple of line explanation, and an invocation relying on py launcher.

## Validation

does not apply. this is just a documentation fix

## Compatibility

does not apply. this is just a documentation fix

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -86,6 +86,12 @@ One setup step: **install Python 3** from
 API gateway are Python scripts (the engine itself is pure C and needs nothing).
 No renaming, no configuration: the launcher finds `colibri.exe` next to itself.
 
+For better understanding, from powershell prompt, a complete invocation line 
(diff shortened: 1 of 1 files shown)


JustVugg:
Retargeted this to `dev` — everything lands there first and `main` is release-only, so a PR opened against `main` gets blocked by the branch ruleset rather than reviewed. Nothing wrong on your side; the contributing docs could say it more loudly, and that is now on my list.

The fix itself is right, and it is the kind of papercut that only gets found by someone actually typing the commands on Windows: `coli` has no extension, so PowerShell will not run it through the interpreter on its own. `python coli ...` is the correct form and the docs assumed it was obvious.

Thanks for taking the time to write it up rather than just working around it locally.

Issue #718, opened 2026-07-30
Zen3 data: SMT collapse (2.3× at 16 threads), ±15% run variance, PIN_GB starves LRU — HIP/gfx1100 + NVMe


dlewis7444:
## Summary

Benchmarking colibrì v1.3.0 with GLM-5.2 (int4, `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`) on a Ryzen 9 5950X + RX 7900 XTX (HIP build) produced three findings that may interest other users: a large SMT-related performance collapse, surprisingly high run-to-run variance, and a `PIN_GB`/LRU interaction that silently starves the expert cache.

First, the good news: the HIP backend works well on gfx1100 — `hip-test` passes, and the VRAM expert tier + `COLI_CUDA_ATTN=1` are solid. `URING=1` was worth +29% decode on NVMe. Overall we went from ~0.13 tok/s (cold, SATA SSD) to ~0.45 tok/s median (NVMe + tuned config). Thank you for the remarkably hackable engine.

## Environment

- CPU: AMD Ryzen 9 5950X (16C/32T, Zen3) · RAM: 62.7 GiB DDR4
- GPU: RX 7900 XTX 24 GiB (gfx1100), Fedora 43 ROCm 6.4.2 packages, `make -C c glm HIP=1 HIP_ARCH=gfx1100`
- Storage: Gen4 NVMe (~4.4 GB/s measured); earlier runs on SATA SSD (~0.5 GB/s) for comparison
- colibrì v1.3.0, GLM-5.2 int4 g64 container, `COLI_CUDA=1 CUDA_EXPERT_GB=20`
- Bench: fixed 24-token prompt, 64-token decode, `PROF=1`, engine otherwise at defaults

## Finding 1: SMT halves decode on Zen3 — `OMP_NUM_THREADS=<physical cores>` was worth 2.3×

With all else identical (config: `URING=1 RAM_GB=55 PIN=auto PIN_FILL=1 PIN_GB=15 CUDA_RELEASE_HOST=1`):

| Threads | Prefill | Decode | Time shares |
|---|---|---|---|
| 32 (SMT default) | 47 s | 0.21 tok/s | matmul 63% |
| 16 (physical) | 30 s | **0.49 tok/s** | matmul 18% |

[...]

PR #786, opened 2026-08-02
feat(inkling): reuse the attention state a previous turn already built

JustVugg:
A chat client resends the whole transcript every turn. `colibri.c` has pinned each conversation to a KV slot since #639, so its turn N prefills only the new text. **`inkling.c`, `kimi_k3.c` and the DeepSeek engine re-process turns 1..N-1 from scratch**, so the cost of a message grows with the length of the conversation — and every replayed position pulls its experts off disk again, which on a streaming engine is the dominant cost.

Nobody reported this because it does not present as a bug. It presents as *"it gets slower the longer we talk"*, and on an engine that streams a 469 GB model from disk, that reads as normal.

| | KV prefix reuse |
|---|---|
| `colibri.c` (GLM) | ✅ #639 |
| `inkling.c` | ❌ → **this PR** |
| `kimi_k3.c` | ❌ → next PR |
| DeepSeek engine | ❌ → local |

It was found by auditing what `colibri.c` has that the others do not — the same shape as **#707** (the Apple OMP exclusion landed in `inkling.c` and never reached `colibri.c`) and **#718** (physical-core sizing added to `kimi_k3.c` and `olmoe.c` and not to `colibri.c`).

## The idea

After a turn the state covers some number of positions. If the next prompt **begins with the token sequence that produced them**, that state already *is* the state at that position: skip the reset and prefill only the tail.

No snapshot, no rewind. A prompt that diverges anywhere starts over. **Either the reused positions are token-identical, or nothing is reused** — the emitted tokens are the same in both cases.

## Why a record and not a counter

Deriving the reusable length from the caller's bookkeeping (`prompt_count + generated - 1`) is tempting and wrong. That invariant differs per engine — whether the last sampled token was fed back, whether a chunked prefill ran to completion, whether generation stopped early — and **getting it wrong does not crash**. It answers the user from an attention state built out of a different conversation, and the reply looks entirely plausible.

So `kv_prefix.h` records the ids **where they are fed**, and that record is the only description of the state anyone consults. The failure mode stops being expressible.

**Audio taints the record.** Every Inkling audio frame carries the same token id (`c->audio_tok`) while the mel payload differs, so an id-only comparison would happily match two different clips. A state that consumed audio — or a request that brings its own — is never reused.

`kv_prefix.h` is shared rather than inlined per engine: one implementation to review, testable without a checkpoint, and ready for the two engines that follow.

## Measured

On the **DeepSeek V4 engine**, where this mechanism landed first (local work, not in this PR):

```
reuse            45 of 55 prompt tokens (82%)
turn 2 cached    61.2s
turn 2 cold     320.0s
speedup         5.23x
identical       YES
```

The speedup tracks the ratio of positions still to prefill, so it grows with conversation length.

**Full disclosure on this PR specifically:** the equivalent end-to-end identity run on the 469 GB Inkling checkpoint is still going on my box — two engine loads at once put the machine into OOM twice, which is a hardware limit here, not a code path. I will post the numbers under this PR when it lands. The correctness argument does not rest on that run: it rests on the structure above and on `tests/test_kv_prefix.c`.

## Tests

`tests/test_kv_prefix.c` covers every rejection rule, including the ones that look paranoid:

- divergence at the first token
- **divergence at the last recorded token** — the boundary an off-by-one `memcmp` length would let through
- a prompt **shorter than** the record (the state cannot be rewound)
- a prompt **equal to** the record (nothing left to prefill, so no final hidden state to sample)
- a write past the end of the buffer — drops the record rather than truncating it, because claiming coverage the state does not have is the one failure that answers wrongly
- taint, and a NULL record

It builds without a checkpoint, so CI carries it on every platform. `make check` picks it up automatically via the derived `TEST_BINS` (#733).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,15 @@ jobs:
           # gate fails the moment a shared-code change regresses the engine.
           python3 tools/make_tiny_inkling.py tiny_inkling
           SNAP=tiny_inkling ./inkling 8 0 tiny_inkling/ref_inkling.json
+      - name: KV prefix reuse is token-identical (serve, two turns)
+        run: |
+          cd c
+          # Reusing a previous turn's attention state must change the timing and
+          # nothing else. A wrong reuse length does not crash -- it answers from
+          # another conversation's state and still reads plausibly -- so this is
+          # the only gate that would catch it. Needs the fixture above, which is
+          # why it runs here and not in `make check`.
+          INKLING_TINY=tiny_inkling python3 -m unittest -v tests.test_inkling_prefix_serve
 
   engine-hip-syntax:
     name: HIP syntax check
--- a/c/Makefile
+++ b/c/Makefile
@@ -743,3 +743,6 @@ clean:
 bench: iobench$(EXE)
 	@if [ -n "$(ARGS)" ]; then ./iobench$(EXE) $(ARGS); else echo "built iobench$(EXE) — run: ./iobench$(EXE) <file> <MB> <iters> <threads> <direct 0|1>"; fi
 .PHONY: all colibri glm iq3 rans fuzz-rans cuda-test hip-test gpu-compile cuda-bench cuda-dll portable test-c test-python test check clean install uninstall bench
+
+tests/test_kv_prefix$(EXE): tests/test_kv_prefix.c kv_prefix.h
+	$(CC) $(CFLAGS) tests/test_kv_prefix.c -o tests/test_kv_prefix$(EXE) $(LDFLAGS)
--- a/c/inkling.c
+++ b/c/inkling.c
@@ -1451,19 +1451,44 @@ static void state_reset(Model *m) {
 static void kv_alloc(Model *m, int max_t) {
     Cfg *c = &m->c;
     if (m->K && max_t <= m->max_t) return;   /* reuse across prompts when big enough */
-    if (m->K) for (int i = 0; i < c->n_layers; i++) { free(m->K[i]); free(m->V[i]); }
-    free(m->K); free(m->V);
-    /* The record is sized with the KV it describes: growing these buffers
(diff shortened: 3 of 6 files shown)


JustVugg:
**Update: the end-to-end identity test now runs in CI**, so this PR no longer rests on a measurement I could not take.

## The test that matters

`tests/test_inkling_prefix_serve.py`, wired into the existing Inkling-oracle job:

- two turns on one engine, the second reusing the first's state, must emit the **same tokens** as that second turn alone on a cold engine
- a **diverging** prompt must not be contaminated by the turn before it

It runs against the tiny random-init fixture that job already builds, so it needs no checkpoint and no fast disk.

## Why it is a fixture and not a benchmark

I tried to take the real number on the 469 GB checkpoint and could not, and the reason is worth recording because it is a fact about the model, not about this change:

```
16,384 experts x ~28 MB          = no useful cache in 24 GB of RAM
one token, minimal prompt        = 79 GB read in 9 minutes, unfinished
cap=8 (the practical maximum)    TIERS 0 0 16384 0.00 0.00
cap=64 would need                117 GB
```

I built a pristine `origin/dev` inkling and it hit exactly the same wall, at the same point, so the behaviour is the hardware and not this PR.

## Why this gate has to exist

A wrong reuse length **does not crash**. It answers the user from an attention state that belongs to a different conversation, and the reply still reads plausibly. Nothing else in the tree would notice — which is also why this defect survived in three engines without a single report.

So the evidence now stands at:

| | |
|---|---|
| decision logic, every rejection rule | `tests/test_kv_prefix.c` — no model needed |
| token-identity, end to end | `tests/test_inkling_prefix_serve.py` — CI fixture |
| speed, on hardware that can show it | DeepSeek V4: **5.23x**, 82% of prompt reused, output byte-identical |

Issue #759, opened 2026-08-01
[Bug]: RAM_GB is not clamped to the #653-corrected MemAvailable on unified memory; CAP_RAISE then inflates against it and the run is OOM-killed mid-prefill


mohamedmastouri2000-boop:
## Summary

On unified-memory hosts the engine computes the correct post-tier memory figure — the `#653` correction — and then does not use it. An explicit `RAM_GB` larger than that figure is accepted without complaint, and `CAP_RAISE` inflates the per-layer cap against the unclamped budget. The result is a plan that exceeds physical memory and a process killed mid-run.

This is distinct from #653 and #687, which are about `auto` sizing. Here the user supplied an explicit value and the engine had already computed the number that would have rejected it.

## Evidence

From a GB10 / DGX Spark, 128 GB unified, `sm_121`, v1.3.0, reported by @RoiArthurB in [VincentMarquez/glm52-gb10-colibri#1](https://github.com/VincentMarquez/glm52-gb10-colibri/issues/1):

```
[CUDA] device 0: NVIDIA GB10, 130.7 GB VRAM, sm_121
[CUDA] hot expert tier: 3061/5007 experts, VRAM 65.00 GB (budget 65.0 GB, reserve 2.0 GB)
[CUDA] integrated/unified memory: expert tier shares physical RAM;
       RAM budget snapshot reduced by 65.00 GB -> MemAvailable=59.0 GB (#653)
[RAM_GB=100.0] cap raised 8->25: budget allows it (projected peak 98.5 GB; set CAP_RAISE=0 to disable)
[PIN] placement: 3061 VRAM + 1946 RAM expert (41.3 GB warm)
loaded in 11.35s | resident dense: 10604.26 MB
prompt: 21 tokens | generating up to 48
[prefill] layer 1/78 ... [prefill] layer 33/78
<process ends; no further output, no completion>
```

The engine states `MemAvailable=59.0 GB`, then accepts `RAM_GB=100.0`, then raises the cap to a **98.5 GB projected peak**. 65 GB of tier plus 98.5 GB of RAM budget is **163.5 GB planned on a 128 GB machine**. The run dies at prefill layer 33 of 78.

A second run on the same box with `CUDA_EXPERT_GB=90 RAM_GB=auto PIN_GB=all` died even earlier — after `MemAvailable=33.9 GB (#653)`, with no `[PIN] placement` line at all.

[...]

Issue #783, opened 2026-08-02
[Bug]: Nvidia GPU's not working with Kimi K3


Limalski:
### Commit or version

1.4.0

### Environment

Debian 13.6, Ryzen 5 5600G, GeForce RTX 4070 Ti &  RTX 5060 Ti (connected to oculink dock). Kimi K3 on 3 different nvme disk, files symlinked.
I have successfully used Colibri with GLM model, Nvidia working, both GPU's utilized. 

### Reproduction steps

git clone https://github.com/JustVugg/colibri
cd colibri/c

compiling:
CUDA_HOME=/usr/local/cuda-13.2 \
PATH=/usr/local/cuda-13.2/bin:$PATH \
make -C . kimi_k3 CUDA=1 CUDA_ARCH="sm_89 sm_90" -j

running:
COLI_CUDA=1 COLI_GPUS=0,1 CUDA_EXPERT_GB=auto CUDA_RELEASE_HOST=1 PIN=auto PILOT=1 PIPE=2 ./coli --ctx 4096 chat --model /kimi

### Expected behavior

Expect NVRAM to fill.
With GLM, VRAM utilization can be seen with nvtop.

### Actual behavior and logs

```shell
Nvtop does not even list the whole kimi_k3 process. 
While running with Kimi model, same VRAM utilization can't be seen with nvtop.
Nvidia GPU's are not used, no error produced either.

doctor:
COLI_CUDA=1 COLI_GPUS=0,1 CUDA_EXPERT_GB=auto CUDA_RELEASE_HOST=1 PIN=auto PILOT=1 PIPE=2 ./c/coli --ctx 4096 doctor --model /kimi
colibri doctor · /kimi
[  ok] model.path         model directory is readable
[  ok] model.config       config.json is valid
[  ok] model.tokenizer    tokenizer.json found
[  ok] storage.persistence model directory can store usage and KV state
[fail] engine.binary      engine is not built
[fail] accelerator.cuda   one or more requested GPUs were not detected
[  ok] model.shards       safetensors headers are valid
[  ok] storage.disk       model backing store is available
[fail] memory.ram         RAM budget cannot hold one expert slot per sparse layer
[warn] placement.plan     RAM budget cannot hold one expert slot per sparse layer; one or more requested GPUs were not detected
[skip] storage.ssd_probe  no cached probe yet; measured on the first Metal+darwin engine start

policy quality · quality-preserving yes
model  96 shards · 1560.9 GB
[...]

PR #788, opened 2026-08-02
feat(windows): enable native HIP with deterministic runtime binding

Kenneth-Javier:
## Summary

Adds a native Windows HIP path so AMD GPUs are reachable from the Windows build:

- **`HIP_DLL=1`** builds `c/coli_hip.dll` from the existing
  `c/backend_cuda.cu` and existing `coli_cuda_*` ABI. No new backend abstraction
  is introduced, and the Linux HIP path remains unchanged.
- The host loads that DLL **from its own directory by absolute path**, using the
  same seam as the existing Windows CUDA split.
- **Deterministic, fail-closed HIP runtime binding.**
  `COLI_HIP_RUNTIME_DIR` names the directory containing `amdhip64_7.dll`; the
  loader refuses to start the GPU tier if a different runtime is already
  mapped, if more than one matching runtime is mapped, or if it cannot establish
  physical file identity.
- **78 no-GPU loader-contract tests**, driven by synthetic stub DLLs, now run on
  Windows CI.
- The path was validated on physical `gfx1151` hardware.
- Teacher forcing was **token-exact against the CPU path at the token-ID level**:
  both CPU runs and both hybrid-HIP runs selected the same predicted token at
  every one of the 32 checked positions.
- CPU and hybrid-HIP free decoding also produced the same exact 20-token
  autoregressive sequence across two independent processes per mode.

## Why

`HIP=1` is Linux-only and is refused elsewhere by an early Makefile error. On
Windows, MinGW GCC cannot compile `.cu` files, while `hipcc` targets the MSVC
ABI. There was therefore no native route from the Windows build to an AMD GPU;
the Windows host could select only the CUDA DLL.

The DLL split alone is insufficient because AMD ships a **versioned** runtime,
and Windows resolves an imported DLL basename against modules already loaded in
the process. A machine may contain several copies of `amdhip64_7.dll`, such as a
system-wide ROCm installation, an SDK runtime, an unpacked development build, or
a copy in a system search location.

Silently binding to an ambient `System32` or unrelated ROCm runtime is not an
acceptable failure mode for a GPU backend. This implementation therefore
selects the runtime explicitly and verifies the resulting module identity.

### Native Windows HIP build

`HIP_DLL=1` is opt-in. The default build remains CPU-only.

The HIP DLL build supports:

- `HIP_DLL=1`
- an explicit `HIP_ARCH`
- `HIP_SDK_ROOT`
- `HIP_BIN_DIR`
- `HIP_INCLUDE_DIR`
- `HIP_LIB_DIR`
- `HIP_DEVICE_LIB_PATH`
- `HIPCC`

`HIP_SDK_ROOT` may default from `HIP_PATH`, but every major SDK component can be
overridden independently because packaged SDK layouts do not necessarily keep
runtime, development, LLVM, and device-library files under one common root.

`HIP_ARCH` must be explicit on native Windows because
`rocm_agent_enumerator` is not available there.

`CUDA_DLL=1` and `HIP_DLL=1` are mutually exclusive.

The existing Linux HIP and Windows CUDA build paths remain supported.

### Backend selection

A Windows host built with `HIP_DLL=1` selects `coli_hip.dll` rather than
`coli_cuda.dll`.

The backend DLL is loaded from the host executable's directory by absolute
path. The existing `coli_cuda_*` ABI remains the host/backend seam, so no
parallel HIP-specific backend interface is introduced.

### Deterministic runtime binding

For HIP DLL mode:

- `COLI_HIP_RUNTIME_DIR` identifies the directory containing
  `amdhip64_7.dll`.
- The configured path must be absolute.
- Existing same-basename runtime modules are enumerated before initialization.
- Runtime modules are compared by **physical file identity**, using volume and
  file index rather than path text alone.
- An alternate spelling or hard link to the same physical file is accepted.
- A different physical `amdhip64_7.dll` is rejected.
- More than one mapped same-basename runtime is rejected, even if one is the
  configured runtime.
- Failure to enumerate modules or establish file identity is treated as an
  initialization failure.
- The configured runtime is loaded by absolute path.
- Runtime identity is verified again after the backend is loaded.
- Backend initialization fails closed if any binding condition is violated.

| Situation | Behaviour |
|---|---|
| A different `amdhip64_7.dll` is already loaded | Refuse |
| More than one same-basename runtime is loaded | Refuse |
| Module inventory cannot be read | Refuse |
| Physical identity cannot be established | Refuse |
| Configured path is relative or drive-relative | Refuse |
| Configured runtime file is absent | Refuse |
| Exactly one matching configured runtime is loaded | Proceed |

There is no bare-name fallback and no implicit `System32` runtime search.

The runtime is verified **after** it is mapped. `LoadLibraryExW` accepts a path,
not an already-open file handle, so a file replacement between preliminary
validation and loading is detected afterward rather than prevented beforehand.

`CUDA_DLL=1` hosts ignore `COLI_HIP_RUNTIME_DIR` and preserve their existing
`coli_cuda.dll` behaviour.

### Diagnostics

Windows backend-load failures now distinguish relevant failure classes,
including:

- backend file not found
- missing transitive dependency
- invalid or incompatible image
- configured HIP runtime unavailable
- conflicting runtime already loaded
- duplicate same-basename runtimes
- runtime identity mismatch

This replaces a generic failure message that could not reliably identify what
had gone wrong.

### Tests

`c/tests/test_backend_loader.py` is the dedicated owner for the Windows
host/backend loading contracts.

Its synthetic DLL fixtures cover:

- CUDA and HIP host selection
- explicit HIP runtime-directory requirements
- absolute-path validation
- runtime and backend path selection
- wrong-runtime rejection
- duplicate same-basename runtime detection
- hard-link and physical-identity behaviour
- Unicode paths
- paths containing spaces
- missing backend files
- missing dependencies
- bad images
- truthful diagnostics
- same-process binding behaviour
- controlled initialization failures
- shutdown behaviour
- isolation from real GPU hardware and installed runtimes

The suite does not require ROCm, TheRock, CUDA, a physical GPU, or a real
`amdhip64_7.dll`.

### Static and regression validation

| Suite | Result |
|---|---|
| `c/tests/test_backend_loader.py` | **78 passed**, 0 failures, 0 errors, 0 skips |
| Full Python suite (`unittest discover -s tests`) | **344 passed**, 0 failures, 0 errors, 37 skipped |
| Full C suite before the later upstream rebase | **37 binaries**, 0 compile errors, runner exit 0 |
| Full C suite after the later upstream rebase | **39 binaries**, 0 compile errors, runner exit 0 |

The original 37 Python skips were environmental, including unavailable optional
Python packages, binaries not built for that test run, and POSIX-only script
coverage. None was loader-, HIP-, or backend-contract-related.

### Physical Windows HIP validation

Validated configuration:

- AMD Ryzen AI Max+ 395
- AMD Radeon(TM) 8060S Graphics
- runtime-reported architecture: `gfx1151`
- TheRock HIP runtime: 7.14.60850
- Visual Studio 2022 MSVC: 14.44.35207
- Windows SDK: 10.0.26100.0

Measured results:

- Zero matching HIP runtime modules before initialization.
- Exactly one matching runtime afterward, at the explicitly configured
  `C:\TheRock\build\bin\amdhip64_7.dll`.
- Loaded runtime physical identity matched the configured file.
- Separate `System32` and ROCm 7.1 copies on the same machine were not selected.
- The runtime module count returned to zero after explicit probe shutdown.
- The physical Radeon 8060S was reported as `gfx1151`.
- A real `quant_matmul` kernel was executed through the production
  host/backend seam.
- Two deterministic matrix cases both had **maximum absolute error 0** against
  the CPU reference.
- CPU teacher forcing scored **32/32** in two independent processes.
- Hybrid Windows HIP teacher forcing scored **32/32** in two independent
  processes.
- Teacher forcing was **token-exact between the CPU and hybrid-HIP paths at the
  token-ID level**.
- The predicted token ID matched position-by-position at all 32 checked
  positions in both CPU runs and both hybrid-HIP runs.
- All four ordered prediction sets were identical, with empty mismatch lists.
- This was an exact comparison of selected token decisions, not a close or
  tolerance-based comparison.
- The teacher-forcing result does **not** claim bit-identical logits.
- Two CPU and two hybrid-HIP free-decode processes produced the same exact
  20-token sequence.
- Prefix equality was independently checked at 1, 2, 4, 8, 16, and 20
  positions.
- **46 dense tensors were resident on HIP** in every hybrid model process.
- Routed experts remained on CPU.
- No disabled-tensor fallback or HIP-refusal diagnostic occurred in the
  qualified hybrid runs.
- Exactly one matching HIP runtime was observed in each hybrid process.

The model workload was a small synthetic GLM MoE:

- model type: `glm_moe_dsa`
- 5 layers
- 8 routed experts
- top-2 routing
- 140 tensors
- generated offline from a fixed seed
- no model-weight download required

The model is intended as a deterministic correctness fixture rather than a
production-model claim.

### Post-rebase validation

The physical validation above was performed while the branch was based on
upstream `dev` at `72ddb673`.

Upstream subsequently advanced through `44d7958` to `7fb1159`, and this branch
was rebased onto the newer base.

The first rebase required one manual conflict resolution in `c/Makefile`:

- upstream's new `tests/test_kv_prefix` target was preserved
- this branch's `hip-dll` `.PHONY` entry was preserved

The resolution was the direct union of the two changes. No semantic alternative
had to be selected.

A later rebase from `44d7958` to `7fb1159` was conflict-free.

The first hosted Windows CI run also exposed two test-harness failures caused by
mixed path-separator conventions in the same process:

```text
.../empty dir\amdhip64_7.dll
.../empty dir/amdhip64_7.dll
```

The loader continues to emit the correct native Windows `\` separator.

The fix was limited to the test and CI harness:

- both path values are normalized before comparison
- the loader was not changed
- the earlier `os.sep` CI assertion was removed because it encoded an incomplete
  diagnosis
- the two security-relevant missing-runtime and identity assertions remain
  strict about path identity rather than textual separator spelling

The resulting branch head is:

```text
1f76e498cc346a0c324a946711ea799ef953250a
```

on upstream base:

```text
7fb11595245b6f320b3cfef9565fdbe6da92442b
```

Post-rebase and post-fix local validation completed with:

- backend-loader contract suite: **78/78**, 0 skips
- loader suite exercised under both tested separator environments: **78/78**
  in each
- full C suite: **39 test binaries**, runner exit 0
- clean merge-tree against the selected `dev` base
- 15 commits
- 0 merge commits

The physical `gfx1151` model validation was **not rerun** on head `1f76e498`.

No W1-authored production patch changed after the physical validation. The later
change was limited to test path normalization and the corresponding CI
diagnostic logic, while newer upstream production changes were incorporated
through the rebase.

## Evidence provenance

Three harness attempts preceded the authoritative model sweep and failed
**before producing qualification data**:

- an empty model-path variable
- a result-echo modification aimed at the wrong launcher
- incorrect quoting generated between Python and PowerShell

None produced an authoritative process-result row, and their temporary logs
were not retained.

The retained qualification data consists of exactly eight final model
processes in one contiguous, correctly ordered sweep:

- CPU teacher forcing ×2
- hybrid HIP teacher forcing ×2
- CPU free decode ×2
- hybrid HIP free decode ×2

Every reported model comparison derives only from those eight final processes.
This is an evidence-provenance limitation, not a runtime or model failure.

## CI boundary

The existing `windows-latest` leg of `engines-all-platforms` runs the synthetic
backend-loader contract suite.

The Windows runner may expose both separator conventions within the same
process: a temporary root may use forward slashes while the native loader
appends the runtime basename with a backslash.

The loader continues to emit the correct Windows separator. The tests normalize
both values before comparison instead of requiring identical textual spelling.

The CI step:

- uses `shell: pwsh`
- makes the required MSYS2 UCRT64 compiler tools available on `PATH`
- also includes the MSYS2 `usr\bin` tools required by the fixture build
- reports the active Python executable and relevant path environment
- removes GPU and runtime selection variables
- fails on any test failure or error
- also fails if any loader test is skipped

The skip check is important because an incomplete fixture toolchain can
otherwise cause a test class to skip while `unittest` still reports an overall
successful exit.

**This PR does not add hosted physical Windows HIP coverage.** The existing
hosted CI configuration has neither the validated Windows HIP SDK setup nor a
physical AMD GPU runner.

`engine-hip-syntax` therefore continues to cover the Linux HIP compile path,
while the Radeon 8060S / `gfx1151` results above were produced externally on
physical hardware and are reported as validation evidence rather than hosted CI
coverage.

## Limitations

- **This is not full-GPU MoE inference.** Routed experts remained on CPU.
- Routed experts become GPU-eligible only through the existing expert-placement
  controls, including `CUDA_EXPERT_GB` and an appropriate pin or usage source.
  That configuration was not validated here.
- Validation used a **small synthetic GLM fixture**, not a production model.
- Teacher-forcing equivalence is a claim of identical selected token IDs at all
  32 checked positions, not bit-identical logits.
- The canonical oracle path does not provide arbitrary generation-length
  control.
- No 32-, 64-, or 100-token free-decode claim is made. The validated
  autoregressive decode length is 20 positions.
- The decode path was exercised recursively for 20 positions without
  divergence. This is not presented as a complete isolated KV-cache proof.
- Normal one-shot model exits rely on Windows process teardown rather than an
  explicit `coli_cuda_shutdown` call.
- That lifecycle behaviour is shared with the existing host path and is not
  presented as explicit backend shutdown.
- Per-tensor upload failure may fall back to CPU while the command still exits
  successfully.
- A run may print a `[CUDA] device 0: ...` line while having no useful GPU tensor
  residency.
- `[CUDA] resident set: N tensors` with `N > 0` is the useful-work signal; the
  device line alone is insufficient.
- No performance, throughput, latency, or memory-scaling claims are made.
- Validation covered one GPU, one HIP SDK configuration, and one MSVC toolchain.
- In the tested configuration, MSVC **14.51.36231** failed because its `<cmath>`
  declarations collided with HIP's `__device__` overloads.
- The validated build therefore pins `VCToolsVersion=14.44.35207`.
- This does not claim that every future HIP/MSVC combination based on 14.51
  will fail.
- The physical model results were produced before the branch was rebased onto
  `dev` at `7fb1159`.
- Post-rebase validation covered the loader contracts and complete C suite, but
  the physical model workload was not rerun on current head `1f76e498`.
- No compatibility claim is made for upstream commits after the selected `dev`
  base used by the current branch.

## Manual operator test

No separate manual end-user/operator walkthrough was performed. The
application was nevertheless executed repeatedly through automated full-host
validation on the physical Windows machine, including model loading, teacher
forcing, and free autoregressive decoding.

## Notes for reviewers

The series contains fifteen commits and keeps them separate intentionally:

- Eleven commits introduce the build support, host/backend selection,
  deterministic runtime binding, diagnostics, synthetic fixtures, loader
  contracts, and supporting tooling incrementally.
- Three commits are documentation and CI-only:
  - `GPU_BACKENDS.md`
  - `docs/windows.md`
  - `.github/workflows/ci.yml`
- The fifteenth commit is the focused test/CI correction prompted by the first
  hosted Windows run:
  - normalize the two affected path comparisons
  - remove the incorrect `os.sep` CI guard
  - leave the production loader unchanged

No W1-authored production source changed after the physical hardware and
full-model validation.

Two PRs touch adjacent areas and may be useful context for maintainers:

- **#670** covers related Windows silent-CPU-fallback behaviour.
- **#399** also touches `c/backend_loader.c`.

Neither implementation was contained in the original selected base for this
branch.

Issue **#759** is relevant background for integrated-GPU owners. HIP already
reports `integrated=1` for the validated device, so this change does not alter
that classification.


JustVugg:
**CI approved and running** — it was stuck in `action_required`, which GitHub applies to a first contribution from a fork. Nothing you did; sorry it sat.

Read the description properly and there is a lot here I like, so let me say what stands out before the review itself.

**You reused the existing seam instead of inventing one.** `HIP_DLL=1` builds from the same `backend_cuda.cu` and the same `coli_cuda_*` ABI, through the same host-side split the Windows CUDA path already uses. No new backend abstraction. That is the difference between a contribution that adds a platform and one that adds a platform *and* a fork in the code everyone else then maintains.

**Fail-closed runtime binding.** Refusing to start the GPU tier when a different `amdhip64_7.dll` is already mapped, when more than one is, or when file identity cannot be established — that is the right default, and it is the opposite of what this project has been shipping. Two of today's issues were exactly the other failure mode: #783, where `make kimi_k3 CUDA=1` compiled `-DCOLI_CUDA` and linked cudart on a source with no CUDA in it, and reported success at every observable point while the GPUs stayed idle. Silent wrong-thing is worse than a loud refusal.

**78 loader-contract tests with synthetic stub DLLs, running on CI without a GPU.** That is the part I would not have thought to ask for. Most GPU PRs here can only be verified by whoever owns the silicon; you made most of yours checkable by anyone.

## What I will be reading for

Not a rubber stamp — 4,273 lines touching the build and the loader:

- **`backend_cuda.cu` shared with CUDA.** Anything that compiles differently under hipcc has to leave the NVIDIA path byte-identical. The HIP syntax job covers compilation; I want to see the guards.
- **`c/Makefile`** — `HIP=1` is Linux-only and refused early elsewhere. I want to check `HIP_DLL=1` cannot be reached in a state that half-configures the build, which is the shape of #783.
- **`backend_loader.c`** — the absolute-path load and the identity check are the security-relevant part. Loading a DLL from the host's own directory by absolute path is right; I want to confirm there is no path where a relative or search-order load survives.
- **`.github/workflows/ci.yml`** — how much runner time the 78 tests add, and whether they can fail for reasons unrelated to the loader.

## One question

You validated on physical `gfx1151` including teacher forcing and free decode. Was the teacher-forced run **token-exact against the CPU path**, or close-but-not-identical? The distinction matters here more than the throughput: this project makes no promise about speed and a hard one about semantics, so a backend that changes which tokens come out is opt-in by policy (that is what #763 was about, merged today).

If it is token-exact, say so in the PR body — it is the strongest claim you can make and it is currently not stated.

JustVugg:
CI ran (it was held in `action_required` — a first contribution from a fork, and nothing surfaces that on the PR page). **349 of 351 pass.** Two fail, on Windows only, and both are the same one-character thing:

```
FAIL test_configured_runtime_foundation
FAIL test_configured_runtime_reports_a_missing_runtime

'…/coli helper yhk99_d1/empty dir\amdhip64_7.dll'
                                 ^
'…/coli helper yhk99_d1/empty dir/amdhip64_7.dll'
```

The loader joins with `\`, the expectation is built with `/`. Under MSYS2 both conventions are live in the same process — the temp root arrives as `D:/a/_temp/msys64/tmp/...` with forward slashes while `os.path.join` gives a backslash — so a string comparison of paths cannot hold. I would not change the loader for this: `\` is correct for Windows. Compare with `os.path.normcase(os.path.normpath(...))`, or `pathlib.PurePath`, on both sides.

Worth noting the two tests that failed are exactly the ones checking that a *missing* runtime is reported and that identity resolution is sound — the security-relevant pair. Good place for the harness to be strict; it just needs to be strict about the right thing.

**Also now CONFLICTING** — five PRs merged to `dev` tonight, including #785 which touched `c/Makefile` and `c/coli`. A rebase on current `dev` and the path fix should take it green.

Everything I said in the earlier comment stands, and the failure does not touch any of it: the fail-closed binding, the reuse of the existing `coli_cuda_*` seam and the 78 no-GPU loader tests are the parts that make this reviewable at all, and 349/351 on the first CI run of a 4,273-line PR is a good showing.

The question from before is still the one I care most about: **was the teacher-forced `gfx1151` run token-exact against the CPU path?** If yes, put it in the PR body — it is the strongest claim available here and it is not currently stated.

Kenneth-Javier:
Yes — it was token-exact against the CPU path at the token-ID level.

In the retained authoritative sweep, CPU teacher forcing ran twice and hybrid
HIP teacher forcing ran twice. All four runs scored 32/32, and the predicted
token ID matched position-by-position between the CPU and hybrid-HIP paths for
all 32 checked positions.

This was not a close or tolerance-based comparison. I am not claiming
bit-identical logits, only identical token decisions at every checked position.

Separately, the free autoregressive decode also produced the same exact
20-token sequence across two CPU processes and two hybrid-HIP processes.

I updated the PR body to make the teacher-forcing token-exactness explicit.

Kenneth-Javier:
### Post-rebase validation

After the physical validation above, upstream `dev` advanced from `72ddb673`
to `44d7958`, and this branch was rebased onto that newer base.

The only manual conflict resolution was in `c/Makefile`, where upstream's new
`tests/test_kv_prefix` target and this branch's `hip-dll` `.PHONY` entry were
both preserved. The remaining commits were range-diff identical.

Post-rebase validation on head `e5ff643` completed with:

- backend-loader contract suite: 78/78, 0 skips
- full C suite: 39 test binaries, runner exit 0

The physical `gfx1151` model results reported above were produced before this
rebase and were not rerun on `e5ff643`.

Kenneth-Javier:
Thanks — and yes, the teacher-forced `gfx1151` run was token-exact against the
CPU path at the token-ID level.

The retained authoritative sweep contained two CPU teacher-forcing processes
and two hybrid-HIP teacher-forcing processes. All four runs scored 32/32, and
the predicted token ID matched position-by-position at every one of the 32
checked positions in both runs per mode.

This was an exact comparison of the selected token decisions, not a
close or tolerance-based comparison. I am not claiming bit-identical logits.

Separately, two CPU and two hybrid-HIP free-decode processes produced the same
exact 20-token autoregressive sequence, with prefix equality checked at
positions 1, 2, 4, 8, 16 and 20.

You were also right about the Windows CI failure. My earlier `os.sep`
explanation was incomplete: the temp root and the loader-appended filename can
use different separator conventions inside the same process.

The fix is test-only. The loader continues to emit the correct Windows `\`
separator, while the two assertions now compare normalized paths instead of
raw strings. I also removed the `os.sep` CI guard because it encoded the wrong
diagnosis.

The branch is now rebased onto `dev` at `7fb1159`, with head
`1f76e498cc346a0c324a946711ea799ef953250a`. The loader suite passes locally
with 78/78 and zero skips under both exercised environments; the new hosted CI
run remains the authoritative confirmation for the runner configuration.

JustVugg:
#165 (DeepSeek V4 target-only CPU engine) has landed on `dev` and this PR now conflicts. It is small — two hunks:

- `c/Makefile` — one hunk, both sides appending to the same list; keep both.
- `c/tools/clean.py` — one hunk, the artifact-removal globs.

I resolved this class of conflict directly on the other affected branches and pushed the merges, so their authors did not have to do anything. I could not do that here: **"Allow edits by maintainers" is off on this PR**, so I have no write access to `Kenneth-Javier:feature/windows-hip-native`.

If you tick that box I will push the merge for you. Otherwise `git merge dev` on your branch takes about a minute — neither hunk is contested.

Separately, and more importantly: **your CI had never run.** The workflow sat in `action_required` — a state that does not show anywhere on the PR page — since the day you opened it. I approved the blocked runs across the repo earlier today, so this PR has real check results now for the first time. Worth reading them before you rebase.

Kenneth-Javier:
Thanks — I merged the current `dev` as suggested.

The new PR head is `aa60f7b`, with `1f76e49` as the first parent and
`c024a46` as the merged `dev` parent. I used a merge rather than another
rebase so the existing 15 commit SHAs, review context and previously unlocked
CI history remain intact.

Both conflicts were resolved as direct unions:

- `c/Makefile`: retained the DeepSeek V4 targets and `hip-dll`
- `c/tools/clean.py`: retained both artifact-cleanup blocks

Local post-merge validation completed with:

- backend-loader suite: 78/78, zero skips under both exercised separator environments
- `make test-c`: exit 0, 45 binaries
- clean merge-tree against `dev`

The new `CI` and `check` workflow runs for `aa60f7b` are currently awaiting
maintainer approval again, so no jobs have run on this head yet.

I have also enabled “Allow edits by maintainers” for future small `dev`
integration conflicts.

Kenneth-Javier:
### Current dev integration

Upstream later advanced to `c024a46868859c795a1596f40707086056c53d43`,
including the DeepSeek V4 target-only CPU engine.

That `dev` head was merged into the PR branch rather than rebased, preserving
the existing 15 W1 commit identities and their review history.

The merge required two manual conflict resolutions, both direct unions:

- `c/Makefile`: upstream's DeepSeek V4 targets and this branch's `hip-dll`
  target were both preserved
- `c/tools/clean.py`: both artifact-removal blocks were preserved

The resulting branch head is:

```text
aa60f7b632f443bf784d5b7870b8696c69905feb

Kenneth-Javier:
All checks are green now on `aa60f7b`.

Thanks for pointing out both the hidden `action_required` state and the exact
conflict locations.

I merged `dev` at `c024a46` into the branch and resolved both conflicts as
direct unions:

- `c/Makefile`: retained both the DeepSeek V4 targets and `hip-dll`
- `c/tools/clean.py`: retained both artifact-cleanup blocks

The hosted checks now pass, including the Windows UCRT64 job and the 78
backend-loader contract tests.

Local post-merge validation also passed:

- backend-loader suite: 78/78, 0 skips under both exercised separator environments
- `make test-c`: exit 0, 45 test binaries

The PR body now also states the semantic result explicitly: teacher forcing on
`gfx1151` was token-exact against the CPU path at all 32 checked positions at
the token-ID level, without claiming bit-identical logits.

The branch is ready for the remaining code review. Thanks again for the detailed
feedback and for getting the blocked workflows running.

Issue #768, opened 2026-08-02
KV cache is stored in fp32 and kvb_all materialises the decompressed K/V (x3.2 vs llama.cpp)


ThefloorMiner:
For GLM-5.2 the KV cache costs **0.316 MB/token** versus **0.098 MB/token** for llama.cpp with `-ctk f16` on the same checkpoint and the same latent attention — a factor **3.2**. At `--ctx 262144` that is 82.8 GB instead of ~26 GB.

**Two independent causes.**

**1. fp32 storage.** MLA keeps `kv_lora_rank 512 + rope 64 = 576` values per token per layer; x78 layers = 44,928 values/token. At 4 bytes -> 179.7 KB/token -> **47.1 GB** at 262,144, matching the engine's reported `KV 1x262144 47.7`. Confirmed at `colibri.c:2277`:

```c
m->kv_dev_L[layer] = (float*)coli_cuda_pipe_alloc(dev, (size_t)m->max_t*kvl*4);
```

**2. `kvb_all`.** `colibri.c:2766` allocates `falloc(Tk * kvb_dim)` with `kvb_dim = H*(qk_nope+v_head) = 64*(192+256) = 28,672` floats = **114.7 KB per context position** -> **30.1 GB** at 262,144, matching the reported `kvb 30.1`. This materialises the full per-head K/V, i.e. precisely what MLA exists to avoid. The code already documents the alternative at `colibri.c:2568`:

> *costo per step ~O(T·kv_lora) invece di O(T·H·(nope+vh)) del matmul kvb_all*

**Impact.** On our host the ~57 GB avoidable covers 83% of the 69 GB by which the model overflows RAM+VRAM at long context. With an fp16 KV and no `kvb_all`, 262,144 context **and** full expert residency would be simultaneously reachable — instead of having to choose (measured: 5.46 tok/s at ctx 32768 vs 0.86 at 262144).

**Suggested fix:** a `KV_BITS`-style knob mirroring `K3_BITS`, and taking the `O(T·kv_lora)` path where available.

---
*Observed on v1.3.0, commit `ecade075cfc2eae684097ea7de5570c3786ce199`, Linux, CUDA build `sm_86`, 4x RTX A6000, 251 GiB RAM, model `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`. Full context in #764.*

ThefloorMiner:
[...]

Issue #766, opened 2026-08-02
The cap computed by --auto-tier drives the process into the OOM killer


ThefloorMiner:
Applying the value `--auto-tier` itself computed (`--cap 113`) drives anonymous RSS to 257 GB on a 251 GiB host. The global OOM killer terminates the engine. The documented guard (*projected peak > MemAvailable -> exit(2)*, overridable via `COLI_RAM_OVERCOMMIT`) does not fire.

**Reproduce:** `coli web --auto-tier --ctx 32768 --cap 113 --vram 176 --ram 212`

**Observed**

```
Out of memory: Killed process (colibri) total-vm:293752600kB, anon-rss:257583016kB
```

**Measured scaling on this host:** `cap 64` -> 118 GB RSS, `cap 113` -> 257 GB.

**Suggested fix:** have the projection account for the same terms the runtime actually allocates, and fail fast with a clear message rather than relying on the kernel.

---
*Observed on v1.3.0, commit `ecade075cfc2eae684097ea7de5570c3786ce199`, Linux, CUDA build `sm_86`, 4x RTX A6000, 251 GiB RAM, model `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`. Full context in #764.*

ThefloorMiner:
**Version note / partial retraction.** I reported this against `ecade075` (2026-07-30) without noticing `dev` was 185 commits ahead.

The reproduction used an explicit `--cap 113`. On current `dev`, `--cap` defaults to auto (see #765, now closed), so **the practical impact is much smaller**: a user has to pass the value by hand to hit it.

What I believe still stands: an explicitly-supplied cap that projects past `MemAvailable` should fail fast with a clear message rather than reach the kernel OOM killer. Feel free to close if you consider it out of scope now that the cap auto-sizes.

JustVugg:
Labelled, and reading your five reports together shows something none of them says on its own — so I want to put that here rather than answer each one separately.

## #766, #767 and #759 are one bug wearing three coats

| | the budget that is computed | what it does not account for | how it ends |
|---|---|---|---|
| **#766** (yours) | `--auto-tier` says `--cap 113` | what cap 113 actually costs in RSS — you measured 64→118 GB, 113→257 GB | OOM killer on a 251 GiB host |
| **#767** (yours) | `CUDA_EXPERT_GB=auto` | prefill scratch for a realistic prompt | a device permanently disabled mid-session |
[...]

Issue #305, opened 2026-07-16
[Bug]: Keeps terminating itself


Ne00n:
### Commit or version

550ddcba83afd27a892dba92c587bfcc1d30f020

### Environment

Debian 12

### Reproduction steps

I run a prompt and in the middle of the generation, it just terminates itself.
I just get

  [engine terminated]
  goodbye — the hummingbird returns to its nest 🐦

No error message.

### Expected behavior

To actually finish the task

### Actual behavior and logs

```shell
[engine terminated]
  goodbye — the hummingbird returns to its nest 🐦
```

bokiko:
A silent mid-generation exit with a clean `[engine terminated]` banner and no error is the signature of the kernel OOM killer — the engine gets SIGKILLed, so it can't print anything, and the wrapper just sees the process end (same class as #229). Two quick checks:

1. `dmesg | grep -i -E "oom|killed process"` right after it happens — if glm is there, that's it.
2. How much RAM does the box have, and what `--ram` are you passing? `--ram` sizes the expert cache; the engine needs headroom above it (KV, activations, page cache), so `--ram` at or near physical RAM will OOM mid-generation as the KV grows.

If dmesg is clean, the next suspects are different: paste `echo $?` after the exit and your full command line + `free -g`, and we can bisect from there.

JustVugg:
Thank you — this was a real bug on our side, and the missing error message *was* the bug.

When the kernel's OOM-killer takes a process it uses SIGKILL: no error, no log, stdout just closes. `coli` saw that EOF and printed `[engine terminated]` without ever reaping the child, so an OOM-kill looked exactly like a clean exit. You had no way to tell what happened, and neither did we.

Two fixes are on `main` now (d57955e):

- **`coli` reports the cause**: the signal or exit code, and when it was SIGKILL it says so explicitly — nothing in the engine sends itself SIGKILL, so that's the kernel — plus the tail of the engine's stderr.
[...]

PR #793, opened 2026-08-02
fix(plan): size the expert slot by the widest width the container holds

ThefloorMiner:
Closes the diagnosis you asked for in #766 — *"even naming which term is missing would shorten this considerably"*. **No term is missing.** The shape of `cap_for_ram()` is right and one constant in it is wrong by 1.78×.

## The bug

`expert_bytes_probe()` probes expert 0 of `c->first_dense`, the first MoE layer, and returns that size. On a container that **mixes widths** it returns the narrowest one — while a slot actually costs the widest, because `ws[]` slots and the per-layer LRU slabs are reused **across layers** and `slab_cap` grows to the largest expert a slot has ever held and is never shrunk. The `ESlot` comment at `colibri.c:314` already says so:

> gli slot `ws[]` sono riusati **TRA layer** e gli expert **non hanno tutti la stessa taglia** (layer MTP int8 = 2x i layer int4)

`cap_for_ram()` divides by this value, so the error goes straight into the cap.

## Measured, two independent ways

**From the container**, reading the safetensors headers of `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`:

| layer | width | expert 0 |
|---|---|---|
| 3 | int4 routed | **21.23 MB** ← what the probe returned |
| 78 | int8 MTP | **37.79 MB** ← what a slot costs |

**From outside**, the two RSS points in #766 on 4× RTX A6000 / 251 GiB:

```
cap  64 -> 118 GB           cap 113 -> 257 GB -> OOM-kill
slope = 139 GB / (49 · 77) = 36.84 MB per slot per layer
```

36.84 against 37.79 is **2.5 %**. Against the 21.23 MB the probe returned it is **1.74×**.

And the corrected projection lands exactly where it should:

```
113 · 21.23/37.79 = 63.5
```

**`cap 64` is the measured point that survives at 118 GB.** Two independent derivations of the same constant, and the corrected formula reproduces the known-good configuration without any term being added.

## What this also settles

- **Nothing in `cap_for_ram()` is super-linear.** Every term is strictly linear in `max_ctx`, and the `slack` decomposition reproduces the 82.8 GB quoted in #768 exactly (`kv_pool_bytes` 47.7 + `kvb_b` 30.1 + `ws_b` 1.3 + page cache 2.5 + activations 1.2). Two points fit a straight line with the wrong slope, not a curve.
- **`nsp += 2` already nods at the MTP row costing double** — but that corrects one row out of `nsp`, while the slabs grow on all of them.

## The change

`expert_bytes_probe()` takes the **max** over the widths the container actually holds — the first MoE layer and, when `has_mtp`, the MTP layer — instead of the first alone. The probe body is factored into `expert_bytes_layer()`; the fallback when no container is present is untouched.

- **Single-width containers: byte-identical.** The max is taken over one value.
- **Mixed-width containers: the cap roughly halves.** That is the direction that keeps the process alive — over-reserving costs cache slots, under-reserving costs the run to `SIGKILL` with no log, which is the failure `cap_for_ram()` exists to prevent (#305).

Second, smaller change: the `cap=1` warning now prints `eb`. It is the one term a user cannot check from outside, so today a wrong width and a wrong budget produce an identical message.

```
[RAM_GB=215.0] WARNING: cap=1 is the floor, projected peak 216.7 GB is 1.7 GB OVER
the budget (resident 132.2 GB + reserve 82.8 GB, expert 37.8 MB).
```

## Testing, and its limits — please read

Builds clean (`make colibri`, clang, `-Wall -Wextra`, no new warnings) on current `dev` at `44d7958`.

**I have not run this on the 4× A6000 host yet.** It is mid-campaign on other measurements as I write, and I would rather send you the reasoning now than sit on it. The specific thing I owe you before this should merge is: `--auto-tier` on that host, confirming it now returns ~64 rather than 113 and that the run survives. I will post that number here.

Two honest caveats:

1. **One checkpoint, one mixed-width layout.** The fit is very good (2.5 % from the container, 1.74× vs 1.78× predicted from the two RSS points) but I have not tested a container whose MTP head is *narrower* than the routed layers, nor one with three widths. The `max` is the safe direction in all of those, but "safe" is not "measured".
2. **The mechanism is inferred from the `ESlot` comment plus the arithmetic**, not from instrumenting `slab_cap` directly. If slabs do *not* in fact grow the way that comment describes, the numbers still say the constant is wrong by 1.74× and this patch still fixes the cap — but my explanation of *why* would be wrong, and the comment I added should be corrected rather than kept.

Happy to hold this until I can post the `--auto-tier` run, if you would rather not have it in the queue in the meantime.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -8314,10 +8314,13 @@ static void cap_for_ram(Model *m, double ram_gb, int ebits, int max_ctx){
      * non entra nemmeno nella RAM realmente disponibile misurata all'avvio. */
     if(floored){
         double peak = (double)m->resident_bytes + (double)capmax*nsp*eb + slack;
+        /* eb is printed because it is the one term nobody can check from outside:
+         * every projection here divides by it, so a wrong width is indistinguishable
+         * from a wrong budget in this message (#766). */
         fprintf(stderr,"[RAM_GB=%.1f%s] WARNING: cap=1 is the floor, projected peak %.1f GB is "
-            "%.1f GB OVER the budget (resident %.1f GB + reserve %.1f GB).%s\n",
+            "%.1f GB OVER the budget (resident %.1f GB + reserve %.1f GB, expert %.1f MB).%s\n",
             ram_gb,auto_b?" auto":"",peak/1e9,(peak-ram_gb*1e9)/1e9,
-            m->resident_bytes/1e9,slack/1e9,
+            m->resident_bytes/1e9,slack/1e9,(double)eb/1e6,
             getenv("PIN_GB")?" PIN_GB is inflating the resident set: lower it or drop it.":"");
 #ifdef COLI_CUDA
         /* #686: on a single GPU that also holds host copies of the VRAM tier, the
--- a/c/telemetry.h
+++ b/c/telemetry.h
@@ -10,18 +10,44 @@ static int64_t tbytes(int O,int I,int bits){
     return (int64_t)O*((I+1)/2) + (int64_t)O*4;
 }
 
-static int64_t expert_bytes_probe(Model *m, int ebits){
-    Cfg *c=&m->c; int64_t eb=0; char nm[256];
-    snprintf(nm,sizeof(nm),"model.layers.%d.mlp.experts.0.gate_proj.weight",c->first_dense);
-    if(st_nbytes(&m->S,nm)>0){
-        const char *suf[3]={"gate_proj","up_proj","down_proj"};
-        for(int k=0;k<3;k++){
-            snprintf(nm,sizeof(nm),"model.layers.%d.mlp.experts.0.%s.weight",c->first_dense,suf[k]);
-            eb+=st_nbytes(&m->S,nm);
-            snprintf(nm,sizeof(nm),"model.layers.%d.mlp.experts.0.%s.weight.qs",c->first_dense,suf[k]);
-            int64_t q=st_nbytes(&m->S,nm); if(q>0) eb+=q;
-        }
+/* Container bytes of expert 0 of one layer (weights + .qs scales); 0 if absent. */
+static int64_t expert_bytes_layer(Model *m, int layer){
+    int64_t eb=0; char nm[256];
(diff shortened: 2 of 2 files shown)


JustVugg:
This is the answer to the question, and it is better than the question deserved.

I asked in #766 whether you could name the missing term. You came back with **there is no missing term** — the shape of `cap_for_ram()` is right and one constant in it is wrong by 1.78×. That is a harder thing to establish than finding an omission, because it means checking that everything else is correct.

**Two independent derivations landing on the same number** is what makes it merge-on-sight:

```
container headers   layer 3  int4 routed  21.23 MB   ← what the probe returned
                    layer 78 int8 MTP     37.79 MB   ← what a slot costs

RSS slope (#766)    139 GB / (49 · 77)  = 36.84 MB per slot per layer
```

36.84 against 37.79. One reading comes from safetensors headers, the other from two OOM data points on 4× A6000 — neither can be tuned to fit the other, and they agree to 2.5%.

And it retires something I had written down wrong. In #766 I called the growth **super-linear** and said that was "exactly what a projection assuming linearity would miss". It is not super-linear. `cap 64 → 118 GB` and `cap 113 → 257 GB` sit on a straight line through a slope of 36.84 MB; I was reading a wrong constant as a wrong shape. `113 × 21.23/37.79 = 63.5` against a measured 64 closes it.

I verified the two things the fix rests on rather than taking them from the description: `has_mtp` exists and is set from the container (`colibri.c:1828`), and the MTP layer is indexed at `c->n_layers` (`colibri.c:1830`), so `expert_bytes_layer(m, c->n_layers)` reads the row you mean.

The `nsp += 2` observation is the part I would have missed. It corrects one row for the MTP width while the slabs grow on all of them — a partial fix sitting next to the real one, which is exactly how a constant stays wrong for months.

Printing `eb` in the floor warning is the right companion change: every projection in that function divides by it, so without it a wrong width and a wrong budget produce the same message.

Merging. #766 and #759 stay open — the GB10 path in #759 is a different term (`RAM_GB` unclamped on unified memory) and should not be closed by this.

Issue #814, opened 2026-08-03
[Bug]: 我下载最新的windows release 1.4.0和glm5.2 g64模型测试，怎么推理能力为0啊？


MineloaderGuo:
### Commit or version

release 1.4.0

### Environment

操作系统win11，从这里下载的预编译版，非自行编译，内存32g，磁盘是ssd，跑的是ram，没用cuda

### Reproduction steps

我下载最新的windows release 1.4.0和g64模型测试，本机32G内存，13th intel cpu，跑ram，用”我想去洗车，洗车店距离50米但堵车，我该开车过去还是走过去？“提问，得到的回答不忍直视，推理能力基本为0。多次重试，均不能正常推理。不知道是我本地环境有问题，还是这个release没能保真glm5.2原有精度？我不在乎有多慢，只希望不论花多长时间，能得到有价值的回答。我没有linux，不知道linux上是不是正常的。有谁能用上面同样的提问测试一下吗？谢谢。

### Expected behavior

推理应至少达到小学生水平。

### Actual behavior and logs

```shell
回答可笑。
```

MichaelFomenko:
please write in English, we don't understand your language. I could also write in German or Russian here, this is not respectful gains other people here.

JustVugg:
你的环境没有问题，模型精度也没有损失。**GLM-5.2 的推理模式默认是关闭的**，而你的问题恰恰需要它。

Your environment is fine and the release does not lose GLM-5.2's precision. **GLM-5.2's reasoning mode is off by default**, and your question is exactly the kind that needs it.

`c/coli:676`:

```python
tk = "<think>" if os.environ.get("THINK", "0") == "1" else "<think></think>"
```

Without `THINK=1` the engine writes an **already-closed** `<think></think>` into the prompt. The model is told, in its own template, not to think — so it answers from reflex. On a question with a trap in it ("50 metres away, but there is traffic") reflex is exactly what produces a laughable answer.

请这样重试 / Please retry like this:

```
set THINK=1
coli chat --model <your-model-dir>
```

or in one line under PowerShell:

```powershell
$env:THINK=1; .\coli.exe chat --model <your-model-dir>
```

You should now see a `<think>` block before the answer, and on your car-wash question the model should reason about walking 50 m versus the time cost of driving in traffic.

## 两点补充 / Two things worth knowing

**Reasoning costs tokens, and you said you do not mind the wait** — good, because the think block can be several times longer than the answer. That is the trade.

[...]

PR #834, opened 2026-08-05
fix(v4): add missing sys/resource.h header for getrusage

QQSHI13:
Fixes a compilation break on Linux/WSL platforms where struct rusage and getrusage storage layouts are unrecognized.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -1,6 +1,8 @@
 /* Amalgamated deepseek_v4.c — GLM-style source; compile with -DCOLI_V4_UNIT_* per object */
 /* Umbrella API: deepseek_v4.h (included by units) */
 
+#include <sys/resource.h>
+
 #ifdef COLI_V4_UNIT_ST
 /* Shared st.h adapter and V4 tensor materialization helpers. */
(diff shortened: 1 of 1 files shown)


JustVugg:
You found this independently and you were right — thank you.

It is already on `dev`, and only by a few hours. When #165 landed I hit the same failure and folded the include into the merge commit rather than landing a broken tree, so the fix went in without its own PR:

```c
#if defined(__APPLE__) || defined(__linux__) || defined(__FreeBSD__)
#include <sys/resource.h>   /* getrusage/RUSAGE_SELF for v4_serve_rss_gb;
                             * on Windows compat.h supplies the shim. */
#endif
```

The guard is there because Windows has no `<sys/resource.h>` — `compat.h` provides `getrusage` via `GetProcessMemoryInfo`. That is also why the failure was Linux-only and one of fourteen checks went red while the rest stayed green.

Closing as already fixed, not as wrong. Two people finding the same missing header on the same day is a good sign for the project, and the next one you find will very likely not already be in flight.

Separately, and unrelated to the code: thank you for what you wrote on #814. That reporter came back, retested, and confirmed the model reasons correctly — which would not have happened if he had been argued out of the thread first.

PR #851, opened 2026-08-06
fix: exit cleanly on SIGTERM

### LINKED ISSUE #810: [Bug]: the engine exits 1 on SIGTERM, so every clean systemctl stop leaves the unit failed
## Summary

A normal `systemctl stop` on a healthy engine leaves the unit in `failed`:

```
août 03 10:07:38 systemd[1]: Stopping colibri.service...
août 03 10:07:49 systemd[1]: colibri.service: Main process exited, code=exited, status=1/FAILURE
août 03 10:07:49 systemd[1]: colibri.service: Failed with result 'exit-code'.
août 03 10:07:49 systemd[1]: Stopped colibri.service.
```

Nothing went wrong. The engine had served for 2 h 50 min, systemd sent SIGTERM, and the process exited **1**. systemd correctly records that as a failure.

## Consequences

**`systemctl is-failed` is useless as a health check.** After any planned stop the unit reports `failed`, so the state cannot distinguish "was stopped on purpose" from "crashed". On a box where the engine is stopped and started routinely — benchmarking, model swaps, maintenance — the unit is in `failed` most of the time it is not running.

**Every start needs `reset-failed` first.** With `StartLimitBurst` set, a few stop/start cycles trip the rate limiter and the unit refuses to start at all until manually reset. Every automation script I wrote against this engine ends up carrying:

```bash
systemctl stop colibri
...
systemctl reset-failed colibri    # otherwise the next start may be refused
systemctl start colibri
```

which is a workaround for the exit code, not for anything real.

## Two ways to fix it, in preference order

**1. Exit 0 on SIGTERM.** The engine already has a soft-stop path — `colibri.c` documents SIGINT closing the current turn *"per la stessa via del tetto NGEN (stats, usage_save, KV append, sentinella END tutti normali)"*, i.e. stats flushed, `.coli_usage` saved, KV appended. SIGTERM should land in the same place and return 0. That also makes the shutdown *correct*, not just quiet: a SIGTERM today does not obviously go through `usage_save`, which on a long-lived serve process is the learned routing profile.

**2. Failing that, document `SuccessExitStatus=1`** in the systemd example. This is a smaller fix and it is honest — it tells systemd that 1 is an expected shutdown code. But it also makes a genuine exit-1 failure invisible, so it is second best.

I would rather send the first as a PR than the second, but SIGTERM handling in a serve loop touches shutdown ordering and I would want to know whether you consider the SIGINT path safe to share before I go near it.

## Reproduce

```
systemctl start colibri     # let it load and serve one request
systemctl stop colibri
systemctl is-failed colibri # -> failed
```

---
*Observed on `dev` @ `7fb1159`, Ubuntu 24.04, CUDA build sm_86, 4× RTX A6000, `coli web` under systemd. Not GLM-specific as far as I can tell — the exit path is in the serve loop, not the model code.*

dyKiU:
Fixes #810.

  SIGTERM now wakes the mux loop through a POSIX self-pipe, runs the existing soft-stop cleanup, and
  exits with status 0. SIGINT behavior, request framing, Windows behavior, and dependencies remain
  unchanged.

  Adds a regression test for worker-thread SIGTERM delivery and clean exit.

  ## Validation

  - [x ] make -C c check
  - [x] CUDA testing not applicable
  - [x] No performance claims

  Full C tests passed on Ubuntu. The focused regression and CPU build passed on macOS.

  Manual test on Ubuntu/Acer using three terminals:

  Terminal 1:

```
  systemctl --user start colibri-810.service
  journalctl --user -fu colibri-810.service
```

  Terminal 2:

```
  python3 - <<'PY' >/tmp/request.json
  import json
  print(json.dumps({
      "model": "glm-tiny",
      "messages": [{"role": "user", "content": "Keep generating output. " * 200}],
      "max_tokens": 4096,
      "stream": True
  }))
  PY
```

```
  curl -N -H 'Content-Type: application/json' \
    --data-binary @/tmp/request.json \
    http://127.0.0.1:8765/v1/chat/completions
```

  Terminal 3, while the request is streaming:

```
  systemctl --user stop colibri-810.service
  systemctl --user show colibri-810.service \
    -p Result -p ExecMainStatus -p ActiveState
```

  Observed:

```
  finish_reason: stop
  data: [DONE]
  Result=success
  ExecMainStatus=0
  ActiveState=inactive
```

  Usage and KV state were persisted. No timeout or SIGKILL occurred.

  ## Compatibility

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -732,6 +732,9 @@ iobench$(EXE): iobench.c compat.h
 tests/test_serve_sentinel$(EXE): tests/test_serve_sentinel.c compat.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
+tests/test_sigterm$(EXE): tests/test_sigterm.c serve_signal.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
 tests/test_ue8m0$(EXE): tests/test_ue8m0.c st.h json.h compat.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -6043,25 +6043,16 @@ static int grammar_draft(GrDraft *g, int *draft, int cap){
 /* STOP MORBIDO (serve/chat): SIGINT chiude il turno CORRENTE per la stessa via
  * del tetto NGEN (stats, usage_save, KV append, sentinella END tutti normali)
  * invece di uccidere il motore; :more puo' continuare la risposta interrotta.
+ * SIGTERM usa la stessa chiusura morbida, poi termina il serve-loop con successo.
  * Il flag e' armato solo nei serve-loop (intr_install): nei run one-shot e in
- * validazione SIGINT resta il default (morte immediata). Solo POSIX: su
+ * validazione SIGINT/SIGTERM restano default. Solo POSIX: su
  * Windows il comportamento di Ctrl-C non cambia.
  * EN: soft stop (serve/chat): SIGINT ends the CURRENT turn through the same
  * path as the NGEN cap — stats/usage/KV/END sentinel all normal — instead of
- * killing the engine; :more can continue the interrupted answer. Armed only
- * in the serve loops; one-shot runs keep default SIGINT. POSIX only. */
-static volatile sig_atomic_t g_intr=0;
-#if defined(__APPLE__) || defined(__linux__) || defined(__FreeBSD__)
-static void intr_sig(int s){ (void)s; g_intr=1; }
-static void intr_install(void){
-    struct sigaction sa; memset(&sa,0,sizeof(sa));
-    sa.sa_handler=intr_sig; sigemptyset(&sa.sa_mask);
-    sa.sa_flags=SA_RESTART;              /* getline/pread non devono vedere EINTR */
-    sigaction(SIGINT,&sa,NULL);
-}
-#else
-static void intr_install(void){}
-#endif
+ * killing the engine; :more can continue the interrupted answer. SIGTERM uses
+ * that same soft-stop path, then exits the serve loop successfully. Armed only
+ * in serve loops; one-shot runs keep default SIGINT/SIGTERM. POSIX only. */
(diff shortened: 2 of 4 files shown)


JustVugg:
Thank you for this, and for the end-to-end systemd validation — that is more than most signal-handling PRs arrive with, and your #852 went in a few minutes ago.

I merged #850 for #810 instead, and I owe you the reason, because the two PRs do **not** fix the same case.

The engine has two serve loops, and `run_serve` — the one blocking in `getline()` — is the default. `run_serve_mux` runs only under `SERVE_BATCH=1`:

```c
if(getenv("SERVE_BATCH") && atoi(getenv("SERVE_BATCH"))) run_serve_mux(&m,snap);
else run_serve(&m,snap);            /* colibri.c:9420 */
```

The self-pipe wakes `select()`, which covers the mux loop. The default loop is not woken, because `SA_RESTART` is kept and the blocked `read` under `getline` resumes after the handler returns. I measured it rather than reasoning about it — same handler, signal delivered while blocked on a pipe nobody writes to, with a second un-restarted signal at 2.00 s purely as an escape hatch so the process could not hang the test:

```
plain     getline returned after 0.20s   <- SIGTERM broke the read
RESTART   getline returned after 2.00s   <- it did not; that is the escape hatch
```

So `while(!g_term && (nr=getline(...))>0)` is reached only once a request arrives. Your manual test passes because you stopped the unit **mid-stream**, where generation is running and `g_intr` ends the turn — the one path that works either way. `systemctl stop` on an idle server, which is the ordinary case in #810, still waits for `TimeoutStopSec` and escalates to SIGKILL.

Two smaller things, for whatever you send next rather than as a verdict on this one:

- `test_sigterm.c` asserts `(sa.sa_flags&SA_RESTART)!=0` as required behaviour. That pins the flag that prevents the fix on the default loop.
- It also asserts four `strstr()` matches against `colibri.c`'s source text, including `fread(raw,1,(size_t)sub.bytes,stdin)`. Those go red on any unrelated refactor of that line, and green on code that does not run.

Closing as superseded. The reporting and the systemd harness were genuinely useful here — please keep sending work.

PR #860, opened 2026-08-06
Fixed release workflow to include deepseek

remielowik:
## Summary

## Validation

- [ ] `make -C c check`

## Compatibility

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,13 +64,18 @@ jobs:
           - os: ubuntu-latest
             name: linux
             shell: bash
+            engines: colibri inkling kimi_k3 olmoe deepseek-v4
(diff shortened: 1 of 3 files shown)


JustVugg:
Thank you for picking this up, and sorry for the wasted round trip — the fix landed on `dev` in #868 before I got to this one.

Two notes for next time, neither of them about the substance:

- **Target `dev`, not `main`.** `main` only ever receives merges from `dev`, so a PR opened against it shows as `blocked` and cannot go in on its own. That is on us for not saying so plainly enough in CONTRIBUTING.
- The diff also reflowed quote style across `ci.yml` (`'12.6.2'` → `"12.6.2"` and similar) in places unrelated to the change. Repo-wide formatting is being handled separately in #798; keeping a fix limited to its own lines makes it much faster to review.

Closing as superseded, not as unwanted. Please do send more.

PR #870, opened 2026-08-07
deepseek_v4: compute RoPE values on demand

waizuichougou:
## Summary

The previous implementation generated RoPE tables from position zero through
the current position, even when only one position or a small contiguous range
was required.

This change computes RoPE values only for the positions consumed by each
caller. The frequency calculation and position semantics remain unchanged.

## Motivation

For DeepSeek-V4-Flash, the RoPE dimension is 64, which requires 256 bytes per
position for the cosine and sine tables.

At position 1,000,000, the previous implementation could allocate
approximately 256 MiB for RoPE tables alone. Most of those values were not
used by the current operation.

The new implementation reduces this to:

- one RoPE row for single-position operations;
- one row per item for batched operations;
- a contiguous range for range-based callers.

## Implementation

- Added a single-position RoPE API.
- Added a range-based RoPE precomputation API.
- Preserved the existing zero-based precomputation API.
- Updated attention, compressor, and indexer callers to use only the required
  rows.
- Added position-boundary validation to prevent integer overflow.
- Preserved the existing YaRN frequency calculation.

The change affects only RoPE table construction and indexing. It does not alter
the model weights, attention equations, or frequency parameters.

## Correctness

The new position and range paths produce the same RoPE values as the previous
full-table implementation.

The generated model output was unchanged in the end-to-end comparison. Expert
request counts, cache hit/miss counts, and bytes read were also identical
between HEAD and this change.

## Performance

The comparison used the same checkpoint, prompt, generation length, and runtime
settings for both versions.

| Metric | HEAD | This change | Improvement |
|---|---:|---:|---:|
| Time to first token | 12.26 s | 11.57 s | 5.6% |
| Total generation time | 25.47 s | 24.06 s | 5.5% |

The largest improvement comes from removing RoPE work and temporary allocation
that scaled with the absolute position. The end-to-end improvement is smaller
because model loading and the remaining inference computation still dominate
runtime.

## Compatibility

- Existing RoPE outputs remain unchanged.
- The existing zero-based RoPE API remains available.
- No model files or generated artifacts are included.
- The default CPU build remains dependency-free.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -1077,6 +1077,7 @@ int coli_v4_expert_store_open_planned(
 /* ######## deepseek_v4_math.c ######## */
 #include "deepseek_v4_internal.h"
 
+#include <limits.h>
 #include <math.h>
 #include <stdlib.h>
 
@@ -1232,15 +1233,11 @@ int coli_v4_rmsnorm(float *output, const float *input, const float *weight,
     return 0;
 }
 
-int coli_v4_rope_precompute(float *cosines, float *sines,
-                            int dimension, int sequence_length,
-                            int original_sequence_length, float base,
-                            float factor, int beta_fast, int beta_slow) {
-    if (!cosines || !sines || dimension < 2 || (dimension & 1) ||
-        sequence_length < 1 || !(base > 1.0f) || !(factor > 0.0f))
-        return -1;
-    int pairs = dimension / 2;
-    int low = 0, high = -1;
+static void rope_yarn_bounds(int dimension, int original_sequence_length,
+                             float base, int beta_fast, int beta_slow,
+                             int *low, int *high) {
+    *low = 0;
+    *high = -1;
     if (original_sequence_length > 0) {
         const float two_pi = 6.2831853071795864769f;
         float denominator = 2.0f * logf(base);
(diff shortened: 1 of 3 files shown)


JustVugg:
Checked this rather than taking the description on trust, and the diagnosis is right — the waste is worse than the numbers in your table suggest.

Four call sites do exactly what you describe: allocate up to the current position, fill every row, then index one:

```c
size_t table_count = (size_t)(rope_position + 1) * pairs;   /* 2625, 4129 */
size_t count       = (size_t)(position + 1) * pairs;        /* 2806, 4369 */
...
coli_v4_rope_apply(rope, 1, state->rope_dim,
                   cosines + (size_t)rope_position * pairs,   /* <- one row */
                   sines   + (size_t)rope_position * pairs, 0);
```

At position 1M with `rope_dim=64`:

```
allocation per call     256 MB
cosf+sinf per call      64 million
rows actually used      1 of 1,000,001
```

Your 5.5% end-to-end is a floor, not the headline. The work is **linear in absolute position**, so it is nearly free at short context and dominates at long — which is the context length this engine exists to serve. A 5.5% win at 25 s of generation is a different animal at 1M tokens.

**Verified token-exact here**, on the generated tiny V4 fixture:

```
PASS target short: teacher forcing and greedy token-exact
PASS target compressed: teacher forcing and greedy token-exact
PASS target long: teacher forcing and greedy token-exact
PASS target session short: exact IDs and exact length
PASS target CLI: prompt beyond the old 512-token cap
PASS target serve: persistent SUBMIT/DATA/DONE protocol is token-exact
PASS prefix reuse: 11 tokens reused, output identical to a cold prefill
PASS DeepSeek V4 KV prefix reuse: all checks completed
```

Two things before this can go in, neither about the substance:

**1. Retarget to `dev`.** `main` only ever receives merges from `dev`, which is why GitHub shows this as `blocked`. `gh pr edit 870 --base dev`, or the base dropdown in the UI.

**2. Rebase.** The branch sits on `8f512fc` (the v1.5.0 tag); `dev` has moved since, including changes to `deepseek_v4.c`. I would rather you rebase than have me resolve it — the conflicts are in your hot path and you know what you meant.

One review note for when you do: the PR adds position-boundary validation "to prevent integer overflow". Please make sure a rejected position fails loudly rather than silently producing a zero table — a wrong RoPE row is not a crash, it is a quietly wrong answer, and this engine's whole test story is token-exactness. If that path is already covered in `test_deepseek_v4.c`, say which case and I will stop worrying.

Nice find. This is the kind of thing that hides indefinitely because it is invisible at the context lengths people test at.

JustVugg:
Merged. Thank you — this is a good first contribution and the reasoning in it was sound.

For the record, what I checked before taking it, since some of it is worth having written down:

**Bit-for-bit equivalence.** Both new APIs against the row they replace, on the real DeepSeek V4 Flash geometry (`dim=64`, YaRN on, `osl=4096`, `factor=40`):

```
  pos 0       single vs full table: identical
  pos 1       ...                   identical
  pos 4095    ...                   identical
  pos 4096    ...                   identical
  pos 65535   ...                   identical
  pos 262143  ...                   identical
  range [1000,1257) vs full table:  identical
  position(-1) returns -1 (rejected)
```

That last line is the thing I asked about: rejection is loud, not a silently zeroed table. A wrong RoPE row is not a crash, it is a quietly wrong answer, so it mattered.

**Token-exact, on the merge with `dev` rather than on your branch alone** — your base was the v1.5.0 tag and `dev` has moved:

```
PASS target short / compressed / long: teacher forcing and greedy token-exact
PASS target session short / long: exact IDs and exact length
PASS target CLI: prompt beyond the old 512-token cap
PASS target serve: persistent SUBMIT/DATA/DONE protocol is token-exact
PASS prefix reuse: 11 tokens reused, output identical to a cold prefill
```

I retargeted the PR to `dev` and approved the workflow run — first-time contributions need that, which is why CI showed nothing for the first hour. Not something you did wrong.

One note on how I would frame this if you write it up: **your 5.5% is a floor, not the headline.** The work is linear in absolute position, so it is nearly free at short context and dominant at long — which is the context length this engine exists for. At 1M tokens the old path was 256 MB and 64 million transcendental calls per call site, to use 256 bytes. Four call sites did it.

Please send more.

Issue #879, opened 2026-08-07
coli launcher can't run OLMoE: missing model_arch() branch + CHAT=1 never set


shujaanazhar:
## Summary

`coli run` / `coli chat` cannot run OLMoE, even though the prebuilt release ships a dedicated `olmoe` engine binary and `tools/convert_olmoe_merged.py` to produce its weights. Two separate bugs compound to cause this.

## Environment

- colibri v1.5.0 (prebuilt `colibri-v1.5.0-linux-x86_64.tar.gz`)
- Ubuntu 24.04, x86_64
- Model: `allenai/OLMoE-1B-7B-0125-Instruct`, converted with `python tools/convert_olmoe_merged.py --repo allenai/OLMoE-1B-7B-0125-Instruct --out ./olmoe_merged` (completed cleanly, `DONE.`, 0 experts incomplete)

## Bug 1 — `model_arch()` has no `"olmoe"` branch, so OLMoE always dispatches to the GLM engine

In `coli`, `model_arch()` only recognizes `inkling`, `kimi`, and `deepseek_v4` in `config.json`'s `model_type` and defaults everything else to `"glm"`:

```python
def model_arch(model):
    ...
    if "inkling" in model_type: return "inkling"
    if "kimi" in model_type: return "kimi"
    if "deepseek_v4" in model_type or (...): return "deepseek_v4"
    return "glm"
```

`OLMoE`'s `config.json` has `"model_type": "olmoe"`, which falls through to `"glm"`. `engine_for()` then hands the request to the GLM binary, which fails immediately because an OLMoE `config.json` doesn't have GLM's expected fields:

```
$ ./coli run --model ./olmoe_merged "hi"
this engine requires n_group=1 (GLM-5.2)
```

Note `_BANNER_MODELS` (used only for the banner text) *does* already have an `("olmoe", "OLMoE", "7B")` entry, so the banner correctly prints `OLMoE · 7B · 7.4 GB on disk` right before the engine dispatch silently picks GLM anyway — that mismatch is what made this confusing to debug (see #191-style "the requirements contradict" pattern, but here it's the *engine choice* that contradicts the banner, not the docs).

## Bug 2 — even with the right engine forced, `env_for_engine()` never sets `CHAT=1` for OLMoE

Forcing the correct binary via `COLI_ENGINE=./olmoe` gets further, but still fails:

```
$ COLI_ENGINE=./olmoe ./coli run --model ./olmoe_merged "hi"
ref.json: No such file or directory
[...]

PR #880, opened 2026-08-07
test: assert the launcher dispatches every engine it ships (#879)

JustVugg:
**Draft on purpose: this fails on `dev` today, for OLMoE, which is the point.** It goes green when @shujaanazhar's launcher fix for #879 lands.

## The test that should have caught #879 was mine, and it passed

#868 added this to the release verification two days ago:

```python
missing = [n for n in expect if not os.path.exists(...)]
if missing:
    sys.exit("FAIL: coli would not resolve these next to itself: " ...)
```

It is *named* as a dispatch check and it is a file-existence check. It went green on every platform while `coli run --model <olmoe>` sent the request to the GLM engine and died with `this engine requires n_group=1 (GLM-5.2)`.

A green test that proves the weaker property is worse than no test, because the name is what people read — including me, when I wrote it.

## What this adds

`tests/test_launcher_dispatch.py` states the invariant once, over the launcher's **own** roster:

```
for every model_type in _BANNER_MODELS:
    model_arch() must not fall through to the "glm" default (except glm)
    engine_for() must not raise, and must not collide with another family
```

Self-updating by construction: adding a family to the banner without teaching the dispatcher fails at PR time rather than shipping and being found by someone who downloaded a release. There is a guard asserting the roster is non-empty, so the loops cannot pass vacuously if `_BANNER_MODELS` is ever renamed.

The release check now does the same against the **packaged** launcher: it loads `coli` out of the archive, hands it a fabricated `config.json` per family, and asserts where the request would go. Presence is still checked — it is step 1 of 2 rather than the whole thing.

## Current output on `dev`

```
DeepSeek V4 Flash    model_type=deepseek_v4  -> arch=deepseek_v4
Inkling              model_type=inkling      -> arch=inkling
Kimi K3              model_type=kimi         -> arch=kimi
OLMoE                model_type=olmoe        -> arch=glm      <-- #879
GLM-5.2              model_type=glm          -> arch=glm
```

```
AssertionError: 'glm' == 'glm' : the banner prints 'OLMoE' for model_type 'olmoe'
and model_arch() returns 'glm' -- the launcher names a family it cannot dispatch

AssertionError: GLM-5.2 (glm) and OLMoE both dispatch to 'colibri' -- one of them
is being misrouted
```

That is the message a maintainer would have got on the v1.5.0 tag build.

## Division of work

@shujaanazhar traced #879 to the line, found the working invocation, and offered the launcher fix. That is theirs: `c/coli` — `model_arch()`, `engine_for()`, `env_for_engine()`.

This is ours: `.github/workflows/release.yml` plus a new test. **No overlapping files, no conflict, and no need to coordinate beyond merge order.**

Refs #879, #868

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -185,13 +185,44 @@ jobs:
           # now, and one hardcoded list that silently omitted an engine is what #858
           # was. Keep the expectation in a single place, above.
           COLI_EXPECT="$SIBLINGS" python3 - <<'PYCHK'
-          import os, sys
+          import importlib.machinery, importlib.util, json, os, sys, tempfile
           exe = ".exe" if os.name == "nt" else ""
           expect = os.environ["COLI_EXPECT"].split()
+
+          # 1. the files are here
           missing = [n for n in expect if not os.path.exists(os.path.join(".", n + exe))]
           if missing:
-              sys.exit("FAIL: coli would not resolve these next to itself: " + ", ".join(missing))
-          print("OK: every engine is where coli's engine_for() looks for it:", " ".join(expect))
+              sys.exit("FAIL: engines missing from the archive: " + ", ".join(missing))
+
+          # 2. and coli can actually SELECT them.
+          #
+          # This check used to stop at step 1 and was named "coli would not resolve
+          # these next to itself", which reads as a dispatch check and is not one.
+          # It passed for the whole of v1.5.0 while `coli run --model <olmoe>` sent
+          # the request to the GLM engine, because model_arch() had no olmoe branch
+          # (#879). The binary was in the archive and unusable.
+          #
+          # Presence is not dispatch. Load the packaged launcher, hand it a
+          # config.json per family, and assert where the request would go.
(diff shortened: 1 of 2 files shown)


terrizoaguimor:
@JustVugg I think I have a solution for this one.

JustVugg:
Ready. @terrizoaguimor's #881 landed, so the invariant this file asserts is now satisfied:

```
DeepSeek V4 Flash    deepseek_v4  -> arch=deepseek_v4  engine=deepseek_v4
Inkling              inkling      -> arch=inkling      engine=inkling
Kimi K3              kimi         -> arch=kimi         engine=kimi_k3
OLMoE                olmoe        -> arch=olmoe        engine=olmoe       <-- was 'glm'
GLM-5.2              glm          -> arch=glm          engine=colibri
```

Four tests, all green, and the strengthened release check passes against the same table.

The point of merging it is what happens **next** time: adding a family to `_BANNER_MODELS` without teaching `model_arch()` about it now fails at PR time, rather than shipping in a release archive and being found by @shujaanazhar trying to run the model.

PR #891, opened 2026-08-08
fix: report Vulkan expert residency in telemetry and brain map

MasterCATZ:
## What

The web UI brain map and `/health` telemetry showed no Vulkan VRAM usage even when Vulkan experts were resident on AMD. This happened because:
- `emap_emit()` only marked CUDA experts as VRAM tier; Vulkan fell through to disk.
- `tiers_emit()`/`hwinfo_emit()` counted only `COLI_CUDA` GPU stats.
- `/experts` was auth-gated, so the UI could not poll the map directly.

## Why

Vulkan residency lives in a separate registry (`vk_reg_at`), not in `m->pin`/`m->ecache`, so the existing tier-classification logic never saw it.

## Fix

- `c/telemetry.h`: Vulkan-aware `tiers_emit`, `hwinfo_emit`, and `emap_emit`.
- `c/colibri.c`: set `vk_eligible=1` on registered Vulkan expert tensors.
- `c/openai_server.py`: make `/experts` public.

## VRAM budget note

`VK_EXT_memory_budget` under-reports device-local budget on RX 6000-series RADV. The existing `COLI_VK_IGNORE_BUDGET=1` override is unchanged. `COLI_VK_RESERVE_GB` is still user-settable and remains the real reserve limit.

## AI-generated

This PR was generated with AI assistance.

### DIFF
--- a/c/backend_vulkan.c
+++ b/c/backend_vulkan.c
@@ -163,7 +163,9 @@ static int alloc_hostvis_mt(size_t bytes, VkBuffer *buf, VkDeviceMemory *mem, vo
 void coli_vk_alloc_priority(float p) { G.prio = p < 0 ? 0 : p > 1 ? 1 : p; }
 
 /* Device-local heap usage/budget in GB (VK_EXT_memory_budget). Returns 0 when the
- * extension is absent — callers then keep their count-based caps unchanged. */
+ * extension is absent — callers then keep their count-based caps unchanged.
+ * Set COLI_VK_IGNORE_BUDGET=1 to use raw heap size as the budget instead of the
+ * driver-reported heapBudget, so COLI_VK_RESERVE_GB remains a real limit. */
 int coli_vk_mem_budget(double *used_gb, double *budget_gb) {
 #ifdef VK_EXT_memory_budget
     if (!G.has_budget || !G.phys) return 0;
@@ -175,7 +177,10 @@ int coli_vk_mem_budget(double *used_gb, double *budget_gb) {
     double u = 0, b = 0;
     for (uint32_t i = 0; i < mp2.memoryProperties.memoryHeapCount; i++)
         if (mp2.memoryProperties.memoryHeaps[i].flags & VK_MEMORY_HEAP_DEVICE_LOCAL_BIT) {
-            u += (double)bud.heapUsage[i]; b += (double)bud.heapBudget[i];
+            u += (double)bud.heapUsage[i];
+            b += getenv("COLI_VK_IGNORE_BUDGET")
+                     ? (double)mp2.memoryProperties.memoryHeaps[i].size
+                     : (double)bud.heapBudget[i];
         }
     if (used_gb) *used_gb = u / 1e9;
     if (budget_gb) *budget_gb = b / 1e9;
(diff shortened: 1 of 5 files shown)


JustVugg:
The diagnosis is right and I want this fixed: Vulkan residency really does live in a separate registry (`vk_reg_at`), so `emap_emit()` never saw it and the Brain map showed disk for experts sitting in VRAM. #892 is a real bug and the `vk_eligible` plumbing plus the new `[VK] tier summary` line are both good.

Two things stop it going in as it stands.

## 1. It reverts a published security advisory

```diff
-  # Liveness is always public; hardware/scheduler internals only when a
-  # request is authed (or no key set), so a configured key isn't leaked
-  # past a bare 200 to an unauthenticated probe. (#SEC-8)
-  if self._is_authed():
-      payload["scheduler"] = self.server.scheduler.snapshot()
-      payload["kv_slots"] = self.server.kv_slots
```

and on `/experts`:

```c
if self._is_authed() and eng and getattr(eng, "emap", None):   # (#SEC-8) hide routing telemetry unless authed
```

**Both of those gates are advisory SEC-8**, published with v1.5.0 three days ago and credited to its reporter. Opening them makes scheduler state, KV slot count, tier sizes and the full routing telemetry readable by anything that can reach the port, on a server the user may have bound beyond localhost.

The problem you hit is real — the dashboard cannot poll a gated endpoint. But the direction is inverted: **the dashboard should authenticate, not the endpoint should open.** The page is served by the same process that holds the key; it can send it. If that turns out to be awkward in the current UI, say so and we will work out where the key should live, but silently undoing a published advisory to make a panel populate is not a trade we can make.

## 2. CI

11 of 17 checks are failing. I have not dug into which, because the SEC-8 question has to be settled first and the answer changes the diff.

## What I would take today

Split it. The `c/backend_vulkan.*`, `c/colibri.c` and `c/telemetry.h` half — Vulkan-aware `tiers_emit`/`hwinfo_emit`/`emap_emit`, `vk_eligible` on registered tensors, the tier summary line — is the actual fix for #892 and touches no auth. Send that on its own, green, and it goes in.

The `openai_server.py` half is a separate conversation about how the dashboard authenticates, and it is worth having properly rather than as a side effect.

Thank you for the `VK_EXT_memory_budget` note on RX 6000 RADV — that under-reporting is exactly the kind of thing that is invisible until someone with the card writes it down.

JustVugg:
Closing this as landed, and the diagnosis in it was right.

Checked against the current dev tree rather than against the merge list: `vk_eligible` is in `c/colibri.c` in three places, `vk_reg_served` in seven, and `c/telemetry.h` now has ten Vulkan-aware references. #1235 carried the telemetry half and the plumbing followed. The bug you found, that `emap_emit`/`tiers_emit`/`hwinfo_emit` were CUDA-only and reported nothing for a Vulkan tier, is fixed.

The one piece not taken is making `/experts` public, and that is deliberate rather than an oversight: `c/openai_server.py` now gates it behind `_is_authed()` with a comment tying it to a security item. Routing telemetry tells an unauthenticated caller which experts a prompt touched.

Thank you for the report.

JustVugg:
Adding one thing to the close above, because it matters more than the supersession and I missed it first time round.

This diff also removes the `_is_authed()` gate from both `/health` and `/experts` in `c/openai_server.py`, deleting the comments that tie them to a security item. That is an unauthenticated information-disclosure change, it is unrelated to the Vulkan telemetry the PR is about, and it is almost certainly an artifact of an old base rather than something you intended.

Flagging it so that nobody merges this branch later on the strength of the title. The Vulkan half is already on dev; the auth gates are staying.

Issue #892, opened 2026-08-08
[Bug]: Vulkan Brain Map not reported


MasterCATZ:
### Commit or version

dev

### Environment

 Vulkan free VRAM space not being reported correctly and the Brain Map was not using Vulkan telemetry
looks like only cuda was in the list ? 

I did have AI play around with a fork to get it working ,
 not 100% sure the AI did the pull request correctly think it just sent it to me ? 

https://github.com/MasterCATZ/colibri/tree/vulkan-brainmap-fix

PR is open.                                                      
                                                                         
    • Repo: JustVugg/colibri              
    • PR: #891 — fix: report Vulkan expert residency in telemetry and brain map by MasterCATZ · Pull Request #891 · JustVugg/colibri · GitHub
    • Branch: MasterCATZ/colibri:vulkan-brainmap-fix                                                                                                             
    • Base: JustVugg/colibri:dev
    • Title: fix: report Vulkan expert residency in telemetry and brain map
    • AI-generated flag: included in PR body
                                                                                                                                                             
    What’s in the PR:                                                                                              

    • c/telemetry.h — Vulkan-aware tiers_emit, hwinfo_emit, emap_emit                                                                                            
    • c/colibri.c — set vk_eligible=1 on registered Vulkan expert tensors                                                                                        
    • c/openai_server.py — /experts public
                                                                                                                                                                 

    VRAM reporting: still not fully correct at the driver level. VK_EXT_memory_budget under-reports device-local budget on RX 6000-series RADV. The PR leaves
    COLI_VK_IGNORE_BUDGET=1 unchanged, and COLI_VK_RESERVE_GB is still manually settable as the real reserve limit.

    Current local branch is at b92aa0c and pushed to your fork.

### Reproduction steps

use Vulkan

### Expected behavior

VRAM usage to be shown 

### Actual behavior and logs

```shell
only CUDA is looked up
```

JustVugg:
Confirmed, and your diagnosis in #891 is right: Vulkan residency lives in its own registry (`vk_reg_at`), not in `m->pin`/`m->ecache`, so `emap_emit()` never saw it and classified those experts as disk-tier.

[...]

Issue #887, opened 2026-08-07
[Bug]: Vulkan not working on deepseek-v4 (Intel GPU)


BranBushes:
### Commit or version

8f512fc8c2f48ffa18cd624cd4a5bcaae4a4abfc

### Environment

OS: Archlinux,
CPU: AMD Ryzen 5 3500X (6) @ 3.66 GHz,
GPU: Intel Arc B570 (10GB VRAM)
RAM: 16GB,
Compiler: gcc 16.1.1 20260728
Storage: NVME 256GB SSD

### Reproduction steps

Steps to reproduce: 
> ➜  c git:(main) make deepseek-v4 VK=1
> ➜  c git:(main) ✗ vulkaninfo | grep deviceName
	deviceName        = Intel(R) Arc(tm) B570 Graphics (BMG G21)
> ➜  c git:(main) ✗ COLI_VULKAN=1 COLI_VK_DENSE=1 COLI_VK_ATTN=1 \
COLI_NO_OMP_TUNE=1 \
PIN=$HOME/Documents/ai-models/DeepSeek-V4-Flash/.coli_usage PIN_GB=0 \
COLI_MODEL=$HOME/Documents/ai-models/DeepSeek-V4-Flash \ 
./coli chat

### Expected behavior

The expected outcome would be it using my gpu to start the inference. Instead it relies only on my cpu and storage.

### Actual behavior and logs

```shell
➜  c git:(main) ✗ COLI_VULKAN=1 COLI_VK_DENSE=1 COLI_VK_ATTN=1 \
COLI_NO_OMP_TUNE=1 \
PIN=$HOME/Documents/ai-models/DeepSeek-V4-Flash/.coli_usage PIN_GB=0 \
COLI_MODEL=$HOME/Documents/ai-models/DeepSeek-V4-Flash \ 
./coli chat

     ▄▀▀▀▄  ▄        colibri v1.5.0
  ▄▄▄▄▀▀▀▀▄▀▀        tiny engine, immense model
      ▀▀▀▀▀▀▀        DeepSeek V4 Flash · 284B MoE · 167 GB on disk
        ▀▀▀▀         chat · deepseek-v4-colibri · local server
          ▀          
  ──────────────────────────────────────────────────────────
  ⠸ loading deepseek-v4-colibri… 4sram_tiers available=11.54GiB dense=streamed(0.00GiB) target_slots=15 target_cache=8.03GiB head=resident-bf16 projected=11.35GiB
[...]

Issue #885, opened 2026-08-07
v1.5.0 leaves ~60 GB of RAM unused and loses 13 points of expert hit rate: 18% slower decode than v1.4.0 on a 128 GB single-GPU host


mohamedmastouri2000-boop:
v1.5.0 is **17–19% slower at decode than v1.4.0** on this box, on both an int4-g64 and an E8/IQ3 container. Same machine, same hour, same frozen usage history, two replicates per arm on each version.

The interesting part is not the throughput, it is the resident set. **v1.4.0 uses 106 GB of the 128 GB available. v1.5.0 uses 47 GB and never comes back for the rest.** Expert hit rate falls with it, 97.4% → 84.4%. The engine is declining memory it has.

## Setup

RTX 5080 / sm_120 / 16 GB · Core Ultra 9 285K · 128 GB DDR5-4400 · SK Hynix PCIe 4.0 NVMe · Windows · GLM-5.2 744B.

Both versions built from source with the same toolchain (nvcc `-arch=sm_120`, `make colibri.exe CUDA_DLL=1 ARCH=native`) — not the release zip, which carries no sm_120 `coli_cuda.dll`.

```
COLI_CUDA=1 CUDA_DENSE=1 CUDA_EXPERT_GB=4 CUDA_RESERVE_GB=1
PIPE=1 DIRECT=1
CACHE_ROUTE=1 ROUTE_J=3 ROUTE_M=16 ROUTE_ALPHA=0.5
USAGE_SAVE=0
--ngen 64 --temp 0, same prompt throughout
```

**Preconditions recorded in both logs, because my first attempt at this was invalid without them** — other GPU tenants had clamped the expert tier and the whole comparison was worthless:

```
VRAM  16303 MiB total, 1739 MiB used, 14239 MiB free
RAM   127.6 GB total / 107.6 GB free
```

`.coli_usage` is restored from a master copy before every single run **and** `USAGE_SAVE=0`, so the history is identical for all fourteen runs and no run can mutate it mid-flight. Warm-up discarded on each arm. v1.5.0 ran first, then the v1.4.0 binaries were reinstalled and the identical script re-run — same session, nothing else changed.

## Result

| arm | v1.4.0 `b085b48` | v1.5.0 `8f512fc` | delta |
|---|---:|---:|---:|
| int4-g64 decode | 1.21, 1.23 → **1.22** | 1.00, 1.01 → **1.005** | **−17.6%** |
[...]

Issue #856, opened 2026-08-06
[Bug]: Possible Performance Regression for GLM5.2 in new 1.5.0 release over 1.4.0 release


brad-evony:
### Commit or version

1.5.0

### Environment

i9-10940x  256gb ram, 4tb nvme ssd, Windows 10 pro

### Reproduction steps

set OMP_NUM_THREADS=14
set COLI_MODEL=E:\AI-MODELS\COLI-GLM52-FP4
python coli web --ram 242 --gpu none --ngen 256

### Expected behavior

Version 1.4.0 gives a speed of 0.7 tokens per second with GLM5.2, but version 1.5.0 gives a speed of 0.5 tokens per second, using the exact same settings and command listed above.  This is a regression, meaning that the performance went down significantly in the newer release.  Perhaps the behavior of one or more of the environment variables or command flags was changed, or perhaps there is a more serious bug involved.

<img width="1182" height="906" alt="Image" src="https://github.com/user-attachments/assets/b1aa6742-f133-4b42-a561-c75cb6531739" />
<img width="1182" height="906" alt="Image" src="https://github.com/user-attachments/assets/5d3ce559-a033-4a07-871b-26764a39adec" />

### Actual behavior and logs

```shell
The 1.5.0 version produced slower output.
```

MasterCATZ:
I was just about to make a bug report 


possibly releated to this as well ? 

# Colibri dev regression: expert-disk I/O ~4x slower on AMD/Vulkan + disk-backed experts

> This issue was drafted with AI assistance from Hermes Agent. The reproduction steps, measurements, and environment details were provided by the reporter; the structure and wording were generated with AI help.

## Summary

On AMD Radeon RX 6800 XT 16GB with Vulkan backend, GLM-5.2 int4 expert service time regressed from ~86s to ~350s between older dev builds and current `origin/dev` (`a54b4da`). The slowdown is concentrated in `expert-disk` service time, not GPU compute.

## Repro

```bash
COLI_VULKAN=1 COLI_NO_OMP_TUNE=1 SNAP=/ZFS3WAY24B/nosnapshot/llm/LM.Studio/colibri/GLM-5.2-int4 PROMPT="Hello" NGEN=32 /mnt/Fluff/colibri/c/colibri /ZFS3WAY24B/nosnapshot/llm/LM.Studio/colibri/GLM-5.2-int4
```

## Hardware

- GPU: AMD Radeon RX 6800 XT 16GB
[...]

PR #893, opened 2026-08-08
Experiment: validate placement and backend residency assumptions

terrizoaguimor:
## Summary

This PR is the local-first Phase 0 and placement-contract work requested before a runtime CNRE admission policy is considered. It combines the offline residency simulator with small fixes that make placement accounting, backend selection, and fallback claims more truthful.

It does not add an external cache service, change routing, add prefetch, or change model outputs by policy. It also does not yet implement a runtime frequency-admission policy.

## Why These Changes Are Connected

Several reports describe the same user-visible failure family: the plan says that a GPU/cache tier is available, but the actual execution falls back to CPU/disk, or the computed budget omits a memory term and fails later.

Related reports:

- #887: Intel Vulkan/DeepSeek V4 reports CPU/storage execution instead of GPU use.
- #892: Vulkan residency is absent from Brain Map/telemetry.
- #813: Metal advertises routed-expert GPU execution but rejects the published fmt=4 gs64 container.
- #585: grouped-int4 Metal attention/backend support remains the broader follow-up.
- #759 and #653: RAM/VRAM accounting on unified-memory devices.
- #766 and #767: planner/runtime memory projection and prompt-dependent fallback.
- #687: lazy CUDA tensor uploads can disable GPU work after the expert tier consumes the card.
- #856: mixed expert widths caused an observed cache-capacity regression, fixed in #869.
- #885: a related v1.5 residency under-utilization report remains under investigation.
- #848: Kimi Vulkan tier fill was gated by RAM misses, fixed in #877.
- #662: AMD/ROCm discovery is the precedent for backend-neutral accelerator discovery.
- #783: Kimi CUDA is not implemented and should not be implied by generic GPU flags.

### Offline simulator

- Strict GLM ROUTE_TRACE parsing and validation.
- First-seen batch unions plus per-row selection multiplicity.
- GLM 64-expert block promotion semantics.
- LRU, LFU/frequency admission, SLRU, and synthetic pin policies.
- Uniform and policy-specific dynamic capacity allocation.
- Held-out category gate: at least 10% mean gain and no category worse than 3%.
- Layer-specific felt-cost sensitivity.
- JSON output retaining specs, allocations, trace sources, gates, and sensitivity.
- 29 focused simulator/backend tests.

### Placement accounting

- `resource_plan.py` now records per-layer expert widths and maximum width.
- Shared working-set sizing uses maximum expert width instead of median width.
- Unified-memory fixtures constrain RAM and VRAM against one physical pool.
- Explicit RAM requests are reported when clamped by a shared GPU pool.
- `doctor` reports a backend-neutral `accelerator.gpu` check instead of calling every device CUDA.
- Apple Silicon planning prefers performance-core count when available.

### Backend truthfulness

- `COLI_VK_DEV` selects the primary Vulkan physical-device enumeration index for reproducible multi-GPU/hybrid testing.
- DeepSeek V4's Makefile warns that it has no Vulkan backend instead of implying that `VK=1` enables one.
- Metal batched routed MoE accepts grouped int4 and carries the group size into the shader.
- Metal grouped MoE tests cover g64 and g32 ragged execution.
- Metal startup wording now says that device initialization is format-dependent rather than promising universal routed-expert dispatch.

## Real-trace evidence

Before this PR, a DigitalOcean H200 pilot collected 13 GLM-5.2 traces: 8 train and 5 held-out across coding, chat, reasoning, multilingual, and long-context prompts.

The simulator predicted the following held-out felt-wait gains against uniform LRU:

| Expert budget | Candidate | Mean gain | Worst category |
|---:|---|---:|---:|
| 120 GB | Uniform frequency | +39.91% | +34.84% |
| 120 GB | Dynamic frequency | +40.45% | +35.41% |
| 160 GB | Uniform frequency | +53.35% | +49.92% |
| 160 GB | Dynamic frequency | +54.50% | +50.66% |
| 200 GB | Uniform frequency | +65.31% | +62.64% |
| 200 GB | Dynamic frequency | +66.29% | +63.07% |

These are offline predictions, not runtime speedup claims. The pilot also showed why backend and overlap telemetry must be correct: `RAM_GB=120, PIPE=0` measured 1.66 tok/s at 57.2% hit, while `RAM_GB=120, PIPE=1` measured 2.00 tok/s at the same hit rate. Hit rate alone is not the objective.

## Verification

- Full Python suite: `407 passed, 57 skipped`.
- `make metal-test`: passed, including grouped-int4 batched MoE.
- Focused planner/doctor/backend tests: passed.
- `python3 -m py_compile`: passed.
- `git diff --check`: passed.

## Scope and follow-up

The simulator and methodology are intentionally included so the next runtime policy can be measured against the real held-out gate rather than selected on synthetic or in-sample wins.

The next PR should be a separate, small, GLM-only opt-in admission policy with no routing change, no new prefetch, fixed memory budget, and ABBA measurements for tok/s, p50/p99 latency, felt wait, physical bytes, admissions, rejections, evictions, migration bytes, and output identity.

The Intel Vulkan report in #887 still needs hardware confirmation. This PR improves device selection and removes misleading generic CUDA wording, but it does not claim DeepSeek V4 Vulkan support that the engine does not currently implement.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,11 @@ c/tiny_inkling/
 
 # Claude Code working directory (agent worktrees, session scratch) — never a repo artifact
 .claude/
+
+# CNRE local experiment outputs (the source tools and methodology are tracked;
+# generated traces, profiles, and result dumps are not repository artifacts).
+cnre-results/
+c/cnre-results/
+*.trace
+*.coli_usage
+.coli_kv*
--- a/c/Makefile.deepseek-v4
+++ b/c/Makefile.deepseek-v4
@@ -62,6 +62,13 @@ V4_OBJS = $(TARGET_OBJS)
 .PHONY: deepseek-v4 deepseek-v4-clean deepseek-v4-test-objs
 deepseek-v4: $(V4_BINARY)
 
+# Vulkan support is currently implemented by the GLM/Kimi sibling engines, not
+# by the amalgamated DeepSeek V4 runtime. Keep `make deepseek-v4 VK=1` from
+# implying that COLI_VULKAN=1 can dispatch V4 work.
+ifeq ($(VK),1)
+$(warning DeepSeek V4 does not currently include a Vulkan backend; VK is ignored)
+endif
+
 $(V4_BINARY): $(V4_OBJS)
 	$(CC) $(V4_OBJS) -o $@ $(LDFLAGS)
 
--- a/c/backend_metal.h
+++ b/c/backend_metal.h
@@ -139,12 +139,13 @@ int coli_metal_resset_stats(double *flush_s);
  *  g/u/d[e]    = pointers to expert e's gate/up/down quantized weights (in RAM slabs)
  *  gs/us/ds[e] = pointers to expert e's scales (per-row or per-group)
  *  fmt         = quant format (shared across experts), including fmt=4 grouped int4.
+ *  group_size  = common fmt=4 group size (0 for non-grouped formats).
  *  xg          = packed activations [total_rows, D]; xoff[e] = row offset of expert e
  *  nr[e]       = rows for expert e; rows[]/rw[] map packed rows back to out positions
  *  out         = [S, D] accumulate target
  * Returns 1 on success, 0 to signal the caller to fall back to the CPU path.
  */
-int coli_metal_moe_block(int nb, int D, int Iinter, int fmt,
+int coli_metal_moe_block(int nb, int D, int Iinter, int fmt, int group_size,
                          const void *const *g, const void *const *u, const void *const *d,
                          const float *const *gs, const float *const *us, const float *const *ds,
                          const float *xg, const int *xoff, const int *nr,
@@ -159,7 +160,7 @@ int coli_metal_moe_block(int nb, int D, int Iinter, int fmt,
  * end returns 0 on GPU fault (caller redoes those experts on CPU).
  */
 typedef struct ColiMetalMoeHandle ColiMetalMoeHandle;
-ColiMetalMoeHandle* coli_metal_moe_block_begin(int nb, int D, int Iinter, int fmt,
+ColiMetalMoeHandle* coli_metal_moe_block_begin(int nb, int D, int Iinter, int fmt, int group_size,
                          const void *const *g, const void *const *u, const void *const *d,
                          const float *const *gs, const float *const *us, const float *const *ds,
                          const float *xg, const int *xoff, const int *nr,
--- a/c/backend_metal.h
+++ b/c/backend_metal.h
@@ -137,12 +137,8 @@ int coli_metal_resset_stats(double *flush_s);
  *
  *  D           = hidden size, Iinter = moe intermediate size
  *  g/u/d[e]    = pointers to expert e's gate/up/down quantized weights (in RAM slabs)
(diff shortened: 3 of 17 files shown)


JustVugg:
Green on all 17, and the individual observations look sound. I am still going to ask you to split it, and the reason is in your own summary:

> Several reports describe the same user-visible failure family: the plan says that a GPU/cache tier is available, but the actual execution falls back to CPU/disk.

That several reports share a *family* is an argument for splitting, not for bundling. #887, #892, #813, #585, #759, #653, #766, #767, #687 are nine issues, on four backends, reported by different people on hardware none of us has. One PR that touches `colibri.c`, `inkling.c`, `backend_metal.mm`, `backend_vulkan.c`, `resource_plan.py` and `doctor.py` and adds a simulator means:

- **no reporter can confirm their own issue.** @BranBushes cannot tell whether his Intel Vulkan case is fixed without also carrying the Metal changes; @ThefloorMiner cannot A/B the Metal claim without the Vulkan ones. The people who can actually verify are each blocked on code irrelevant to them.
- **a revert is all-or-nothing.** If one of the nine turns out wrong on hardware we do not have, undoing it takes the other eight with it.
- **and 2,060 lines across four backends is more than I can review honestly.** I would be approving the ones I can reason about and waving through the rest, which is exactly how #856 and #885 got in — both from a single constant nobody enumerated the consumers of.

The natural cut is the one your description already makes:

1. **The offline residency simulator** (`tools/residency_sim.py`, its test, `docs/experiments/`) — self-contained, adds no runtime behaviour, and can go in on its own merits today.
2. **Placement/accounting truthfulness** (`resource_plan.py`, `doctor.py`, `docs/ENVIRONMENT.md`) — testable here, no hardware needed.
3. **Vulkan** (#887, #892) — @BranBushes and @MasterCATZ can confirm. Note #891 is also touching Vulkan telemetry; worth coordinating so you two are not writing the same thing twice.
4. **Metal** (#813, #585) — @ThefloorMiner and @RDouglasSharp have the hardware, and #587 is already green and touching `backend_metal.mm`.

Send 1 and 2 and I will take them without ceremony. 3 and 4 want their reporters attached.

One correction while you are here, since you cite it: #856 was fixed in #869, but the same constant was **also** feeding `autopin_lru_reserve()`, which is #885 — 60 GB of resident set declined on a 128 GB host. Same defect, second consumer, and I did not notice the second one when I fixed the first. If your placement work touches that path, that is the shape to watch for.

terrizoaguimor:
Per your review, I split PR #893 into two focused PRs from `dev`:\n\n1. Offline simulator: https://github.com/JustVugg/colibri/pull/902\n2. Placement/accounting: https://github.com/JustVugg/colibri/pull/903\n\nThe simulator PR contains only `residency_sim.py`, its tests, and the experiment methodology. The placement PR contains only `resource_plan.py`, `doctor.py`, their tests, and the environment documentation.\n\nI intentionally left the Vulkan and Metal changes out of both PRs so #887/#892 and #813/#585 can be coordinated with the reporters and existing PRs #891/#587. The original PR remains open temporarily as the superseded bundle; I will close it after the split PRs are visible.

terrizoaguimor:
Superseded by the focused PRs requested in review: #902 (offline simulator) and #903 (placement/accounting). Backend-specific Vulkan and Metal changes will be proposed separately with the relevant hardware reporters.

Issue #865, opened 2026-08-06
Degradation policy for prefetch misses: zero-fill beats substitution (measured)


outtodata:
**TL;DR**: when a routed expert is not yet loaded, the cheapest safe fallback is to contribute *nothing* (zero the slot), not to substitute a "similar" expert. Measured on OLMoE-1B-7B (ppl, 15% of slots affected): **zero 26.78 < random 30.09 < nearest-neighbor substitution 36.35** (baseline 13.91). Substitution's error is *coherent* across layers and accumulates linearly; random error cancels as ?L. Substituting a neighbor is worse than a random expert - the similarity map lies consistently.

## Numbers (ppl, 15% slots replaced, fixed eval set)

| policy | ppl | note |
|---|---|---|
| exact | 13.91 | baseline |
| zero (skip slot) | 26.78 | best fallback |
| random expert | 30.09 | error cancels as ?L |
| nearest expert (activation-similarity map) | 36.35 | coherent error, accumulates ~linearly |

Zero with a gate-weight threshold is nearly free: slots with w<0.03 (21.8% of all slots) zeroed -> ppl 14.32 (**+2.9%**). w<0.05 -> +41% (too much). The weight threshold, not the count, is the right control.

## Bonus datapoint: the 8->4 bit cliff on OLMoE experts

Same harness: int8 row-absmax = **+0.1%** ppl (13.928), while 4-bit RTN = **+67%** (group-64) / +92% (row-absmax). W_down alone accounts for the damage (+63% when only W_down is 4-bit). Naive RTN at <=4 bits is dead for these experts; int8 (the converter's choice) is validated.

## Proposal

An opt-in degradation flag, e.g. `--degrade=zero` with a weight threshold (default tau=0.03), as the documented-safe fallback when an expert read misses its deadline. Reproduction scripts + logs available.

outtodata:
**Follow-up: completed the quant cliff anatomy + final measured stack**

Since posting, we finished the bit-width map and found the working point. All same harness, baseline ppl 13.9144 validated by a control run, OLMoE-1B-7B-0924:

| scheme | ppl | delta |
|---|---|---|
| int8 RTN | 13.93 | +0.1% |
| 6-bit RTN g64 | 15.09 | +8.5% |
| 5-bit RTN g64 | 17.71 | +27% |
| 4-bit RTN g64 | 23.25 | +67% |
| 4-bit RTN absmax-row | 26.66 | +92% |
[...]

PR #906, opened 2026-08-09
feat(moe): add DEGRADE_ZERO miss-slot zero-fill policy (issue #865)

kritikagarg:
Introduce DEGRADE_ZERO=1 / DEGRADE_TAU=<f> (default 0.03): when a prefetch deadline is missed, experts whose aggregate gate weight across the batch is below tau are zero-filled instead of triggering a blocking demand-load. Issue #865 measures tau=0.03 zeroes 21.8% of slots for only +2.9% perplexity — well within acceptable quality budget on NVMe-bound workloads where the stall cost dominates.

Design constraints respected:
- Decode-only (S<=4 guard): same invariant as EXPERT_BUDGET — dropping experts during prefill corrupts the KV cache (#292)
- Miss-only: pin/LRU hits are never dropped; they load for free
- Rescue rule: no position is left with zero routed experts
- Renormalises surviving weights for both norm_topk=1 (GLM-5.2) and norm_topk=0 (other MoE configs) to preserve output magnitude
- Opt-in only (default OFF): output changes, must never be silent
- Stats footer reports tau and cumulative zeroed slot count when active

Adds tests/test_degrade_zero.c: 8 properties, 20 checks, standalone (no model or weights needed), auto-discovered by the Makefile test gate.

## Summary

## Validation

- [x] `make -C c check`
- [x] CUDA changes were tested with `make -C c cuda-test` (if applicable)

## Compatibility


kritikagarg:
Updated the implementation to address the two semantic points raised in #865:

  **1. Per-position τ gate (was: aggregate)**
  The original code summed gate weight across all batch positions before
  comparing to τ. This coupled positions — at S>1 a slot could survive because
  *other* positions in the batch pushed the aggregate over the threshold, even
  if every individual position was sub-τ. The gate now tests each position's
  own weight independently. At S=1 (the common decode case) the two are
  identical; at S=2–4 the per-position gate is stricter and matches the
  measurement methodology the +2.9% ppl number was taken under.

  **2. No renorm (was: renorm after drop)**
  The previous apply step renormalised surviving weights — either to
  sum-1×routed_scale (norm_topk=1) or by old_sum/new_sum (norm_topk=0). Both 
  paths redistribute the dropped mass upward rather than losing it, which is a 
  different approximation with no measured quality curve. Survivors now keep 
  their original weights unchanged. The approximation is the dropped mass.

  Tests: replaced P4 (renorm-normtopk) and P5 (renorm-nonnorm) with P4 
  (no-renorm: survivor weight is byte-identical to input) and P5 
  (per-position-tau: S=2 shared expert kept when one position meets τ, dropped 
  when all are below). All 20 checks pass.

JustVugg:
Thank you for taking this — @outtodata's #865 is one of the better-measured proposals in the tracker and it deserved an implementation.

I want to read it properly rather than fold it into a batch, because it is the first change here that makes the engine **produce different tokens on purpose**. Everything else in this repo is either exact or explicitly a quantisation choice made at conversion time; this is a runtime decision to skip work and accept the output moving. That is a legitimate capability and #865's numbers support it — zeroing slots below τ=0.03 costs +2.9% ppl on OLMoE while τ=0.05 costs +41%, so the operating point is real and narrow.

Three things I will be looking for, said up front so you can point me at them or fix them before I do:

**Off by default, and loud when on.** A run that silently approximates is indistinguishable from a run that is subtly broken, and six months later someone files a quality bug we cannot reproduce. The `[PROF]` line should report how many slots were zeroed, not just that the flag was set.

**The gate on weight, not on count.** #865 is explicit that the weight threshold is the control and the slot count is not: τ=0.03 zeroes 21.8% of slots for +2.9%, but "zero 21.8% of slots" chosen by any other rule would not give that number.

**A token-exactness assertion with the flag off.** The existing oracles must stay bit-identical, so that enabling this is the only way to change output.

+473 lines across 3 files and CI has not reported yet — I have approved the workflow run. I will come back to it with a real review rather than a checklist.

One thing worth flagging because it affects where this pays off: #865's measurements are on OLMoE, and @outtodata says so plainly. This engine's own profile shows expert **I/O** dominating expert **matmul** by roughly 5:1 on a streaming box, so a policy that avoids waiting for a late read is aimed at exactly the right cost — but the ppl numbers that justify τ=0.03 are one model's. Worth stating in the docs that the default is calibrated on OLMoE, not measured on GLM or Kimi.

outtodata:
Thanks for reading it this way — the three gates you named are exactly the ones the measurements support. Tying each back to the data, plus two places where an implementation can silently drift from them:

**Off by default, loud when on.** Agreed, and the [PROF] counter is mandatory from our side too. One refinement worth asking for: report the zeroed share **per layer**, not just a global count. τ is a global knob but the drop share is not uniform across layers — a global "21.8% zeroed" can hide one layer zeroing far more, and that is where the ppl cost concentrates. A per-layer min/max next to the hit/miss line makes a silently-miscalibrated run visible instead of indistinguishable from a healthy one. I can pull the per-layer drop-share table from our grid run if that helps the review.

**Gate on weight, not on count.** This is the load-bearing point, and I'm glad it's called out up front. The +2.9% at τ=0.03 comes from the *weight threshold selecting which slots to drop* — the slots contributing least to the MoE output under `norm_topk_prob=False`. A count-based rule ("drop 21.8% of slots, chosen any other way") selects a different set, and there is no reason to expect it lands anywhere near +2.9%. Related, from our renorm A/B on the same harness: the reason zeroing is cheap at all is that OLMoE's router contract lets the top-k weights sum to less than 1, so dropping a slot only ever *removes* a small amount of mass — a bounded, subtractive deviation. Renormalizing survivors instead was catastrophically worse at every τ (+6253% ppl at τ=0.03). Keep the flag's semantics exactly "skip slots with gate weight < τ"; any rescaling of survivors changes the regime, not the degree.

**Token-exactness with the flag off.** Should hold naturally: the gate sits before `pipe_dispatch` / the blocking load, so with the flag off the path is untouched and the existing oracles stay bit-identical. The one place to look hard is the PIPE path: a zeroed slot must still produce a well-defined QT view (or be excluded from the matmul consumer list) so the pipelined consumer never reads a half-loaded expert. The serial path is trivial; slot lifetime under PIPE=1 is where this breaks subtly.

**OLMoE-calibrated default.** Fair, and worth stating in the docs exactly as you put it. For the record, the numbers behind τ=0.03: OLMoE-1B-7B, fp16 offload, ppl 13.9144 → 14.3194 (+2.9%), task_acc unchanged at 0.889, ~22% of slots dropped; τ=0.05 → +41% ppl, so the operating point is real and narrow. GLM (`norm_topk=1`) and Kimi have different router contracts and expert counts — until someone measures them, the default should ship documented as calibrated on OLMoE, with per-model τ left to the operator. Happy to re-run any τ point or ablation if you want one more data point before merge.

kritikagarg:
@JustVugg @outtodata — follow-up on the three gates and the open items from this thread.

 **Changes in this update** (commit `37590cc`):

 - **Per-layer drop reporting** (outtodata's request, echoed by @JustVugg):
 `g_degrade_dropped_by_layer[layer]` now accumulates per-layer, and the `[PROF]` footer shows
 the top-3 layers by drop count alongside the global total — e.g. `| DEGRADE_ZERO tau=0.030
 (zeroed 1234 miss slots; top layers: L12:234 L8:198 L31:156)`. A miscalibrated tau that
 concentrates drops in one layer is now visible rather than hidden behind a global number.

 - **Stale comment fix**: the `g_degrade_zero` declaration comment said "aggregate gate weight"
 — left over from before `ecd369c`. Corrected to "per-position gate weight" to match the
 implementation.

 - **`docs/ENVIRONMENT.md`**: added entries for `DEGRADE_ZERO` and `DEGRADE_TAU` with the
 measured numbers, the per-position / no-renorm semantics, and an explicit note that tau=0.03 is
 OLMoE-calibrated.

 - **`docs/tuning.md`**: added a `DEGRADE_ZERO` section with the tau table, the OLMoE-calibrated
 caveat (@JustVugg's ask — "worth stating in the docs that the default is calibrated on OLMoE,
 not measured on GLM or Kimi"), and a pointer to ENVIRONMENT.md.

 **PIPE path**: the DEGRADE_ZERO block compacts `uniq[]` before the FASE C/D phase where `nmiss`
 and `pipe_dispatch` are computed — dropped experts are never submitted to the pipe workers.

 **All three of @JustVugg's gates**: off by default with a `[DEGRADE]` announcement when on;
 gate is on per-position weight not count; P9 asserts token-exact output with the flag off.

  (GLM-5.2 tau sweep tracking in #865.)

outtodata:
@kritikagarg Solid update — the per-layer drop reporting is exactly what was needed. L12/L8/L31 concentration is visible now, which is the difference between "21.8% zeroed overall" and "L12 is silently eating 3× the median drop share."

The docs additions (ENVIRONMENT.md + tuning.md with the OLMoE-calibrated caveat) close the loop on the maintainer's ask.

One remaining question from my side, same as in #865: any cold-start A/B planned? The current numbers are all warm-cache; on a cold start with an empty ecache, DEGRADE_ZERO at τ=0.03 might behave differently (more candidates, less resident mass to protect). Not a blocker for merge, but worth documenting as a known gap if you don't have the data.

From my side: approved, pending CI green.

kritikagarg:
@outtodata  Good catch on the warm-cache assumption — added a paragraph to `docs/tuning.md` and a note to the
  `ENVIRONMENT.md` entry in 9a39018 documenting that the calibration numbers are steady-state and
  that cold-start transient behavior hasn't been characterized. Thanks for the approval!

PR #914, opened 2026-08-09
fix: account real row widths in pin budgets

bherald:
## Summary

Fixes the remaining mixed-width `PIN_GB` GB-to-count conversion identified in #885.

- Preserve the popularity-ranked prefix.
- Charge each candidate using its real per-layer `expert_bytes_row()` width.
- Keep the transient `CUDA_RELEASE_HOST` VRAM prefix outside the disjoint RAM budget.
- Record and report resident/warm bytes using the selected experts' actual widths.
- Extend the mixed-width regression fixture to reproduce the widest-divisor under-pinning and verify budget and ranking boundaries.

This does not change the Kimi K3 engine; K3 uses its separate `kimi_k3.c` cache and pin-seeding path.

Addresses #885 without closing it pending the reporter's hardware sweep.

## Validation

- [x] `make -C c check`
- [ ] CUDA changes were tested with `make -C c cuda-test` — no CUDA-capable Linux host was available; the `COLI_CUDA` conditional path passes compilation/syntax validation.
- [x] Performance claims include hardware, commands, and repeatable measurements — not applicable; this PR makes no new performance claim and leaves final throughput validation to the reporter's established #885 sweep.

## Compatibility

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -8128,6 +8128,33 @@ static int pin_rec_cmp(const void *a,const void *b){
     const PinRec *x=a,*y=b; return x->c<y->c?1:x->c>y->c?-1:0;
 }
 
+/* A pin budget buys a popularity-ranked PREFIX, but mixed-width containers do
+ * not have one honest bytes-per-expert divisor.  GLM-5.2's routed rows are int4
+ * while its MTP row is int8: pricing every candidate at the widest row made a
+ * 51.1 GB budget buy the same 1,352 experts in both g64 and E8 containers even
+ * though their routed experts cost only 21.2 and 14.6 MB (#885).
+ *
+ * Keep the ranking semantics -- never skip a hot wide candidate to admit a
+ * colder narrow one -- and stop immediately before the first candidate that
+ * would cross the byte ceiling.  `from` lets CUDA_RELEASE_HOST exclude the
+ * transient VRAM prefix before spending the disjoint RAM budget. */
+static double pin_range_bytes(Model *m, const PinRec *r, int from, int to){
+    double used=0.0;
+    for(int a=from;a<to;a++) used+=(double)expert_bytes_row(m,r[a].l,m->ebits);
+    return used;
+}
+static int pin_count_for_budget(Model *m, const PinRec *r, int from, int n,
+                                double budget_b){
+    double used=0.0; int count=0;
+    if(budget_b<=0.0 || from>=n) return 0;
+    for(int a=from;a<n;a++){
+        double need=(double)expert_bytes_row(m,r[a].l,m->ebits);
+        if(need<=0.0 || used+need>budget_b) break;
+        used+=need; count++;
+    }
+    return count;
+}
+
 #ifdef __linux__
 /* #419: bind the pinned hot-store as ONE arena per layer instead of one mbind
  * per slab. Per-slab policies cost ~2 unmergeable VMAs each; a PIN_GB=all load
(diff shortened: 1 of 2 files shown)


bherald:
Thank you @mohamedmastouri2000-boop for the controlled frozen-history sweeps, especially the exact 51.1 GB / 37.8 MB count match that isolated the remaining call site changed by this PR. Thanks @JustVugg as well for #869 and #899, which made the budget terms observable.

If you are willing, the same g64/E8 sweep against this branch would be the strongest hardware validation before merge.

mohamedmastouri2000-boop:
Hardware validation for this branch, as asked. Full write-up with preconditions, controls and raw profile numbers is in #885; the short version:

**Placement is fully restored on both containers.** g64 pins 2406 experts (v1.4.0: 2431), E8 pins 3536 (v1.4.0: 3588), RSS 105.3 / 104.7 GB against v1.4.0's 105.8 / 106.5, hit rate 93.5% / 97.6% against 93.3% / 97.4%. The divisor is per-container again — 21.2 MB and 14.5 MB — and warm bytes now equal the plan instead of falling short. `pin`/`lru` split, `swap %`, `route_agree` and `route_kl` all land on v1.4.0 within replicate noise.

**Measured against its own merge-base** (dev @47a749d built without `bc99ebf`, same script, same session):

| | dev @47a749d | **this PR** |
|---|---:|---:|
| g64 decode | 0.95, 0.99 | **1.02, 0.99** |
| E8 decode | 1.23, 1.22 | **1.28, 1.27** |
| g64 RSS | 84.67 GB | **105.3 GB** |
| E8 RSS | 75.43 GB | **104.7 GB** |
| g64 hit | 88.7 / 89.8% | **93.5 / 93.1%** |
| E8 hit | 91.9 / 91.8% | **97.6 / 97.3%** |

+21 GB and +29 GB resident, +5 points of hit rate, ~+4% decode, nothing regressing. **From this box, merge it.**

Decode is still ~19% below v1.4.0, but that is a second and unrelated regression in the expert matmul path — it is present at `47a749d` *without* this commit and absent at `0f4598b`. Detail and the ruled-out candidates are in #885. It is not this PR.

Box: RTX 5080 / sm_120 / 16 GB · Core Ultra 9 285K · 128 GB DDR5-4400 · Windows · GLM-5.2 744B, int4-g64 and E8/IQ3 containers.

mohamedmastouri2000-boop:
**Correction to my validation comment above.** In it I said decode was still ~19% below v1.4.0 and blamed a second regression in the expert matmul path. **I am withdrawing that half.** Full detail in #885; the short version is that I rebuilt the "fast" anchor `0f4598b` on this box today and it read 29.3 / 29.1 s of `expert-matmul` against the 23.6 / 23.5 s the same commit gave in the earlier session — same source, same script, same resident set, same hit rate. The split I reported was between sessions, not between commits, and I had lined up numbers from different sessions in one table.

Ruled out along the way: a background process pinning one core (killed it, no change), build flags (byte-identical `gcc` line), and the launcher (same binary, `c/coli` swapped between revisions, no difference). I also stated in that comment that the harness never goes through the launcher — that was wrong, it calls `python coli run` with `COLI_ENGINE` pinned, which is why I tested it rather than leaving the claim standing.

**Nothing changes for this PR.** Its evaluation was always same-session — this branch and its merge-base `47a749d` built and run an hour apart on the same box:

| | dev @47a749d | **this PR** |
|---|---:|---:|
| g64 decode | 0.95, 0.99 | **1.02, 0.99** |
| E8 decode | 1.23, 1.22 | **1.28, 1.27** |
| g64 RSS | 84.67 GB | **105.3 GB** |
| E8 RSS | 75.43 GB | **104.7 GB** |
| g64 hit | 88.7 / 89.8% | **93.5 / 93.1%** |
| E8 hit | 91.9 / 91.8% | **97.6 / 97.3%** |

Placement fully restored, +21 / +29 GB resident, +5 points of hit rate, ~+4% decode, nothing regressing. **Merge it.** The v1.4.0 comparison is a separate question and I do not currently have a measurement of it I would defend.

Issue #848, opened 2026-08-05
[Bug]: Kimi-K3 Can not allocate nearly all free VRAM


Limalski:
### Commit or version

1.4.0

### Environment

Debian 13.6, Ryzen 5 5600G, GeForce RTX 4070 Ti & RTX 5060 Ti (connected to oculink dock). 

There seems to be some 4,3 GB maximum with Kimi-K3, even I have plenty (10-11 GB) free VRAM. if I specify K3_VK_GB=3 it goes below 4.3 GB and is quite close to 3 GB, but if I try to go K3_VK_GB=5 or more, I get only 4.3 GB (4360 MB to be precise) NVRAM allocated.

It is always that at maximum. And I can see that coli gives parameter "8" for kimi_k3 binary. I not sure what is that, it can't be context size, because larger contexts work fine.

### Reproduction steps

Compile Kimi with Vulkan support
Then give command

test 1
COLI_TEMP=0.65 NUCLEUS=0.9 K3_THINK=0 K3_VK=1 OMP_NUM_THREADS=4 K3_MAXT=16384 K3_VK_GB=2 K3_VK_UP=4 K3_EXPERT_GB=16 c/coli chat --model /kimi --ngen 10000

test 2
COLI_TEMP=0.65 NUCLEUS=0.9 K3_THINK=0 K3_VK=1 OMP_NUM_THREADS=4 K3_MAXT=16384 K3_VK_GB=8 K3_VK_UP=4 K3_EXPERT_GB=16 c/coli chat --model /kimi --ngen 10000

### Expected behavior

test 1
expected to reserve 2 GB of VRAM

test 2
expected to reserve 8 GB of VRAM

### Actual behavior and logs

```shell
test 1
works as intended 

test 2
does not work, reserves only 4360 MB of VRAM (seen with nvtop)

4360 MB seems to be some kind of limit, but no error is given.

More VRAM in my system is easily allocated by using GLM (with CUDA) or llama.cpp.

All values larger than K3_VK_GB=4 allocate the same 4360 MB.
```

Limalski:
More info when starting. Kimi K3

  ⠦ loading kimi-k3-colibri… 272s[VK] VRAM pressure-proofing: memory_priority on, memory_budget on
[VK] ready: NVIDIA GeForce RTX 4070 Ti, compute qfam 0, memtype 4, fused gate+up, absorb attention
[...]

PR #917, opened 2026-08-10
Scan vulkan devices in reverse order, to avoid main GPU

xxxajk:
## Summary

Chose a Vulkan device in reverse order, so that the main display GPU is not chosen by default in a multi-GPU setup.
Simple one-line fix.

## Validation

- [X] `make -C c check`

## Compatibility


JustVugg:
One line, and it changes device selection for everyone:

```diff
- for (uint32_t i = 0; i < count; i++)
+ for (int i = (int)count - 1; i >= 0; i--)
```

The problem is real — on a box with an iGPU and a discrete card, or two discrete cards, Vulkan enumerates in an order the driver chooses and colibri takes the first. If that is your display GPU, you get VRAM contention with the desktop. @Limalski hit an adjacent version of this in #848.

But reverse order is not a fix, it is a different arbitrary choice. It works on your machine because your enumeration happens to put the card you want last. On the next reporter's machine it will be first again, and the one-line change will need reverting for them.

What the engine should do is **pick**, not iterate:

- prefer `VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU` over `INTEGRATED_GPU` over the rest — the property Vulkan exposes precisely so callers do not have to guess from ordering
- break ties on device-local heap size, which is what actually matters for an expert tier
- and honour an explicit override, so someone with two identical cards can still say which

The last of those may already be most of the way there: please check whether `COLI_VK_DEVICE` or similar exists before writing anything, because selecting by index would let you solve your case today without changing anyone else's default.

If you want to take the typed selection, I would take that PR. If you would rather just report the environment, post `vulkaninfo | grep -A2 deviceType` and the enumeration order you see — with two of those from different vendors the heuristic writes itself, and right now nobody here has a multi-GPU Vulkan box to test on.

3 of 5 checklist items and CI has not reported yet; I have approved the workflow run so it will.

xxxajk:
> One line, and it changes device selection for everyone:
> 
> ```diff
> - for (uint32_t i = 0; i < count; i++)
> + for (int i = (int)count - 1; i >= 0; i--)
> ```
> 
> The problem is real — on a box with an iGPU and a discrete card, or two discrete cards, Vulkan enumerates in an order the driver chooses and colibri takes the first. If that is your display GPU, you get VRAM contention with the desktop. @Limalski hit an adjacent version of this in #848.
> 
> But reverse order is not a fix, it is a different arbitrary choice. It works on your machine because your enumeration happens to put the card you want last. On the next reporter's machine it will be first again, and the one-line change will need reverting for them.
> 
> What the engine should do is **pick**, not iterate:
> 
>     * prefer `VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU` over `INTEGRATED_GPU` over the rest — the property Vulkan exposes precisely so callers do not have to guess from ordering
> 
>     * break ties on device-local heap size, which is what actually matters for an expert tier
> 
>     * and honour an explicit override, so someone with two identical cards can still say which
> 
> 
> The last of those may already be most of the way there: please check whether `COLI_VK_DEVICE` or similar exists before writing anything, because selecting by index would let you solve your case today without changing anyone else's default.
> 
> If you want to take the typed selection, I would take that PR. If you would rather just report the environment, post `vulkaninfo | grep -A2 deviceType` and the enumeration order you see — with two of those from different vendors the heuristic writes itself, and right now nobody here has a multi-GPU Vulkan box to test on.
> 
> 3 of 5 checklist items and CI has not reported yet; I have approved the workflow run so it will.

How about an option to choose which device then?

```
        deviceType        = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
        deviceName        = AMD Radeon RX 5700 XT (RADV NAVI10)
        pipelineCacheUUID = 81083e57-af5d-e002-1230-8be7d6da98cf
--
        deviceType        = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
        deviceName        = Tesla P40
        pipelineCacheUUID = 7cebaaca-c4fb-2b03-58ee-f08a3305aec7
--
        deviceType        = PHYSICAL_DEVICE_TYPE_CPU
        deviceName        = llvmpipe (LLVM 20.1.2, 256 bits)
        pipelineCacheUUID = 32352e32-2e38-2d30-7562-756e7475302e
```

xxxajk:
I propose options can cure the issues, leaving defaults as is, and it's pretty simple.
A flag that specifies which GPU to use, if the GPU is not there, or does not have the resources, we fall back to scanning.

The scanning can rank each choice based on a few metrics, and choose the best one as an additive ranking.
During scan each ranking can simply add 1, and the highest overall value wins.
Each criteria could contain:
Is this CPU only? Skip it, that's the default.
Can the device do compute? Add 1
Dedicated, no video (like the P40)? Add 1
Minimum amount of free VRAM available that we can work with? add 1

When there is a tie amongst multiple devices, the device with the most free VRAM wins.
If there is STILL a tie, choose the first one.

So that means:
An if flag, we can default to something that's not going to be ever present, -1 for example would cause it to fall through to scanning.
A new int or byte array to hold the scanned ranking.
No extra loops, we only collect capabilities with a few if statements.
Finally three if statements for the scan validation, breaks any the ties and if nothing is good, we just use CPU.

That's how I think it could be solved, with minimal effort, and it would provide a few safety checks.

JustVugg:
@xxxajk — eight days without an answer to a design proposal you wrote in good faith is on us, sorry. Here is where it stands, with the part you will want to know first: **half of what you proposed is already on `dev`.**

`backend_vulkan.c` now ranks devices instead of taking `devs[0]`:

```c
int rank = p.deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU   ? 4 :
           p.deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU ? 3 :
           p.deviceType == VK_PHYSICAL_DEVICE_TYPE_VIRTUAL_GPU    ? 2 :
           p.deviceType == VK_PHYSICAL_DEVICE_TYPE_OTHER          ? 1 : 0; // CPU last
```

That is your "is this CPU only? skip it / can it do compute? +1 / dedicated? +1" additive ranking, in the shape it ended up taking — and it means the **original symptom of this PR is likely fixed for the iGPU + discrete case**, without either arbitrary enumeration order. Worth testing on your box before anything else: if `dev` already picks the right card for you, the remaining work is smaller than when you wrote the proposal.

**What is still missing from your design, and is still wanted:**

1. **An explicit primary-device selector.** There is `COLI_VK_DEV2` for the second-device expert tier, but nothing to say "use device N as the primary" — so a user with two *discrete* cards (your P40 case, or two identical GPUs) still cannot choose, because the type ranking ties. This is the piece that actually needs your proposal.
2. **Free VRAM as the tie-break.** The `VK_EXT_memory_budget` extension is already detected at init for allocation pressure — the budget it exposes is exactly the number a tie-break would rank on, so the plumbing exists.
3. **"Dedicated, no display output"** as a distinct signal from `DISCRETE_GPU`. A P40 and a display-attached 4090 both report discrete; only the first is free of desktop contention. Vulkan does not expose "has a display" directly, so this one needs thought — that is a design question worth its own comment rather than an assumption.

**So: this PR as written should close** — the one-line reverse is superseded by the ranking, and we agreed it was a different arbitrary choice rather than a fix. **Your proposal should become a new PR**, scoped to (1) and (2): an explicit selector plus a free-VRAM tie-break on top of the existing ranking. That is small, testable, and does not need (3) to be useful.

Would you like to write it, or would you rather we do and have you test it on the multi-GPU box? Either is fine — you found the problem and designed the fix, so the choice is yours. Closing this one only when you have said which.

xxxajk:
Cool, no rush, I'll give it a look asap, right now my GPUs are involved in a long horizon task.

xxxajk:
BTW, finding out if the GPU is compute only, I did a search, and this is what I found:
To determine if a Vulkan physical device has video output (display presentation capability) versus being compute-only, inspect its queue families using vkGetPhysicalDeviceQueueFamilyProperties and check for presentation support via vkGetPhysicalDeviceSurfaceSupportKHR.

A compute-only device exposes queue families with compute/transfer flags but lacks support for presenting images to a window or screen surface. 

HTH :-)

ZacharyZcR (commented):
The loop bound has an off-by-one that changes the semantics beyond the stated goal: `for (uint32_t i = nd-1; i > 0; i--)` never evaluates `devs[0]` at all — the condition excludes index 0, so device 0 is not merely deprioritized on ties, it is excluded from ranking entirely. Two concrete consequences:

- On a machine where `devs[0]` is the **only discrete GPU** (common: dGPU enumerates first, iGPU second), this selects the integrated GPU unconditionally — the rank comparison that is supposed to prefer discrete never sees the discrete device.
- `nd == 1` works only by accident: `i = 0`, loop body never runs, `bestrank` stays `-1`, and the `devs[0]` fallback catches it.

If the intent is 'on equal rank, prefer the later-enumerated device' (avoid the display GPU), the full-coverage form does that with no exclusion:

```c
for (uint32_t i = nd; i-- > 0;) {
```

— iterates nd-1 .. 0 inclusive, and with the existing strict `>` comparison the first-seen (highest-index) device wins ties, which is exactly the avoid-main-GPU behavior this PR wants, while a lone discrete GPU at index 0 still outranks everything. An explicit `VK_DEVICE=<index>` override might serve the underlying use case (compute on the non-display GPU) more predictably than enumeration-order heuristics, but that's the maintainer's call; the bound fix stands either way.

xxxajk:
@ZacharyZcR nice catch.

JustVugg:
Thanks for this, and the intent is right: on a multi-GPU box the display GPU should not win by default. The one-line form does not do it though, for two reasons.

**The loop never reaches device 0.**

```c
for (uint32_t i = nd-1; i > 0; i--)
```

`i > 0` stops before `i == 0`, so `devs[0]` is never ranked. With two devices, a discrete GPU at index 0 loses to an integrated one at index 1. With a single device the body never runs at all: `G.phys = devs[0]` still holds from the line above and the result happens to be right, but nothing was actually ranked. If the goal were only to reverse the tie-break, the loop would have to be written so that index 0 is still visited, for example `for (uint32_t i = nd; i-- > 0;)`.

**Reverse order is not the same as avoiding the display GPU.** Enumeration order is not defined to put the display adapter first, so on another machine this reverses the tie-break without changing which device is picked, or picks the display GPU anyway.

If the aim is "prefer a GPU that is not driving a display", that wants an explicit signal rather than an index heuristic: an env override naming the device, or a rank penalty for the adapter that owns the presentation surface. An override is also testable, which an ordering heuristic is not.

Happy to look at a revised version. Leaving this open rather than merging it, since as written it changes behaviour on single-GPU machines too.

JustVugg:
Closing this one, with the reason rather than silence. Two things: the loop as written (`for (i = nd-1; i > 0; i--)`) never examines device 0, so a machine whose only discrete card enumerates first would lose it; and the selection the PR wants already exists as an explicit knob, `COLI_VK_DEV=<index>` (docs/ENVIRONMENT.md), which picks the device by enumeration index without changing the default ranking for everyone else. If the default ranking itself is wrong on your machine (a discrete card ranked below an integrated one), that is a bug worth its own issue with the `[VK] ready:` line, and I would take that fix.

xxxajk:
Just a quick comment here, the second selection for vulkan gpu doesn't seem to do anything.
plan, doctor, etc also don't show anything on vulkan. There's still no "good way" to choose 2 vulkan GPUs, but that's really not so important. On my system, the gpu access is actually slower than running on cpu, because I do have enough ram to do everything, and the extra hop to gpu incurs an extra latency penalty. 
The only case that I have observed that has a benefit is if you don't have enough system ram, and that's fine.
Once Qwen3.8-Flash-Next gets GPU support, that particular model might benefit, but nobody knows, so it is an unknown unknown.

Issue #910, opened 2026-08-09
[Bug]: Last line of reply disappears in chat, Kimi-K3


Limalski:
### Commit or version

47a749d8225b815adc9622c964ba386639320973

### Environment

Debian 13, development branch, Kimi-K3.

### Reproduction steps

Make a query, copy paste the reply from console to your favorite editor.
After query ends the last line of Kimis reply disappears. (painted in picture)

<img width="1687" height="413" alt="Image" src="https://github.com/user-attachments/assets/e322a6db-18ad-4f89-a651-5c17db0f7561" />

The following reply is not correct imho (but that is a different thing and not related to Colibri) 

### Expected behavior

Expected to see full query after it is finished.

### Actual behavior and logs

```shell
See attachment photo.
```

JustVugg:
Reproduced by reading, but I do not have the full diagnosis yet and would rather say where I got to than guess.

The TUI's markdown renderer buffers a partial line and only emits on a newline, so a reply that ends without a trailing newline leaves its last line in the buffer. There is a `close()` that flushes exactly that, and it *is* called — but at two specific sites: `cmd_run`, and the **GLM branch** of `cmd_chat`.

You are on Kimi K3, which does not take that branch: non-GLM chat goes through `openai_server.py` and a different display loop. That is where I stopped, and it is the obvious place for a missing flush.

Two things that would narrow it fast if you have a minute:

1. Does the same happen on `coli web` in the browser, or only in the terminal? If the browser shows the full reply, it is the TUI path and my reading above is probably right.
2. Does the missing line reappear if the model's answer happens to end with a blank line?

You are also on `dev` at `47a749d`, which included a change to the web chat pane but not to the terminal renderer, so this is not new breakage from that — worth ruling out explicitly since the timing invites the assumption.

Thank you for the screenshot with the missing line marked; "copy the console into an editor" is a good repro because it separates what was printed from what the terminal is showing.

JustVugg:
Fixed by #942 (dcutugno) — the markdown stream now closes before the spinner erases the line. Root cause matched the diagnosis above: the TUI renderer emits only the delta past what it already printed, so the spinner's `\r\033[K` erased the partially-printed tail and `close()` never re-emitted it — gone from screen, present in history, exactly what you captured. Functionally verified under a pty with the real classes: old order loses the tail, new order keeps it. In dev now, in the next release within days. Thanks for the clean report and the picture.

PR #942, opened 2026-08-11
fix(cli): close the markdown stream before stopping the spinner

dcutugno:
## Problem

In `coli chat` (HTTP mode), the tail of a streaming reply visibly **vanishes the moment the `~N tok · Ns` footer appears** — the answer looks cut off mid-sentence, yet the token count and saved history still include the missing text.

## Root cause

`chat_attached` in `c/coli` called the spinner stop *before* closing the markdown stream:

```python
sp.stop(); md.close()
```

`Spinner.stop()` writes `\r\033[K` (carriage return + erase-to-end-of-line). Since a reply rarely ends with a newline, the cursor still sat on the reply's **last partial line**, so `sp.stop()` erased that whole line. `md.close()` then had nothing to re-print: `MDStream._line()` only emits `line[self.printed:]`, and that text was already marked as printed.

The final partial line of the reply is therefore wiped from the display exactly when the summary prints. `Spinner.stop()` is safe *only* when it runs on a line that is already terminated — the GLM serve chat already used the correct order (`md.close()` before `sp2.stop()`); the HTTP path had it reversed.

## Fix

```python
md.close(); sp.stop()
```

`md.close()` terminates the partial line with a newline first; `sp.stop()` then clears only the empty line and can no longer erase reply text.

Introduced by `8bf4cb9a` ("coli: chat attaches to a running serve").

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -917,7 +917,7 @@ def chat_attached(a, base, model_id):
         except OSError as e:
             sp.stop()
             print(f"\n  {C.yel}[server unreachable: {e}]{C.r}"); break
-        sp.stop(); md.close()
+        md.close(); sp.stop()
         if reply: msgs.append({"role":"assistant","content":"".join(reply)})
         else: msgs.pop()                              # turno vuoto: non sporcare la history
         el=time.time()-t0


JustVugg:
Verified beyond reading: drove the real `MDStream` + `Spinner` classes under a pty with a minimal terminal interpreter (`\r`, `\033[K`, `\n` semantics), feeding a reply that ends without a newline:

```
old order (sp.stop; md.close):  tail visible = False
new order (md.close; sp.stop):  tail visible = True
```

The mechanism matches your diagnosis exactly — `_line()` emits only the delta past `self.printed`, so once the spinner's erase removes the partially-printed line, `close()` re-emits only the held-back marker characters and the text is gone from screen while the history keeps it. This is the same defect reported in #910 on Kimi chat, so this PR closes that too.

CI was waiting on first-contributor approval — approved, running now. Merging on green.

JustVugg (approved):
Functionally verified under a pty with the real MDStream/Spinner classes (see comment above): old order loses the tail, new order keeps it. One-line fix, correct diagnosis, closes #910 as well.

PR #944, opened 2026-08-11
fix: reject GLM top-k larger than expert count

benmaster82:
## Summary

- Reject GLM configurations where `num_experts_per_tok` exceeds `n_routed_experts`.
- Add regression coverage for valid and invalid expert and top-k combinations.
- Register the new test in the portable C test gate.

## Why

Routing data structures are sized using the number of routed experts. Allowing the configured top-k value to exceed that count can cause out-of-bounds access in downstream routing code.

The configuration loader is the central validation point, so invalid configurations are now rejected before model initialization continues.

## Testing

- `make -C c test-c`: passed
- `make -B -C c tests/test_cfg_topk.exe`: passed
- `tests/test_cfg_topk.exe`: passed
- `git diff --check`: passed

The Python portion of `make check` reports unrelated Windows environment and toolchain failures in backend-loader fixtures, one resource-plan fixture that raises `MemoryError`, and FP8 tests with Windows path or compiler-flag issues. The complete C test gate passes.


JustVugg (approved):
Verified against dev before the PR existed: the loader reads topk and n_experts with no relational check, and topk > n_experts drives the selection loop to best=-1 → si[kk]=-1 → lg[-1], an out-of-bounds read from a corrupt or malicious config — squarely inside the untrusted-model-file threat model. The exit(1) style matches the loader's existing validations, and forking a child per case is the right way to test a rejection that exits. Green across the matrix.

PR #945, opened 2026-08-11
fix(metal): rename test-local fp8_nblk to avoid quant.h collision

gouravkargwal:
Fixes #940

The test file defined its own `static int fp8_nblk` while `quant.h` now provides `static inline int64_t fp8_nblk`. Same name and parameter list with different return types breaks compilation of the Metal test (`make metal-test`): "functions that differ only in their return type cannot be overloaded".

Renamed the test-local helper to `ref_fp8_nblk`. Header API and production behavior are untouched.

Reproduced and verified on Apple M3, 16 GB: `make metal-test` passes in full.

Issue #940, opened 2026-08-11
[Bug]: make metal-test failed


xeonai44:
### Commit or version

make metal-test failed

### Environment

In the c folder: make metal-test
clang++ -x objective-c++ -std=gnu++17 -fobjc-arc -O3 -c backend_metal.mm -o backend_metal.o
make: Circular colibri <- colibri dependency dropped.
clang -O3 -Xclang -fopenmp -I/opt/homebrew/opt/libomp/include -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function -DCOLI_METAL colibri.c  backend_metal.o  -o colibri -lm -L/opt/homebrew/opt/libomp/lib -lomp -framework Metal -framework Foundation -lc++
ganzomini@192 c % make metal-test
clang++ -x objective-c++ -std=gnu++17 -fobjc-arc -O3 tests/test_backend_metal.mm backend_metal.mm -framework Metal -framework Foundation -o backend_metal_test
In file included from tests/test_backend_metal.mm:321:
tests/../quant.h:483:23: error: functions that differ only in their return type cannot be overloaded
  483 | static inline int64_t fp8_nblk(int n){ return ((int64_t)n + FP8_BLOCK - 1) / FP8_BLOCK; }
      |               ~~~~~~~ ^
tests/test_backend_metal.mm:99:12: note: previous definition is here
   99 | static int fp8_nblk(int n){ return (n+127)/128; }
      |        ~~~ ^
1 error generated.
make: *** [metal-test] Error 1

### Reproduction steps

build for macos, in c folder  _make colibri METAL=1_ and than _make metal-test_

### Expected behavior

finish the tests with success

### Actual behavior and logs

```shell
make metal-test
clang++ -x objective-c++ -std=gnu++17 -fobjc-arc -O3 -c backend_metal.mm -o backend_metal.o
make: Circular colibri <- colibri dependency dropped.
clang -O3 -Xclang -fopenmp -I/opt/homebrew/opt/libomp/include -Wall -Wextra -Wno-unused-parameter -Wno-misleading-indentation -Wno-unused-function -DCOLI_METAL colibri.c  backend_metal.o  -o colibri -lm -L/opt/homebrew/opt/libomp/lib -lomp -framework Metal -framework Foundation -lc++
ganzomini@192 c % make metal-test
[...]

PR #947, opened 2026-08-11
ci: run Metal backend tests on macOS

gouravkargwal:
**Add the Metal backend test suite to macOS CI.**

The Metal backend tests are the standalone `metal-test` target (`c/Makefile:622`), which compiles and runs `backend_metal_test`. They are **not** part of `make check`: `TEST_BINS` is derived from `tests/test_*.$(EXE)` build rules, and the metal target does not match that pattern. So neither the `check.yml` macOS job (`make check`) nor the `engines-all-platforms` build matrix has ever executed them — which is how issue #940 (a compile error in `test_backend_metal.mm`) shipped with CI green.

This adds a macOS-only step after the engine build:

```yaml
- if: matrix.os == 'macos-latest'
  name: Metal backend tests
  run: |
    cd c
    make metal-test
```

GitHub's `macos-latest` runners are Apple Silicon with a Metal device, so the suite runs there as-is (libomp is already installed for the build step).

**Expected CI status: red until #945 lands.** The new step fails on current `main` with the exact #940 regression — that failure is the proof the gap was real. It goes green once the fix in #945 merges. If you'd prefer a single green PR, I can fold the #940 fix (commit `6ee57c3`) into this branch and close #945 — just say the word.

**Proof — reproduced locally on Apple M3, 16 GB, clang 17.0.0:**

`make metal-test` on current `main` (before #945):
```
tests/../quant.h:483:23: error: functions that differ only in their return type cannot be overloaded
tests/test_backend_metal.mm:99:12: note: previous definition is here
1 error generated.
make: *** [metal-test] Error 1
exit_code=2
```

`make metal-test` with the #940 fix (after #945):
```
top8 E=257 (>256, auto-serial-fallback) ok (serial==parallel bitwise, S=1 E=257)
metal backend tests: ok
exit_code=0
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,17 @@ jobs:
             echo "::endgroup::"
           done
           exit $rc
+      - if: matrix.os == 'macos-latest'
+        name: Metal backend tests
+        # The Metal backend suite is its own target (metal-test) and is not
+        # part of `make check`: TEST_BINS is derived from tests/test_*.$(EXE)
+        # rules, and backend_metal_test is built/run by the metal-test target
+        # itself. Nothing in CI exercised it until issue #940 shipped a
+        # compile error in it, so run it explicitly on the Apple Silicon
+        # runners.
+        run: |
+          cd c
+          make metal-test
 
       # The Windows backend-DLL loader contract (c/tests/test_backend_loader.py)
       # had no CI anywhere: the `python` job below is Linux-only, where these
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,11 @@ jobs:
         # itself. Nothing in CI exercised it until issue #940 shipped a
         # compile error in it, so run it explicitly on the Apple Silicon
         # runners.
+        # The runner's Apple Paravirtual device hangs on Metal submissions
+        # (observed: 47 min vs a 5-6 min normal run), so the test harness
+        # detects it and skips GPU execution with an explicit line; this
+        # timeout is the backstop that turns any residual hang into a red X.
+        timeout-minutes: 10
         run: |
           cd c
           make metal-test
--- a/c/tests/test_backend_metal.mm
+++ b/c/tests/test_backend_metal.mm
@@ -4,6 +4,7 @@
 // reference (cpu_ref_grouped) and harness (run_grouped) below -- unlike fmt 1-3 the
 // group scale is per-GROUP, not per-row, so it can't share cpu_ref's [O] scale layout.
 #include "../backend_metal.h"
+#include <Metal/Metal.h>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
@@ -650,6 +651,17 @@ static int run_attn_grouped(int S, int pos_base, int gs, const char* name){
 
 int main(void) {
   if (!coli_metal_init()) { printf("Metal unavailable (skipping)\n"); return 0; }
+  /* GitHub Actions Apple Silicon runners expose Metal through the Apple
+   * Paravirtual device, where Metal submissions never complete (hangs -- the
+   * #947 CI observation). The shader compile above (coli_metal_init) still
+   * runs, keeping this suite's compile coverage -- the class of bug #940
+   * shipped -- while GPU execution is honestly skipped here. (#947 review) */
+  id<MTLDevice> dev = MTLCreateSystemDefaultDevice();
+  if (dev && [[dev name] containsString:@"Apple Paravirtual device"]) {
+    printf("SKIPPED: paravirtual device (%s) -- compile-only\n",
+           [[dev name] UTF8String]);
+    return 0;
+  }
   printf("Metal backend kernel tests:\n");
   int fail=0;
   fail |= run(I8, 2048,6144,1, "int8 gate/up S=1");


JustVugg:
This is the right fix for the right reason — #940 shipped precisely because `metal-test` matches no `tests/test_*` rule and therefore no CI gate ever ran it; your diagnosis of the `TEST_BINS` derivation is exact. CI approved and running.

Two things this review is watching:
1. **Does the runner's paravirtual GPU actually satisfy the suite?** GitHub's Apple Silicon runners expose Metal through the Apple Paravirtual device; if any test in the suite needs capabilities it lacks, the honest follow-up is a device-capability skip *inside* the test, not dropping the CI step.
2. **Merge sequencing**: #946 (arm64 macOS DeepSeek V4) edits the same region of `ci.yml`; whichever merges second rebases. Trivial either way.

If your step goes green on the runner, this merges — it's the vaccine for the entire class of bug you just fixed a duplicate of.

gouravkargwal:
The failing check is the step working as designed. `CI / All engines (macos)` failed at 32s inside the new step, with the exact #940 regression:

```
clang++ -x objective-c++ -std=gnu++17 -fobjc-arc -O3 tests/test_backend_metal.mm backend_metal.mm -framework Metal -framework Foundation -o backend_metal_test
tests/../quant.h:483:23: error: functions that differ only in their return type cannot be overloaded
  483 | static inline int64_t fp8_nblk(int n){ return ((int64_t)n + FP8_BLOCK - 1) / FP8_BLOCK; }
tests/test_backend_metal.mm:99:12: note: previous definition is here
```

That is the same error my local "before" run produced on main — the step now catches this class of regression in CI within seconds, which is the entire point of the PR.

On the two review points:

1. **Paravirtual GPU**: agreed — this failure is at compile time, so it has not yet exercised the runner's paravirtual Metal device. The capability question gets its real test once the fix lands. If any test then fails on the paravirtual device, I'll add a device-capability skip inside the test (per your policy) rather than touch the CI step.

2. **Merge sequencing with #946**: happy to rebase whenever needed.

Green path: the step stays red until the #940 fix (your `c7aaaf7`, shipping in v1.6.0) reaches `main`. Two options from here:
- **Rebase on `main` after v1.6.0 lands** → step goes green automatically, PR becomes merge-ready (my recommendation, keeps release hygiene intact), or
- **Fold the identical fix commit into this branch now** to go green immediately — same code as `c7aaaf7`, zero risk, and I'll drop the now-closed #945.

Either way, once green on the runner I'll follow up on the paravirtual capability question with the full suite results.

JustVugg:
Half-confirming your reading: the red **is** the proof of value — the step caught the #940 error class in 32 seconds, which is exactly what it exists to do. But the reason it fired is that your branch is based on `main`, where the bug still lives; the rename (`ref_fp8_nblk`, `c7aaaf7`) has been on `dev` for six days — same root as your #945 duplicate.

Done on our side: retargeted the PR to `dev` (the repo's flow — everything lands there first) and updated the branch with dev's contents. The step now compiles against the fixed test, so the next CI round finally answers the real open question: whether the runner's Apple Paravirtual device satisfies the suite at *runtime*. If it does, this merges; if a capability is missing, the follow-up is the in-test skip we discussed.

JustVugg:
The open question answered itself, in the worst of the three possible ways: on the runner's Apple Paravirtual device the suite neither passes nor fails — it **hangs**. The macos job sat at 47 minutes against a typical 5–6 (started 15:02, cancelled 15:49), which is the signature of a Metal submission that never completes. I've cancelled the run; while hung it was also holding one of the five macOS runner slots hostage, slowing every other PR's queue.

So the step needs two guards before this can merge, both in the spirit we discussed:

1. **A `timeout-minutes` on the CI step** (e.g. 10) — the backstop that turns a hang into a red X instead of a 6-hour stall.
2. **A device-capability check in the test harness itself**: detect the paravirtual device (`MTLDevice.name` contains "Apple Paravirtual device") and either skip the submission-dependent tests with an explicit `SKIPPED: paravirtual device` line, or run compile-only there. A skip that names its reason keeps the *compile* coverage — which is what #940 needed — while being honest about what the runner cannot execute.

With those two, this PR still delivers its core value: the #940 class (compile breakage in the Metal test TU) dies in CI within seconds, on every PR. The GPU-execution half of the suite stays a real-hardware concern — your M3 from #949 is where that half actually runs.

Not release-gating either way — v1.6.0 proceeds in parallel.

gouravkargwal:
Both guards from your review are in (commit `1654f39`):

1. **`timeout-minutes: 10`** on the step — a hang is now a red X at 10 minutes, not a 47-minute stall holding a runner slot hostage.

2. **Device-capability check in the harness** (`c/tests/test_backend_metal.mm`): after `coli_metal_init()` (shader compile runs first, so the #940-class compile coverage is preserved on every PR), the test queries `MTLDevice.name` and, if it contains `Apple Paravirtual device`, prints `SKIPPED: paravirtual device (...)` and exits 0 — explicit, named, and honest that the runner cannot execute the GPU half.

Verified on real hardware (Apple M3, 16 GB): the suite compiles and passes in full (`metal backend tests: ok`, exit 0); the skip does not fire on a real device, only on the paravirtual runners.

Branch is updated; the run needs your approval on the fork-PR workflow as before.

gouravkargwal:
CI is fully green on the current head (`279379d`), and the macOS job log shows both guards working exactly as designed:

```
make metal-test
SKIPPED: paravirtual device (Apple Paravirtual device) -- compile-only
```

The step compiles the harness (and the runtime shader compile runs first), detects the Apple Paravirtual device, prints the explicit skip, and exits in ~5 s — the macOS job finished in 1m16s with no hang and no runner-slot hostage. The #940-class compile regression still dies in CI within seconds on every PR, which is the value this step exists to deliver.

PR #948, opened 2026-08-11
feat(v4): DSML tool calling for DeepSeek V4 (#916)

### LINKED ISSUE #916: [Feature]: Implement tool use with DeepSeek V4
### Problem

When our agent is calling tools we get the following error:

```
InvalidStatusCodeWithMessage(                                                                                                                                                                          
    400,                                                                                                                                                                                               
    "{\"error\":{\"message\":\"Tool use is not wired up for DeepSeek V4 yet.\",\"type\":\"invalid_request_error\",\"param\":\"tools\",\"code\":\"unsupported_parameter\"}}",
),                                                                                                                                                                                                     
```

It seems like some implementation for DeepSeek is missing.

### Proposed solution

With GLM5.2 tool usage is working fine.

It would be nice to have tool working with DeepSeek V4 too.

### Alternatives considered

Using GLM5.2 instead of DeepSeek V4.
But DeepSeek seems to have some performance advantages.

### Scope and compatibility

Current dev branch on 47a749d8225b815adc9622c964ba386639320973

dcutugno:
Closes #916.

Wire the official DeepSeek-V4 DSML tool format (`encoding/encoding_dsv4.py`) into the OpenAI and Anthropic paths. Tool use was previously rejected for `deepseek_v4` with `"Tool use is not wired up for DeepSeek V4 yet."`

## What this adds

- `render_chat_v4` now accepts `tools`/`tool_choice`: schemas render via the official `## Tools` block onto the first system/developer message (or an empty system message, byte-matching the encoder). Assistant `tool_calls` render as `<｜DSML｜tool_calls>`/`invoke`/`parameter` blocks, and tool results merge into user turns as `<tool_result>` blocks (V4 has no standalone tool role).
- `reasoning_effort` is honored in thinking mode by prepending the official `REASONING_EFFORT_PROMPTS` (high/max) after BOS; developer messages are wrapped in `<｜User｜>` per the encoder.
- New tolerant `parse_dsv4_tool_calls` maps DSML `string=true|false` params back to OpenAI `tool_calls`, drops the block and thinking/EOS markers from content, and cuts at the first DSML marker when output is length-truncated so raw syntax never leaks.
- `parse_arch_tool_calls` dispatches the 3 parse sites; streaming suppression (OpenAI + Anthropic SSE) uses arch-aware tool markers.

## Verification

- Byte-equivalence against the checkpoint's `encoding_dsv4.py`: 8 render cases, 5 parse cases.
- Unit tests: `test_openai_server` 131 OK, `test_anthropic_messages` 27 OK.
- Live tool-call round trip on the GPU engine.

GLM path unchanged.

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -374,6 +374,173 @@ def parse_tool_calls(reply, tools=None):
     return text.strip(), calls
 
 
+# ---- DeepSeek V4 tool calling (DSML) -------------------------------------------------------
+# V4 expresses tool calls as DSML blocks (see encoding/encoding_dsv4.py):
+#   <｜DSML｜tool_calls>\n<｜DSML｜invoke name="fn">\n
+#   <｜DSML｜parameter name="k" string="true">v</｜DSML｜parameter>\n</｜DSML｜invoke>\n</｜DSML｜tool_calls>
+# preceded by "\n\n". There is no standalone "tool" role: results are <tool_result>{content}
+# </tool_result> blocks merged into the following user turn. DSML = U+FF5C (｜) + ASCII.
+DSV4_DSML = "｜DSML｜"
+DSV4_EOS = "<\uff5cend\u2581of\u2581sentence\uff5c>"
+_DSV4_BLOCK_RE = re.compile(
+    r"<" + DSV4_DSML + r"tool_calls>(.*?)</" + DSV4_DSML + r"tool_calls>", re.DOTALL)
+_DSV4_INVOKE_RE = re.compile(
+    r"<" + DSV4_DSML + r"invoke name=\"(.*?)\">(.*?)</" + DSV4_DSML + r"invoke>", re.DOTALL)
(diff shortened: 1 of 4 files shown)


JustVugg (approved):
This is the shape #916's reply asked for, delivered in full: template rendering (declaration placement, tool_choice none, developer wrapping, thinking-mode interplay), the round trip (assistant tool_calls → DSML → back to OpenAI shape), and — the part that earns the merge — the failure modes as first-class tests: `test_parse_truncated_block_leaks_no_markers` is exactly the 'plausible but corrupted' class that made tool-use the deliberate 400 until now. The parser is defensively built: non-greedy DOTALL matching, every `json.loads` guarded with fall-back-to-string, and streamed marker hold-back so partial tags never leak into visible text. 18/18 green. If the live model's DSML drifts from the fixtures, the parser degrades to text rather than corrupting arguments — that is the property that makes shipping this safe.

PR #951, opened 2026-08-11
docs: add Apple M3 16 GB OLMoE datapoint row

gouravkargwal:
Closes #949

Adds the base-M3 datapoint to the benchmark table in docs/benchmarks.md — the table had M5 Max 128 GB rows but no base M3 (16 GB unified), the low end of what runs this project on Apple Silicon.

**Row added:**

| machine | disk | config | result |
|---|---|---|---|
| Apple M3 (base, 4P+4E) · macOS · **16 GB unified** · internal Apple SSD ([#949](https://github.com/JustVugg/colibri/issues/949)) | 3.18 GB/s cold F_NOCACHE (post-eviction) · 7.27 GB/s buffered — no O_DIRECT on macOS, caveat #86 | **OLMoE int8** · cap 16 · TEMP=0 · CPU-only | **3.69 tok/s cold → 4.18 tok/s warm** · RSS 1.5–1.81 GB · load 0.7 s |

Measured with the one-command datapoint runner (c/tools/datapoint.py, separate PR); two capped 128-token runs agree to 0.01 tok/s (4.09 / 4.08 in the runner's latest run — the table uses the values captured at commit time). Full machine detail and measurement notes are in #949.

Issue #86, opened 2026-07-12
[Bug]: iobench test command output misleading


mkiesel:
### Commit or version

(current)

### Environment

iobench (without O_DIRECT, i.e., cached) does random reads from a file with bounded size, which means its reported speed depends on installed memory size.

*With* O_DIRECT, it's better, but macOS doesn't support O_DIRECT (which leads to weird reported I/O speeds like the current 14.2 GB/sec one - the Apple M5 Max in reality manages about 4GB/sec).

### Reproduction steps

(see above)

### Expected behavior

I/O speed measurements should not depend on RAM size.

### Actual behavior and logs

```shell
-
```

JustVugg:
Sharp catch, and you're right on both counts — thank you. Fixed in the README (928fd67):

**1. macOS has no O_DIRECT.** iobench already falls back to `fcntl(F_NOCACHE)`, but F_NOCACHE only stops *new* caching — it can't evict pages a prior buffered run already resident-mapped in the unified buffer cache. So a macOS "O_DIRECT" figure measured right after a buffered run is still reading cache. **The 14.2 GB/s M5 Max number in the README was exactly this artifact** — I've corrected it to ~4 GB/s cold with a note (the tok/s and hit-rate in that row are unaffected; only the disk column was wrong).

**2. Buffered reads on a big-RAM box measure the page cache, not the disk**, because iobench reads a bounded ~1 GB shard that fits entirely in RAM on re-read. Added a caveat block to the README explaining: use the O_DIRECT run for a true number, on a shard you haven't touched this session, and on macOS reboot / use a fresh shard for a real cold read.

The deeper reason this matters for the project: the *engine's* actual workload streams 370 GB of experts that mostly won't be cached, so the relevant disk number is always the cold/random one — which is exactly what iobench's O_DIRECT path (arg 1) is trying to isolate. Your report makes that honest. If you're on an M-series and can post a true-cold number (fresh shard, first read), it goes straight into the corrected table.

monotophic:
*Authored by Fable 5 in Claude Code, analysis in partnership with @Monotophic.*

For the record on this closed issue, since future readers will land here from the README history:
a measured datapoint from Apple hardware quantifying exactly how large the page-cache illusion was,
and confirmation that the `dev` fix closes it.

**Coordinates:** measured 2026-07 on a MacBook Pro M5 Max (128 GB, internal Apple NVMe AP4096Z),
macOS 26.5, engine `main` @ `72d3d37`.

[...]

PR #952, opened 2026-08-11
tools: add one-command datapoint runner

gouravkargwal:
Adds `c/tools/datapoint.py` — a stdlib-only, one-command runner for the standard machine datapoint that docs/benchmarks.md asks for (engine load, capped greedy decode tok/s, iobench disk figures), printing a ready-to-paste markdown block.

What it measures, and how:

- **Exact decode tok/s** — generation is capped at `--max-new` tokens, so the token count is known without re-tokenizing output. TEMP=0 greedy.
- **True cold vs warm** — before measuring, the page cache is evicted (macOS `purge` if permitted, otherwise a portable temp-file write larger than RAM). iobench cold runs right after eviction, before decode re-warms the container. `--no-evict` skips it.
- **Disk** — iobench 19 MB × 64, 8 threads, cold + buffered, with the macOS F_NOCACHE caveat #86 called out in the emitted block.
- Machine info auto-detected (CPU, RAM, cores, OS) per platform.

docs/benchmarks.md Quick Start now points at it as step 3 (steps renumbered).

The M3 datapoint row it produced is in a separate PR (docs: add Apple M3 16 GB OLMoE datapoint row, #951).

### DIFF
--- a/c/tools/datapoint.py
+++ b/c/tools/datapoint.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+"""datapoint.py - run the standard colibri machine datapoint.
+
+Measures what docs/benchmarks.md asks for - engine load, capped greedy
+decode tok/s, and iobench disk figures - and prints a ready-to-paste
+markdown datapoint block.
+
+The decode number is exact: generation is capped at --max-new tokens,
+so the token count is known without re-tokenizing the output.
+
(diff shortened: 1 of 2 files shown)


JustVugg (approved):
One-command datapoint runner — tooling only, doesn't touch the engine. Lowers the bar for the hardware datapoints we keep asking for.

PR #953, opened 2026-08-11
ci: stop running every dev commit twice, and bound every job

JustVugg:
**Measured, on a single commit (`33302b8`): 36 jobs where 18 suffice.**

```
CI     event=pull_request  -> 14 jobs
CI     event=push          -> 14 jobs   <- exact duplicate
check  event=pull_request  ->  4 jobs
check  event=push          ->  4 jobs   <- exact duplicate
```

**Cause**: the release PR is `dev -> main`, so its head *is* `dev`. With `main` listed under `pull_request`, every push to dev fired both events on the same SHA and ran the whole matrix twice. That is half the project's CI wall-clock, and it is also why release PRs show every check name doubled.

Dropping `main` from the PR trigger costs nothing: check runs attach to the **SHA**, so the release PR still displays the push run's results. `main`'s ruleset requires a **review**, not status checks, so nothing merge-gating is lost — and a PR mistargeted at main now visibly gets no CI, which is exactly the signal that it needs retargeting to dev.

Two more gaps, both things `check.yml` has had all along and `ci.yml` never did:

- **`concurrency` + `cancel-in-progress`** — superseded runs were never cancelled; they ran to completion burning slots nobody was waiting on.
- **`timeout-minutes` on all 12 jobs** — #947's macOS job hung *after* its step failed and held one of the five macOS runner slots for ~50 minutes, queueing every other PR behind it. Budgets (15–30) are hang backstops against a measured worst case of ~10 min, not targets.

**No check is removed and no job's content changes.** The same work runs, once instead of twice, with an upper bound.

Held for the maintainer's go-ahead until after the v1.6.0 tag — `release.yml` is a separate file and is untouched either way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,14 +2,31 @@ name: CI
 
 on:
   pull_request:
-    branches: [dev, main]
+    # dev only, deliberately. The release PR is dev->main, so its head IS dev:
+    # with `main` listed here every push to dev ran the whole matrix TWICE on
+    # one commit -- once for `push`, once for the open release PR -- 36 jobs
+    # where 18 suffice, and the duplicate pair is what filled release PRs with
+    # doubled check names. Check runs attach to the SHA, so the release PR
+    # still displays the push run's results. main carries no required status
+    # checks (its ruleset requires a review), so nothing merge-gating is lost,
+    # and a PR mistargeted at main visibly gets no CI -- which is the signal
+    # that it needs retargeting to dev.
+    branches: [dev]
   push:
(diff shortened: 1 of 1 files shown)


JustVugg:
Parked as draft at the maintainer's call — the CI is fine as it is for now. The findings stay recorded here for whenever it's worth picking up: 36 jobs per dev commit where 18 suffice (the release PR's head *is* dev, so `push` and `pull_request` both fire on one SHA), no `concurrency` on `ci.yml`, and no `timeout-minutes` on any of its 12 jobs.

PR #956, opened 2026-08-11
fix(api): cancel a generation that disconnects before the first frame (#908)

### LINKED ISSUE #908: The colibri server is still running with high CPU when the Anthropic EP is disconnected
The server should detect disconnection and shut down properly to wait for another connection.
Same might be true for the OpenAI EP.

Version: 1.5
FreeBSD 15.1

gouravkargwal:
Fixes #908 — the pre-first-token half of the disconnect spin, per the mechanism @terrizoaguimor pinned down in the thread.

**Root cause:** `Engine.generate()` only polled the `cancelled()` callback in the `kind == "data"` branch. A client that disconnects while the engine is still prefilling (before the first `DATA` frame) therefore never sent `CANCEL <id>` — the engine kept generating to its token limit and the generate thread stayed blocked in `events.get()` until the engine emitted something.

**Fix** (c/openai_server.py): the wait loop now uses `events.get(timeout=0.05)` and, while idle, polls `cancelled()`; a pre-first-frame disconnect sends the same `CANCEL <id>` the data branch would have, and the loop resolves via the engine's `CANCELLED` error exactly as before. `stopped()`/decode behavior is untouched; the idle poll costs nothing on the happy path (frames still return immediately).

**Regression test** (c/tests/test_openai_server.py): submit to `FakeProcess`, flip `cancelled()` to true before any `ACCEPT`/`DATA` frame, assert `CANCEL <id>` is written and the generate thread unblocks with `ClientCancelled` — model-free, mirroring the existing post-first-frame cancel test.

**Verified:** `test_openai_server.py` 123/123 ok; full `test_*.py` suite 430 tests OK (35 dependency skips).

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -1723,7 +1723,21 @@ class Engine:
                     on_accept(info)
 
         while True:
-            kind, value = events.get()
+            try:
+                kind, value = events.get(timeout=0.05)
+            except queue.Empty:
+                # #908: cancelled() is only polled in the "data" branch, so a
+                # client that disconnects before the engine's first DATA frame
+                # (it is still prefilling) never cancels: the CANCEL never went
+                # out, the turn ran to its token limit, and this thread stayed
+                # blocked until the engine emitted something. Poll the callback
+                # while idle so a pre-first-frame disconnect cancels too.
+                if not cancel_sent and not stop_sent and cancelled and cancelled():
+                    cancel_sent = True
+                    with self.write_lock:
+                        self.process.stdin.write(f"CANCEL {request_id}\n".encode())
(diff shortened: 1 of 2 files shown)


JustVugg (approved):
Verified the scope and the mechanism: `cancel_sent`/`stop_sent` are defined before the loop (openai_server.py:1710) and the Empty-handler mirrors the existing CANCEL send exactly (same write_lock, same frame). The 0.05s poll only spins while the engine is silent — 20 wakeups/s during prefill, nothing after the first DATA frame. The fake-engine test pins the exact regime (#908): emits nothing until it sees CANCEL, so the pre-first-frame disconnect must cancel or the test hangs.

PR #958, opened 2026-08-11
deepseek_v4: let the launchers delegate OpenMP team sizing to the runtime

Blakeolson21:
Follow-up to #897, prompted by @terrizoaguimor's report on the merged policy: both launch funnels (c/coli env_for_engine and c/openai_server.py tune_child_env) synthesize OMP_NUM_THREADS from physical cores before V4 starts, so v4_omp_reserve_loader_cpus() treats the launcher default as an explicit user override and returns. The merged sizing never runs on the launch paths users normally take.

This change leaves OMP_NUM_THREADS unset for deepseek_v4 in both funnels while preserving explicit user values, COLI_NO_OMP_TUNE=1, the wait and affinity defaults, and the #805 physical-core default for Inkling, Kimi, and OLMoE.

Verification: the focused launcher suites pass (12 tests, including a new guard that fails if either launcher sizes the team for V4), make check passes (C suite all pass, 424 Python tests, 57 skipped), and a direct V4 probe on 16 logical CPUs selects 13 with no override, honors an explicit OMP_NUM_THREADS=4, and stays untouched under the kill switch. No C runtime code changed.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -362,7 +362,7 @@ def env_for_engine(a, arch):
     if arch == "glm":
         return env_for(a)
     env = os.environ.copy()
-    # OMP_NUM_THREADS for the sister engines too.
+    # OMP_NUM_THREADS for the sister engines that do not size their own team.
     #
     # #805 set this from physical cores "on every platform" -- but only inside
     # env_for(), which this function calls only for glm. inkling, kimi_k3,
@@ -375,12 +375,18 @@ def env_for_engine(a, arch):
     # Same defect shape as the one #805 fixed, one level out: a mechanism that
     # reaches one engine and not its siblings.
     #
-    # setdefault, so an explicit OMP_NUM_THREADS still wins -- which is also
-    # how to A/B it: OMP_NUM_THREADS=12 reproduces the old behaviour.
+    # DeepSeek V4 is now the exception: its C runtime sizes the team from
+    # logical CPUs minus its expert-loader reservation. Setting the physical
+    # count here makes that runtime mistake a launcher default for a user
+    # override and bypass its loader-aware policy. The other engines still
+    # need the physical-core default below.
+    #
+    # setdefault, so an explicit OMP_NUM_THREADS still wins.
     if not env.get("COLI_NO_OMP_TUNE"):
-        from resource_plan import physical_cpu_count
-        env.setdefault("OMP_NUM_THREADS", str(physical_cpu_count()))
-        if arch == "deepseek_v4":
+        if arch != "deepseek_v4":
+            from resource_plan import physical_cpu_count
+            env.setdefault("OMP_NUM_THREADS", str(physical_cpu_count()))
+        else:
             env.setdefault("OMP_WAIT_POLICY", "active")
             env.setdefault("GOMP_SPINCOUNT", "200000")
             env.setdefault("OMP_DYNAMIC", "FALSE")
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -1490,8 +1490,9 @@ def tune_child_env(env, arch):
     if arch != "deepseek_v4":
         return env
     if not env.get("COLI_NO_OMP_TUNE"):
-        from resource_plan import physical_cpu_count
-        env.setdefault("OMP_NUM_THREADS", str(physical_cpu_count()))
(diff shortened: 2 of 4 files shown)


JustVugg:
The diagnosis is correct and I verified it independently before reading your patch: `c/coli:382` and `openai_server.py:1494` both `setdefault("OMP_NUM_THREADS", physical_cpu_count())` for every engine, and `v4_omp_reserve_loader_cpus()` returns on `getenv("OMP_NUM_THREADS")`. So #897's reservation has never fired through either launcher — it only fires when the engine is invoked directly, which is how it was measured. Good catch, and the `side_effect=AssertionError` test is a nice way to assert the launcher does not even *compute* a team for V4.

**One arithmetic concern before this merges, and it is the reason I am not merging it today.** The runtime's team is:

```c
int logical = omp_get_max_threads();          /* libgomp default = nproc = LOGICAL */
int team    = logical - COLI_V4_EXPERT_LOADER_COUNT;
```

On a 6-physical / 12-logical box the launcher currently yields **6** threads; after this PR the runtime yields **12 - 3 = 9**. That is 9 threads over 6 physical cores — the SMT over-subscription of a memory-bound GEMV that #805 removed and #718 documents ("l'SMT dimezza il decode su alcune CPU"). So the patch makes the reservation fire, but the thing it makes fire counts the wrong CPUs.

The engine has no physical-core counter today (`omp_tune.h`'s `coli_physical_cores()` is not reachable from `deepseek_v4.c`). Two ways out, both fine by me:

1. **Runtime learns physical cores** — reserve from physical, not logical: `team = physical - loaders` (on the box above: 6 - 3 = 3... which is probably too few, so see 2).
2. **Launcher keeps sizing but subtracts the loaders** — pass `OMP_NUM_THREADS = max(2, physical - COLI_V4_EXPERT_LOADER_COUNT)`, keeping the physical-core basis *and* the loader reservation. The constant would need exporting to the launcher, which is precisely the "one constant, two consumers" shape we have been bitten by (#856), so it should come from one place.

Worth noting the loaders block in `pread` and do not need whole cores, so subtracting all of them from a physical count may over-correct — a measurement on an SMT box would settle whether the right basis is physical, physical minus one, or logical minus loaders.

Not release-gating: today's behaviour is not a regression, it is an inert feature, and v1.6.0 ships as measured. Happy to take this right after the tag once the CPU basis is settled — and thanks to @terrizoaguimor for the report that surfaced it.

Blakeolson21:
0.7414 tok/s at logical minus loaders (13 threads), 0.7304 at physical cores (8), and 0.7223 at physical minus one (7). The sample SDs were 0.0009, 0.0007, and 0.0002 tok/s respectively across three 256-token runs. Relative to the best mean, physical was 1.49% lower and physical minus one was 2.58% lower.

I measured this on a Ryzen 7 6800H box, 8 physical cores and 16 logical CPUs with SMT 2:1, 32 GB RAM, three V4 expert loaders, and colibri 1.5.0. I used the same fixed prompt for every run and calculated decode rate as 255 tokens divided by the engine's `after_first` time. Each run was sequential and finished in under ten minutes. The box also hosts an inference service; it stayed up but idle, with no requests scheduled during any run. The box was also under memory pressure the whole time. Swap use rose from about 4.4 GB during the first runs to about 7.5 GB during the last ones. That rise tracks run order rather than thread count, and if memory pressure were setting the pace the last runs would be the slowest. They were the fastest, so I don't think swap drove the ranking, but I would rerun on a box with real headroom before trusting the absolute rates.

Two caveats. The unset/default case averaged 0.6439 tok/s with a 0.0224 SD, but its effective 13-thread team is inferred from the runtime policy and it ran first, so I think that result is dominated by warm-state effects rather than team size. And the conditions ran in a fixed order (default, then 8, 7, 13), so the winner ran last. I don't think ordering explains the explicit comparison though: the 7-thread condition ran between 8 and 13 and was the slowest of the three, which is the opposite of what a warm-up trend would produce.

Based on the explicit settings, I would keep logical minus loader count. Reserving a whole core per loader overcorrects on this box, which is consistent with the loaders spending time blocked in `pread`. Physical cores was close, but it did not beat 13 threads in any run. This is one box and one workload, so I would treat the 1.5% gap as directional rather than universal.

PR #960, opened 2026-08-11
convert_inkling: support output locking on Windows

Blakeolson21:
## Summary

- use `fcntl.flock` when it is available
- fall back to `msvcrt.locking` on native Windows Python
- keep the lock handle alive until conversion exits

## Problem

`convert_inkling_int4.py` imports `fcntl` unconditionally while acquiring its output-directory lock. Native Windows Python does not provide that module, so the converter exits before it can process a shard.

The GLM converter already carries the required platform split. This change gives the Inkling converter the same behavior without changing the lock contract.

## Validation

- `python3 -m unittest tests.test_inkling_converter_lock -v` passed 3 tests on macOS. The suite exercises the Unix `fcntl` path, a simulated native Windows `msvcrt` path, and the existing contention error without importing optional model dependencies.
- `python3 -m py_compile tools/convert_inkling_int4.py` passed on macOS.
- `make check` passed on macOS. The Python suite ran 424 tests with 57 dependency or platform skips, and the C tests passed. The build emitted the existing unused `KV_GB` warning in `tests/test_k3_ram_budget.c`.
- Native Windows execution was not run during implementation. The earlier Windows 10 pc1 audit found that `fcntl` was unavailable, `msvcrt.locking(..., LK_NBLCK, 1)` acquired the converter's empty lock-file shape, and a competing handle received `PermissionError`.

### DIFF
--- a/c/tests/test_inkling_converter_lock.py
+++ b/c/tests/test_inkling_converter_lock.py
@@ -1,0 +1,63 @@
+import ast
+import builtins
+import os
+import sys
+import tempfile
+import types
+import unittest
+from pathlib import Path
+from unittest import mock
+
+
+CONVERTER = Path(__file__).resolve().parent.parent / "tools" / "convert_inkling_int4.py"
+ERROR = "ERROR: another converter is already using this output directory."
+
+
+def load_lock_helper():
+    tree = ast.parse(CONVERTER.read_text(encoding="utf-8"), filename=str(CONVERTER))
+    helper = next(node for node in tree.body
+                  if isinstance(node, ast.FunctionDef) and node.name == "acquire_output_lock")
+    namespace = {"os": os, "sys": sys}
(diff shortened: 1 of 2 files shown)


JustVugg (approved):
Right shape: `fcntl` → `msvcrt.locking` → unlocked last resort, identical error message and non-blocking semantics on both real paths, and the handle stays alive with the caller so the lock outlives the helper. The tests exercise the actual function extracted via AST, so they cannot drift from it. Native Windows Python could not run this converter at all before — `import fcntl` raised at lock time.

Blakeolson21 (commented):
Thanks for the careful read, and for verifying the AST-extraction approach on the tests. The unlocked last resort was the part I went back and forth on; keeping the error message and non-blocking semantics identical on both real paths felt like the safest way to make the fallback invisible to callers. Glad the shape holds up.

PR #961, opened 2026-08-11
coli: select python.exe from Windows project venvs

Blakeolson21:
## Summary

- select `mio_env\Scripts\python.exe` on Windows
- keep `mio_env/bin/python3` on POSIX systems
- share the selection logic between `coli bench` and `coli convert`

## Problem

Both commands currently append `python3` after choosing the Windows `Scripts` directory. Standard Windows venvs create `Scripts\python.exe`, so the existence check misses the project environment and silently falls back to the interpreter running `coli`.

That fallback can bypass the environment containing the converter or benchmark dependencies. It also makes the existing comment about preferring the project environment untrue on Windows.

## Validation

- `python3 -m unittest tests.test_launcher_dispatch tests.test_cli_output -v` passed 31 tests on macOS. The launcher test creates a temporary Windows venv layout and verifies that both bench and convert select `Scripts\python.exe`. It also covers the unchanged POSIX path and the missing-venv fallback.
- `python3 -m py_compile coli` passed on macOS.
- `make check` passed on macOS. The Python suite ran 424 tests with 57 dependency or platform skips, and the C tests passed. The build emitted the existing unused `KV_GB` warning in `tests/test_k3_ram_budget.c`.
- Native Windows execution was not run during implementation. The earlier Windows 10 pc1 audit created a standard-library venv that contained `Scripts\python.exe` and did not contain `Scripts\python3`.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -1372,12 +1372,19 @@ def cmd_web(a):
     print(f"dashboard: {url}  (opens automatically when the engine is ready)")
     cmd_serve(a)
 
+def project_python():
+    if sys.platform == "win32":
+        candidate = os.path.join(HERE, "mio_env", "Scripts", "python.exe")
+    else:
+        candidate = os.path.join(HERE, "mio_env", "bin", "python3")
+    return candidate if os.path.exists(candidate) else sys.executable
+
+
 def cmd_bench(a):
     need_model(a.model)
     banner("bench")
     # python con `tokenizers`: l'ambiente del progetto se c'e', altrimenti quello corrente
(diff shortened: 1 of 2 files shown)


JustVugg (approved):
Correct and minimal: standard Windows venvs create `Scripts\python.exe`, never `Scripts\python3`, so both call sites were resolving to a path that cannot exist and silently falling back to `sys.executable` — the project venv was ignored. One shared helper for both commands, with a test.

Blakeolson21 (commented):
Thanks. The silent fallback to sys.executable was the sneaky part; everything looked like it worked while the project venv sat unused. Extracting the one shared helper felt safer than fixing two call sites that could drift apart again.

PR #962, opened 2026-08-11
convert: make multistream shard writes portable on Windows

Blakeolson21:
## Summary

- open the multistream part file with `O_BINARY` where Windows exposes it
- keep `os.pwrite` on platforms that provide it
- add a locked `lseek` and complete `os.write` fallback for native Windows Python
- handle short writes without advancing segment progress too far

## Problem

The FP8 converter uses `os.pwrite` for files at least 256 MiB when more than one download stream is active. Native Windows Python does not expose `os.pwrite`.

The resulting `AttributeError` is caught by the downloader's retry block, so each worker retries the same unsupported operation instead of failing cleanly. The file descriptor is also opened without `O_BINARY`. On Windows, CRT text translation can change arbitrary model bytes written through that descriptor.

## Validation

- `python3 -m unittest tests.test_convert_positioned_write -v` passed 3 tests on macOS. The suite forces the fallback, forces short writes, writes disjoint chunks containing LF and byte `0x1a`, compares the complete file byte for byte, covers the POSIX `pwrite` path, and verifies zero-length writes raise `OSError`.
- `python3 -m py_compile tools/convert_fp8_to_int4.py` passed on macOS.
- The existing converter and repack suites were invoked separately. `test_int3_convert.py` skipped because NumPy is not installed in the available system Python. `test_fp8_repack.py` skipped because Torch and Safetensors are not installed. `test_fp8_e2e_repack_load.py` skipped because Torch is not installed.
- `make check` passed on macOS. The Python suite ran 424 tests with 57 dependency or platform skips, including the optional converter and repack suites above, and the C tests passed. The build emitted the existing unused `KV_GB` warning in `tests/test_k3_ram_budget.c`.
- Native Windows execution and the Windows UCRT64 check were not run during implementation. The earlier Windows 10 pc1 audit found no `os.pwrite`, found `os.O_BINARY`, observed newline translation without `O_BINARY`, and observed exact storage of `41 0a 42 1a 43` with `O_BINARY`.

### DIFF
--- a/c/tests/test_convert_positioned_write.py
+++ b/c/tests/test_convert_positioned_write.py
@@ -1,0 +1,89 @@
+import ast
+import os
+import tempfile
+import threading
+import types
+import unittest
+from pathlib import Path
+
+
+CONVERTER = Path(__file__).resolve().parent.parent / "tools" / "convert_fp8_to_int4.py"
+
+
+def load_positioned_write(os_module):
+    tree = ast.parse(CONVERTER.read_text(encoding="utf-8"), filename=str(CONVERTER))
+    helper = next(node for node in tree.body
+                  if isinstance(node, ast.FunctionDef) and node.name == "_positioned_write")
+    namespace = {"os": os_module, "_positioned_write_lock": threading.Lock()}
+    exec(compile(ast.Module(body=[helper], type_ignores=[]), str(CONVERTER), "exec"), namespace)
+    return namespace["_positioned_write"]
+
+
+class ShortWriteFallback:
+    SEEK_SET = os.SEEK_SET
+    pwrite = None
+
+    @staticmethod
+    def lseek(fd, offset, whence):
+        return os.lseek(fd, offset, whence)
+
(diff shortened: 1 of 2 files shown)


JustVugg (approved):
Reviewed the mechanism, not just the platform claim: the old call discarded `os.pwrite`'s return value, so a short write silently truncated a shard — on **any** platform, not only Windows. `_positioned_write` now loops until the buffer is drained, and takes the lock only on the lseek+write fallback, which is correct: pwrite is atomic with respect to file position and needs none. `O_BINARY` via getattr is 0 on POSIX. Tests load the real helper via AST rather than restating it. This is a correctness fix wearing a portability hat.

Blakeolson21 (commented):
Correctness fix wearing a portability hat is exactly right, and honestly the short-write truncation surprised me more than the Windows part when I traced it. Good point on pwrite needing no lock; keeping the lock scoped to the lseek fallback was the piece I most wanted a second pair of eyes on.

PR #964, opened 2026-08-11
v4: pluggable expert-store backend registry

8PotatoChip8:
## Summary

Right now the DeepSeek-V4 engine opens its expert store by calling the on-disk/mmap opener directly. That works for the default case, but it means anyone who wants the engine to read experts from somewhere else — a different SSD layout, a remote store, whatever — has to fork the engine and keep that diff alive as upstream moves.

This adds a small registry so the engine asks for a backend by name instead. The on-disk opener registers itself as `"auto"` at link time, and `COLI_EXPERT_STORE` unset (or `"auto"`) lands on it — so nothing changes unless someone opts in. Set `COLI_EXPERT_STORE=foo` and the engine routes to whatever `foo` backend a build links in; a name that isn't registered errors out cleanly with a message naming it.

The aim is the smallest change that lets a downstream build plug in its own store without touching the engine: a name→open-function table, register/lookup, and the one call site in `coli_v4_engine_open` (plus the standalone CLI path) going through `coli_expert_store_backend_open_selected`. No new dependencies, and the default CPU path is byte-for-byte the same.

## Validation

- [x] `make -C c check` — 421 tests, 0 failures (52 skipped, the same set that skips on `dev` without a model/CUDA)
- [x] CUDA — N/A, no CUDA paths touched
- [x] Performance — N/A, default path unchanged

## Compatibility

- [x] The default CPU build remains dependency-free — the registry is plain C with no new libs, and `COLI_EXPERT_STORE` unset is the exact prior call
- [x] No model files, generated binaries, or benchmark artifacts — only `.c`/`.h` sources and a unit test

## Notes

`coli_v4_expert_store_open_planned` picks up a `config` argument so a backend that lives outside the engine (linked in by a downstream build) still receives the parsed config. The two existing tests that reference the old direct call — `test_v4_ownership.c` (stub signature) and `test_deepseek_v4_dspark_source.py` (source-order grep) — are updated to the new call site. `Makefile.deepseek-v4` also gains a `deepseek-v4-objs` target that prints the unit object list, so a downstream tree can link exactly what the plain binary links without recompiling the units.

### DIFF
--- a/c/.gitignore
+++ b/c/.gitignore
@@ -3,8 +3,12 @@ glm_tiny/
 olmoe_hf/
 olmoe_i4/
 .build-config
+# DeepSeek-V4 amalgamation build artifacts (make -f Makefile.deepseek-v4)
 build/
 COLI_V4_UNIT_*.o
+expert_store_registry.o
+deepseek_v4
+test_expert_store_registry
 tests/test_st_mirror
 tests/test_st_pread
 tools/libiq3.so
--- a/c/Makefile
+++ b/c/Makefile
@@ -1068,7 +1068,8 @@ V4_OWNERSHIP_OBJS = \
 	$(V4_OWN_DIR)/COLI_V4_UNIT_RUNTIME.o \
 	$(V4_OWN_DIR)/COLI_V4_UNIT_CONFIG.o \
 	$(V4_OWN_DIR)/COLI_V4_UNIT_ST.o \
-	$(V4_OWN_DIR)/COLI_V4_UNIT_NATIVE_QUANT.o
+	$(V4_OWN_DIR)/COLI_V4_UNIT_NATIVE_QUANT.o \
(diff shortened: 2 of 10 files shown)


JustVugg:
Status ping: this is currently `dirty` against `dev`, which has moved a good deal in the last day — the CUDA tier for DeepSeek V4 landed (#1054, #1055), the dual-SSD mirror (#988), and #1063 made model families registry-owned across `coli` / the gateway / `doctor.py` / `resource_plan.py`.

A rebase would put it back in the queue. I'm not asking as a formality — the change is still wanted, it just can't be reviewed or merged in this state.

If you'd rather not carry it any further, say so and I'll close it with the reasoning recorded, so the work is findable rather than silently dropped. Either answer is fine; what I'd like to avoid is it sitting here indefinitely because nobody said anything.

Issue #826, opened 2026-08-04
[Feature]: optional streamable dense trunk, so colibri runs where RAM is the hard blocker


michael-denyer:
## The floor

colibri streams the routed experts but keeps the dense and attention trunk resident and
mlock'd, read from RAM on every layer of every token (`colibri.c`, the resident-dense load
path, comment: "attention/shared stream from RAM every token"). That default is right, and it
is why decode is fast. The trunk is small next to the expert set, so holding it resident spends
a fixed few GB to save a per-token disk read.

But it is a hard floor. On a box whose RAM cannot hold trunk + KV + a useful expert cache,
colibri does not run slowly. It does not run at all. Open issues show the floor hitting real
machines: #766 (auto-tier drives the process into the OOM killer), #759 (RAM_GB not clamped to
MemAvailable, OOM-killed mid-prefill), #707 (32 GB host). The biggest models on the smallest
boxes have nowhere to put the trunk.

## kimi-k3-in-c already does it

`kimi-k3-in-c` (https://github.com/FareedKhan-dev/kimi-k3-in-c) runs Kimi K3, 2.78T
parameters, on one CPU in 8.24 GB. It streams the dense trunk to a configurable depth and holds
only the top of it resident, and its output is byte-identical at every RAM budget from 8 GB to
224 GB. It is CPU-only with no expert cache, so it is far slower than colibri (10-32 s/token
against colibri's ~3 tok/s on a comparable streaming model). But it does the one thing colibri
cannot do today. It fits a model of any size into a RAM budget of any size by trading the trunk
to disk.

## Proposal

An opt-in knob, say `TRUNK_RESIDENT_LAYERS=N` (default = all, so nothing changes), that keeps
the top N dense/attention layers resident and streams the rest on demand through the same
O_DIRECT path the experts already use. Off by default. When set, colibri spends seconds per
token to run at all on a box that currently OOMs.

This is a "run at all vs run fast" lever, not a speed win. The dense trunk, unlike the sparse
experts, is read on every layer of every token, so streaming it re-reads those bytes every
token with no cache hit to amortize them. On a machine with RAM to spare it is strictly worse
and should stay off. Its only job is to move the floor.

## What it would take, before anyone writes code

This reverses a core assumption, so it is worth agreeing on before a PR. Streaming the trunk
turns the dense and attention matmul sites (~99 references to resident layer-weight pointers in
`colibri.c`) from read-a-resident-pointer into fetch-then-compute, adds resident-set accounting,
[...]

PR #965, opened 2026-08-11
feat(k3): map prepared weights read-only

bherald:
## Summary

Add an opt-in, CPU-only `K3_MMAP=1` path for Kimi K3 containers whose dense,
MLA, shared-expert, and head tensors are already stored in their final U8 data
+ F32 scale representation.

- add a small cross-platform read-only mapping primitive that carries the
  aligned OS view and Windows mapping handle required for exact cleanup;
- map prepared int8 and int4-g64 tensor bytes plus their F32 scale sidecars
  instead of copying them into private heap;
- leave the existing loader and all defaults unchanged;
- refuse Vulkan, CUDA, load-time conversion, incompatible scale formats, and
  mapping failures instead of silently falling back; and
- test unaligned tensor offsets, byte-exact mapped-versus-copied CPU math, the
  backend refusal contract, and Windows file/directory cleanup.

This is the focused `st_map` + K3 vertical slice discussed in #826. It does not
implement `TRUNK_RESIDENT_LAYERS`, prefetching, page locking, or an OS working-
set policy.

## Field result

On a Ryzen 9 7845HX / 64 GB Windows host using the official Kimi K3 text model,
int4/MLA8/Head8, a 1-GiB expert cache, 1K context, `K3_VK=0`, and `K3_CUDA=0`,
the mapped path reported 1,906 prepared views / 33.8 GB file-backed. Mapping
reduced peak private memory from the prior heap-copy path's roughly 35.36 GB to
about 7.32 GB.

Mapping alone did not bound the active working set. The host therefore applied
its own verified hard working-set maximum to only the exact K3 engine; that
limit is external to this PR. Three isolated 26-token smoke repeats at 16 GiB
completed in 72.945-73.892 seconds with 2,126-2,578 MiB pagefile growth and
zero final K3 engines.

A follow-up exposed an important limit. This runtime uses `K3_CHUNK=32`; prompts
above 32 total chat tokens require another full 93-layer prefill traversal. A
40-token synthetic prompt repeatedly crossed the unchanged 3,072-MiB pagefile
guard even with a 19-GiB host cap. Five concise 31-32-token synthetic cases did
complete safely at 19 GiB in 86.491-92.282 seconds with 1,759-1,881 MiB
pagefile growth and zero final K3 engines, but only four returned the expected
token. The NoFallback case returned an unexpected token. These results show
that prepared read-only mapping makes host-managed residency possible inside a
narrow single-prefill envelope; they are not a production-readiness,
throughput, broad-quality, or full-model-equivalence claim.

## Validation

- [x] `make -C c check`
- [x] `make -C c test-asan`
- [x] Windows native suite and 484 Python/API tests passed with 49 expected
      Windows skips
- [x] Two Windows builds were byte-identical
- [x] Exact copied-versus-mapped int8 and int4 CPU math tests passed
- [x] Three new bounded full-model repeats returned the same one-token content
      hash and clean final process state
- [x] Five bounded single-prefill quality cases completed safely; four returned
      the expected token and the mismatch failed closed
- [x] CUDA changes were not made; mapped mode explicitly refuses enabled CUDA
      or Vulkan backends
- [x] Performance claims include the host, exact settings, and repeatable
      measurements above

## Compatibility

- [x] No model files, generated binaries, receipts, or benchmark artifacts are
      included
- [x] `K3_MMAP` defaults off and changes no existing loader behavior

Closes no issue; contributes the mapped-residency slice toward #826 while
leaving the broader `TRUNK_RESIDENT_LAYERS` design open.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -768,6 +768,9 @@ tests/test_tok_kimi_tiny$(EXE): tests/test_tok_kimi_tiny.c tok.h tok_unicode.h t
 tests/test_st_pread$(EXE): tests/test_st_pread.c st.h json.h compat.h
 	$(CC) $(CFLAGS) -DST_PREAD_CHUNK=7 $< -o $@ $(LDFLAGS)
 
+tests/test_st_map$(EXE): tests/test_st_map.c st.h json.h compat.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
 tests/test_st_shape$(EXE): tests/test_st_shape.c st.h json.h compat.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/Makefile
+++ b/c/Makefile
@@ -763,6 +763,9 @@ tests/test_tok_o200k$(EXE): tests/test_tok_o200k.c tok.h tok_unicode.h tok_unico
 tests/test_k3_ram_budget$(EXE): tests/test_k3_ram_budget.c kimi_k3.c st.h tok.h quant.h compat.h omp_tune.h route_trace.h kv_prefix.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
+tests/test_k3_mmap$(EXE): tests/test_k3_mmap.c kimi_k3.c st.h tok.h quant.h compat.h omp_tune.h route_trace.h kv_prefix.h
+	$(CC) $(NOCUDA_CFLAGS) $< $(VK_OBJ) -o $@ $(NOCUDA_LDFLAGS)
+
 tests/test_tok_kimi_tiny$(EXE): tests/test_tok_kimi_tiny.c tok.h tok_unicode.h tok_unicode_o200k.h json.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 tests/test_st_pread$(EXE): tests/test_st_pread.c st.h json.h compat.h
--- a/c/compat.h
+++ b/c/compat.h
@@ -11,6 +11,15 @@
 #ifndef COMPAT_H
 #define COMPAT_H
 
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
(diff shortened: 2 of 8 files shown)


JustVugg:
Merged. I held this one back for a day specifically to check it against a bug in this project's history, so it's worth recording what I looked for and why it didn't apply.

**The concern:** colibrì previously had an mmap reader that defeated the whole low-RAM goal — mapped pages for *streamed experts* stayed resident when they should have been dropped, and the fix was `pread` + `fadvise(DONTNEED)` (13.6 GB → 6.0 GB). Any new mmap in `st.h` gets measured against that.

**Why this is the inverse case, not a repeat.** You map the *prepared, resident* tensors, not the streamed ones. Without mapping, those bytes exist twice — once in page cache, once in the heap copy. Mapping them read-only collapses that to one copy the kernel can evict, which is why your numbers go the right way: **peak private memory 35.36 GB → 7.32 GB**, with 1,906 views / 33.8 GB file-backed.

Four things I verified before merging:

1. `g_k3_mmap = getenv("K3_MMAP") ? atoi(...) : 0` — **off by default**.
2. `CHECK(!copied.mapped, "%s changed the default loader")` — you wrote a test that *fails if someone flips the default*. That is the part that convinced me; the risk with a change like this isn't today, it's the refactor in three months.
3. `k3_mmap_backend_allowed()` refuses when Vulkan or CUDA are active, so mapped host storage never meets an upload path that assumes private memory.
4. The streamed-expert path is untouched — no `fadvise`/`DONTNEED`/expert-load hunks anywhere in the diff. It still does `pread` + `DONTNEED` exactly as before.

**One docs request, non-blocking.** Those 33.8 GB stay visible in RSS as clean shared pages, so `top` will still show a large number — what collapses is *private* memory, which is what actually gets a process OOM-killed. Worth stating explicitly in `docs/kimi_k3.md`, otherwise the first person to check RSS will conclude the flag does nothing and open an issue saying so.

Thanks for the care on this one — twelve checked boxes, two new tests, and a measurement that made the decision for me rather than requiring me to trust it.

PR #966, opened 2026-08-11
fix(build): compile CUDA backend on glibc >= 2.41

martinm86867-ops:
glibc >= 2.41 (Debian trixie, Ubuntu 26.04, ...) declares the C23 math additions (rsqrt/rsqrtf, sinpi/cospi, ...) via __THROW, which expands to noexcept(true) in C++. CUDA's crt/math_functions.h declares the same functions without an exception specification, and since C++17 noexcept is part of the function type, nvcc's host pass fails with:

  mathcalls.h: error: exception specification is incompatible with that
  of previous function "rsqrt" (declared at ... crt/math_functions.h)

Force-include c/glibc_c23_math_compat.h into the host pass on Linux (NVCCFLAGS -Xcompiler=-include). It strips glibc's __THROW so the C declarations carry no exception specification, matching CUDA's. The device pass is unaffected (-Xcompiler flags never reach nvcc's device frontend).

Also ignore the V4 unit-test artifacts (COLI_V4_UNIT_*.o, build/) that `make test` produces on Linux.

### DIFF
--- a/c/glibc_c23_math_compat.h
+++ b/c/glibc_c23_math_compat.h
@@ -22,8 +22,8 @@
 #define COLI_GLIBC_C23_MATH_COMPAT_H
 
 #if defined(__cplusplus) && defined(__linux__)
-#include <sys/cdefs.h> /* glibc's own guards; forces __THROW into existence */
-#if defined(__THROW)
+#include <features.h> /* defines __GLIBC__ on glibc; musl ships one too */
+#if defined(__GLIBC__) && defined(__THROW)
 #undef __THROW
 #define __THROW
 #endif
--- a/c/glibc_c23_math_compat.h
+++ b/c/glibc_c23_math_compat.h
@@ -1,0 +1,32 @@
+// glibc_c23_math_compat.h — glibc >= 2.41 C++17 noexcept clash for CUDA host passes.
+//
+// glibc >= 2.41 (Debian trixie, Ubuntu 26.04, ...) declares the C23 math
+// additions (rsqrt/rsqrtf, sinpi/cospi, ...) via __THROW, which expands to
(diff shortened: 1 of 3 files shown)


c/glibc_c23_math_compat.h
```diff
@@ -0,0 +1,32 @@
...
+// pass is unaffected: -Xcompiler flags never reach nvcc's device frontend,
+// and it includes CUDA's math_functions.h directly.
+//
+// Deliberately Linux-only: Windows builds use MSVC cl.exe, which never sees
+// glibc headers.
+
+#ifndef COLI_GLIBC_C23_MATH_COMPAT_H
+#define COLI_GLIBC_C23_MATH_COMPAT_H
+
+#if defined(__cplusplus) && defined(__linux__)
```
rofl0r, line 24: this should be `__GLIBC__`, not `__linux__`. other libcs than GLIBC exist on linux. and they don't have sys/cdefs.h.

martinm86867-ops: Good catch — fixed in 23b2905: the header now includes <features.h> first and only neutralizes __THROW when __GLIBC__ is defined, so non-glibc libcs are untouched. Verified the CUDA build still passes locally.

JustVugg (approved):
Compiles the CUDA backend on glibc >= 2.41 (C23 math compat header, Linux-only), and — bonus — adds build/ and COLI_V4_UNIT_*.o to .gitignore, the recurring gap. Fixes a real build break on recent distros.

JustVugg (approved):
Compiles the CUDA backend on glibc >= 2.41 (C23 math compat, Linux-only) and adds build/ + COLI_V4_UNIT_*.o to .gitignore. Real build break on recent distros.

Issue #509, opened 2026-07-21
[Bug]: TEMP env var conflicts with ROCm on AMD GPUs causing SIGSEGV


icedream:
### Commit or version
dev branch, commit 8c8c798

### Environment
- **OS:** CachyOS
- **CPU:** AMD Ryzen 9 7950X 16-Core
- **RAM:** 64 GB
- **GPU:** AMD Radeon RX 7900 XTX (gfx1100), 24 GB VRAM
- **ROCm:** 7.2.4
- **Compiler:** gcc 16.1.1 20260625
- **Storage:** NVMe SSD

### Reproduction steps
1. Build the engine with HIP support:
   ```bash
   make -C c clean all HIP=1 HIP_ARCH=gfx1100
   ```

2. Run with `TEMP` set to a non-path value (e.g., sampling temperature):
   ```bash
   # Direct binary — crashes when TEMP env var is set to anything that isn't a valid temp-directory path:
   TEMP=0.6 COLI_CUDA=1 CUDA_DENSE=0 CUDA_EXPERT_GB=0 CUDA_EXTEND=0 \
     SNAP=<snap_path> CTX=512 NGEN=1 ./c/colibri 512

   # Through the c/coli script with --temp:
   COLI_CUDA=1 CUDA_DENSE=0 CUDA_EXPERT_GB=0 CUDA_EXTEND=0 \
     COLI_MODEL=<model_path> \
     ./c/coli chat --ctx 512 --temp 0.6
   ```

   Replace `<snap_path>` and `<model_path>` with a real GLM-5.2 colibri model checkpoint from Hugging Face.

### Expected behavior
The engine should start normally: detect the GPU, init CUDA/HIP backend, load the model, allocate hot tier experts into VRAM, and begin processing the prompt. Setting `--temp 0.6` should change sampling behaviour, not crash initialisation.

### Actual behavior and logs
```
[CUDA] device 0: AMD Radeon RX 7900 XTX, 25.8 GB VRAM, sm_110
[engine terminated: killed by SIGSEGV]
```

GDB backtrace:

```gdb
#0  std::__atomic_base<unsigned int>::fetch_sub (at atomic_base.h:658)
    → `referenceCount_.fetch_sub(1, ...)` on NULL `this`
#1  amd::ReferenceCountedObject::release (runtime.cpp:137)
#2  amd::HostQueue::Thread::Release (commandqueue.hpp:187)
#3  amd::HostQueue::terminate (commandqueue.cpp:96)
[...]

PR #968, opened 2026-08-12
fix(olmoe): COLI_TEMP channel, and never evict an in-flight slot

terrizoaguimor:
### 1. Temperature: OLMoE carried #509 after GLM was cured

`olmoe.c` read `TEMP` raw, and the launcher exported `TEMP=<temp>` for olmoe (`coli --temp`):

- On Windows / ROCm stacks `%TEMP%` is a **directory path**; `atof("C:\...")` is `0.0`, so every olmoe chat silently forced **greedy decoding** — with the user setting apparently in effect.
- Conversely, `coli chat --temp 0.7` on olmoe **overwrote the child process's own `%TEMP%` with "0.7"** for anything that honored it downstream.

This is exactly why #509 moved GLM to `COLI_TEMP`. OLMoE now follows the same contract: **`COLI_TEMP` primary, `TEMP` as a legacy alias only when fully numeric** (same full-parse rule as `temp_from_env`). The launcher stops writing `TEMP` for olmoe — `COLI_TEMP` was already set for every arch.

`tests/test_v4_cli.py`'s olmoe block asserted `env["TEMP"] == "0.25"` — it was pinning the poisoning. It now pins the contract (`COLI_TEMP` set, `TEMP` never written by the launcher).

### 2. `expert_get`: the all-in-flight last resort raced the pilot's writer

When every cache slot was reserved (`eid==-1` — a pread physically in progress, usually the pilot's), the eviction fallback did:

```c
if (lru < 0) lru = 0; /* absolute last resort: all in-flight, evict slot 0 */
```

then loaded into that slot's slab **outside the lock** — while the pilot worker was mid-`load_expert_merged` into the same slab. Two writers racing one buffer, and whichever published last decided which expert id the resident bytes answered to → silently wrong weights from cache. It is the one place in the tree contradicting the in-flight exclusion every sibling honors (kimi_k3 keeps the read pending, V4 skips referenced slots, colibri.c marks reservations never-evictable).

The fallback now **waits for an in-flight publish (1 ms poll) and rescans**. The wait always drains: a load either finishes or the host is already dead in the water. Normal-pressure eviction is unchanged — pass 1 (skip pinned+in-flight) and pass 2 (allow pinned, never in-flight) are byte-identical.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -392,7 +392,10 @@ def env_for_engine(a, arch):
     if arch == "olmoe":
         env["CHAT"] = "1"
         env["MAX_NEW"] = str(ngen_for(a))
-        if a.temp is not None: env["TEMP"] = str(a.temp)
+        # olmoe reads COLI_TEMP like every other engine (line above). Setting the
+        # legacy TEMP alias here also clobbered %TEMP% with "0.7" for every child
+        # of the chat process on Windows — and olmoe never read COLI_TEMP, so the
+        # ONLY working channel was a poisoned system variable (#509's exact shape).
     # --ram reaches the engines that read it. kimi_k3 gained a RAM budget in
     # #855; before that `grep -c RAM_GB c/kimi_k3.c` returned 0, so `coli chat
     # --ram 242` on Kimi K3 set an environment variable nobody looked at and the
--- a/c/olmoe.c
+++ b/c/olmoe.c
@@ -468,7 +468,22 @@ static void expert_get(Model *m, int layer, int eid, Slot **out) {
                 if (lru < 0 || lc->slots[i].used < lc->slots[lru].used) lru = i;
             }
         }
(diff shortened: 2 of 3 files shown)


JustVugg (approved):
Both halves verified, and both matter *more* because of this release: v1.6.0 fixes the launcher dispatch (#879), so launcher users reach the olmoe engine for the first time — straight into these two. The TEMP half: %TEMP% is a directory on Windows, atof gives 0.0, silent greedy decoding, while the launcher clobbered the system variable for every child; COLI_TEMP primary with a fully-numeric-only legacy fallback is the right shape and matches #509's convention. The eviction half: the old `lru=0` last resort stole a slab owned by an unlocked in-flight pread — two writers, resident bytes answering to the wrong expert id, the exact silent-corruption class we've met in the slab-overflow work. Wait-and-rescan is correct since a load either publishes or the process is already gone. Green across the matrix.

PR #969, opened 2026-08-12
feat(v4): the expert history lives in route_trace.h now — #700 completed

terrizoaguimor:
## #700 left one engine out

V4 kept a **private `.coli_usage` reader/writer** after the telemetry unification: same file format hand-duplicated (header records, FNV-1a identity, tmp+rename), but:

- `"deepseek_v4"` was never registered in `rt_engine_names`, so a V4-written history announced itself as **"an unknown engine"** to every sibling's refusal message;
- its copy had already drifted from the header: no `COLI_USAGE_DECAY` on save, no legacy-layout path (a stricter dims-record requirement **silently dropped headerless histories** the siblings accept), and the duplicated loops were one more place for the recurring "landed in one engine, never reached the siblings" defect to live.

## The change

V4 adopts `route_trace.h`, which is also where the *next* fix in this class reaches V4 automatically:

- `rt_init("deepseek_v4", layers, experts)` at hot-store creation; `rt_count` feeds the shared counters at each store lookup, next to `policy->usage` (same site, same semantics: store lookup == selection);
- **save**: `rt_save` — cumulative across sessions, same header records byte-for-byte (`-1 dims`, `-2 1 fnv("deepseek_v4")`), the standard `[STATS]` line, and `COLI_USAGE_DECAY` now applies to V4 like everyone else. The `COLI_V4_SAVE_USAGE=0` opt-out is preserved;
- **load**: `rt_read` with a callback into `policy->usage` — same bounds admission as the private reader, plus the standard dimension/identity refusals and the `PIN=<path>` trusted-override semantics its private copy lacked. Existing V4 histories carry over unchanged (formats agree byte for byte);
- `hot_engine_id()` and the hand-rolled parse/serialize loops go away (**−91 LOC**);
- `test_route_trace` section 6 now proves the `deepseek_v4` id resolves to its name;
- `Makefile.deepseek-v4` objects depend on `route_trace.h`.

## One thing deliberately NOT wired

Per-row `ROUTE_TRACE` routing lines stay GLM/K3/Inkling/OLMoE-only. The amalgam builds one object per `COLI_V4_UNIT_*`, so the header's statics get a single owner — the hot-store unit — while the router runs in other units. `rt_init` opens the stream, so when `ROUTE_TRACE` is set, V4 prints one explicit note that the stream will hold no rows rather than silently suggesting otherwise. Wiring the router units is a clean follow-up for whoever wants V4 in the route-pairs pipeline.

### DIFF
--- a/c/Makefile.deepseek-v4
+++ b/c/Makefile.deepseek-v4
@@ -98,6 +98,7 @@ $(V4_BINARY): $(V4_OBJS)
 
 $(TARGET_OBJS) $(TEST_UNIT_OBJS): %.o: deepseek_v4.c deepseek_v4.h \
 	deepseek_v4_internal.h deepseek_v4_dspark.inc st.h json.h compat.h tensor.h quant.h \
+	route_trace.h \
 	native_quant.h native_quant_batch.h native_quant_dual.h \
 	native_quant_fp4_rows16.h
 	$(CC) $(CFLAGS) -D$* -c deepseek_v4.c -o $@
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -5318,6 +5318,15 @@ void coli_v4_attention_snapshot_destroy(ColiV4AttentionSnapshot *snapshot) {
 #include <stdlib.h>
 #include <string.h>
 
+/* The .coli_usage expert history lives in route_trace.h (#700) — one format,
+ * one reader, one writer for every engine. Included inside THIS unit only:
+ * the amalgam builds one object per COLI_V4_UNIT_*, and the shared header's
+ * statics must have a single owner, which is the unit that loads, counts
+ * (lookup_hot) and saves the history. The router runs in other units, so V4
+ * does not emit per-row ROUTE_TRACE lines; if the stream is enabled the
+ * banner says so below at store creation. */
+#include "route_trace.h"
+
 
 #ifdef COLI_V4_EXPERIMENTAL_PREFETCH_BATCH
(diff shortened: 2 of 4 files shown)


JustVugg (approved):
Consolidation done right: the .coli_usage expert history now has one format, one reader, one writer in route_trace.h — the anti-recurring-defect pattern (a mechanism that reaches every engine, not one) this repo keeps reaching for. 56+/91- (more deleted than added — the good direction), a dedicated test_route_trace.c, cumulative-across-sessions preserved, green on all platforms. Closes #700.

JustVugg:
Approved the approach — this is exactly the consolidation we want (one format/reader/writer for .coli_usage, closing #700). It just picked up a conflict from today's V4 merges (#966 and others touched Makefile.deepseek-v4 / deepseek_v4.c). A quick rebase on `dev` and I'll merge it straight away.

terrizoaguimor:
@JustVugg Mi dispiace per il ritardo, Hyphae mi ha tenuto estremamente occupato.

JustVugg:
Merged — thank you, Mario. This closes out #700 properly: the expert history now lives in `route_trace.h` where the other routing telemetry already is, and the diff is net **-35 lines** while doing more. Removing a duplicated home for the same state is exactly the kind of change that keeps this codebase reviewable as the engine count grows.

Your contributions have been consistently useful here — you don't just add code, you take things out and leave the structure cleaner than you found it. That's genuinely valuable and it's noticed. 🙏

PR #973, opened 2026-08-12
feat(v4): emit dashboard telemetry (EMAP/HITS/TIERS/PROF/HWINFO)

martinm86867-ops:
## Problem

With the DeepSeek-V4 engine, the web dashboard's **Expert Cortex (Brain)** page stays on "waiting for engine" and the **Profiling** panel stays empty, even after running inference.

The dashboard fills those panels only from `EMAP`/`HITS`/`PROF`/`TIERS`/`HWINFO` protocol lines on the engine's stdout (parsed in `openai_server.py`). Only the GLM engine emitted them (`telemetry.h`); the V4 expert store has its own layout and never wrote them, so `/experts` returns `{rows:0,...}` forever.

## Fix

Wire the V4 engine into the same wire format:

- **Expert store** (`COLI_V4_UNIT_EXPERT_STORE_HOT_ROWS16`): track per-expert routing selections (`eheat`, capped at 63 to match the EMAP low-6-bit heat field), the per-turn routed bitmap (`ehit`), and measured expert-disk read time (`disk_sec`) — hooked into `lookup_hot`.
- **Emitters** `coli_v4_expert_store_emit_{tiers,emap,hits}` + `coli_v4_expert_store_disk_sec`, exported across the unit split and called from the serve loop at each turn boundary (plus one startup emission for the initial cortex).
- **Serve unit**: emits `HWINFO` (cores/RAM/CPU) and per-turn `PROF` (wall, token counts, `expert_disk_s`). The V4 runtime has no other per-phase instrumentation, so the remaining PROF phases report zero and fold into "other" in the frontend.
- V4 is CPU-only: the VRAM tier is always 0; "ram" means resident in the expert-store cache slots.

## Verification

Against DeepSeek-V4-Flash-0731 (`coli web`):
- `/experts` returns a live 43×256 map: routing heat accumulates per turn (capped at 63), the per-turn hit bitmask pulses (`seq` increments), cache-resident experts show the RAM tier.
- `/profile` reports per-turn wall and disk time (e.g. 36.6 s disk of 246.7 s wall on a disk-bound turn).
- Inference output unchanged.

Known cosmetic limitation (frontend, not addressed here): `Brain.tsx` tooltips compute `layer = row + 3`, which is GLM-specific; the V4 cortex renders correctly, only the hover layer number is off.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -5358,6 +5358,9 @@ typedef struct {
     unsigned active_leases;
     ColiExpertStoreStats stats;
     pthread_mutex_t mutex;
+    double disk_sec;   /* cumulative wall time spent reading expert bytes from disk */
+    uint8_t *ehit;     /* layers*experts_per_layer: experts routed in the current turn */
+    uint8_t *eheat;    /* layers*experts_per_layer: cumulative routing selections, capped 63 */
 } V4ExpertStoreState;
 
 static int set_error(char *error, size_t size, const char *format, ...) {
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Chosen over #882 (same feature, both emit HWINFO/TIERS/EMAP/PROF/HITS): this one covers all five markers, is purely additive (261+/0-), and is green across the matrix, while #882 is red on four platforms. Merging this; closing that with credit.

Issue #923, opened 2026-08-10
[datapoint] i5-12450H · 16 GB DDR5-4800 · Solidigm QLC NVMe PCIe 4.0 · Windows 11 native · no dGPU — build + iobench results


YashDate31:
Sharing numbers from a budget Lenovo laptop — probably the low end of what people are trying to run this on. No discrete GPU, 16 GB RAM, QLC storage. Posting partly because the existing Windows datapoints (#113, #128, #273) all have more RAM or a dGPU, and I was curious whether this machine is even viable as a floor.

---

### Hardware

| component | detail |
|---|---|
| machine | Lenovo IdeaPad 5i (model 83ER) |
| CPU | Intel Core i5-12450H — 4P + 4E cores, 12 threads, up to 4.4 GHz turbo, AVX2 + AVX-VNNI (DL Boost), **no AVX-512** |
| RAM | 16 GB DDR5-4800 (8× Micron 2 GB — soldered; confirmed 4800 MHz configured) |
| GPU | Intel UHD Graphics (iGPU only, no discrete GPU) |
| storage | Solidigm SSDPFINW512GZL — 512 GB NVMe PCIe 4.0 x4, **QLC NAND, DRAM-less** (uses HMB) |
| OS | Windows 11 Home 26200, native (no WSL) |
| colibri | `81f08a0` (v1.1.1, current `main` as of Aug 2026) |

Quick note on the storage: this drive is QLC and DRAM-less. It uses HMB from system RAM instead of an on-die DRAM cache. Sequential reads are okay (~2.5–3 GB/s) but random read performance on small blocks is heavily HMB-dependent and throttles fast under sustained load. I expected this to show up in iobench.

---

### Build

The MinGW install on this machine (MinGW.org 6.3.0, via an old `C:\MinGW`) doesn't include pthreads-w64, so the build fails at `pthread.h: No such file or directory` — this is the standard MinGW-org vs MinGW-w64 split documented in the repo. The fix is w64devkit or MSYS2 per `docs/windows.md`. I'm in the process of setting that up.

[...]

PR #978, opened 2026-08-12
mgua: added info about cpu optimizations

mgua:
## Summary

clarifications about cpu optimizations when compiling and how the compiler optimizes for your cpu when using arch=native. 
(documentation only change)


JustVugg (approved):
Useful practical additions for Windows builders — CPU flag discovery and the gcc -march=native introspection command are exactly what #923-class reports need to self-serve.

PR #982, opened 2026-08-12
deepseek_v4: print the effective OpenMP team when the sizing policy is silent

Blakeolson21:
I added an OpenMP startup diagnostic to the DeepSeek V4 engine for the cases where the team sizing is silent today. `v4_omp_reserve_loader_cpus()` already prints its policy line when it sizes the team, but when `OMP_NUM_THREADS` is set, `COLI_NO_OMP_TUNE=1` is set, or the machine is too small for the reservation, nothing reports what team the engine actually got. This prints `omp_get_max_threads()` on stderr with the existing `[OMP] deepseek-v4:` prefix in exactly those cases, so the default path still prints one line, not two.

The override case matters most on the launcher paths: today both launch funnels set `OMP_NUM_THREADS` before V4 starts, so launcher-driven runs currently start with no OpenMP report at all. With #958, explicit user overrides remain the silent case this line covers.

I kept the launcher and sizing policy unchanged. Built with OpenMP and verified three startups: unset (policy line only), `OMP_NUM_THREADS=3` (effective team 3 reported), and `COLI_NO_OMP_TUNE=1` (effective team reported). The focused DeepSeek V4 C tests pass. I did not run `deepseek-v4-tiny-check`; the change is a print with no numeric path touched.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -8889,23 +8889,26 @@ static int v4_serve_main(void) {
  * rationale omp_tune.h records for the spin-wait half of the GLM tuning: a
  * busy team steals cores from the I/O pool). An explicit OMP_NUM_THREADS or
  * COLI_NO_OMP_TUNE=1 wins, exactly like the other engines' tuning. */
-static void v4_omp_reserve_loader_cpus(void) {
-    if (getenv("COLI_NO_OMP_TUNE")) return; /* family-wide kill-switch */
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Right follow-up to the #958 discussion: when the sizing policy stays silent, print what the team actually is — this gives every future SMT report the number we currently have to ask for. Return-value refactor is clean.

PR #983, opened 2026-08-12
fix(v4): move hot rows16 packing out of store mutex

### LINKED ISSUE #900: deepseek_v4: rows16 pin-packing runs under the expert-store mutex and costs ~7 s of prefill (v1.5.0 TTFT +24%)
## Summary

On v1.5.0 the DeepSeek-V4 engine gains ~22% decode throughput but loses ~24% time-to-first-token against the pre-merge tree (`f60e3be`, the PR #165 side of merge `c024a46`). The decode win is real and worth keeping. The TTFT loss is not an I/O effect: the largest single contributor is `hot_pack_slot_locked()` doing rows16 repacking of pinned experts **while `state->mutex` is held**, on the miss path in `lookup_hot()`.

This matters more than the numbers below suggest, because prefill streams experts exactly like decode, so the cost scales with prompt length.

## Hardware / build

Ryzen 7 6800H (8C/16T, AVX2, no AVX-512), 32 GB DDR5-4800, weights on a WD Green SN350 (Gen3 x4, DRAM-less). Ubuntu 26.04, gcc, stock `make -f Makefile.deepseek-v4 deepseek-v4`. DeepSeek-V4-Flash-0731, `CTX=4096`, `OMP_NUM_THREADS=14`, `memory=auto` (~27.7 GiB plan), `target_only=1`.

All runs restore a pristine `.coli_usage` and set `COLI_V4_SAVE_USAGE=0`, so the warm-pin history is frozen and the arms are genuinely independent (verified by hashing the file before and after every run). Prompt is a fixed 45 tokens, and every arm is 2 rounds run in opposite order on an idle box.

`--max-tokens 1` is used to isolate prefill: the engine measures TTFT as `first_at - setup_done`, so TTFT at 1 token matches TTFT at 160 tokens to within noise (67.6 vs 68.1).

## The regression

| binary | TTFT (s), 2 runs | hit rate | misses | `packed_slots` |
| --- | --- | --- | --- | --- |
| `f60e3be` (pre-merge) | 54.595, 54.811 | 59.49% | 4703 | 0 |
| v1.5.0 (`8f512fc`) | 67.570, 68.341 | 60.35% | 4604 | 742 |

**+13.3 s (+24%).** Note that v1.5.0 has the *better* hit rate and *fewer* misses and is still much slower, so this is not cache behaviour or I/O volume.

## What it is not

| arm | TTFT (s), 2 runs | vs v1.5.0 default |
| --- | --- | --- |
| v1.5.0 default | 67.570, 68.341 | baseline |
| `COLI_V4_DIRECT=0` (buffered pread) | 73.692, 74.399 | **+6.1 s worse** |
| `COLI_V4_AUTOPIN=0` | 68.476, 66.744 | -0.35 s (noise) |
| both off | 72.041, 73.382 | +4.8 s worse |

O_DIRECT is a net win in prefill as well, and disabling it costs another ~6 s. `COLI_V4_AUTOPIN=0` changes nothing measurable, because it only disables history *seeding*; pinning still runs from live usage, so `packed_slots` stays around 700.

Warm-pin history loading cannot be the cause either, by construction: `hot_usage_load()` is an `fscanf` of a ~70 KB text file and the opening `hot_repin_locked()` sweep does no I/O, and both run at expert-store open, before `setup_done`, so they sit outside the measured TTFT window entirely.

## What it is

`lookup_hot()` on the miss path relocks `state->mutex`, then calls:

```c
    if (hot_is_pinned(policy, key.layer, key.expert))
        hot_pack_slot_locked(policy, state, record, slot);
```

`hot_pack_slot_locked()` does a `malloc`, three `coli_fp4_pack_rows16_v10()` passes and six `memcpy`s over the full expert weight and scale buffers (~12.6 MB per expert here), then `free`, all with the store-wide mutex held. Every other OMP thread that needs any expert in any layer blocks behind it, and batched prefill is the worst case because it is the most parallel phase.

Instrumenting the function with a wall-clock accumulator (instrumentation only, no control-flow change; the unpatched rebuild of v1.5.0 in this checkout is byte-identical to the released binary, so the toolchain is known to reproduce upstream):

```
v4_pack calls=743 total_s=5.603
v4_pack calls=741 total_s=5.690
```

Adding a `COLI_V4_NOPACK` knob to that same binary, so pack-on and pack-off share one compilation, gives the whole cost of the feature. Skipping the pack also leaves `policy->packed[]` clear, so `hot_fill_view` stops raising `block_rows` to 16:

| arm (same binary) | TTFT (s), 2 runs | mean |
| --- | --- | --- |
| packing on | 66.169, 66.988 | 66.58 |
| packing skipped | 59.793, 59.626 | 59.71 |

**6.87 s**, of which 5.70 s is the lock-held packing and ~1.2 s the `block_rows=16` kernel path. Hit rate is unchanged between the two arms (60.64% vs 60.62%), so this is not a caching side effect.

**Honest scope: this is roughly 55% of the regression, not all of it.** With packing skipped, v1.5.0 is still ~5.0 s slower than `f60e3be` (59.71 vs 54.70). I have not attributed that remainder, and bisecting `f60e3be..v1.5.0` is the obvious next step that I have not done. `origin/dev` (`98cab3b`) measures identical to v1.5.0 here (TTFT 66.59 vs 66.94, decode 0.862 vs 0.866 tok/s at 160 tokens), which is expected since it does not touch this code.

`packed_slots` is 740 at one generated token but 1704 at 160, so packing continues through decode, which is where the rows16 layout buys the +22%. The feature looks like a decode optimisation whose cost is billed to prefill.

The slot-reuse path also clears `policy->packed[slot_index] = 0`, so an evicted and refetched pinned expert is repacked from scratch and the cost recurs.

## Suggested directions

1. Pack outside `state->mutex`. The slot already holds a reference at that point, so the packing work does not need the store-wide lock; only the `policy->packed[]` flag update does.
2. Consider not packing during prefill at all. The rows16 layout pays off across many decode steps, but a batched prefill touches each fetched expert once per batch, so the repack is not amortised there.
3. `COLI_V4_PREWARM=1` is a useful escape hatch today and is worth documenting: it moves the work out of the TTFT window (61.4 / 63.5 s vs 67.6 / 68.3 s) at roughly equal total wall time, because prewarm runs before `setup_done`.

## Side note

`docs/ENVIRONMENT.md` documents no V4-specific variables, though `deepseek_v4.c` reads 22 of them (`COLI_V4_DIRECT`, `COLI_V4_AUTOPIN`, `COLI_V4_PREWARM`, `COLI_V4_SAVE_USAGE`, `V4_DRAFT`, and others). Happy to send a docs PR separately if that would help. Related: #894 asks for exactly this kind of per-engine environment matrix.

aniketshukla1:
## Summary
- Move rows16 packing behind a per-slot preparation mutex instead of the store-wide mutex.
- Convert in place only while the requesting lease is exclusive; an existing row-major view keeps its layout until a later lookup can safely repack the slot.
- Publish the packed state under the store mutex, so consumers of the same slot wait while unrelated expert loads continue.
- Add a hot-store regression test that verifies deferred conversion and concurrent packed lookups; it is part of `make check`.

## Validation
- `make check`
- `make tests/test_deepseek_v4_hot_store && ./tests/test_deepseek_v4_hot_store`

Resolves #900

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -898,6 +898,16 @@ tests/test_deepseek_v4$(EXE): tests/test_deepseek_v4.c deepseek_v4.c deepseek_v4
 		COLI_V4_UNIT_RESOURCE_PLAN.o COLI_V4_UNIT_PROMPT.o COLI_V4_UNIT_NATIVE_QUANT.o
 	$(CC) $(CFLAGS) $< $(V4_TEST_LINK_OBJS) -o $@ -pthread $(LDFLAGS)
 
+V4_HOT_STORE_TEST_OBJS = \
+	COLI_V4_UNIT_ST.o \
+	COLI_V4_UNIT_EXPERT_STORE_HOT_ROWS16.o \
+	COLI_V4_UNIT_NATIVE_QUANT_ROWS16.o
+
+tests/test_deepseek_v4_hot_store$(EXE): tests/test_deepseek_v4_hot_store.c \
+		deepseek_v4.c deepseek_v4.h Makefile.deepseek-v4.units
+	$(MAKE) -f Makefile.deepseek-v4 ARCH=$(ARCH) $(V4_HOT_STORE_TEST_OBJS)
+	$(CC) $(CFLAGS) $< $(V4_HOT_STORE_TEST_OBJS) -o $@ -pthread $(LDFLAGS)
+
 V4_OWN_DIR = build/ownership
 V4_OWN_CFLAGS = $(CFLAGS) -DCOLI_V4_TEST_HOOKS
 V4_OWNERSHIP_OBJS = \
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -5733,6 +5733,7 @@ typedef struct V4HotPolicy {
     uint64_t *layer_requests;
     int *pins;
     unsigned char *packed;
+    pthread_mutex_t *pack_mutexes;
     uint64_t packed_slots;
     uint64_t history_total;
     int history_seeded;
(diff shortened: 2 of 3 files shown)


JustVugg:
Closing as a duplicate of **#1023**, which targets the same issue (#900) and is currently green while this branch is red on four jobs (Engine Linux, Linux, ASan+UBSan, Windows UCRT64).

To be clear about the credit: you and the author of #1023 converged independently on the same diagnosis — that the rows16 pin-packing has no business running under `state->mutex` — and that convergence is itself useful evidence the fix is right. #1023 is smaller (+79/-2) and follows the direction the issue itself recommended, so it's the one I'll take.

Thanks for the analysis; please don't read this as the work being unwanted.

PR #986, opened 2026-08-12
openai_server: use COLI_TEMP as the gateway default temperature

Blakeolson21:
## Summary

- use `COLI_TEMP` as the gateway default when a request omits `temperature`
- seed `COLI_TEMP` from `--temp` in `coli serve`, whose gateway runs in-process and never saw the launcher env
- keep an explicit request temperature authoritative
- fall back to the existing `0.7` when `COLI_TEMP` is absent, malformed, or outside the gateway's 0..2 range

## Problem

The launcher publishes `--temp` through `COLI_TEMP`, following #509 and #968. The API gateway then replaced that value with its own hard-coded `0.7` whenever request JSON omitted `temperature`.

The built-in non-GLM chat client omits the field, so `coli chat --temp 0.25` silently generated at `0.7`. `coli serve` and `coli web` had the same mismatch twice over: the gateway ignored an exported `COLI_TEMP`, and `--temp` never reached the in-process gateway at all because `env_for_engine` builds a copy for the engine child only.

Use `COLI_TEMP` at the point where the gateway chooses a missing request default, and have `cmd_serve` seed it from `--temp` for the in-process gateway. Explicit API values still win, and an env value that would not pass the gateway's own temperature validation falls back to `0.7` instead of failing requests that never sent the parameter.

## Validation

- `python3 -m unittest tests.test_openai_server.TemplateTest.test_coli_temp_is_the_default_for_requests_that_omit_temperature -v`
- `python3 -m py_compile openai_server.py`
- `make check`

The focused test covers the environment default, an explicit request value of zero, and malformed or out-of-range values (`malformed`, `5`, `-1`, `nan`, `1e999`), each falling back to `0.7`. `make check` passed all C tests and 446 Python tests, with 57 dependency or platform skips.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -1321,6 +1321,7 @@ def cmd_serve(a):
                             "olmoe-colibri" if arch=="olmoe" else
                             "glm-5.2-colibri")
     try:
+        if a.temp is not None: os.environ["COLI_TEMP"] = str(a.temp)
         openai_server.serve(a.model,a.host,a.port,model_id,a.api_key,
               a.cap,ngen_for(a,interactive=True),engine,env_for_engine(a,arch),a.cors_origin,
               a.max_queue,a.queue_timeout,a.kv_slots,allowed_hosts=a.allowed_host)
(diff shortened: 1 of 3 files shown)


JustVugg (approved):

PR #988, opened 2026-08-12
feat(v4): dual-SSD mirror (COLI_MODEL_MIRROR) for the DeepSeek V4 expert store

dcutugno:
Add SSD mirror streaming support to the DeepSeek V4 engine, allowing expert reads to be split across multiple drives for doubled I/O bandwidth on cold cache misses.

## Commits
- **2e65c13** feat(v4): dual-SSD mirror (COLI_MODEL_MIRROR) for the expert store — Ports the COLI_MODEL_MIRROR mechanism from the GLM engine to deepseek_v4.c. At startup probes each drive's bandwidth and splits expert reads via deterministic hash.
- **725ead5** fix(v4/mirror): use linear-index hash for balanced expert routing — Replaces XOR-based hash with linear-index multiply-and-shift, fixing hot-expert clustering on one drive.
- **bb92d80** fix(v4/mirror): count primary direct reads in mirror stats — v4_read_direct_window was missing primary direct reads, making the split appear as ~33/67 when it was actually ~49/51.

Files changed: c/deepseek_v4.c, c/deepseek_v4_internal.h (+367/-30 lines)

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -13,6 +13,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 
 static int set_error(char *error, size_t size, const char *format, ...) {
     if (error && size) {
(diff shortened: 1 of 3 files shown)


dcutugno:
Pushed 3d96376 with three review follow-ups:

- **Serve-mode telemetry**: the cumulative per-drive read split now prints on stderr after every DONE (`v4_mirror primary=…GiB/… mirror=…GiB/…`), emitted as a single write so it stays greppable next to the gateway's log lines. The CLI-only print left the long-lived serve mode blind to whether the split actually works.
- **Per-replica O_DIRECT**: `coli_st_streaming_direct_available_rep()` decides direct-I/O availability for the *routed* replica (same `COLI_V4_DIRECT` gate) — a mirror with an O_DIRECT twin no longer falls back to buffered reads just because the primary lacks one, and vice versa.
- **Config-sized routing**: the flat-index route hash uses the store's `experts_per_layer` (threaded through `coli_st_mirror_setup`) instead of a hardcoded 256. Identical routing for V4-Flash; exact for any future expert-count geometry.

Measured in production serve (2× NVMe, probe 4.96/4.88 GB/s): **51.6/48.4 split** over ~22 GiB of expert reads in a live session — including bank-refill streaming, i.e. the hot path, not just cold start. Engine builds clean, `test_deepseek_v4_dspark_source` 9/9.

dcutugno:
Refreshed with current `dev` (7a671f0). Two stacked PRs build on this branch: #1053 (prefix checkpoints / segmented prefill) and #1055 (CUDA tier), so their diffs against `dev` include the mirror until this merges.

PR #989, opened 2026-08-12
cuda: keep async expert slots alive through completion

### LINKED ISSUE #933: CUDA async group path can recycle an ESlot while a GPU read is in flight (needs per-slot refcount)

ZacharyZcR:
Fixes #933.

Async CUDA expert groups can retain borrowed zero-copy views after `moe()` returns. This change adds per-slot in-flight references, acquires them before every async group issue, and releases them only after the corresponding stream-completing `take()`.

Eviction, pilot reload, io_uring reload, and RSS reclamation now skip referenced slots. If every LRU slot is busy, cache promotion is skipped with a one-time warning instead of recycling live storage or waiting indefinitely.

A regression test covers victim selection while slots are referenced.

Verification:
- `make -C c test-c`
- `make -C c -j2`

CUDA hardware/toolkit is not available locally; CUDA-specific compilation and runtime coverage are left to CI and maintainer hardware.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -941,6 +941,9 @@ tests/test_cap_precedence$(EXE): tests/test_cap_precedence.c colibri.c st.h urin
 tests/test_cap_mixed_width$(EXE): tests/test_cap_mixed_width.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h route_trace.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
+tests/test_eslot_inflight$(EXE): tests/test_eslot_inflight.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h route_trace.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+
 tests/test_ssd_probe$(EXE): tests/test_ssd_probe.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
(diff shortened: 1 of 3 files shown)


JustVugg (approved):
The per-slot refcount from #933's analysis, implemented: async GPU readers borrow an ESlot, in_flight guards it from LRU recycling, and the underflow abort turns a lifetime bug into a loud crash instead of silent corruption. Test on CPU covers the refcount logic. This is the fix that makes zero-copy expert views safe.

PR #992, opened 2026-08-13
deepseek_v4: wire tool calling via checkpoint-native DSML encoding

gcap0n1:
## What

Tool calling for the DeepSeek V4 engine (`/v1/chat/completions` with `tools` →
`tool_calls`), closing the "tools and grammar are rejected" serving gap for V4.

Instead of inventing a prompt format, this vendors the checkpoint's **own DSML
encoding** (`encoding/encoding_dsv4.py` inside `deepseek-ai/DeepSeek-V4-Flash-0731`)
into `c/v4_dsml.py`:

- tool schemas render into the system turn with the official `## Tools` DSML template
- assistant `tool_calls` history renders as `<｜DSML｜invoke>` blocks
- `tool` role results merge into user turns as `<tool_result>` blocks
  (V4 has no standalone tool role — this is the native merge)
- reply parsing splits the `\n\n<｜DSML｜tool_calls>` block and decodes
  `string="true|false"` parameters back to JSON arguments

## Changes

- **`c/v4_dsml.py`** (new): vendored reference encoding. Encode side kept byte-exact
  (verified against the checkpoint's own fixtures); parse side made fault-tolerant
  for gateway use — malformed/truncated DSML degrades to visible content + stderr
  log, never raises to the client (same philosophy as GLM salvage-off).
- **`c/openai_server.py`** (+~90 lines):
  - `render_chat_v4_tools()`: used when `tools` are declared or the history contains
    `tool`/`tool_calls` messages; plain chat rendering is byte-identical to before
  - `parse_tool_calls_any()`: routes reply parsing per arch (GLM unchanged)
  - streaming tool-marker suppression now uses the per-arch opener
    (`<｜DSML｜tool_calls>` for V4)
  - `tool_choice`: `auto`/`none` supported; `required`/forced returns an honest 400
- **`c/tests/test_openai_tools_v4_e2e.py`** (new): mirrors `test_openai_tools_e2e.py`
  against the deepseek_v4 arch with a DSML-speaking mock engine (4 tests).

The Anthropic `/v1/messages` endpoint gets `tool_use` blocks for free, since it
translates onto the same OpenAI path.

## Validation

- `test_openai_tools_v4_e2e`: **4/4 pass** (non-stream call, `<tool_result>` round
  trip, streamed marker suppression across 12 chunks, plain chat unchanged)
- `test_openai_tools_e2e` (GLM regression): **6/6 pass**
- `test_openai_server`: 122/123 — the single error (`resource_plan.physical_cpu_count`
  mock) reproduces identically on unpatched main
- **Real checkpoint probe**: `tools/try_tool_calling.py` PASS against
  DeepSeek-V4-Flash-0731 (284B) served CPU-only on consumer hardware (6-core Ryzen 5,
  32 GB RAM, `--ram 22`, greedy): turn 1 `get_weather({"city": "Rome"})`, turn 2
  answer consumes the tool result ("sunny, 31°C")

## Scope / follow-ups

- `tool_choice: required` / forced function: 400 for now, prompt-level support is a
  small follow-up
- grammar/`response_format` on V4 remains rejected (unchanged)
- serving stays greedy + 1 KV slot (unchanged)

## Nota strategica

Questa PR è la **SPEC-41** (prima PR mergiata) e chiude il cerchio con il commento su #175:
arriviamo al progetto con un dataset pubblico E con codice funzionante misurato sul
modello reale. Se il maintainer chiede modifiche, si itera sul branch — è il gioco.

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -19,6 +19,8 @@ import sys
 import threading
 import time
 import uuid
+
+import v4_dsml                      # DeepSeek V4 DSML tool-call encoding (vendored)
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from pathlib import Path
 from urllib.parse import unquote, urlsplit
@@ -308,6 +310,13 @@ def _unclosed_tail(reply, tools):
     declared = {(t.get("function", t) if isinstance(t, dict) else {}).get("name")
                 for t in (tools or []) if isinstance(t, dict)}
     return inner if inner.strip() in declared else None
+
+
+def parse_tool_calls_any(reply, tools):
+    """Route reply parsing to the active engine's tool-call format."""
+    if ARCH == "deepseek_v4":
+        return v4_dsml.parse_completion_text(reply)
+    return parse_tool_calls(reply, tools)
 
 
 def parse_tool_calls(reply, tools=None):
@@ -670,9 +679,16 @@ def render_chat_v4(messages, enable_thinking=False, reasoning_effort=None, tools
     """
     if not isinstance(messages, list) or not messages:
         raise APIError(400, "`messages` must be a non-empty array.", "messages")
-    if tools or tool_choice not in (None, "none"):
-        raise APIError(400, "Tool use is not wired up for DeepSeek V4 yet.",
-                       "tools", "unsupported_parameter")
+    if tool_choice == "none":
+        tools = None
+    elif tool_choice not in (None, "auto"):
+        raise APIError(400, "`tool_choice` values other than auto/none are not wired up "
+                            "for DeepSeek V4 yet.", "tool_choice", "unsupported_parameter")
+    history_has_tools = any(isinstance(m, dict) and
+                            (m.get("role") == "tool" or m.get("tool_calls"))
+                            for m in messages)
+    if tools or history_has_tools:
+        return render_chat_v4_tools(messages, enable_thinking, reasoning_effort, tools)
     bos = "<\uff5cbegin\u2581of\u2581sentence\uff5c>"
     user = "<\uff5cUser\uff5c>"
     assistant = "<\uff5cAssistant\uff5c>"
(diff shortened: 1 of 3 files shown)


dcutugno:
This one seems better than my previous #948, so better drop mine or merge this...

gcap0n1:
Thanks @dcutugno — and congratulations on #948, we independently converged on the exact same design (vendored official DSML encoding, tolerant parse, per-arch dispatch). No point duplicating: I'm happy to close this in favor of the merged work.

What this PR still adds on top: a standalone e2e suite (4 tests, DSML-speaking mock engine over the SERVE wire protocol, incl. streamed marker suppression across chunks) and a live validation of the full two-turn tool loop on the real 284B checkpoint served CPU-only on consumer hardware (6-core, 32 GB RAM, greedy) — try_tool_calling.py
PASS. If useful, I'll rebase the e2e suite onto dev and open a tests-only follow-up PR, plus file the CPU-serving numbers where they belong.

dcutugno:
What i recommend is keep my 948, but refactor it with your vendor encoder, i'm currently doing it, stay tuned!

dcutugno:
Done this PR can be closed!

JustVugg:
Welcome, and thank you — this is real work. One thing before it can merge: it overlaps #1003 (@dcutugno), who authored the already-merged tool-calling base (#948) and is refactoring the same `v4_dsml.py` toward checkpoint-native DSML. Two implementations of the same thing can't both land.

Please sync with @dcutugno on which carries — if your checkpoint-native encoding is cleaner or more correct than the #948 base, that's exactly the kind of thing worth grafting in. I'll merge whichever you settle on, on green. Not a rejection: a collision to resolve.

JustVugg:
Closing in favour of #1003 (now merged) — you and @dcutugno converged on the same insight: use the checkpoint's native DSML encoding instead of a re-implementation. #1003 vendors the reference primitives from the author of the already-merged base (#948) and is green; this one targets the same files and is currently red on four checks. No wasted work — parallel discovery of the right idea, and #1003 credits the direction you both saw. Please keep contributing; the instinct was correct.

gcap0n1:
Thanks @JustVugg — fair resolution, and congratulations @dcutugno on #1003. For the record our angle was validated on a 284B V4-Flash running CPU-only in 32GB RAM (Contabo VPS), which is where #948's GPU-tested path hadn't been exercised. I'll port our 4 standalone e2e tests (mock-engine, OpenAI wire level) onto dev as a tests-only PR — they're arch-gated and complement the vendored primitives. If CI is green, it's yours; if not, I'll fix before asking for review.

Issue #971, opened 2026-08-12
[Bug]: Wrong CUDA test in 1.6.0?


ailabai:
### Commit or version

1.6.0, from https://github.com/JustVugg/colibri/archive/refs/tags/v1.6.0.tar.gz

### Environment

Ubuntu 22.04; NVIDIA RTX A6000, 50.9 GB VRAM, sm_86

### Reproduction steps

In v. 1.6.0 -- `setup.sh` worked fine, but `make cuda-test CUDA=1` gives error 1

### Expected behavior

With v. 1.5.0
```sh
/colibri-1.5.0/src/c$ make cuda-test CUDA=1
"/usr/local/cuda/bin/nvcc" -O3 -std=c++17 -arch=native -Xcompiler=-Wall,-Wextra backend_cuda.cu tests/test_backend_cuda.cu -o backend_cuda_test 
./backend_cuda_test
[CUDA] device 0: NVIDIA RTX A6000, 50.9 GB VRAM, sm_86
[CUDA] tensor allocation: out of memory
cuda backend: q8/q4/q2/f32/e8 correctness ok on 1 device(s)
"/usr/local/cuda/bin/nvcc" -O3 -std=c++17 -arch=native -Xcompiler=-Wall,-Wextra backend_cuda.cu tests/test_ragged_attention.cu -o ragged_attention_test
./ragged_attention_test
[CUDA] device 0: NVIDIA RTX A6000, 50.9 GB VRAM, sm_86
ragged_relative_rms=0 ragged_g64_relative_rms=0 warm_ms=2.222 cross_page_growth_ms=0.099 grow_1024_ms=2.587 steady_1024_ms=0.167
```

### Actual behavior and logs

```shell
/colibri-1.6.0/c$ make cuda-test CUDA=1
"/usr/local/cuda/bin/nvcc" -O3 -std=c++17 -arch=native -Xcompiler=-Wall,-Wextra backend_cuda.cu tests/test_backend_cuda.cu -o backend_cuda_test 
./backend_cuda_test
[CUDA] device 0: NVIDIA RTX A6000, 50.9 GB VRAM, sm_86
make: *** [Makefile:677: cuda-test] Error 1
```

ailabai:
It is
```cpp
if (coli_cuda_tensor_upload(&bad, q8, s8, 7, 4, 2, d0)) return 1;
```
this place

JustVugg:
Look closely at your own v1.5.0 log — it contains the real story:

```
[...]

PR #995, opened 2026-08-13
test(cuda): link the MXFP4 OpenMP reference

Blakeolson21:
## Summary

- link the MXFP4 CPU oracle OpenMP runtime in the CUDA and HIP test binaries
- keep the setup OpenMP probe in the writable source directory
- add dry-run checks for both GPU compiler recipes

Issue #971 reached the MXFP4 correctness test on an A6000, then failed to link `omp_get_num_threads`, `omp_get_thread_num`, and `GOMP_parallel`. The host compiler builds the oracle with OpenMP, but the final GPU compiler command did not link the matching host runtime.

## Validation

- `python3 -m unittest c.tests.test_cuda_test_makefile -v`
- `bash -n c/setup.sh`
- `make -C c check`: all C tests and 450 Python tests passed, with 57 expected skips

I could not run `make -C c cuda-test` on this machine because it has no CUDA runtime or GPU. Actual CUDA execution is unverified.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -325,6 +325,9 @@ endif
 GPUCC      = $(NVCC)
 GPUFLAGS   = $(NVCCFLAGS)
 GPUCC_NAME = nvcc (set CUDA_HOME or NVCC)
+# tests/mxfp4_ref.c is compiled by $(CC) with OpenMP. Its object therefore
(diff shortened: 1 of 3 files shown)


JustVugg (approved):

PR #996, opened 2026-08-13
fix(download): fail when checkpoint remains incomplete

Blakeolson21:
## Summary

- return a nonzero status when shard retries are exhausted or metadata is missing
- return zero only for a complete checkpoint
- keep `--source ms` from silently falling back to Hugging Face
- switch automatic downloads to Hugging Face when ModelScope returns an empty manifest
- return 130 on Ctrl-C

The downloader could print that shards were still missing and exit successfully. Automation could then continue with an incomplete checkpoint.

## Validation

- `python3 -m unittest tests.test_download_fp8 -v`
- `python3 -m py_compile download_fp8.py tests/test_download_fp8.py`
- `python3 -m pyflakes download_fp8.py tests/test_download_fp8.py`
- `make check`: all native tests and 454 Python tests passed, with 57 expected skips

### DIFF
--- a/c/download_fp8.py
+++ b/c/download_fp8.py
@@ -4,7 +4,7 @@ Usage: python download_fp8.py
        python download_fp8.py --parallel 4
        python download_fp8.py --source hf  (force HuggingFace)
 """
-import os, sys, time, threading, argparse, subprocess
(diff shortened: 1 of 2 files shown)


JustVugg (approved):

PR #997, opened 2026-08-13
fix(kimi): guard the Vulkan-only state

Blakeolson21:
## Summary

- declare the Vulkan liveness flag only when Vulkan is compiled in
- keep the ordinary Kimi build warning-free

The non-Vulkan build declared `g_k3_vk` even though every use was compiled out. Apple clang reported it as an unused variable.

## Validation

- `make -s -C c clean && make -s -C c kimi_k3 EXTRA_CFLAGS=-Werror`
- `make check`: all native tests and 447 Python tests passed, with 57 expected skips


JustVugg (approved):
Moves g_k3_vk inside the COLI_VULKAN guard; unused global in non-Vulkan builds. Green across platforms.

PR #1000, opened 2026-08-13
fix(supervisor): stop only its own converter

Blakeolson21:
## Summary

- find the converter that owns this supervisor's exact output directory
- stop or restart only that converter
- handle model paths containing spaces

The supervisor used global `pgrep` and `pkill` patterns. Finishing or stalling one model conversion could stop a separate conversion running on the same host.

## Validation

- Linux behavior test passed in a local Python 3.12 container with two concurrent converters
- `bash -n c/scripts/supervisor.sh`
- `shellcheck -S error c/scripts/supervisor.sh`
- `make check`: all native tests and 448 Python tests passed, with 58 expected skips

The Linux behavior test is skipped by the macOS full gate because it requires `/proc` and `flock`.


JustVugg (approved):

PR #1002, opened 2026-08-13
fix(stop): scope shutdowns to the requested port

Blakeolson21:
## Summary

- stop only the `coli serve` wrapper for the requested port
- tag engine workers with their owning serve port and scope engine shutdowns to it
- keep pidfile shutdown working on native Windows when `/proc` is absent
- close `/proc` and pidfile handles after discovery

`coli stop --port 8000` scanned for every Colibri server and every serving engine. On a host serving multiple ports, it could stop unrelated instances.

## Validation

- Linux behavior test passed in a local Python 3.12 container with two servers on different ports and resource warnings treated as errors
- `python3 -m unittest tests.test_stop_scope -v`
- `python3 -m py_compile coli tests/test_stop_scope.py`
- `make check`: all native tests and 455 Python tests passed, with 58 expected skips

The Linux process test is skipped by the macOS full gate because it requires `/proc`.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -1302,6 +1302,34 @@ def cmd_chat(a):
 
 def serve_pidfile(port): return os.path.join(tempfile.gettempdir(), f"coli-serve-{port}.pid")
 
+def _serve_cmdline_matches_port(raw, port):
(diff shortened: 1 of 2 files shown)


JustVugg (approved):

PR #1003, opened 2026-08-13
refactor(v4): vendor the checkpoint's DSML primitives behind #948's tool calling

dcutugno:
## What

#948 wired DeepSeek V4 tool calling with an in-gateway re-implementation of the checkpoint's DSML encoding — byte-verified against `encoding_dsv4.py` at review time, but a re-implementation drifts silently if DeepSeek revises the encoding. This vendors the reference primitives instead and delegates to them, so a future encoding revision becomes a mechanical re-vendor + diff.

## Changes

- **`c/v4_dsml.py`** (new): verbatim excerpts of the checkpoint's `encoding/encoding_dsv4.py` covering exactly the surfaces the gateway speaks — the `## Tools` declaration template, the `<｜DSML｜tool_calls>` block encoder, and the strict block parser. Two gateway adaptations, both marked ADAPTED in place: `arguments` accepted as JSON string or dict, and a never-raises wrapper (`parse_completion_text`) around the strict parser.
- **`c/openai_server.py`**: `_dsv4_tools_block` / `_dsv4_tool_calls` / `parse_dsv4_tool_calls` now delegate to the vendored module. The hand-maintained template string and the three parse regexes are deleted; parsing gains the reference's strict validation (duplicate parameter names, malformed invokes) while keeping the gateway's degrade-to-content behaviour for truncated output (raw DSML still never leaks into visible content).

## Deliberately NOT vendored (rationale in the module docstring)

- `render_chat_v4`'s turn structure: the gateway always terminates the prompt with an assistant generation cue (the reference leaves a trailing assistant `tool_calls` turn uncued) and renders `reasoning_content` wherever present.
- The tuned reasoning-effort prompts with an explicit token budget — the reference's unbounded variants routinely consumed the whole completion budget with reasoning.

## Verification

- `test_openai_server`: 135/135, `test_anthropic_messages`: 27/27 — identical before and after (no wire-format change).
- The DSML bytes produced are unchanged: the vendored tools template is byte-identical to the string it replaces, and the block encoder produces the same layout the reference emits.

Related: this supersedes the vendoring idea from #992 — same goal, but implemented as a delegation refactor on top of #948 instead of a parallel second implementation, so there is a single render/parse path.

### DIFF
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -19,6 +19,8 @@ import sys
 import threading
 import time
 import uuid
+
+import v4_dsml                      # vendored DeepSeek V4 DSML reference primitives
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from pathlib import Path
 from urllib.parse import unquote, urlsplit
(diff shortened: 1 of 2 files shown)


JustVugg:
Heads-up on an overlap: #992 (@gcaponi, first-timer) targets the same thing — checkpoint-native DSML for V4 tool calling — and touches the same `c/v4_dsml.py` + `openai_server.py`. You authored #948 (the merged tool-calling base) so this refactor is the natural continuation, but I don't want to merge one and strand the other's work.

Could you two coordinate on which lands? If #1003 is the vendor-the-primitives cleanup on top of #948 and #992 is a parallel rewrite, the clean path is probably #1003 with anything genuinely better from #992 grafted in. Whichever way — one of these, not both, and I'll merge on green once you've settled it.

JustVugg (approved):
This is the right resolution of the tool-calling overlap: instead of the in-gateway re-implementation from #948 (byte-verified once, but silently drift-prone if DeepSeek revises the encoding), it vendors the checkpoint's reference DSML primitives and delegates to them — a future encoding change becomes a mechanical re-vendor + diff instead of a silent divergence. Green, from the author of the merged base. This supersedes #992 (same goal, currently red).

PR #1016, opened 2026-08-14
security: fix deepseek_v4 DSpark drafter: MTP expert slab sized from expert 0 only, per-expert (GHSA-8p69-257x-6fp6)

zh-Processor:
Private security fix for advisory GHSA-8p69-257x-6fp6.

**deepseek_v4 DSpark drafter: MTP expert slab sized from expert 0 only, per-expert copy unchecked — heap buffer overflow write with file-controlled content/length**

See the private advisory for root cause, PoC and discussion: https://github.com/JustVugg/colibri/security/advisories/GHSA-8p69-257x-6fp6

Fix follows the advisory's suggested patch, adapted from v1.5.0 to current v1.6.1 code.


JustVugg:
Thanks for the detailed report and fix. All six advisories were consolidated into a single reviewed change on `dev` (#1018, merged) and are shipping in **v1.6.2** (release #1019). Closing this in favour of #1018 — full credit noted in the advisory. 🙏

Issue #1009, opened 2026-08-14
Proposal: opt-in HQQ 5-bit AXIS=0 MoE weights — −25% bytes, +36% E2E on disk-bound hardware, quality within +2.4% ppl of int8


outtodata:
## Proposal: opt-in HQQ 5-bit AXIS=0 MoE weights — −25% bytes, +36% E2E on disk-bound hardware, quality within +2.4% ppl of int8

**TL;DR.** A converter-side, opt-in 5-bit weight format for MoE expert FFNs (attention stays int8), using HQQ AXIS=0 grouping so scales/zeros are per *input channel*. On a Core i5-7300U with weights streamed from disk: snapshot 7.0 GB → 5.4 GB (−25%), end-to-end 2.42 → 3.28 tok/s (+36%, medians of 3; up to 2.46× in more disk-bound conditions), HF-corpus perplexity 14.2420 vs 13.9144 int8 control (+2.36%), greedy token streams bit-identical to the int8 engine in all 8 test runs. Everything sits behind `bits==5` in the converter + a snapshot field; default builds and int8 snapshots are untouched. Related: the axis finding in [issue draft: HQQ axis asymmetry].

**Motivation.** On disk-bound CPUs the expert pipe, not matmul, is the wall (profiled: `pipe_wait` ≈ 45% of decode time at Q8_0). The cheapest FLOP is the one you never load: cutting weight bytes 25% cuts pipe traffic 25%. But 5-bit only works if the grouping axis is chosen correctly — HQQ's AXIS=1 mixes 32 heterogeneous input channels under one scale and costs +9% ppl (separate issue); AXIS=0 keeps scales per input channel and lands at +2.4%.

### Design (mirrors the existing i4 pattern)

- **Converter** (`convert_olmoe_hqq5.py`): explicit `--bits 5` choice at conversion time, per the repo's "quantization choice lives in the converter" policy. Writes `merged_weight5_ax0` tensors + a snapshot header field. Weight layout is the natural flat order of the HQQ `q` tensor (no repacking needed — AXIS=0 groups are strided slabs, so the packed stream is identical in form to the axis=1 variant: 20 B per 32 weights = 16 B nibbles + 4 B hi-plane). Meta (`scales|zeros`) is contiguous per row, one 64 KB slab per 32 rows.
[...]

PR #1024, opened 2026-08-14
feat(moe): FUSED3=1 opt-in AVX2 expert matmul — 40% less matmul time, bit-identical output, off by default

outtodata:
## FUSED3: vectorized activation quantizer + 4-way IDOT + gate/up pair matmul — 40% less matmul time, bit-identical output, off by default

**TL;DR.** Three opt-in kernel upgrades in `c/fused_simd.h` (env `FUSED3=1`, default off): (1) AVX2-vectorized activation quantization replacing the scalar tail of `quant_x_q8`, (2) a 4-accumulator IDOT matmul variant with prefetch, (3) a paired gate/up matmul that shares the activation stream. On a Core i5-7300U (AVX2, 2C/4T), time inside expert matmul drops ~40%, greedy token streams are bit-identical to master, and end-to-end throughput on our production config goes 2.08 → 2.38 tok/s. No change to file formats, no approximation anywhere: all arithmetic is still exact integer IDOT.

**Setting.** OLMoE-1B-7B Q8_0 snapshot, our production flags, Windows + WSL pipes (i.e. the *noisy* setting — numbers below are conservative; on a quiet machine the deltas are larger). MoE weights streamed from disk via the async expert pipe.

### What changes (all under `FUSED3`, default OFF)

| Kernel | Before | After |
|---|---|---|
| `quant_x_q8` | scalar loop for tail + row quant | `quant_x_q8_avx2`: vector abs-max + quant, exact same round-to-nearest-even semantics |
| `matmul_q_idot_v2` | 2-way unroll | `matmul_q_idot_v3`: 4 independent int32 accumulators + `_mm_prefetch` of the next weight row |
| gate/up (proj0/proj2) | two separate matmuls, activation re-read | `matmul_q_idot_pair_v3`: single pass, both outputs, shared activation cache traffic |

Dispatch is a single `if (g_fused3)` in `olmoe.c` at the MoE FFN call site; nothing else is touched.

### Measured (bench_fused3_inproc, in-process, isolated kernels; medians of ≥10)

| Metric | master | FUSED3 | Δ |
|---|---|---|---|
| time inside expert matmul | baseline | −40% | matmul_ms 7664→4172 per 200-token decode (bench_fused3_tokens, 8 experts, dim 2048) |
| E2E (quiet machine, all our prod flags) | 2.08 tok/s | 2.38 tok/s | +14% |

Output equivalence: md5 of the 200-token greedy stream identical with and without the flag across three independent runs (`d6cb4e151fa83994d62982f3cbba4190`), and identical to a separately-built reference binary. This is expected — every path is exact integer arithmetic; the flag only changes *instruction scheduling*, never values.

### How this answers the review gates from #906

- **(a) off by default, loud when on:** `FUSED3` defaults to 0; when set, startup prints `MODE FUSED3 ...` banner plus per-layer `[PROF]` counters already present in the build.
- **(b) gate semantics exact:** the flag selects between two *exact* implementations of the same integer kernel. No approximation, no numeric mode change. Documented as such at the dispatch site.
- **(c) token-exact with flag off:** when off, the code path is byte-for-byte today's master (the new kernels are not even linked into the call path). Verified by identical md5 on flag-off vs master build.

### What this is NOT

- Not a quantization change: weights stay Q8_0, activations stay Q8_0. Pairs naturally with the int8 path; orthogonal to any 4/5-bit weight work.
- We also tested a 2-worker parallel expert matmul (`EXPERT_WORKERS=2`) in the same series: it is a **NO-GO** from us (pipe interference eats the gain, and it introduces thread-count-dependent scheduling). Not proposed.

### Caveats

- Numbers are from one machine (i5-7300U, AVX2). No ARM/NEON variant of the new kernels — the flag should refuse or fall back on non-AVX2 builds (current code: compile-time `#if defined(__AVX2__)` around v3, scalar/v2 fallback otherwise).
- The +14% E2E is on a disk-bound setup; on NVMe the relative gain is smaller (matmul is a smaller slice). The −40% matmul-time figure is the portable number.

Diff is ~200 lines in `c/fused_simd.h` + a dispatch flag + a bench binary. Happy to open the PR if the direction looks right, or to carve it differently (e.g. three separate flags) if you prefer finer-grained gating.

### Related

- #1009 — HQQ 5-bit AXIS=0 weights RFC (orthogonal: that one is a converter-side format proposal; this PR keeps weights Q8_0).
- The bench harness for the correctness gates is included as `c/tests/bench_fused3.c` (`v3_quant_bitexact=yes`, `v3_output_bitidentical_v2=yes` on this branch).


JustVugg:
Cross-linking a decision that affects how this PR should be described: #1044 established that olmoe's IDOT branch quantizes the **activations**, so it is not token-exact against the fp32 path — and the decision there is to return `IDOT` to **off by default** on x86.

Two consequences for FUSED3:

1. **Correctness is unaffected.** Your kernels accelerate the IDOT path and are bit-identical to it; I verified that independently (I rebuilt `bench_fused3` on an AVX2 box: `v3_quant_bitexact=yes`, `v3_output_bitidentical_v2=yes`, and ~1.21× on the kernels here).

2. **The wording needs scoping.** I also compared FUSED3 against the *stock* `matmul_q` IDOT branch bit-for-bit and got **915/1024 outputs differing** (ulp-level; the FMA plus the two accumulator chains change the rounding). So "bit-identical output" reads as "token-exact", which it isn't — it's "bit-identical to the IDOT path". With `IDOT=0` becoming the default, it's also no longer the shipped configuration, so the 40% figure needs that caveat.

Please retitle/reword along those lines and this is good to go — the engineering is sound, the isolation is exemplary (one new header, one engine, opt-in, `__AVX2__`-guarded), and the speedup is real. It's only the claim I want precise, because exactness is the one thing this project can't be sloppy about.

JustVugg:
Status ping: this is currently `dirty` against `dev`, which has moved a good deal in the last day — the CUDA tier for DeepSeek V4 landed (#1054, #1055), the dual-SSD mirror (#988), and #1063 made model families registry-owned across `coli` / the gateway / `doctor.py` / `resource_plan.py`.

A rebase would put it back in the queue. I'm not asking as a formality — the change is still wanted, it just can't be reviewed or merged in this state.

If you'd rather not carry it any further, say so and I'll close it with the reasoning recorded, so the work is findable rather than silently dropped. Either answer is fine; what I'd like to avoid is it sitting here indefinitely because nobody said anything.

JustVugg:
This PR just became more strategic than when it was opened, and we would like to get it in — could you rebase onto current `dev`?

Context: we have been auditing the expert hot path across engines, and the audit's conclusion matches what FUSED3 does — the win is fusing gate+up+silu into fewer passes while keeping bit-identical output. Your measured "40% less matmul time, bit-identical, off by default" on olmoe is exactly the shape we want to replicate on the GLM engine next, so having yours land first gives us a reviewed reference in-tree.

Two things while you rebase, if you don't mind:
1. `dev` moved today (#1036, #1017, #1068 merged) — the conflict is likely trivial but real.
2. If you have the A/B numbers handy (tok/s before/after on your box, and the bit-exactness check you used), putting them in the PR description helps the review go fast.

JustVugg:
Update, and an explanation you deserve: **we rebased this ourselves and opened #1082 carrying your commit, authorship intact.** Nothing was changed in your work — we resolved the one rebase conflict (dev's `OLMOE_TESTING` hook and your `fused_simd.h` include coexist) and re-ran your bench on the rebased branch: `v3_quant_bitexact=yes`, `v3_output_bitidentical_v2=yes`, 1.22–1.27× on our box.

Why we didn't just wait: we are mid-execution on a performance roadmap across the engines (the activation-hoist series #1071–#1077, the K1 planar int4 layout #1079, all merged this week), and the next CPU item on it is precisely the fused gate/up pass you built here. Your PR went from "nice to have" to **a dependency of our own sequencing** — the colibri.c fusion will copy your reviewed shape instead of inventing a second one. So the rebase was on us, not on you.

#1082 closes this one on merge and GitHub will credit you as the commit author. If you'd rather carry it yourself after all, say the word there and we'll step back. Thank you for writing a piece of our roadmap before we published it.

JustVugg:
Merged via #1082 with your commit and authorship intact — GitHub now lists you as a colibri contributor. Thank you again; the colibri.c fusion on our roadmap will cite this as its reference shape.

outtodata:
Thanks for carrying this over the line — and for the careful rebase plus re-verification on your own hardware, the authorship handling is much appreciated.

Glad the fused gate+up+silu shape matches what the audit concluded; if #1024 works as the reviewed reference for the GLM engine fusion, that is exactly where I hoped it would be useful. If the colibri.c fusion work hits anything the FUSED3 kernels can inform (or if you want the same treatment profiled on another engine), ping me — happy to help.

Cheers!

PR #1035, opened 2026-08-15
refactor: extract shared routed-expert FFN helper (byte-identical)

lineape:
## Summary

`moe()` inlines the four-step routed-expert FFN (`expert_gate_up → silu(gate)*up → [fmt=6 down-input rotation] → down matmul`) six separate times, and those six copies were written inconsistently: four omit the fmt=6 down-input rotation, two apply it. This extracts the sequence into one `expert_ffn` helper and routes all six call sites through it.

The helper takes the expert's own tensors plus buffers and a row count — no `moe()` locals — so a downstream distributed worker can call the same helper instead of carrying a parallel copy. Input arrives already rotated by the caller via `E8_XE` (the per-call cache is preserved, #452); the fmt=6 down-input rotation lives inside the helper and is applied whenever `d->fmt==6`.

## Validation

- `make` → 0 warnings.
- `make check` → green (C suite `ALL PASS (0 failures)`, Python `Ran 473 tests OK`).
- `gcc -fsyntax-only -DCOLI_CUDA -DCOLI_VULKAN -DCOLI_METAL -Wall -Wextra c/colibri.c` → 0 warnings (the `#ifdef`-gated call sites still compile).

## Compatibility

No new knobs, no behaviour change on any single-format model (the oracle `SNAP=./glm_tiny TF=1 ./colibri 64 16 16` is byte-identical before and after). The helper applies the fmt=6 down-input rotation unconditionally — it is a pure function of the down tensor's format. Of the four copies that previously skipped it, the two Vulkan device-lost fallbacks provably cannot see fmt=6 (the Vulkan tier loads only fmt 2/4/5), and the other two (Vulkan CPU-share, CUDA early-issued-take fallback) can reach fmt=6 only in a mixed-format container, where this now matches the canonical path. Single-machine path unchanged.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -1221,6 +1221,27 @@ static void softmax(float *x,int n){ float m=-1e30f; for(int i=0;i<n;i++) if(x[i
 static inline float sigmoidf(float x){ return 1.f/(1.f+expf(-x)); }
 static inline float siluf(float x){ return x/(1.f+expf(-x)); }
 
+/* FFN di un expert MoE (routed): gate+up -> silu(gate)*up -> down. Un solo helper per
+ * i ~6 siti inline di moe() e (piu' avanti) il worker distribuito, che la chiama con i
+ * propri buffer e i tensori dell'expert — nessun locale di moe(). xg arriva GIA' ruotata
+ * (Q^T x) dal chiamante via E8_XE (#452: una copia ruotata per chiamata, MAI per expert).
+ * La rotazione dell'input del down e' per-expert e vive qui: `d->fmt==6` la applica
+ * SEMPRE — e' la forma CANONICA, pura funzione del fmt del down-tensor. Perche' resta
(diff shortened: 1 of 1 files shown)


lineape:
Quick context for reviewers: this is a straight extraction — the four-step FFN body is moved verbatim into expert_ffn, six call sites now call it, and the fmt=6 down-input rotation is applied unconditionally inside the helper (a pure function of the down tensor's format). The only behavioral nuance is that two of the six copies previously skipped that rotation; both of those (Vulkan device-lost fallback, CUDA early-take fallback) either provably can't see fmt=6 or only could in a mixed-format container, so output is unchanged on any single-format model and the oracle is byte-identical. No new knobs, single concern.

I'm also shipping a follow-up PR (#1036) that consumes this helper from a distributed worker — mentioned only so the "so a downstream worker can call it" framing doesn't read as speculative generality.

PR #1056, opened 2026-08-16
build(v4/cuda): fetch the pinned DeepGEMM sm120 headers on first build — cuda-dsv4-dg-dll / DEEPGEMM=1 need no external checkout, nothing vendored

dcutugno:
Reworked per review (fetch script, nothing vendored). One commit on current `dev`; 7 files, +208/−21, zero third-party lines in the tree.

### Motivation
The DeepGEMM flavour of the DeepSeek V4 CUDA tier (#1055) replaces the generic fp32-from-fp8/fp4 GEMMs with sm_120a block-scaled tensor-core GEMMs for the batched dense projections and the MoE expert bank. Measured on an RTX 5080 16 GB (2× NVMe, current `dev` engine, DLL forced with `COLI_DSV4_DLL`, runs alternated):

| prompt | DeepGEMM DLL (sm_120a) | generic DLL | ratio |
|---|---|---|---|
| 3324 tokens, TTFT | **101.3 s / 102.4 s** (90 s in the docs' earlier run) | 231.5 s / 232.1 s | **2.28×** |
| 821 tokens, TTFT | **44.3 s** | 76.9 s | **1.74×** |
| 32 decode tokens | 22.2–22.7 s | 21.6–24.0 s | ≈ 1.0× (disk-bound) |

Text identical between the two flavours (same md5). It reaches sm_120a only, so it stays an opt-in build; this PR just removes the "clone DeepGEMM yourself" step.

### What it does
- **`c/tools/fetch_deepgemm.sh [<pinned-sha>] [<dest>]`** — shallow git checkout of the community sm120 port (bvolpato/DeepGEMM `sm120-full` @ `39fb4447`, MIT; upstream DeepGEMM has no sm120 kernels) plus its CUTLASS/CuTe submodule (BSD-3, gitlink `f3fde583`) into `c/third_party/deepgemm/` (**gitignored**); verifies both commit ids against the pins (content-addressed, so they are the checksums); applies `patches-deepgemm-sm120-msvc.patch` (MSVC ignores `alignas(64)` on by-value TMA descriptors); writes a marker so a re-run is a no-op and a tree at another commit is refetched. Forces an LF checkout regardless of `core.autocrlf` and normalises the patch, so it behaves the same from MSYS2 / Git for Windows / Linux. ~40 MB, one command, `git` required.
- **Makefiles**: `DEEPGEMM_HOME ?= third_party/deepgemm`, `DEEPGEMM_PIN`, and a stamp header (only present in the sm120 port) as a prerequisite of `cuda-dsv4-dg-dll` and of `backend_cuda_dsv4.o` under `DEEPGEMM=1`, so the fetch runs exactly when the tree is absent; `make deepgemm-fetch` runs it alone. `DEEPGEMM_HOME=<dir>` still points at an existing tree; bumping the pin is a one-line change.
- **CI** (Windows CUDA job): runs the fetch (pin check + patch + idempotency check) and caches `c/third_party/deepgemm` keyed on the pin + patch hash. The sm_120a compile itself still needs CUDA ≥ 12.8 and a Blackwell card, which CI does not have.
- **`THIRD_PARTY_NOTICES.md`** stays in-tree (ds4/ggml router adaptation as before; the DeepGEMM/CUTLASS section now describes what the fetch brings in and its licences).
- Docs updated.

### Verified
- Windows: `make cuda-dsv4-dg-dll` from a clean tree — fetch 37 s, DLL byte-size identical to the previously vendored build; the fetched+patched headers are byte-identical (modulo line endings) to the 78 files this PR used to vendor.
- Linux (WSL2, CUDA 13.3): `make -f Makefile.deepseek-v4 deepseek-v4 CUDA=1 DEEPGEMM=1` from a clean tree fetches and builds; second `make` is a no-op; `make deepgemm-fetch` reports "already at <pin>".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -535,7 +535,30 @@ jobs:
           path-type: inherit
           install: >-
             make
+            git
             mingw-w64-ucrt-x86_64-gcc
+      # The DeepSeek V4 DeepGEMM flavour compiles against a pinned checkout
+      # that tools/fetch_deepgemm.sh makes on first use (nothing vendored).
+      # CI cannot compile it (sm_120a needs CUDA >= 12.8; this job pins 12.6)
+      # but it does exercise the fetch — pin verification + MSVC patch — and
+      # caches the tree keyed on the pin so a bump is the only cache miss.
+      - name: DeepGEMM pin (for the cache key)
+        id: deepgemm-pin
+        shell: bash
+        run: echo "pin=$(sed -n 's/^DEEPGEMM_PIN ?= //p' c/Makefile | head -1)" >> "$GITHUB_OUTPUT"
+      - name: Cache the pinned DeepGEMM checkout
+        uses: actions/cache@v4
+        with:
+          path: c/third_party/deepgemm
+          key: deepgemm-${{ steps.deepgemm-pin.outputs.pin }}-${{ hashFiles('c/patches-deepgemm-sm120-msvc.patch') }}
+      - name: make deepgemm-fetch (pinned checkout + MSVC patch, no compile)
+        shell: msys2 {0}
+        run: |
+          cd c
+          make deepgemm-fetch
+          test -f third_party/deepgemm/deep_gemm/include/deep_gemm/impls/sm120_bf16_gemm.cuh || { echo "fetch_deepgemm produced no sm120 headers" >&2; exit 1; }
+          # second run must be a no-op (idempotency)
+          make deepgemm-fetch | grep -q "already at" || { echo "fetch_deepgemm is not idempotent" >&2; exit 1; }
       - name: make cuda-dll (nvcc + MSVC host)
         shell: msys2 {0}
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -121,6 +121,10 @@ c/tiny_inkling/
 # Claude Code working directory (agent worktrees, session scratch) — never a repo artifact
 .claude/
 
+# DeepGEMM sm120 checkout made by c/tools/fetch_deepgemm.sh (pinned commit,
+# not vendored): the DeepSeek V4 DeepGEMM flavour's headers.
+c/third_party/deepgemm/
+
 # DLL flavours + MSVC import/export artifacts of the V4 CUDA tier (regenerated
 # by cuda-dsv4-dll / cuda-dsv4-dg-dll); the loader picks the DLL at runtime.
 c/coli_cuda_dsv4*.dll
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -1,0 +1,50 @@
+# Third-party notices
+
+Portions of `backend_cuda_dsv4.cu`, including the DeepSeek-V4 GPU router
+selection algorithm, are adapted from `ds4_cuda.cu` in the ds4 project:
+https://github.com/antirez/ds4
+
+MIT License
+
+Copyright (c) 2026 The ds4.c authors
+Copyright (c) 2023-2026 The ggml authors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+The pinned DeepSeek-V4 reference runner in `c/tools/dsv4_vllm_reference.py`
+executes an unmodified vLLM checkout (commit
+`ffd46bfab2128bb84146050e98b51a617c6575ab`) as a behavioural oracle for the
+native port; no vLLM code is vendored.
+
+## DeepGEMM sm120 headers (fetched, not vendored: `c/third_party/deepgemm/`)
+
+The DeepSeek V4 CUDA tier's DeepGEMM flavour (`make cuda-dsv4-dg-dll`,
+`Makefile.deepseek-v4 DEEPGEMM=1`) compiles against headers that
+`c/tools/fetch_deepgemm.sh` checks out at a pinned commit into the gitignored
+`c/third_party/deepgemm/`; nothing from them is committed to this repository.
+What that checkout contains, and the licences that apply when you build with it:
+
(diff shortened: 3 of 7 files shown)


JustVugg:
I want to be straight with you about this one rather than let it sit, because the work is fine and the decision isn't about the work.

**The licensing you did correctly**, and that's the part people usually get wrong: `c/third_party/deepgemm/cutlass/LICENSE.txt`, `c/third_party/deepgemm/deep_gemm/LICENSE`, and a `THIRD_PARTY_NOTICES.md` that names the ds4/ggml MIT adaptation. Nothing to fix there.

**The problem is the ratio, and it comes from your own numbers.** From `docs/deepseek-v4.md` as merged:

| build | TTFT | 3 decode tokens |
|---|---|---|
| `CUDA=1 DEEPGEMM=1` (sm_120a) | 119.9 s | 32.4 s |
| `CUDA=1 CUDA_ARCH=portable` (generic) | 122.1 s | 33.2 s |

**DeepGEMM is ~2% faster than the generic kernels** — 1.8% on TTFT, 2.4% on those decode tokens. And it only applies to `sm_120a`, so that 2% reaches Blackwell owners and nobody else.

Against that: **~35,000 lines of vendored CUTLASS and DeepGEMM headers in the tree, permanently.** Repo weight for every clone, a vendored dependency to track against upstream, and a third-party surface inside a project whose entire pitch is "one C file per engine, zero dependencies". The alternative it removes is a single `git clone` plus `DEEPGEMM_HOME=<dir>` — real friction, but one command, paid once, by the small set of people who want the last 2%.

So my read is: **not worth merging as-is**, and I'd rather tell you now than after you rebase it again.

What I'd take instead, if you want the friction gone:

1. **A fetch step rather than a vendor.** A make target that clones DeepGEMM at a pinned commit into `c/third_party/` (gitignored) when `DEEPGEMM=1` is requested. Same "no manual checkout", zero lines added to the repo, and the pin keeps it reproducible.
2. **Or a submodule**, if you prefer it declarative — same effect, opt-in on clone.
3. **Or just documentation**: the two-line snippet in `docs/deepseek-v4.md` telling people exactly what to clone and where to point `DEEPGEMM_HOME`. Given a 2% payoff, honestly this might be the right size of solution.

None of this reflects on the tier itself — **#1055 is merged** and it's the substantial piece. This is only about what lives in the repository.

If you think I'm undervaluing DeepGEMM — for instance if the gap widens a lot at longer context, or on a different card, or with a bigger batch — show me that measurement and I'll reconsider on the spot. A 2% delta from one run on one machine is a thin basis for a 35k-line decision in either direction, including mine.

dcutugno:
on which card you tested this? i'm currently running some more tests with and without

dcutugno:
Fair call on the vendoring, but the 2 % is the wrong number — it comes from the *WSL2* rows of the Validation table, where both binaries were bound by the 9p file system streaming experts over `/mnt/c` (the caption says so); the GPU kernels barely ran there. The Windows number in the same doc (Performance table: 3324-token prefill 90 s DeepGEMM vs 225 s generic) is the real comparison.

Re-measured just now to be sure, same machine (RTX 5080 16 GB, 2× NVMe, current `dev` engine), DLL forced with `COLI_DSV4_DLL`, order alternated, 3.3k runs repeated:

| prompt | DeepGEMM DLL (sm_120a) | generic DLL | ratio |
|---|---|---|---|
| 3324 tokens, TTFT | **101.3 s / 102.4 s** | 231.5 s / 232.1 s | **2.28×** |
| 821 tokens, TTFT | **44.3 s** | 76.9 s | **1.74×** |
| 32 decode tokens after either | 22.2–22.7 s | 21.6–24.0 s | ≈ 1.0× |

Greedy text identical between the two DLLs (same md5), no refill fallbacks. So the gap is entirely in GPU prefill (block-scaled tensor-core GEMMs for the batched dense projections and the MoE expert bank replacing the fp32-from-fp8/fp4 generic kernels), it widens with prompt length (an 8k agent system prompt: ~9 min → ~4 min once), decode is disk-bound and equal, and it reaches Blackwell only — the loader picks it per device, everyone else is unaffected either way.

That said, I agree the ratio question is *lines-in-tree vs who benefits*, and I'm happy with any of your three options. My preference is (1): a `make deepgemm-fetch` target that clones the sm120 port at the pinned commit into a gitignored `c/third_party/` and applies `patches-deepgemm-sm120-msvc.patch`, wired as the default `DEEPGEMM_HOME` for `cuda-dsv4-dg-dll` / `DEEPGEMM=1` — zero repo lines, same one-command experience, reproducible. Tell me which one you want and I'll rework this PR into it (or close it in favour of the docs snippet, which is already in `docs/deepseek-v4.md`).

JustVugg:
You are right about the number, and I withdraw it: the 2% came from the WSL2 rows, where both binaries were 9p-bound and the GPU kernels barely ran — that was my misreading of the table, not a property of DeepGEMM. The Windows comparison (90 s vs 225 s on the 3324-token prefill, 2.5×) is the relevant one, and at 2.5× the cost-benefit of this PR looks completely different from what I argued.

The vendoring concern was always a separate axis from the speedup, so let me restate it at its actual size now that the gain is real: 35k vendored lines is a maintenance surface we would carry forever, but there is a middle path that keeps your build-with-no-external-checkout property without carrying the source in-tree — a pinned-commit fetch script (`tools/fetch_deepgemm.sh <pinned-sha>` + checksum) that `DEEPGEMM=1` invokes on first build, with the vendored tree in `.gitignore`. CI caches it; users get exactly your current UX; the repo stays 35k lines lighter. The license attribution you already did right stays as-is.

If the numbers from your with/without runs hold up (curious which card, too — you asked me the same and the honest answer is I have no sm120 hardware here, which is exactly why your measurements decide this, not mine), I'm in favour of taking this in the fetch-script form. If you'd rather keep it vendored, make the case for why the pinned-fetch fails your use case and we'll weigh it — the 2.5× has earned the discussion.

dcutugno:
@JustVugg so do you prefer the fetch script way?

JustVugg:
Yes — fetch script, confirmed. Concretely what we'd merge:

- `tools/fetch_deepgemm.sh <pinned-sha>` — clones/downloads at the pinned commit, verifies a checksum, drops the tree under `c/third_party/deepgemm/` (which goes in `.gitignore`)
- `DEEPGEMM=1` invokes it on first build if the tree is absent — your "no external checkout" UX stays exactly as it is today
- CI caches the fetched tree keyed on the pinned SHA
- keep the license attribution file **in-tree** (that part you already did right, and it should not depend on the fetch)
- the 90 s vs 225 s Windows numbers go in the PR description as the motivation

Bump the pin in one-line PRs whenever you need a newer DeepGEMM. If any of this fights your setup, say so — but if it works for you, rework it in that shape and we merge on green CI.

dcutugno:
Reworked in that shape and force-pushed (`467bec4`, one commit on current `dev`, 7 files, +208/−21, no third-party lines in the tree):

- `c/tools/fetch_deepgemm.sh [<pinned-sha>] [<dest>]` — shallow checkout of the sm120 port @ `39fb4447` + its CUTLASS submodule (`f3fde583`) into gitignored `c/third_party/deepgemm/`, verifies both commit ids (content-addressed = the checksum), applies the MSVC patch, marker for idempotency / refetch on a pin change; LF-forced so MSYS2 / Git for Windows / Linux behave the same.
- `DEEPGEMM=1` and `cuda-dsv4-dg-dll` invoke it on first build through a stamp-header prerequisite; `make deepgemm-fetch` alone; `DEEPGEMM_HOME=<dir>` still works; `DEEPGEMM_PIN` bumps the commit.
- CI: the Windows CUDA job runs the fetch (pin + patch + idempotency) and caches the tree keyed on pin + patch hash (it can't compile sm_120a on 12.6, so this is what CI can honestly verify).
- `THIRD_PARTY_NOTICES.md` stays in-tree; the 90 s vs 225 s numbers (and today's re-run: 101 s vs 232 s at 3.3k, 44 s vs 77 s at 821 tokens, decode equal) are in the description.

Verified from clean trees on Windows (`make cuda-dsv4-dg-dll`: fetch 37 s, DLL byte-identical in size to the vendored build, fetched+patched headers identical to the 78 files previously vendored modulo EOL) and Linux/WSL2 (`CUDA=1 DEEPGEMM=1`), second runs no-ops. Card, since you asked: RTX 5080 16 GB.

PR #1059, opened 2026-08-16
fix(windows): make coli stop find the serve, and stop orphaning engines (#1049)

### LINKED ISSUE #1049: [Bug]: `coli stop` crashes on Windows (FileNotFoundError: /proc) and leaves the server running
﻿
### Summary

`coli stop` is unusable on Windows. It walks `/proc` to find the running engine; that path
does not exist, so it throws â€” and the server it was asked to stop keeps running and holding
its port. The failure mode that matters is the silent one: the operator reasonably believes
the server stopped when it did not.

### Environment

colibri v1.6.0 (`72cd9f7`) Â· Windows 11 Enterprise 26200 (25H2), native (no WSL) Â·
MSYS2 20260611 Â· Python 3.12.10 Â· OLMoE engine (`olmoe.exe`), built `ARCH=native` with
mingw-w64 gcc 16.2.0

### Repro

```powershell
cd colibri\c
$env:SNAP="D:/models/olmoe_merged"
$env:COLI_MODEL="D:/models/olmoe_merged"
$env:COLI_API_KEY="local-secret"
python coli serve --model D:/models/olmoe_merged --host 127.0.0.1 --port 8000
# then, from another shell:
python coli stop
```

### Actual

```
    for pd in os.listdir("/proc"):
              ^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [WinError 3] The system cannot find the path specified: '/proc'
```

and the listener survives:

```powershell
PS> Get-NetTCPConnection -LocalPort 8000 -State Listen | Select LocalAddress,OwningProcess
LocalAddress OwningProcess
127.0.0.1            49976
```

### Expected

Either shut the server down, or fail with an explicit "not supported on this platform"
message rather than a traceback.

### Workaround

```powershell
$pid8000 = (Get-NetTCPConnection -LocalPort 8000 -State Listen).OwningProcess
Stop-Process -Id $pid8000 -Force
Get-Process -Name colibri,olmoe -ErrorAction SilentlyContinue | Stop-Process -Force
```
Verify with `Get-NetTCPConnection -LocalPort 8000 -State Listen` returning nothing.

### Suggested direction

`compat.h` already establishes "every platform difference lives here". Process discovery could
follow the same rule â€” on Windows via `CreateToolhelp32Snapshot`, or `psutil` where available.

A simpler option that removes the platform question entirely: have `coli serve` write the
engine PID (and port) to a small state file next to the model or in the run directory, and
have `coli stop` read it back. That would also be more robust than scanning `/proc` on Linux,
and it makes `coli stop` work when several models are served from one checkout.

Happy to send a PR once you've indicated which shape you'd prefer.

JustVugg:
Fixes the three defects @sami7969 measured in #1049. Credit where it's due: this PR is his diagnosis — I only wrote the code.

## 1. `os.kill(pid, 0)` is not a liveness probe on Windows

The one neither of us saw coming. `os.kill()` has no signal semantics on win32, and for a pid the calling process did not spawn it raises `OSError [WinError 87] "The parameter is incorrect"` rather than reporting liveness. `cmd_stop` caught that as "not running":

```
$ python coli stop --port 8000
$ # exit 0, port 8000 still LISTENING, ~5.2 GB of engines still resident
```

New `_pid_alive()` (`c/coli`) asks the OS directly on win32 via `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` + `GetExitCodeProcess`, and keeps the POSIX idiom everywhere else. Documented caveat: a process that exited with code 259 (`STILL_ACTIVE`) reads as alive — the resulting failure mode is a redundant `SIGTERM` on a dead pid, never a missed target.

This also corrects a comment of mine: `# native Windows has no /proc; the pidfile still works` was wrong. It didn't.

## 2. `signal.SIGKILL` escaping the handler

`AttributeError` is not `OSError`, so on Windows this aborted the command *after* the `SIGTERM`s had already gone out. Now uses `getattr(signal, "SIGKILL", signal.SIGTERM)` — the form this same file already uses at `:752`. On Windows `os.kill(pid, SIGTERM)` is `TerminateProcess` anyway.

## 3. The orphaned engine — 2,617 MB per serve/stop cycle

The measurement in #1049:

| process | pid | parent | RSS |
|---|---|---|---|
| `python.exe` — `coli serve`, the pid in the pidfile | 692712 | shell | 30 MB |
| `olmoe.exe` — direct child | 432160 | 692712 | 1,854 MB |
| `olmoe.exe` — **OMP re-exec** | 689052 | 740652 — *already exited* | **2,602 MB** |

The engine re-execs itself for OMP tuning and the re-exec's parent exits immediately, so the survivor is **orphaned at birth**. It is not a descendant of anything `cmd_stop` can walk to, and it is not the pid the server recorded — the pidfile is written in `cmd_serve` before `openai_server` spawns the engine at all. (This is the same failure that OOM'd a box on 2026-07-16 with two 17+5 GB ghosts, where `pkill -x glm` matched nothing because the re-exec renames itself.)

Rather than chase pids, this ties the engine to the server's lifetime at the OS level: a **Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`**. Job membership is **inherited by child processes**, so the OMP re-exec stays inside the job despite its parent being gone, and when the last handle closes — normal exit, `TerminateProcess`, or the server crashing — Windows terminates everything in it.

Chosen over the three options offered in the issue because:

- *appending the engine pid to the pidfile* doesn't reach it — the re-exec has a **different pid** than the process that was spawned;
- *tagging processes* requires reading another process's environment block (`NtQueryInformationProcess` + `ReadProcessMemory`), which @sami7969 rightly called out as too large for this;
- *documenting the caveat* leaves the RAM leaked.

No pidfile contract change, no foreign memory reads, stdlib `ctypes` only, and it fails closed: any error returns `None` and restores today's behaviour.

## Verification

- **Python suites green: 210 passed** (`test_openai_server`, `test_launcher_dispatch`, `test_cli_output`, `test_doctor`, plus the new file).
- **C suite green**, exit 0.
- New `c/tests/test_stop_liveness.py` — 6 cases covering the probe (self is alive, a reaped child is not), that `cmd_stop` no longer uses the raw POSIX idiom, the `SIGKILL` guard, and the job wiring.

**One honest caveat: the Windows behaviour itself is unverified.** This was developed on Linux, where the `ctypes` paths compile but never execute. @sami7969 — you have the machine and the reproduction; a run of your exact serve/stop cycle would confirm both that `coli stop` now finds its targets and that no `olmoe.exe` survives it. If the job assignment misbehaves under MSYS2 vs native Python, that's exactly the sort of thing only your setup will show.

Closes #1049

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -1370,6 +1370,36 @@ def cmd_serve(a):
         try: os.unlink(serve_pidfile(a.port))
         except OSError: pass
 
+def _pid_alive(pid):
+    """Is `pid` a live process?
+
+    `os.kill(pid, 0)` is the POSIX idiom, but on Windows os.kill() has no signal
+    semantics: for a pid this process did not spawn it raises
+    OSError [WinError 87] "The parameter is incorrect" instead of reporting
+    liveness. cmd_stop caught that as "not running", so `coli stop` printed
+    "nothing running" while the port was still LISTENING and ~5 GB of engines
+    were resident (#1049, reported with a reproduction on Windows 11 native).
+    On win32 ask the OS directly instead."""
+    if sys.platform != "win32":
+        try: os.kill(pid, 0); return True
+        except OSError: return False
+    import ctypes
+    from ctypes import wintypes
+    PROCESS_QUERY_LIMITED_INFORMATION, STILL_ACTIVE = 0x1000, 259
+    k32 = ctypes.WinDLL("kernel32", use_last_error=True)
+    k32.OpenProcess.restype = wintypes.HANDLE
+    handle = k32.OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, False, pid)
+    if not handle: return False
+    try:
+        code = wintypes.DWORD()
+        if not k32.GetExitCodeProcess(handle, ctypes.byref(code)): return False
+        # STILL_ACTIVE is also a legal exit code; a process that exited with 259
+        # reads as alive here. SIGTERM on a dead pid is harmless, so the failure
+        # mode is a redundant kill, never a missed one.
+        return code.value == STILL_ACTIVE
+    finally:
+        k32.CloseHandle(handle)
+
 def cmd_stop(a):
     """Shut down a running `coli serve` AND its engine — one command, no pkill.
     The engine re-execs itself for OMP tuning, so its process is named `exe`,
@@ -1382,7 +1412,7 @@ def cmd_stop(a):
     pf=serve_pidfile(a.port)
     try:
         with open(pf) as f: pid=int(f.read().split()[0])
-        os.kill(pid,0); targets[pid]=f"coli serve (pidfile, port {a.port})"
+        if _pid_alive(pid): targets[pid]=f"coli serve (pidfile, port {a.port})"
     except (OSError,ValueError,IndexError): pass
     try: proc_entries=os.listdir("/proc")
     except OSError: proc_entries=[]       # native Windows has no /proc; the pidfile still works
@@ -1408,7 +1438,11 @@ def cmd_stop(a):
         except OSError: pass
     time.sleep(2.0)
     for pid in targets:
-        try: os.kill(pid, signal.SIGKILL); print(f"  {pid}: forced (SIGKILL)")
+        # signal.SIGKILL does not exist on win32 and AttributeError is not OSError,
+        # so the bare `except OSError` let it escape and abort the command AFTER
+        # the SIGTERMs above (#1049). Line 752 in this file already uses the
+        # getattr form; os.kill(SIGTERM) is TerminateProcess on Windows anyway.
+        try: os.kill(pid, getattr(signal, "SIGKILL", signal.SIGTERM)); print(f"  {pid}: forced")
         except OSError: pass       # gia' morto: bene
     try: os.unlink(pf)
     except OSError: pass
--- a/c/openai_server.py
+++ b/c/openai_server.py
@@ -1685,6 +1685,87 @@ def tune_child_env(env, arch):
     return env
 
 
+def _win_kill_on_close_job(pid):
+    """Tie an engine process to this server's lifetime, on Windows.
+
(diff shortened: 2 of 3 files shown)


JustVugg:
The Windows CI failure was **the code working and the test being wrong** — worth writing down, because the failure is better news than a pass would have been.

```
FAIL: test_helper_is_a_noop_off_windows
  self.assertIsNone(openai_server._win_kill_on_close_job(os.getpid()))
AssertionError: 1392 is not None
```

On Windows UCRT64 the helper **created a job object and returned its handle (1392)**. That is the one thing I said in the PR description I could not verify from Linux, and CI just verified it: `CreateJobObject` → `SetInformationJobObject(KILL_ON_JOB_CLOSE)` → `AssignProcessToJobObject` all succeeded against a live process.

**And the test was worse than wrong.** It passed `os.getpid()`, which assigned **the test runner itself** to a kill-on-close job. Nothing died only because a raw `HANDLE` is a plain int that Python never closes for you — one `CloseHandle` (ours, or a GC'd stray reference) away from terminating the CI runner. That was my bug and it deserved to be caught here rather than on someone's machine.

Fixed in `ca657ce`:

- the no-op assertion is **skipped on win32**, where returning a handle is the correct behaviour;
- win32 gets the test that actually matters instead: **spawn a throwaway child, assign it to the job, close the handle, assert the child dies**. That exercises the mechanism the whole fix depends on, rather than asserting a platform detail;
- an explicit comment about never passing `os.getpid()`, so nobody reintroduces it;
- the files the suite opens are closed (CI flagged `ResourceWarning`s).

Everything else in the run was green — 582 tests, this was the single failure.

@sami7969 this strengthens the ask rather than weakening it: CI now proves the job gets *created* on Windows and that closing it kills a child. What it still cannot show is the real scenario — that `coli stop` finds its targets on your box, and that **no `olmoe.exe` survives** a serve/stop cycle with the OMP re-exec in play. That part is still yours.

Issue #1060, opened 2026-08-16
Nix flake packages only the GLM engine — coli can't dispatch to OLMoE / DeepSeek V4 / Inkling / Kimi K3


jerome-benoit:
## Problem

The flake's `packages.colibri` (`flake.nix`, `main`@4ef9a99 / `dev`@03e8677) packages **only the GLM engine**, so `coli` installed via the flake can run just GLM-5.2 — although the README says "Five families run today" and `coli` already dispatches all five.

- **buildPhase**: `make -C c colibri ARCH="$ARCH"` — only `c/colibri` is built; `olmoe`, `deepseek_v4`, `inkling`, `kimi_k3` are not.
- **installPhase**: only `cp c/colibri $out/lib/colibri/colibri` — the other engine binaries never reach `$out`.
- **wrapper**: `--set-default COLI_ENGINE "$out/lib/colibri/colibri"`. In `c/coli`, `if os.environ.get("COLI_ENGINE"): GLM = os.environ["COLI_ENGINE"]`, and `engine_for()` does `if arch == "glm" or os.environ.get("COLI_ENGINE"): return GLM` — so setting `COLI_ENGINE` forces **every** model to the GLM engine regardless of `model_arch`.

Net effect: pointing flake-installed `coli` at an OLMoE / DeepSeek V4 / Inkling / Kimi K3 model can't work — the engine binary is absent from `$out`, and the pin would force GLM anyway.

(Distinct from the already-fixed #595, which was about the GLM build / `apps.glm` path itself, not multi-engine dispatch.)

## Proposed solution

Build + stage every engine the platform supports, and stop pinning `COLI_ENGINE` so `coli` dispatches per model:

1. **buildPhase** — build the engines, e.g. `make -C c colibri olmoe deepseek-v4`. Note: `make -C c install` already covers `colibri` + `olmoe` + `deepseek-v4` (the latter when `COLI_V4_SUPPORTED`), but **not** `inkling` / `kimi_k3` — those need their targets added explicitly where their build deps are available.
2. **installPhase** — stage each built engine beside `coli` in `$out/lib/colibri/`; `coli` resolves engines relative to its own directory (`HERE`), so co-location is enough.
3. **wrapper** — **remove** `--set-default COLI_ENGINE` entirely (not just repoint it — `engine_for` short-circuits on `or os.environ.get("COLI_ENGINE")`). Unset, it dispatches to `$HERE/<engine>` per the model's arch.

## Alternatives considered

[...]

PR #1063, opened 2026-08-16
refactor(core): make model families registry-owned

terrizoaguimor:
## Summary

- add one mandatory, immutable family registry for GLM, Inkling, Kimi K3, OLMoE, and DeepSeek V4
- make `coli`, `openai_server.py`, `doctor.py`, and `resource_plan.py` consume the same exact `model_type`, engine, build, API, capability, context, and planner contract
- reject unknown families instead of silently treating them as GLM
- model family identity separately from the engine artifact, so a future family such as MiniMax-M3 may deliberately share `colibri` without becoming a fallback
- add test-local Qwen3.6 and MiniMax-M3 descriptors reproducing the two same-day integration failures described in Discussion #1057
- make unproven family planners refuse explicitly instead of returning plausible zero-byte KV/expert figures
- extend GLM planning with DSA inventory detection and KV-slot-aware state accounting
- enforce build/install/CI/release coverage and package the registry in source installs, release archives, and Docker

## Why

The current model-family contract is distributed across launcher branches, gateway branches, doctor/planner assumptions, Make targets, install rules, CI, and release packaging. That made a complete integration audit a manual requirement for every family, and the same planner bug reached both Qwen3.6 and MiniMax-M3 independently.

This is the first implementation step approved in [Discussion #1057](https://github.com/JustVugg/colibri/discussions/1057#discussioncomment-18041674): registry first, no forward/kernel changes, and no optional bypass around the registry.

## Behavior and safety

- No engine `.c` or hot path changes.
- Existing GLM planning remains available, now charging DSA state from the actual indexer tensor inventory and multiplying state by KV slots.
- Inkling, Kimi K3, OLMoE, and V4 planning now reports an explicit unsupported reason until measured family adapters exist; it no longer applies an MLA-shaped formula silently.
- The slim Docker image remains GLM-only and rejects sibling families explicitly.
- `response_format` is rejected before writing a grammar payload to engines that cannot consume it.
- Context flags use each family's registered environment channel and maximum.

## Validation

- `make check`
- 518 Python tests passed, 59 skipped
- targeted registry/launcher/gateway/doctor/planner/E2E suite: 307 passed
- `python3 -m py_compile family_registry.py resource_plan.py doctor.py openai_server.py tests/test_family_registry.py`
- `git diff --check`
- multiple independent review passes; final result: no blocking findings

Refs #1057

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -135,6 +135,7 @@ jobs:
           cp c/version.py dist/
           cp c/openai_server.py dist/
           cp c/v4_dsml.py dist/        # openai_server imports v4_dsml (vendored DeepSeek V4 DSML primitives); without it the packaged server won't import
+          cp c/family_registry.py dist/ # authoritative model-family dispatch/capability registry
           cp c/resource_plan.py dist/
           cp c/doctor.py dist/
           cp c/autotune.py dist/
@@ -184,11 +185,9 @@ jobs:
           # fine. The check was wrong, not the artifact.
           python3 -c "import ast; ast.parse(open('tools/k3_tokenizer.py', encoding='utf-8').read())" \
             || { echo "FAIL: packaged k3_tokenizer.py does not parse"; exit 1; }
-          # COLI_EXPECT rather than a literal tuple: the list is platform-dependent
-          # now, and one hardcoded list that silently omitted an engine is what #858
-          # was. Keep the expectation in a single place, above.
           COLI_EXPECT="$SIBLINGS" python3 - <<'PYCHK'
           import importlib.machinery, importlib.util, json, os, sys, tempfile
+          from family_registry import all_families
           exe = ".exe" if os.name == "nt" else ""
           expect = os.environ["COLI_EXPECT"].split()
 
@@ -213,16 +212,25 @@ jobs:
           loader.exec_module(cli)
 
           bad = []
-          for model_type, label, _ in cli._BANNER_MODELS:
-              with tempfile.TemporaryDirectory() as d:
+          for family in all_families():
+            if family.id == "deepseek_v4" and "deepseek_v4" not in expect:
+              continue
+            model_type = family.model_types[0]
+            label = family.display_name
+            with tempfile.TemporaryDirectory() as d:
                   with open(os.path.join(d, "config.json"), "w", encoding="utf-8") as f:
                       json.dump({"model_type": model_type}, f)
                   open(os.path.join(d, "tokenizer.json"), "w").close()
                   engine = os.path.basename(cli.engine_for(d))
                   arch = cli.model_arch(d)
-              if model_type != "glm" and arch == "glm":
-                  bad.append("%s (%s) -> the GLM engine; the banner names a family "
-                             "the launcher cannot dispatch" % (label, model_type))
+            expected_file = family.engine_artifact + exe
+            if not os.path.exists(expected_file):
+              bad.append("%s: expected artifact %s is absent" %
+                         (label, expected_file))
+            elif arch != family.id or engine != expected_file:
+              bad.append("%s (%s) -> %s/%s, expected %s/%s" %
+                         (label, model_type, arch, engine,
+                          family.id, expected_file))
           if bad:
               sys.exit("FAIL: packaged launcher misroutes:\n  " + "\n  ".join(bad))
           print("OK: every engine is packaged AND coli dispatches to it:", " ".join(expect))
--- a/c/Makefile
+++ b/c/Makefile
@@ -1170,17 +1170,21 @@ check:
 	$(MAKE) portable
 	$(MAKE) test
 
-install: colibri$(EXE) olmoe$(EXE) $(if $(filter 1,$(COLI_V4_SUPPORTED)),deepseek-v4,)
+install: colibri$(EXE) inkling$(EXE) kimi_k3$(EXE) olmoe$(EXE) \
+	$(if $(filter 1,$(COLI_V4_SUPPORTED)),deepseek-v4,)
 	$(INSTALL) -d $(DESTDIR)$(BINDIR)
 	$(INSTALL) -d $(DESTDIR)$(LIBEXECDIR)
 	$(INSTALL) -d $(DESTDIR)$(LIBEXECDIR)/tools
 	$(INSTALL) -m 755 coli $(DESTDIR)$(BINDIR)/coli
 	$(INSTALL) -m 755 colibri$(EXE) $(DESTDIR)$(LIBEXECDIR)/colibri$(EXE)
+	$(INSTALL) -m 755 inkling$(EXE) $(DESTDIR)$(LIBEXECDIR)/inkling$(EXE)
+	$(INSTALL) -m 755 kimi_k3$(EXE) $(DESTDIR)$(LIBEXECDIR)/kimi_k3$(EXE)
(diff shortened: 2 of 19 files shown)


terrizoaguimor:
@JustVugg At this rate, you may need to turn Colibrì into a company and put me on the payroll before the next phase lands. 😄

Jokes aside, thank you for the clear scope and the priority review offer — it made this a much sharper PR.

JustVugg:
Merged, and it's already earning its keep — but I want to hand you a measurement rather than a compliment, since the RFC said this was step one and the useful thing now is knowing exactly what step two is.

I counted what still carries family knowledge outside the registry on `dev`:

```
c/coli               25 literal family strings
c/openai_server.py   20
c/doctor.py           0   ← done
c/resource_plan.py    1
```

Then I read all 45 and classified them, because "45 literals remain" would be a misleading headline. Two of the groups are genuine drift; the rest are not, and I'd argue against touching them.

## Still drift — worth moving into the registry

**1. The adapter allow-lists, repeated four times in `c/coli`:**

```python
:976   if family.gateway_adapter not in ("glm","inkling","kimi","olmoe","deepseek_v4"):
:1021  if family.cli_adapter     not in ("glm","deepseek_v4","olmoe"):
:1231  if family.gateway_adapter not in ("glm","inkling","kimi","olmoe","deepseek_v4"):
:1424  if family.gateway_adapter not in ("glm","inkling","kimi","olmoe","deepseek_v4"):
```

This is a **second source of truth about which families exist**, duplicated four times. Add a sixth family to the registry and forget these, and it breaks — which is precisely the failure mode the registry exists to make impossible. A declared capability on the descriptor (`has_gateway_adapter` / `has_cli_adapter`, or just letting the adapter field be optional) removes the list entirely.

**2. The prompt templates at `c/coli:984-987`** — four literal strings in a dict inside `cmd_run`:

```python
{"glm": "[gMASK]<sop><|user|>…", "inkling": "<|user|>…",
 "kimi": "K3CHAT1\nM user…",     "olmoe": "<|user|>\n…"}
```

That is data wearing a code costume, and it's family-identity data at that. `cmd_chat` already goes through the gateway renderers; `cmd_run` is the odd one out.

## Not drift — I'd leave these alone

The ~20 in `openai_server.py` are mostly real behaviour, not configuration:

```python
if ARCH == "inkling":  audio_clips = []              # only that family has an audio tower
if ARCH == "olmoe" and enable_thinking: ...          # thinking semantics differ
renderer = render_chat_kimi if ARCH == "kimi" else render_chat_v4 if ...
```

Pushing these into descriptors would make the code *worse*: you'd end up with a table of flags that really mean "call this function", which is harder to follow than the branch it replaced. Your own framing in #1057 is the right test — **specialized below, shared above**. These are the specialized part, and they're where they belong.

The renderer dispatch is the one genuine borderline case: it's a function pointer keyed by family, which is arguably a descriptor field. I'd only move it if it falls out naturally, not as a goal.

## What already paid off

For the record, three things this fixed that I'd found by hand the day before and had no structural answer for:

- the planner returning **`kv_bytes = 0`** for Qwen3.6 and MiniMax-M3 independently — now an explicit refusal, and both are test descriptors in your PR;
- `response_format` reaching an engine that can't parse it and **permanently desyncing its stdin** — now rejected before the payload is written;
- unknown `model_type` silently running the GLM engine — the exact shape of #1060, where a packager knew about one family out of five.

GLM planning also got more accurate rather than merely reorganised: DSA state from the real indexer inventory instead of an assumption, and state multiplied by KV slots.

I've opened **#1066** to track the four per-family planner adapters, with `_glm_geometry` named as the worked example, so the refusals have somewhere to converge.

terrizoaguimor:
@JustVugg Thanks for the classification — the count is more useful than a review, and I agree with the split. The two adapter allow-lists and the `cmd_run` template dict are clearly a second source of truth; the `openai_server.py` branches are behaviour and I won't touch them. I'll leave the renderer dispatch alone unless it falls out on its own.

One thing before I start, since it's the part that's expensive to change afterwards: what's the target field set for the family descriptor — including fields we won't consume yet?

Adding `has_gateway_adapter` / `has_cli_adapter` now means migrating all five families plus the test descriptors. If there are three or four more capability flags you already know you'll want (planner adapter, renderer, audio tower, thinking semantics), I'd rather land them in one schema change than in four. If the shape isn't settled yet, say so and I'll design for extension instead of guessing at it.

Also: do you want the allow-lists + templates first, or the per-family planner adapters in #1066? I can take either, but not both in one PR.

JustVugg:
Both good questions. The second is easy; the first has a better answer than a field list.

## On the field set — I think you're solving the wrong half

I looked at `FamilyDescriptor` before answering, and the expensive part isn't *which* fields exist. It's that all twenty are **positional with no defaults**. That's what makes "one schema change instead of four" feel necessary: today, adding a field means touching five descriptors plus the test ones, and the call sites are already at the edge of readable —

```python
"colibri", ("glm",), "colibri-core", "glm", "colibri",
("colibri", "glm"), "glm-5.2-colibri", "glm", "glm", "glm_mla",
```

Nobody can tell what the twelfth positional argument means without counting.

So the change I'd make first is not a bigger schema, it's a **cheaper one to extend**: give new fields defaults and convert the five construction sites to keyword arguments. Then adding `has_cli_adapter` later costs one line in one descriptor instead of a migration, and the question you're asking stops being expensive to get wrong. It also makes the constructor self-documenting, which matters more as the sixth family arrives.

**On the specific fields:** you have already covered every one that a real bug demanded. I went back through the integration failures from this week to check rather than guess:

| failure | field |
|---|---|
| Qwen absent from `cmd_build` | `build_target` ✓ |
| Qwen absent from `_BANNER_MODELS` | `display_name` / `display_scale` ✓ |
| `coli stop` couldn't kill the engine (the OMP re-exec renames itself) | **`process_names` ✓** — you got there before I did |
| context 8192 vs launcher's 16384 | `limits` + env channel ✓ |
| grammar reaching an engine that desyncs on it | `capabilities` ✓ |
| planner inventing zero-byte KV | `planner_geometry` / `planner_unsupported_reason` ✓ |

That leaves exactly the two you named — `has_gateway_adapter` / `has_cli_adapter` and the `cmd_run` template. I would **not** add renderer, audio tower or thinking-semantics flags speculatively: those are the behaviour half, and a descriptor full of flags that really mean "call this function" is harder to read than the branch it replaced. If one of them later turns out to be data, it'll be obvious, and with defaults in place it'll be cheap.

## On priority — allow-lists and templates first

Take the drift, not the planner. Three reasons:

1. **The allow-lists are the live hazard.** Four copies of `("glm","inkling","kimi","olmoe","deepseek_v4")` are a second source of truth about which families exist. Until they're gone, family six can be added to the registry and still break — which is the exact failure the registry was built to make impossible. Every day they survive is a day the guarantee is only partial.
2. **It has a finish line.** It's a pure refactor with a testable end state, and it's the natural completion of #1063 rather than a new project.
3. **#1066 needs measurements, not a refactor.** Kimi's KDA recurrent state and MXFP4 reserves, Inkling's hybrid GQA/sconv — those need someone with the model and the hardware to measure. That work parallelises across the community; the drift cleanup doesn't. Doing it in that order means when a planner adapter does land, it lands into a registry that is already the single source of truth.

So: keyword-args + defaults, then the two allow-lists and the template dict. I'll review it as a priority, and I'd rather see it as one PR than three.

PR #1076, opened 2026-08-18
perf(deepseek_v4): thread-local scratch for qdq activations (bit-identical)

JustVugg:
## What

Engine-by-engine continuation of #1071 (GLM) and #1075 (kimi_k3), V4 edition.

Every `*_ref` matvec/matmul entry in the `NATIVE_QUANT` unit — `coli_fp4_matvec_ref`, `coli_fp8_matvec_ref`, both `dual_matvec` variants, both `matmul_batch` variants — paid a **malloc/free pair per call** for its qdq activation buffers. On the decode hot path that is ~5 fp8 matvecs per layer per token in the attention chain alone, × 43 layers on V4-Flash: hundreds of heap round-trips per token.

The buffers now come from a growable `_Thread_local` scratch (`v4_qdq_scratch`), never freed per call. Peak footprint = the widest `batch*columns`, a few MB per thread. **Computed values are bit-for-bit unchanged** — only where the buffer comes from changes.

The transformation was script-applied within the unit bounds and verified to leave **zero residue** of `malloc`/`free(activation…)` — a leftover `free` on the scratch would have been a use-after-free, so this was checked mechanically, not by eye.

## Deliberately not in this PR

Deduplicating the `wq_a`/`wkv` qdq of the same attention input (it is quantized twice per token today). `coli_fp8_dual_matvec_ref` requires `dual_same_shape` and those weights have different row counts, so the dedup needs a new `_pre` entry across three attention units — follow-up, kept out so this PR stays mechanical.

## Verification

1. **`make deepseek-v4-tiny-check`** on the patched engine: full PASS — token-exact oracle, serve protocol token-exact, KV prefix reuse (cold/repeat/reset). The oracle is an absolute reference, not a relative A/B.
2. **Real-model A/B** on DeepSeek-V4-Flash (156 GB fp4, streamed from disk on a 25 GB box): dev-base binary vs patched binary, same prompt —
   - generated text **byte-identical**
   - `expert_requests` identical (9,842)
   - timing deltas between the two runs are disk noise (QD3 floor, `.coli_usage` heat evolving between runs) and carry no signal either way; the benefit accrues to RAM-resident deployments where the allocator contends with the loader threads.

Series status: #1071 GLM ✅ merged · #1075 kimi_k3 ✅ merged · this PR = deepseek_v4 · qwen36 flagged on #712 (comment) for the author.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -14628,6 +14628,39 @@ int coli_hadamard_bf16_ref(float *values, size_t length) {
     return 0;
 }
 
+/* Scratch _Thread_local riusabile per le attivazioni qdq (famiglia #1071/#1075):
+ * ogni entry *_ref qui sotto pagava un paio di malloc/free PER CHIAMATA sul
+ * path caldo (attenzione: ~5 matvec fp8 per layer per token; expert: una per
+ * coppia). I valori calcolati non cambiano di un bit — cambia solo da dove
+ * arriva il buffer. I buffer crescono e non si restituiscono mai: il picco è
+ * batch*columns del matmul più largo, pochi MB per thread.
+ * EN: growable thread-local scratch for the qdq activations. Every *_ref entry
+ * below paid a malloc/free pair per call on the hot path; values are bit-for-
+ * bit unchanged — only the buffer's provenance changes. */
+typedef struct {
+    float *activation; size_t activation_capacity;
+    uint8_t *scales; size_t scales_capacity;
+} V4QdqScratch;
+static _Thread_local V4QdqScratch v4_qdq_scratch_state;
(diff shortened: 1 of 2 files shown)


JustVugg:
Correction to the PR description: the first commit's scripted transform was bounded to the `NATIVE_QUANT` unit and only converted the two plain matvecs — the dual and batch entries live in separate translation units (`NATIVE_QUANT_DUAL` / `NATIVE_QUANT_BATCH`) and had been left untouched, while the description claimed all six. The second commit completes the transform (the fp4 batch variant additionally needed a hand edit: its scales size is `batch*columns/128`, which the scripted pattern had missed and left as a leak). Residue re-verified at zero; tiny oracle re-run on the completed transform: full PASS.

Issue #1050, opened 2026-08-16
[perf] olmoe: O(cap) eviction scan under g_pilot_mx makes tok/s non-monotonic in cache size


sami7969:
﻿
### Summary

`olmoe.c`'s expert-cache eviction picks a victim by scanning every slot linearly, on every
miss, while holding the global `g_pilot_mx`. Because the scan cost is O(cap) and the miss rate
saturates near 100% at small caps, total bookkeeping cost peaks in the middle of the range â€”
which makes measured throughput **non-monotonic** in cache size. Shrinking the cache from 4 to
1 slots per layer makes the engine *faster*, which is not what a cache is supposed to do.

### The code

`c/olmoe.c:457-462`:

```c
/* LRU eviction â€” skip pinned and in-flight (eid==-1) slots */
int lru = -1;
for (int i = 0; i < lc->n; i++) {
    if (lc->slots[i].pinned || lc->slots[i].eid < 0) continue;
    if (lru < 0 || lc->slots[i].used < lc->slots[lru].used) lru = i;
}
```

with up to two more full scans in the fallback paths below it, all under `g_pilot_mx`.

### Measurement

`PPL=1` teacher-forced mode, 89 scored tokens, identical `ref.json` for every rung,
`OMP_NUM_THREADS=24`, Windows 11 / MSYS2, mingw-w64 gcc 16.2.0, Core Ultra 9 275HX (24 cores),
OLMoE int8 container (64 experts/layer):

```
SNAP=<model> OMP_NUM_THREADS=24 PPL=1 ./olmoe.exe <cap> 8 ref.json
```

| cap | hit rate | tok/s | miss rate Ã— cap |
|---|---|---|---|
| 64 | 92.7% | **7.36** | 4.7 |
| 32 | 79.5% | 6.33 | 6.6 |
| 16 *(default)* | 53.6% | 5.14 | **7.4** |
| 8 | 27.0% | 4.28 | 5.8 |
| 4 | 3.7% | **3.63** | 3.9 |
| 2 | 0.8% | 3.74 | 2.0 |
| 1 | 0.1% | **3.98** | 1.0 |

Throughput bottoms out at cap=4 and then **rises again** at cap=2 and cap=1, despite the hit
rate falling to 0.1%. The last column is the predicted per-access bookkeeping cost
(miss_rate Ã— scan length); it tracks the shape.

The mechanism: miss rate saturates near 100% by cap=4, so below that the only term still
[...]

Issue #1034, opened 2026-08-15
[Bug]: ram-guard cap size reduction does not work


tonnthuir:
### Commit or version

6546cdde7296f28771e2ba1a1d7c1d4b0cb550aa

### Environment

Debian Linux, AMD Ryzen 5 9600X, 64GB DDR5 5200 RAM, No GPU, 4x 2TB SSD Gen 4 drives

### Reproduction steps

Description:

The rss_guard code decreases the cap to limit memory usage. However, the lowered cap is ignored and memory creeps back up to where it had been. If the cap is too high, then rss_guard will keep finding the memory is too high and will evict experts from the cache. This is sub-optimal and could lead to problems with memory usage.

initial condition:
ecap=N, ecn[layer]=N

ram-gard runs:
ecap=N-1, frees many ESlots, ecn[layer]=N

next miss:
freed ESlot is selected as LRU victim, slot is re-used by the cache even though it brings the number of active slots in the layer above ecap

result:
cache memory returns to N slots and memory usage was not effectively lowered by ram-guard

Severity:
This will not cause any calculation errors. It means that rss_guard does not prevent OOM or tight memory problems as well as it should. Once the cache grows large enough, rss_guard will remove experts from the cache one until the memory limit is met. But the cache then refills back to the same size as it had been.

This is complicated to fix, because many parts of Colibri interact with the LRU cache. The evicted expert could be anywhere in the layer's row, so simply limiting the number of experts to inspect will not work. Re-ordering the EStore's in the row so that the array can be shrunk could cause problems for other parts of the code.

### Expected behavior

When rss_guard is called and detects that memory usage is too high, it evicts some of the EStore entries from the cache and reduces the cap. The new, lower cap should be respected to prevent the cache from growing too large again.

### Actual behavior and logs

```shell
The cap is ignored and the cache fills up again. This leads to memory exceeding the RAM limit again, so rss_guard will evict experts again and keep reducing the cap until it reaches 2. Since the cap is ignored, it makes no difference. But having rss_guard evict experts from the cache every time it runs is probably inefficient.

I wrote some debugging code that measured the size of the LRU cache and the WS cache (based on adding slab_cap+4*fslab_cap for each entry). It normally takes a while for the cache to grow large enough to trigger the rss_guard, so I also increased the cap calculation to make it trigger immediately while diagnosing this problem, but the behavior is the same regardless of how the cap is determined. Here are some of the logs:

[...]

PR #1092, opened 2026-08-18
perf(cache): use second-chance expert eviction

pacocartones:
## Summary

Replace full oldest-timestamp scans in the OLMoE and GLM expert caches with a per-layer clock/second-chance hand.

The change keeps the existing safety rules:

- OLMoE prefers unpinned slots and only considers pinned slots as the existing last-resort fallback.
- In-flight loads are never selected.
- GLM skips async GPU borrows and pilot reservations.
- Failed or RSS-guard-cleared slots become immediate reuse candidates.

Hits and successful publications set a one-bit recent marker. Victim selection clears recent markers while advancing the hand, avoiding a full minimum-timestamp comparison on every full-cache miss.

This does not claim that eviction scanning caused the throughput curve reported in the issue. That causal explanation was withdrawn in the issue thread. The change is scoped to the independently valid hot-path scalability and critical-section cleanup.

## Validation

- [ ] `make -C c check`
- [x] Focused GLM clock test: `make -B tests/test_eslot_inflight && ./tests/test_eslot_inflight`
- [x] Focused OLMoE clock and framing test: `make -B tests/test_olmoe_serve_framing && ./tests/test_olmoe_serve_framing`
- [x] OLMoE math regression: `./tests/test_olmoe_matmul_q`
- [x] Default CPU builds: `make colibri olmoe`
- [x] `git diff --check`

A model-free microbenchmark calling the production GLM selector showed the old minimum scan growing from 2.87 ns/call at one slot to 873.85 ns/call at 256 slots. With second chance and the real replacement pattern, the same sizes measured 14.06, 21.68, 17.88, and 17.24 ns/call respectively. These numbers only validate selector scaling, not end-to-end token throughput.

## Compatibility

- [x] No CUDA behavior is added or required

Refs #1050

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -378,6 +378,7 @@ typedef struct {
  * expert non hanno tutti la stessa taglia (layer MTP int8 = 2x i layer int4). */
 typedef struct { int eid; QT g,u,d; uint8_t *slab; float *fslab;
                  int64_t slab_cap, fslab_cap; uint64_t used;
+                 uint8_t recent; /* second-chance reference bit */
                  unsigned in_flight; /* async GPU readers borrowing this slot */
                  /* pin-arena backing (#419): when set, slab/fslab are interior
                   * slices of a per-layer arena and must never be free()d —
@@ -393,24 +394,38 @@ static void eslot_release(ESlot *s){
 static int eslot_busy(const ESlot *s){ return __atomic_load_n(&s->in_flight,__ATOMIC_ACQUIRE)!=0; }
 static void eslots_acquire(ESlot **slots,int n){ for(int i=0;i<n;i++) eslot_acquire(slots[i]); }
 static void eslots_release(ESlot **slots,int n){ for(int i=0;i<n;i++) eslot_release(slots[i]); }
-/* Victim per una riga piena (#1034): uno slot svuotato da rss_guard (eid=-1,
- * slab=NULL) e' riusabile SOLO finche' gli slab vivi della riga stanno sotto
- * ecap — riusarlo rialloca uno slab, quindi e' crescita, non eviction. Le
- * prenotazioni in volo (eid<-1) contano come vive: stanno per possederne uno.
- * EN: reusing a slab-less slot re-allocates, so it only counts as eviction
- * EN: while the row's live-slab count is under ecap; else pick a slab owner. */
-static int eslot_lru_victim(ESlot *slots,int n,int ecap){
-    int lru=-1, empty=-1, live=0;
+/* Second-chance (clock) victim for a full row, cap-aware (#1034).
+ * Two policies must both hold:
+ *  - clock: skip a slot whose reference bit is set, clearing it (second chance),
+ *    so a hot resident survives one sweep instead of dying on a raw LRU stamp;
+ *  - cap: a slot emptied by rss_guard (eid==-1, slab==NULL) is only reusable while
+ *    the row's live-slab count is under ecap, because reusing it re-allocates a
+ *    slab and that is growth, not eviction. At or over the cap, evict a slab owner.
+ * In-flight reservations (eid<-1) count as live: they are about to own a slab.
+ * The hand persists per layer, so sweeps resume where the last one stopped. */
+static int eslot_clock_victim(ESlot *slots,int n,int *hand,int ecap){
+    if(n<=0) return -1;
+    if(*hand<0 || *hand>=n) *hand=0;
+    int live=0, empty=-1;
     for(int i=0;i<n;i++){
         ESlot *s=&slots[i];
         if(s->slab || s->eid<-1) live++;
-        if(eslot_busy(s) || s->eid<-1) continue;
-        if(!s->slab){ if(s->eid==-1 && empty<0) empty=i; continue; }
-        if(s->eid==-1) return i;              /* slot libero che possiede ancora lo slab */
-        if(lru<0 || s->used<slots[lru].used) lru=i;
+        /* An emptied, idle slot is a candidate only if the cap allows growth. */
+        if(!s->slab && s->eid==-1 && !eslot_busy(s) && empty<0) empty=i;
     }
-    if(empty>=0 && live<ecap) return empty;   /* sotto il tetto: meglio il vuoto che sfrattare */
-    return lru;
+    int owner=-1;
+    /* Two passes: the first may clear reference bits, the second then finds a victim. */
+    for(int pass=0;pass<2 && owner<0;pass++) for(int seen=0;seen<n;seen++){
+        int i=*hand; *hand=(i+1)%n;
+        ESlot *s=&slots[i];
+        if(eslot_busy(s) || s->eid<-1 || !s->slab) continue;  /* busy, reserved, or empty */
+        if(s->eid==-1){ owner=i; break; }        /* free but still owns its slab: best victim */
+        if(!s->recent){ owner=i; break; }        /* reference bit already clear: evict */
+        s->recent=0;                             /* grant the second chance */
+    }
+    if(empty>=0 && live<ecap) return empty;      /* under the cap: growing beats evicting */
+    if(owner>=0) return owner;
+    return live<ecap ? empty : -1;
 }
 
 typedef struct {
@@ -442,7 +457,7 @@ typedef struct {
     uint8_t **Lc8, **Rc8; float **Lsc, **Rsc;    /* alias KV8 (fp8 + scale) della KVState attiva */
     int *kv_start;                               /* prima pos valida nella KV del layer (MTP: parziale) */
     KVState *kv;
-    ESlot **ecache; int *ecn; int ecap;          /* LRU expert per-layer */
+    ESlot **ecache; int *ecn, *ehand; int ecap;      /* second-chance expert cache per layer */
     float **kv_dev_L, **kv_dev_R; int *kv_dev_valid; /* ombra KV su device (decode) */
     float **ln_dev;                              /* in_ln/post_ln cached on device: [layer*2+{0,1}] (Inc.4) */
 #ifdef COLI_VULKAN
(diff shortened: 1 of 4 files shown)


JustVugg:
CI approved and running. Two notes while it goes:

1. Respect for the scoping honesty — *"this does not claim eviction scanning caused the throughput curve; that explanation was withdrawn"* is exactly how we like claims sized here. A clock hand instead of a full oldest-scan per miss is independently valid.
2. What the review will need beyond green CI: eviction POLICY changes shift **which** experts stay resident, so hit rates can move either way even with tokens unchanged. One before/after datapoint on a real run — `Expert cache hit rate` line plus tok/s, same prompt, LRU vs clock — settles it. Also the `make -C c check` box in your list is still unticked; CI covers most of it, but tick it once it's true locally.

If the hit-rate A/B comes out neutral-or-better, this is mergeable.

pacocartones:
Rebased onto `dev`.

The conflict turned out to be semantic rather than textual: while this sat, `b00688b` (@ZacharyZcR, #1034) rewrote the same three victim scans this PR had replaced. Taking my side wholesale would have silently reverted that fix — a genuinely good catch, since reusing a slab-less slot re-allocates its slab, so `rss_guard` lowers `ecap` and the cache grows straight back and the guard fires forever.

So rather than picking a side, `eslot_clock_victim(slots, n, hand, ecap)` now enforces both invariants: clock second-chance keeps hot experts resident, and cap-awareness means a slot emptied by `rss_guard` is only reusable below `ecap` — at or over it, a slab owner is evicted instead. @ZacharyZcR is credited as `Co-authored-by`.

`test_eslot_inflight` covers both invariants together and passes.

mfethe1:
Windows/x86-64 validation of this branch, run today on a real machine — a platform
your validation checklist and CI runs (none present on head `0054162`) do not appear
to cover. Windows 11 (10.0.26200), Ryzen 7 3700X, gcc 15.2.0 (x86_64-w64-mingw32),
Python 3.12.7, no GPU. Fresh clone of `pull/1092/head` at `0054162a7a006190330185e61dfc756b5686534e`,
nothing reused.

**Focused tests from the PR checklist:**
- `make -B tests/test_eslot_inflight && ./tests/test_eslot_inflight` → `test_eslot_inflight: ok`
- `make -B tests/test_olmoe_serve_framing && ./tests/test_olmoe_serve_framing` → `olmoe serve framing baseline: ok`

**Default CPU builds:** `make colibri` and `make olmoe` both link clean with
`-march=x86-64-v3 -fopenmp -static -lpsapi`; `git diff --check` clean.

**Full suite (`make check`):** 678 tests, 8 failures, 49 skipped. All 8 are
pre-existing on base `dev` @ `6896e5fc4ba2dae66e2f03cbf569bcb5367eb9d1` — the FAIL
lists are identical line-for-line on both branches (all 8 in `test_fp8_repack.*`,
i.e. Python-side, environment-related, none attributable to this PR). Full FAIL
lists for both SHAs available on request.

One observation connecting to your scoping note ("this does not claim eviction
scanning caused the throughput curve"): we measured a hard ceiling on the
cache-policy side in a different engine (offline replay, Belady oracle proxy:
80.84% steady-state at 3.41x working-set scaling, 16.84pp below the 97.68% target —
no online policy beats Belady). Under that ceiling, hot-path selector latency like
the 873.85 → ~17-22 ns/call scaling fix here is exactly the layer where headroom
still exists. The two results are complementary, not conflicting: policy is
ceilinged; selector mechanics are not.

pacocartones:
Following up on the earlier short-prompt benchmark, which came out as a tie (LRU 62.4% vs clock 62.3%). That test wasn't fair to this PR: with a short prompt the whole working set stays resident and no eviction decision is ever exercised, so the two policies are indistinguishable by construction. I re-ran the A/B on a reuse workload, which is where second-chance should earn its keep.

**Setup.** OLMoE-1B-7B int8 (64 experts/layer, 16 layers, top-8), CPU-only, 2 threads. Single natural English passage of 704 tokens (no artificial repetition): 200 prompt + 504 decode = 89,984 expert requests. Two decode paths: free-running greedy generation, and teacher-forced over the real tokens (`PPL=1`). Cache pressure at `ecap` 16 and 32 (25% and 50% of experts resident). 4 repetitions per cell, `7f9a1af` (LRU) vs `0054162` (clock).

**Control.** The generated token streams are byte-identical across both policies and both ecaps, and TF-NLL is identical (2.6618 nats / ppl 14.32) in every run. Same expert-request sequence, so the hit-rate delta is attributable to eviction policy alone. Hit/miss counts were bit-exact across all 4 repetitions of each cell.

| workload | ecap | LRU hit rate | clock hit rate | delta | extra misses |
|---|---|---|---|---|---|
| greedy gen | 16 | 51.4% (46217/43767) | 49.9% (44897/45087) | −1.5 pp | +1320 (+3.0%) |
| greedy gen | 32 | 77.1% (69403/20581) | 76.1% (68480/21504) | −1.0 pp | +923 (+4.5%) |
| teacher-forced | 16 | 49.5% (44571/45413) | 48.0% (43202/46782) | −1.5 pp | +1369 (+3.0%) |
| teacher-forced | 32 | 78.5% (70670/19314) | 77.2% (69461/20523) | −1.3 pp | +1209 (+6.3%) |

**Result: clock loses, consistently.** Not a tie — it is behind in all four cells, by 1.0–1.5 pp, which is 3–6% more expert misses on an identical request stream. My reading is that MoE expert access is strongly recency-driven, and collapsing LRU's full timestamp ordering into a single `recent` bit throws away information the workload actually needs: the second chance keeps experts alive that LRU correctly evicts.

Two caveats I want to state explicitly. First, throughput: I could not measure a reliable tok/s difference. The box is a shared 4-vCPU VM and contention noise (occasional runs at 1.4–2.5 tok/s) dominates the signal, so the medians are inconsistent in sign and I'm not claiming a speed regression — only a hit-rate one. Second, scope: this exercises `c/olmoe.c` only. The cap-aware half of the commit lives in `c/colibri.c` and is not covered here, and the #1034 invariant this rebase carefully preserved is independently correct.

Given that, I'd suggest not landing the `olmoe.c` half as-is. If there's a workload where the clock does win, I'm happy to run it — the harness, driver script and raw logs for these 32 runs are reproducible:

```bash
cd colibri && git checkout <rev> && make -C c clean && make -C c olmoe ARCH=native
SNAP=<olmoe_merged> OMP_NUM_THREADS=2 ./c/olmoe 16 8 reuse1.json        # greedy
SNAP=<olmoe_merged> OMP_NUM_THREADS=2 PPL=1 ./c/olmoe 16 8 reuse1.json  # teacher-forced
```

mfethe1:
A correction to something we said earlier in this ecosystem, and a sharper question for #1092 that comes out of it.

## The correction

We had been carrying an interpretation that byte-halving (fp8 minting) changes cache-policy value. Testing it properly, the first version of that claim was **too broad and we're retracting it**: under uniform per-expert halving, the capacity axis just rescales by 2×. That's a homothety, not a policy effect. A test built on it would have passed while proving nothing.

## What we tested instead

Same Kimi-Linear-48B decode trace (256 experts, 26 layers, top-8, 62,400 assignments), byte-capacity LRU/LFU replay, atomic routing events preserved. We held the **compression-ratio multiset fixed** and varied only *which experts* got which ratio — same byte budget, same dispersion, same working set, so any difference is assignment alone:

| ratios assigned | Δ crossover (byte-working-set fraction) |
|---|---:|
| independently of popularity | **+0.16 pp** |
| popular experts compress **better** | **−27.9 pp** |
| popular experts compress **worse** | **+10.6 pp** |

The popularity-independent arm lands on the pure-rescaling control within 0.16 pp. Heterogeneous expert sizes on their own do essentially nothing; correlate the same sizes with reuse and the crossover swings up to 28 points of working-set fraction.

**Narrower conclusion:** policy ranking is sensitive to the **covariance between reuse frequency and physical expert extent** — not to quantization, and not to size heterogeneity by itself.

Stated plainly about what this is: **synthetic, extremal sensitivity arms.** The ratios are a seeded lognormal, and the correlated arms are perfect rank-correlation by construction, so real covariance is weaker and real effects sit between the arms. No throughput, latency, or speedup claim anywhere — this is a simulated cache replay over a measured routing trace.

## The questions about #1092

The clock-vs-LRU A/B there is the cleanest policy comparison published in this ecosystem — bit-exact hit counts, identical token streams, and it correctly reports a *negative* result. That's why it's worth asking what it was matched on:

1. Was the comparison matched by **expert slots** or by **physical bytes**?
2. Were expert extents **uniform** in that run?
3. Is there a **model-matched artifact** where routing identity can be joined to per-record `encoded_payload_bytes`?

Question 3 is the blocker on our side, and I want to be explicit about why rather than imply we have data we don't. We ran `repack_rans.py --census` from #594/#1205 on Windows/x86-64 and the accounting checked out exactly — but that was the **nine-record unit-test fixture**, 1024 logical bytes per record. It validates the accounting logic; it cannot answer this question, because the encoded extents have to belong to the *same experts* as the routing trace. Nine synthetic records don't join to 5,083 experts.

If such an artifact exists — or if the census can be pointed at a real minted directory whose expert identities line up with a routing capture — we'd contribute the byte-capacity replay joining `{layer, expert_id, reuse_count, encoded_payload_bytes}` and report whether the policy crossover survives normalization by byte-working-set fraction. That's the measurement that would turn this from a directional sensitivity result into something with real extents behind it.

One methodological note in case it's useful: we found a crossover has to be defined as the **last** sign change with no return to positive. A first-sign-change rule fired on a −0.0 wobble in our data and returned a crossover 6× off the structural one, which inverted the conclusion until we caught it. And crossovers shouldn't be quoted tighter than the grid supports — ours is ±5 pp of working-set fraction, which is why the reported effects are the 10–28 pp ones.

ZacharyZcR:
Closing this experiment based on its own controlled A/B result: second-chance lost to the current LRU policy in all four measured cells, increasing expert misses by 3–6% on identical request streams, while no reliable throughput benefit was demonstrated. The branch is also conflicting again, and preserving the independently correct cap-aware behavior from #1034 does not justify landing the policy regression. Thanks for the unusually careful negative-result follow-up and scope correction.

PR #1097, opened 2026-08-19
perf(deepseek_v4): expert-loader pool default 3 → 9 lanes (measured 1.41× decode)

JustVugg:
## What

`V4_LOADER_LANES` (48db957) made the expert-loader pool depth runtime-tunable but left the default at the compile constant **3**. This flips the **pool default only** to **9** — the CPU reservation deliberately keeps subtracting the compile constant, per the two-consumers note in the source: lanes block in `pread` and do not consume cores.

## Evidence

Quiet-box campaign on the **real V4-Flash checkpoint** (156 GB streamed on a 25 GB box), 8 interleaved runs (3,9)×4, same prompt, 32 tokens:

| | lanes=3 | lanes=9 | ratio |
|---|---|---|---|
| decode, mean of 4 | 147.3 s | **104.8 s** | **1.41×** |
| decode, median | 152.9 s | 104.4 s | 1.46× |
| TTFT | 55.7 s | 62.6 s | noise |
| completions / OOM | 4/4, none | **4/4, none** | — |

An earlier apparent lanes=9 crash reproduced as *host memory pressure* (a stale 18 GB python3 OOM kill in dmesg), not the pool — the quiet campaign shows 8/8 clean. Consistent with the in-tree 48 ms → 29.6 ms cold-read figure and the QD1=86 MB/s → QD8=696 MB/s `dd` measurement on the same shards. A first noisy attempt at this A/B was **discarded as invalid** (concurrent builds polluted it) — this campaign replaced it rather than averaging it in.

## Gates

- tiny oracle full PASS on the flipped default (token-exact, serve protocol, KV prefix reuse)
- `V4_LOADER_LANES` still overrides for disks that behave differently (the knob's docs unchanged)

Roadmap: closes item 4.1.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -4515,13 +4515,23 @@ typedef struct {
  * only the pool. */
 enum { DUAL_EXPERT_LOADER_COUNT = COLI_V4_EXPERT_LOADER_COUNT };
 enum { DUAL_EXPERT_LOADER_MAX = 16 };
+/* Default del POOL (non della riserva CPU qui sotto: quella resta sul COMPILE
+ * default — le lane bloccano in pread e non consumano core, vedi il commento
+ * sopra). 9 misurato contro 3 su questa classe di macchina: 8 run interfogliate
+ * a box quieto sul V4-Flash reale, decode 147.3s -> 104.8s medi (1.41x, 1.46x
+ * in mediana), TTFT invariato nel rumore, zero OOM/crash su 8/8 run — coerente
+ * con la misura in-tree 48ms -> 29.6ms per lettura fredda. V4_LOADER_LANES
+ * resta la manopola per dischi che si comportano diversamente.
+ * EN: pool default only; the CPU reservation keeps subtracting the compile
+ * constant. 9 vs 3 measured on the real checkpoint: 1.41x decode, no OOM. */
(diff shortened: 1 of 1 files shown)


dcutugno:
Data point from the GPU tier's reference box (RTX 5080 16 GB, 2× NVMe mirror, canonical GPU env, 3324-token prompt, 48 greedy tokens, two interleaved runs per config on v1.7.0):

| config | TTFT | decode (47 tok) |
|---|---|---|
| default 9 lanes | 126.5 s / 127.6 s | **33.9 s / 32.4 s** (0.70 s/tok) |
| `V4_LOADER_LANES=3` | **114.9 s / 116.4 s** | 35.3 s / 33.6 s (0.73 s/tok) |

Your 1.41× decode reproduces in direction but not size there: with the CUDA tier most decode experts hit the VRAM mirrors / RAM cache, so 9 lanes buys only ~4 % decode — while prefill loses ~10 %, because the loader threads contend with the MoE expert bank's own parallel O_DIRECT refill reads for the same two NVMe queues (`COLI_CUDA_MOE_BATCH=1`). Net on a GPU box, 3 is the better setting; on the CPU path the new default 9 is clearly right, exactly as you measured. Sending a two-line docs PR that records this next to `V4_LOADER_LANES` so GPU users know to turn it back down.

PR #1108, opened 2026-08-19
fix(CUDA): set weights_owned before the host-to-device copy so failed uploads free their buffer

monotophic:

In the CUDA weight-upload path, ownership (`weights_owned=1`) is recorded
only after the H2D `cudaMemcpy` succeeds. If the malloc succeeds and the
copy fails, the error path's `coli_cuda_tensor_free` sees an un-owned
pointer and skips the `cudaFree` — the allocation leaks for real, and the
byte ledger never sees it (the tensor was still untracked). Ownership is
a fact of the allocation, not the copy: this sets it between the malloc
and the memcpy. One line; a device-side injection test locks it.

**Behavioral contract**
- A failed upload releases every byte it allocated; repeated failures
  leave device free memory unchanged.
- Success path byte-identical (ownership, frees, ledger all unchanged).

**Capstone matrix**
| claim | decisive evidence |
|---|---|
| the leak is real and consumer-visible | unfixed head, GB10: 8 injected copy-failures leak 538,443,776 B, monotonic ~67 MB/cycle, ledger blind (0 tracked) |
| independently reproduced | second harness (different shape/fmt, ld --wrap seam): monotonic 37,617,664→229,900,288 B on base; flat on fix |
| the fix closes it | same injections on fix head: 0 B cumulative, exit 0 |
| no new free hazard | every upload exit path traced under the new ordering (both reviews): single-free, no free-of-unallocated, compressed/arena path still owned=0 |
| tests bite and suites pass | new test exits 1 on unfixed head; full `make cuda-test` green on fix head; CPU `make check` + METAL suites green |

<details><summary>Fuller matrix and review record</summary>
Two-reviewer roster + one fix round. The fix round was test-infra only:
the injection hook initially used two CUDA symbols outside
backend_gpu_compat.h's mapped surface (would break the HIP compile lane);
rewritten to the mapped surface, product diff untouched. HIP compile is
inspection-verified only (no AMD hardware here) — CI's hip-syntax lane
arbitrates, same as prior PRs. Success-path control: 8 create/free
cycles, zero free-memory delta, ledger restored, both heads.
Pre-existing adjacent observations recorded out of scope: a ledger-counter
race under parallel pin_load; ANS async-failure free ordering.
</details>

**Durable vs current-state:** the ordering fix is durable; leak-rate
figures are current-state (GB10/sm_121, ~64 MB test tensors, 2026-08-18,
base ad79236).

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -824,7 +824,7 @@ rans: $(RANSLIB)
 $(RANSLIB): tools/rans_ctypes.c rans.h
 	$(CC) $(CFLAGS) -fPIC -shared $< -o $@ $(LDFLAGS)
 
-cuda-test: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/test_backend_cuda.cu tests/test_ragged_attention.cu tests/test_mxfp4_cuda.cu tests/mxfp4_ref.c tests/test_fp8_warp_cuda.cu
+cuda-test: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/test_backend_cuda.cu tests/test_ragged_attention.cu tests/test_mxfp4_cuda.cu tests/mxfp4_ref.c tests/test_fp8_warp_cuda.cu tests/test_weights_owned_cuda.cu
 	@command -v "$(GPUCC)" >/dev/null 2>&1 || { echo "$(GPUCC_NAME) not found" >&2; exit 1; }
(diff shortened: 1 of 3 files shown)


JustVugg:
Merged via #1115 with your commits and authorship intact — the rebase was ours, since #1098 (your own fix) conflicted it on the shared `cuda-test` dependency list. Ships in v1.7.0.

PR #1103, opened 2026-08-19
planner: add OLMoE + Kimi K3 + Inkling + DeepSeek V4 geometry adapters (closes #1066)

SebaWag:
Closes #1066 — all four unproven production planners now have Python geometry adapters in `family_registry.py`, each mirroring the engine's actual allocation in C rather than trusting config hints.

---

## 1. `_olmoe_geometry` — OLMoE (7B)

Conventional full-head K/V cache in fp32, no recurrent state.

- **K/V cache sized with `num_attention_heads` (not `num_key_value_heads`)** — `olmoe.c` allocates per q-head (olmoe.c:1019-1020, 1047-1048, 1097-1098, 1412-1413), fp32. Sizing from `n_kv_heads` would under-reserve by the GQA ratio.
- **`head_dim` derived** as `hidden_size / num_attention_heads` (olmoe.c:327).
- **Workspace = 4 attention temporaries** (`q, k, vv, ctx`), each S x hidden floats (olmoe.c:643, 662).

## 2. `_kimi_geometry` — Kimi K3 (2.8T hybrid)

69 KDA (recurrent) + 24 gated MLA layers.

- **MLA cache (Lc/Rc)** per non-KDA layer: `context × (kv_lora + qk_rope) × 4` (kimi_k3.c:1704-1706).
- **KDA recurrent state** per KDA layer: `kda_heads × kda_hd × kda_hd × 4` (kimi_k3.c:691) — **fixed**, does NOT scale with context.
- **KDA membership** from `linear_attn_config.kda_layers` (1-based, kimi_k3.c:552-555).
- **Workspace** = max(KDA temporaries at kimi_k3.c:896-898, MLA at kimi_k3.c:992-994).

## 3. `_inkling_geometry` — Inkling (975B hybrid GQA)

Sliding-window (5 of 6 layers) + global layers + short convs + optional DMel audio tower.

- **KV cache** per layer from L_KV/L_HD (inkling.c:80-82; kv_alloc:1687-1699). Sliding layers are a **ring of `window` rows** (kv_ring_rows:1124-1125) — capped growth past the window; global layers keep full context.
- **Conv states** (inkling.c:888-890, 1646): four depthwise short-conv states per layer, `conv_k-1` rows each — context-independent → `fixed_state_bytes`.
- **Audio tower** (inkling.c:808-832): DMel encoder table `[mel_bins×mel_vocab, hidden]` + RMSNorm in `fixed_state_bytes` when `audio_config` present.
[...]

PR #1109, opened 2026-08-19
fix(arm64): probe dotprod by compiling, prefer armv8.2-a base

### LINKED ISSUE #1104: ARM64: portable build fails with `vdotq_s32: target specific option mismatch` (needs +dotprod)
## Summary

`make check` (which runs `make portable` then `make test`) fails to compile on ARM64 hardware that supports dot products but whose baseline `armv8-a` does not expose `__ARM_FEATURE_DOTPROD`.

## Environment

- Oracle Cloud A1 Flex — **Neoverse-N1** (`aarch64`)
- `gcc 11.4` (Ubuntu 22.04)
- `make check` with default flags

## Error

```
In file included from tests/../quant.h:37,
                 from tests/../colibri.c:494,
                 from tests/test_cluster_protocol.c:4:
/usr/lib/gcc/aarch64-linux-gnu/11/include/arm_neon.h:32134:1: error: inlining failed in call to 'always_inline' 'vdotq_s32': target specific option mismatch
32134 | vdotq_s32 (int32x4_t __r, int8x16_t __a, int8x16_t __b)
      | ^~~~~~~~~
In file included from tests/../colibri.c:494,
                 from tests/test_cluster_protocol.c:4:
tests/../quant.h:651:29: note: called from here
  651 |     for(;i+16<=I;i+=16) acc=vdotq_s32(acc,vld1q_s8(w+i),vld1q_s8(x+i));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make[1]: *** [Makefile:973: tests/test_cluster_protocol] Error 1
```

## Root cause

- `PORTABLE_ARCH = armv8-a` on aarch64 (c/Makefile:954) → compiled with `-march=armv8-a`, which does **not** define `__ARM_FEATURE_DOTPROD`
- `quant.h:651` calls `vdotq_s32` unconditionally in the NEON int8 path
- The `ARCH=native` path (c/Makefile:139-156) has a probe that *re-adds* `+dotprod` when the compiler doesn't report `MATMUL_INT8` (it reads `/proc/cpuinfo` for `asimddp`), but `make portable` overrides `ARCH` to the generic `armv8-a`, bypassing that logic

## Workaround

Compiling with the feature flag works on dot-product-capable silicon:

```bash
make check ARCH=armv8.2-a+dotprod PORTABLE_ARCH=armv8.2-a+dotprod
```

Verified: full suite **547 tests, 543 passed, 36 skipped** on Neoverse-N1 with that flag (the 4 failures are `test_makefile_platform` expecting the default portable flags — an artifact of the override).

## Suggested fix

Either:
1. Have `make portable` on aarch64 probe `/proc/cpuinfo` for `asimddp`/`i8mm` like the `native` path does and append the modifiers, or
2. Guard the `vdotq_s32`/`sdot` kernels in `quant.h` behind `#ifdef __ARM_FEATURE_DOTPROD` with a scalar fallback (keeping `armv8-a` truly portable as the name implies).

Happy to submit a PR for either approach.

SebaWag:
Fixes #1104

## Problem

`make check` fails to compile on ARM64 hardware that supports dot products (e.g. Oracle Cloud A1 Flex, **Neoverse-N1**) when the default `ARCH=native` branch kicks in:

```
/usr/lib/gcc/aarch64-linux-gnu/11/include/arm_neon.h:32134:1: error: inlining failed in call to 'always_inline' 'vdotq_s32': target specific option mismatch
tests/../quant.h:651:29: note: called from here
  651 |     for(;i+16<=I;i+=16) acc=vdotq_s32(acc,vld1q_s8(w+i),vld1q_s8(x+i));
```

## Root cause

The aarch64 `ARCH=native` branch of `c/Makefile` probed for dot-product support with a **macro-only** check:

```make
$(CC) -march=armv8-a$(ARCH_EXT) -dM -E - </dev/null | grep -cE '__ARM_FEATURE_DOTPROD|__ARM_FEATURE_MATMUL_INT8'
```

But **GCC 11 defines `__ARM_FEATURE_DOTPROD` for `-march=armv8-a+dotprod` yet cannot actually emit `vdotq_s32` for that base** ("target specific option mismatch"). The macro is defined, so the guard in `quant.h:642` selects the `vdotq_s32` path — and then the compiler rejects the intrinsic. Every test binary that pulls in `colibri.c` → `quant.h` fails to build.

On Neoverse-N1 (`asimddp` present, an ARMv8.2 core), the fix that *works* is `-march=armv8.2-a+dotprod` — dotprod is a first-class ISA feature there.

## Fix

Replace the macro-only probe with a **real compile** of the `vdotq_s32` intrinsic from a temporary file:

1. Try `armv8.2-a$(ARCH_EXT)` first (compiles on GCC 11 + Neoverse-N1).
2. Fall back to `armv8-a$(ARCH_EXT)` if the compiler accepts the modifiers there too.
3. Otherwise degrade to the existing warning (idot kernels compile out, int4 S=1 decode falls back to f32) — unchanged behavior from before, #631.

`make portable` is untouched (still plain `armv8-a`, no modifiers), so `test_makefile_platform`'s portable-flag expectations hold.

## Validation

On Oracle Cloud A1 Flex (Neoverse-N1, gcc 11.4, Ubuntu 22.04), plain `make check` with **no override** now passes:

```
Ran 565 tests in 663.153s
OK (skipped=36)
```

0 failures. Previously the build failed to compile at all on the default `ARCH=native`.

---

Note: this is the ARM64 portable-build issue documented in #1104; it is independent of the planner-geometry work in #1103 (different files, no overlap).

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -141,10 +141,25 @@ ifeq ($(shell $(CC) -mcpu=native -dM -E - </dev/null 2>/dev/null | grep -c __ARM
 CPUFEAT  := $(shell grep -m1 '^Features' /proc/cpuinfo 2>/dev/null)
 ARCH_EXT := $(if $(filter asimddp,$(CPUFEAT)),+dotprod)$(if $(filter i8mm,$(CPUFEAT)),+i8mm)
 ifneq ($(ARCH_EXT),)
-ifneq ($(shell $(CC) -march=armv8-a$(ARCH_EXT) -dM -E - </dev/null 2>/dev/null | grep -cE '__ARM_FEATURE_DOTPROD|__ARM_FEATURE_MATMUL_INT8'),0)
+# gcc 11 defines __ARM_FEATURE_DOTPROD for -march=armv8-a+dotprod yet cannot
+# emit vdotq_s32 for that base ("target specific option mismatch"): the macro
+# probe alone is not enough. Verify by compiling the intrinsic for real, and
+# prefer the armv8.2-a base where dotprod is a first-class ISA feature
+# (Neoverse-N1 and friends report asimddp but are armv8.2 cores). Fall back to
+# armv8-a only if the compiler accepts the modifiers there too; otherwise the
+# idot kernels compile out and int4 S=1 decode falls back to f32 (#631, #1104).
+PROBE_C := $(shell mktemp /tmp/coli_v4_probe.XXXXXX.c)
+PROBE_OK := $(shell echo '#include <arm_neon.h>' > $(PROBE_C); echo 'int p(const signed char*w,const signed char*x,int n){int32x4_t a=vdupq_n_s32(0);for(;n>=16;n-=16)a=vdotq_s32(a,vld1q_s8(w),vld1q_s8(x));return vaddvq_s32(a);}' >> $(PROBE_C); $(CC) -c $(PROBE_C) -o /dev/null -march=armv8.2-a$(ARCH_EXT) 2>/dev/null && echo yes; rm -f $(PROBE_C))
+ifeq ($(PROBE_OK),yes)
+ARCHFLAG := -march=armv8.2-a$(ARCH_EXT)
+else
+PROBE_C8 := $(shell mktemp /tmp/coli_v4_probe.XXXXXX.c)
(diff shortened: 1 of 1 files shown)


JustVugg:
CI approved and running. The diagnosis is exactly right — the macro probe trusted `__ARM_FEATURE_DOTPROD` while GCC 11 cannot emit `vdotq_s32` on the `armv8-a` base it was defined for, and probing by *compiling the actual intrinsic* is the only ground truth. Preferring `armv8.2-a` (Neoverse-N1's real ISA level) with the probed `armv8-a` fallback keeps the #631 compile-out path intact.

One honest note for the record: our ARM CI job runs `-mcpu=native` on Graviton, where the outer native probe succeeds — so CI validates *no regression*, not the fix itself; the fix's validation is your Oracle A1. If you can paste the one-line `make check` outcome from the A1 on this branch, that plus green CI is the merge bar. Two quick sanity confirmations from reading: the $(shell) probes only evaluate inside the aarch64-native branch (no cost on x86), and the temp files are cleaned in the same shell — both good.

PR #1114, opened 2026-08-19
fix(CUDA): fmt=8/fmt=6 scale-byte accounting — rebase of #1100 by @monotophic

JustVugg:
@monotophic's #1100, rebased onto `dev` by the maintainers — commits and authorship theirs. It went conflicting the moment #1098 (their own `__syncthreads()` fix) merged, since both add a test to the `cuda-test` target.

**Conflict resolution: pure union, both mechanical** — the `cuda-test` dependency line now lists both `test_absorb_determinism.cu` (from #1098) and `test_fp8_cuda.cu` (this PR), and both recipe bodies are kept in their intended order (fp8 before MXFP4, absorb-determinism last, per each test's own comment about schedule sensitivity).

Fix and its regression test unchanged. Targeting v1.7.0. Supersedes #1100 on merge.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -839,7 +839,7 @@ rans: $(RANSLIB)
 $(RANSLIB): tools/rans_ctypes.c rans.h
 	$(CC) $(CFLAGS) -fPIC -shared $< -o $@ $(LDFLAGS)
 
-cuda-test: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/test_backend_cuda.cu tests/test_ragged_attention.cu tests/test_absorb_determinism.cu tests/test_mxfp4_cuda.cu tests/mxfp4_ref.c tests/test_fp8_warp_cuda.cu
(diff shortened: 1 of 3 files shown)


JustVugg:
Deferred to the next release (v1.7.1/v1.8.0) — **not on merit, on infrastructure**. This PR is 21/22 green; the one job left is `CUDA syntax check`, which downloads the CUDA network installer from NVIDIA's CDN, and that download has been hanging on GitHub's runners all afternoon (three separate jobs stuck on third-party downloads today — the Vulkan job's apt install has the same problem). Retrying twice changed nothing, because the stall is not ours.

Rather than hold v1.7.0 — which ships a sixth engine with its GPU tier — on NVIDIA's CDN, this goes in the moment the runners recover. The branch is already rebased and ready; it needs a green, nothing else.

Follow-up on our side: your PRs are the argument for putting `timeout-minutes` and toolkit caching on these jobs (#953), so a slow mirror fails fast instead of blocking a release.

PR #1105, opened 2026-08-19
feat(tools): full-family FP8 (e4m3) container mint with content-validated resume

monotophic:
Authored by Fable 5 in Claude Code, analysis in partnership with @Monotophic

Extends `repack_fp8_passthrough.py` from experts-only to **full-family**
checkpoint-faithful emission: routed experts, o_proj, attention families,
kv_b_proj (fmt=8 stamped), embed_tokens/lm_head/norms/router raw
pass-through, plus config/generation_config/tokenizer.json — the output
loads standalone. Emission is deterministic (canonical metadata key
order); resume is content-validated (size+sha256 per shard, input
fingerprint that aborts by name if the source changed between runs);
malformed repack-kind tensors refuse loudly instead of dropping silently.
Companion engine backstop: `weight_at` gains an explicit fmt=3 branch and
a device-side `__trap()` fall-through for undecodable formats (refuse,
never fabricate), with host- and device-side test coverage.

**Behavioral contract**
- Byte-faithful: every emitted payload byte identical to the checkpoint;
  aux files byte-copied.
- Deterministic: repeated mints byte-identical, cross-platform.
- Resume never trusts existence: partial/mutated output re-emits;
  mutated input aborts by name.
- Malformed inputs refuse with the tensor name; legitimate inputs never
  false-refuse.
- Engine backstop: unsupported formats in `weight_at` trap instead of
  falling through to a wrong decode.

**Capstone matrix**
| claim | decisive evidence |
|---|---|
| deterministic emission | 8 toy mints byte-identical incl. metadata order; canonicalizer-removed mutant breaks on mint 0. Cross-platform: 140/140 shard SHA-256 identical across x86-64 Linux, aarch64 Linux, and macOS arm64 mints of the full real checkpoint |
| tests bite | vs origin/dev: full-family module 7F+2E/11, repack module 10F+16E/32; suite 58/58 green on head |
| resume is content-validated | truncated "completed" shard re-emitted byte-identical; mutated input → named InputChangedError abort |
| no false refusals on real input | dry-run over the real 755 GB checkpoint: rc=0, 58,059 tensors classified, ambiguous o_proj family (78 tensors) stamped correctly |
| the trap backstop is real | silicon bites (GB10): trap removed → 7/7 probes DECODED, exits 1; dispatch mutated vs predicate-driven expectations → 7 FAILED. Expectations derive from the host predicate, so the test stays correct as formats gain decode support |

<details><summary>Fuller matrix and review record</summary>
[...]

Issue #1099, opened 2026-08-19
cuda: add the __syncwarp() its two siblings carry, after grouped_s4_wmma's store_matrix_sync


monotophic:
Authored by Opus 5 in Claude Code, analysis in partnership with @Monotophic.

`grouped_s4_wmma` writes its WMMA accumulator to a shared tile and then reads it
back from other lanes without a warp barrier. Its two siblings, doing the same
thing a few lines above, both carry one.

| site | after `wmma::store_matrix_sync` | |
|---|---|---|
| `c/backend_cuda.cu:617` | `__syncwarp();` | present |
| `c/backend_cuda.cu:646` | `__syncwarp();` | present |
| **`c/backend_cuda.cu:685`** | **— nothing —** | **missing** |

All three then read `out[warp][...]` across lanes in a `for(int z=lane; z<N; z+=32)`
loop, so the same lane-to-lane visibility requirement applies to all three.

**What `compute-sanitizer` says.** On `backend_cuda_test`, which reaches this
kernel, `--tool=racecheck` reports **3 hazards (0 errors, 3 warnings)**, all of
them `Write access at nvcuda::wmma::store_matrix_sync ... in mma.hpp:546`.
Adding the single `__syncwarp()` takes that to **0 hazards (0 errors, 0
warnings)**, with the suite still passing and no new warnings anywhere:

```
before   RACECHECK SUMMARY: 3 hazards displayed (0 errors, 3 warnings)   suite rc=0
after    RACECHECK SUMMARY: 0 hazards displayed (0 errors, 0 warnings)   suite rc=0
```

Measured on NVIDIA GB10 (sm_121, CUDA 13.0), same binary and flags either side,
patch applied to a throwaway tree.

**Why this is worth a line of code even though it only reports as a warning.**

1. **The suite already drives this path on every run.** `c/tests/test_backend_cuda.cu:399-400`
   sets `COLI_CUDA_TC_INT4=1` and `COLI_CUDA_TC_MIN_ROWS=1` itself, so the kernel
   executes during a plain `make cuda-test`. In production the tensor-core path
   is opt-in; in the test suite it is not.
2. **It is the only thing standing between the CUDA suite and a `racecheck`
   gate.** `ragged_attention_test` is already clean. With this one line, a CI
   target that fails on `errors > 0` becomes possible with no warning noise to
   triage.
3. **The path's only numeric oracle is a 30% relative-RMS tolerance**
[...]

PR #1115, opened 2026-08-19
fix(CUDA): weights_owned before the H2D copy — rebase of #1108 by @monotophic

JustVugg:
@monotophic's #1108, rebased onto `dev` by the maintainers — commits and authorship theirs. It conflicted once #1098 and #1101 landed (all add a test to the same `cuda-test` dependency list).

**Conflict resolution: pure union on `c/Makefile`** — the `cuda-test` target now lists `test_absorb_determinism.cu` (#1098), `test_fp8_cuda.cu` (#1114) and `test_weights_owned_cuda.cu` (this PR), each recipe body preserved in its intended order. Fix and regression test unchanged.

Targeting v1.7.0. Supersedes #1108 on merge.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -840,7 +840,7 @@ rans: $(RANSLIB)
 $(RANSLIB): tools/rans_ctypes.c rans.h
 	$(CC) $(CFLAGS) -fPIC -shared $< -o $@ $(LDFLAGS)
 
-cuda-test: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/test_backend_cuda.cu tests/test_ragged_attention.cu tests/test_absorb_determinism.cu tests/test_mxfp4_cuda.cu tests/mxfp4_ref.c tests/test_fp8_warp_cuda.cu tests/test_fp8_cuda.cu
+cuda-test: backend_cuda.cu backend_cuda.h backend_gpu_compat.h tests/test_backend_cuda.cu tests/test_ragged_attention.cu tests/test_absorb_determinism.cu tests/test_mxfp4_cuda.cu tests/mxfp4_ref.c tests/test_fp8_warp_cuda.cu tests/test_fp8_cuda.cu tests/test_weights_owned_cuda.cu
 	@command -v "$(GPUCC)" >/dev/null 2>&1 || { echo "$(GPUCC_NAME) not found" >&2; exit 1; }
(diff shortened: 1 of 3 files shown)


JustVugg:
@monotophic — disposition of all eight of your PRs, and a thank-you that should not wait for the release notes.

**In v1.7.0:**
- #1099 → #1111 (`__syncthreads()` after `grouped_s4_wmma`'s store) — merged
- #1098 (`__syncthreads()` in the absorb softmax reduction) — merged
- #1100 → **#1114** (fmt=8/fmt=6 scale-byte accounting) — rebased by us, in CI
- #1108 → **this PR** (`weights_owned` before the H2D copy) — rebased by us, in CI
- #1101 (alloc/snprintf guards on the checkpoint-load path) — merged
- #1106 (refuse duplicate tensor names across shards) — merged

**Deferred to v1.8.0, on size not merit:** #1102 (fmt=8 on the kv_b absorb path, +548), #1105 (FP8 container mint, +3007), #1107 (logprob channel + token-ID intake, +818). All three are features that deserve a real review rather than a release-eve skim; the release was already cut when they arrived. They are at the front of the queue after the tag.

The rebases were ours to do: your three CUDA fixes each add a test to the same `cuda-test` dependency list, so merging the first one conflicted the other two. Pure union in every case, fixes and tests untouched — but review the resolutions if you want, they are stated in each PR body.

**And the part that matters:** eight PRs in thirteen hours, every one of them carrying its own regression test — `test_absorb_determinism.cu`, `test_fp8_cuda.cu`, `test_weights_owned_cuda.cu`, `test_798_guards.c`, `test_dup_name_refusal.c` — and every one green on first submission. The `__syncthreads()` report that started this came with a three-site table and `compute-sanitizer` racecheck output, which turned an investigation into a five-minute fix. That is not the normal standard for drive-by contributions to a C codebase, and this repo is measurably better for it. Thank you.

JustVugg:
Deferred to the next release (v1.7.1/v1.8.0) — **not on merit, on infrastructure**. This PR is 21/22 green; the one job left is `CUDA syntax check`, which downloads the CUDA network installer from NVIDIA's CDN, and that download has been hanging on GitHub's runners all afternoon (three separate jobs stuck on third-party downloads today — the Vulkan job's apt install has the same problem). Retrying twice changed nothing, because the stall is not ours.

Rather than hold v1.7.0 — which ships a sixth engine with its GPU tier — on NVIDIA's CDN, this goes in the moment the runners recover. The branch is already rebased and ready; it needs a green, nothing else.

Follow-up on our side: your PRs are the argument for putting `timeout-minutes` and toolkit caching on these jobs (#953), so a slow mirror fails fast instead of blocking a release.

Issue #1091, opened 2026-08-18
Benchmark: GLM-5.2 on Ryzen 7 9850X3D + RTX 5090, native Windows (0.89 tok/s median)


Appie-NL:
## Machine datapoint

| component | detail |
|---|---|
| CPU | AMD Ryzen 7 9850X3D, 8 cores / 16 threads |
| RAM |  64 GB DDR5 4800MT/s |
| DISK | 2TB WD_Black SN8100
| GPU | NVIDIA GeForce RTX 5090, 32607 MiB, driver 591.86 |
| OS | native Windows 11 |
| model | GLM-5.2 Colibri int4 |
| Colibri commit | `4ef9a9920dd4290b4ac26f74e73913fd18379bc9` |
| build | native Windows CUDA DLL, sm_120 |

## Configuration

```text
COLI_CUDA=1
CUDA_DENSE=1
CUDA_EXPERT_GB=auto
CUDA_RESERVE_GB=3
COLI_CUDA_PIPE=2
COLI_CUDA_ATTN=1
PIN=auto
PIN_GB=20
RAM_GB=52
CAP_RAISE=1
DRAFT=0
PIPE=1
PIPE_WORKERS=8
DIRECT=1
PILOT=1
PILOT_REAL=1
PILOT_WORKERS=4
cap=8
TEMP=0
NGEN=32
```

Fixed prompt: `Continue this story: The lighthouse keeper climbed the stairs and saw something impossible in the fog.`

The same fixed prompt was run three times. Existing `.coli_usage` history was retained, so these are warm, usage-trained runs rather than clean-install cold runs.

## Decode results

| run | prefill (21 tokens) | decode (32 tokens) | throughput | expert hit rate | RSS |
|---|---:|---:|---:|---:|---:|
| 1 | 18.77 s | 35.57 s | 0.90 tok/s | 38.1% | 43.10 GB |
| 2 | 18.93 s | 36.15 s | 0.89 tok/s | 38.1% | 43.10 GB |
| 3 | 18.60 s | 36.91 s | 0.87 tok/s | 38.2% | 43.10 GB |
| **median** | **18.77 s** | **36.15 s** | **0.89 tok/s** | **38.1%** | **43.10 GB** |

Additional runtime details:

- CUDA resident set: 20.32 GB VRAM
- CUDA expert tier: 478 resident experts / 10.15 GB
- MTP compiled active, but `DRAFT=0`; measured 1.00-1.03 tokens/forward
[...]

PR #1127, opened 2026-08-20
perf(iobench): sub-MB block sizes, and [threads] was a no-op on Windows

view321:
Two fixes to `c/iobench.c`, both of which change what it measures, plus a small tool that uses them.

Neither touches the engine. `iobench` is standalone and builds from its own make target.

## 1. Sub-MB block sizes were inexpressible

```c
long blk=(argc>2?atol(argv[2]):19)*1024*1024;
```

`atol("0.75")` is 0, so any fraction of a megabyte truncated away. GLM-5.2's experts are ~19 MB so this never mattered, but an expert's on-disk size is `3 · d_model · moe_intermediate_size · bits/8`, and a fine-grained MoE lands between roughly 0.2 and 1 MB — exactly the range the tool could not reach.

Now `atof`, rounded up to 4096 because `O_DIRECT` requires a sector-aligned length. The result line prints `%.4g MB` so a sub-MB block is legible instead of reading `0MB`.

## 2. `[threads]` did nothing on Windows

Every thread shared one fd. `compat_open_direct()` returns a **synchronous** handle — `FILE_FLAG_NO_BUFFERING` without `FILE_FLAG_OVERLAPPED` — and `ReadFile` on a synchronous handle serialises behind the file object lock. So the thread count had no effect and every Windows `iobench` figure was really a queue-depth-1 measurement.

`compat_pread()` itself is fine; the sharing was the problem. Each thread now opens its own fd via a small `bench_open()` helper (which is just the existing platform `#ifdef` block, factored out so it can be called twice).

Measured here, Phison 512 GB NVMe, Windows 11, 768 KiB blocks, `O_DIRECT`:

| | 1T | 2T | 4T | 8T | 16T |
|---|---|---|---|---|---|
| before | 1.09 | 1.07 | 1.07 | 1.02 | 1.02 |
| after | 1.02 | 2.31 | **2.91** | 2.78 | 2.58 |

GB/s. Cross-checked against an independent Python harness using `CreateFileW` + `FILE_FLAG_NO_BUFFERING` with per-thread handles: it reports 2.85–2.95 GB/s at 4–8 threads on the same file. The two agree after the fix and disagree by ~3× before it.

This is worth knowing when reading existing Windows datapoints. #1091's thread-scaling table (1T 9.57, 2T 9.90, 4T 9.85, 8T 10.65, 12T 10.72, 16T 9.57 GB/s on native Windows) is flat, which is the signature of this bug rather than a property of that drive — worth a re-run there.

Linux and macOS behaviour is unchanged: positional `pread` already concurs, and a private fd is correct there too.

## 3. `c/tools/blocksweep.py`

Walks the curve with `iobench` and prints a ready-to-paste markdown table. Stdlib only, like `tools/datapoint.py`.

```bash
python tools/blocksweep.py --shard /path/to/out-00069.safetensors
```

On the same drive (O_DIRECT, 128 reads, median of 3), against a plateau of 3.02 GB/s at ≥1.5 MB:

| expert size | 1 thread | 4 threads | 8 threads | vs plateau | QD1 cost |
|---|---|---|---|---|---|
| 192 KB | 0.45 | 1.70 | 1.95 | **−36%** | 4.33× |
| 384 KB | 0.84 | 2.45 | 2.50 | −17% | 2.98× |
| 768 KB | 1.27 | 2.88 | 2.73 | −10% | 2.15× |
| 1.5 MB | 1.58 | 2.99 | 2.94 | −3% | 1.86× |
| 3 MB | 1.91 | 3.01 | 2.96 | −2% | 1.55× |
| 6 MB | 2.44 | 3.15 | 3.13 | +3% | 1.28× |
| 12 MB | 2.55 | 3.07 | 3.07 | +1% | 1.20× |

Both effects are graded and monotone in expert size: smaller experts lose bandwidth outright, and lose proportionally more of it at low queue depth. There is no sharp knee — the penalty decays smoothly and is under 5% only above roughly 1.5 MB.

That last column is relevant to #441 (PILOT prefetch at queue depth 1): its cost is ~4× for a fine-grained container and ~1.2× for GLM-5.2's ~19 MB experts, so how much that issue matters depends strongly on the model being streamed.

**One drive, one platform.** The position of the roll-off depends on the controller, the NAND and whether the drive has DRAM, so none of this generalises until someone replicates it. That is why the script is here: it needs no model download and no engine build beyond `iobench`, so a datapoint costs a couple of minutes on any machine with an SSD and a large file.

## Checks

- `make iobench` builds with the repo's own CFLAGS, **0 warnings** (gcc 15, MinGW-w64, `-Wall -Wextra`).
- Backward compatible: an integer argument behaves as before (`19` → `19 MB`), the no-argument default is unchanged, and `tools/datapoint.py`'s `IOBENCH_RE` only captures the `-> N GB/s` field, which is untouched.
- Verified on Windows 11 / MSYS2. **Not yet built or run on Linux or macOS** — the changes are inside the existing per-platform `#ifdef` arms and only re-parenthesise them, but a second pair of eyes on those arms would be welcome.

### DIFF
--- a/c/iobench.c
+++ b/c/iobench.c
@@ -2,6 +2,8 @@
  * Misura cio' che fa il motore davvero: N thread che leggono in parallelo
  * (expert_load sotto omp parallel for), buffered oppure O_DIRECT.
  * uso: ./iobench <file_grande> [blocco_MB] [n_letture] [threads] [direct 0/1]
+ * blocco_MB accetta frazioni: 0.75 = 768 KiB. Un expert di un MoE
+ * fine-grained sta sotto 1 MB, e atol() lo troncava a 0.
  * build: gcc -O2 -fopenmp iobench.c -o iobench */
 #define _GNU_SOURCE
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


JustVugg:
Merged. The `atol("0.75") == 0` truncation is the kind of bug that hides behind a well-chosen default: GLM-5.2's experts are ~19 MB, so nobody ever passed a fraction and nobody ever noticed the argument silently became zero. It matters now precisely because the newer engines do not all have 19 MB experts.

The Windows `[threads]` no-op is the more consequential half, and it has a history worth naming: every Windows iobench number in `docs/benchmarks.md` was measured single-threaded regardless of what the reporter typed. That does not invalidate them — they are honest measurements of something — but it does mean the Windows disk figures are not comparable to the Linux ones the way the table implies. Worth a note there, which we will carry rather than ask you to.

`iobench` is standalone and never touches the engine, which is exactly why it earns a fast merge: the blast radius is a measurement tool, and the fix makes the measurements truer.

Issue #1117, opened 2026-08-19
[Bug]: coli doctor/plan report "result error" and 166.9 GB dense on DeepSeek-V4-Flash — expert tensor regex never matches


Crowmen007:
### Commit or version

`4ef9a9920dd4290b4ac26f74e73913fd18379bc9`

### Environment

Ubuntu 24.04, Linux 6.8.0-111-generic, AMD Ryzen 9 5950X (16C/32T, Zen 3, AVX2, no AVX-512), 62.7 GiB RAM, Samsung SSD 980 1TB NVMe (PCIe 3.0 x4, DRAM-less), ext4 on LVM, gcc 13.3.0, Python 3.12.3. CPU-only build (`make deepseek-v4`), no GPU backend.

Model: `deepseek-ai/DeepSeek-V4-Flash-0731`, official checkpoint, 48 shards, all 74 files verified against the HF manifest by size and sha256.

### Reproduction steps

```bash
cd c && make deepseek-v4
python ./coli doctor --model <model_dir>
python ./coli plan   --model <model_dir> --ram 32
```

### Expected behavior

`doctor` passes (or at least does not fail) on a healthy install, and `plan` reports the dense weights as ~6.27 GiB, matching `docs/deepseek-v4.md`:

> Dense weights occupy about 6.27 GiB and a resident BF16 output head about 1.06 GiB.

### Actual behavior and logs

`doctor` exits `result error`, and both commands report the entire 166.9 GB checkpoint as dense with zero experts:

```
[fail] memory.ram         RAM budget cannot hold one expert slot per sparse layer
[warn] placement.plan     RAM budget cannot hold one expert slot per sparse layer; VRAM tier was clamped ...

policy quality · quality-preserving yes
model  48 shards · 166.9 GB
disk   0.0 GB cold experts · 567.1 GB free
RAM    32.0 GB budget · 166.9 GB dense · 3.7 GB runtime · 0.0 GB warm experts · cap 0/layer
hit    100% projected expert residency

result error
```

The cause is the expert-tensor regex in `c/resource_plan.py`:

```python
EXPERT_RE = re.compile(r"model\.layers\.(\d+)\.mlp\.experts\.(\d+)\.")
```

This checkpoint names its expert tensors without the `model.` prefix and with `ffn` instead of `mlp`:

```
layers.0.ffn.experts.0.w1.weight
layers.0.ffn.experts.0.w1.scale
layers.0.ffn.shared_experts.w1.weight
```

[...]

Issue #1110, opened 2026-08-19
[Bug]: resource_plan EXPERT_RE misses DeepSeek V4 Flash naming — all routed experts counted as dense


NiklasWork:
## Summary

`resource_plan.py` detects routed experts with a regex that only matches the
GLM-5.2 tensor naming. DeepSeek V4 Flash uses a different naming, so **every
routed expert is counted as dense**. `coli plan` and `coli doctor` then report
the full checkpoint as the dense set, and `doctor` fails on a machine where the
model actually runs fine.

Still present on `main` (checked today, v1.6.2).

## The mismatch

```python
# c/resource_plan.py:15
EXPERT_RE = re.compile(r"model\.layers\.(\d+)\.mlp\.experts\.(\d+)\.")
```

`deepseek-ai/DeepSeek-V4-Flash-0731` names the same tensors:

```
layers.3.ffn.experts.17.w1.weight        # no `model.` prefix, `ffn` not `mlp`
layers.3.ffn.shared_experts.w1.weight
```

From that checkpoint's `model.safetensors.index.json`: 72317 tensors, of which
11008 match `layers.N.ffn.experts.M.w{1,2,3}.weight` — 43 layers x 256 experts,
exactly what the README documents. None of them match `EXPERT_RE`.

## What it looks like

`coli doctor --model <DeepSeek-V4-Flash-0731> --ram 28`, before:

```
[fail] memory.ram         RAM budget cannot hold one expert slot per sparse layer
[warn] placement.plan     RAM budget cannot hold one expert slot per sparse layer

model  48 shards · 166.9 GB
disk   0.0 GB cold experts · 46.4 GB free
RAM    28.0 GB budget · 166.9 GB dense · 3.7 GB runtime · 0.0 GB warm experts · cap 0/layer
limit  CPU expert compute (fully resident)
hit    100% projected expert residency
auto-tune:
  PIN_GB=all          enough RAM for full expert residency
result error
```

Identical output at `--ram 22`, `--ram 31` and `--ram 33` — the budget makes no
difference, which is the tell that nothing is being classified as an expert.

After widening the regex, same machine, same command:

```
[  ok] memory.ram         RAM budget is viable
[warn] placement.plan     cold expert misses may reach disk; ...

disk   143.5 GB cold experts · 46.4 GB free
RAM    28.0 GB budget · 19.7 GB dense · 4.6 GB runtime · 3.7 GB warm experts · cap 6/layer
limit  disk expert misses
[...]

PR #1133, opened 2026-08-20
fix(family_registry): accept compress_ratios longer than num_hidden_layers

QQSHI13:
## Problem

`coli doctor` fails with:

```
[fail] model.shards       deepseek_v4: compress_ratios must be a list matching num_hidden_layers
```

The DS-V4-Flash model (e.g. `ds-v4-flash-0731`) ships `compress_ratios` with 46 entries for `num_hidden_layers = 43`. The C engine (`deepseek_v4.c:14443`) only requires `compress_ratio_count >= num_hidden_layers` and ignores trailing extras, but the Python planner (`family_registry.py` `_dsv4_geometry`) demanded an exact `len == num_hidden_layers` match, breaking every doctor/plan run for this model.

## Fix

- Relax the check from exact match to `len(ratios) >= num_hidden_layers`.
- Iterate only `ratios[:layers]` when projecting compressor/indexer memory, mirroring the C engine's per-layer reads.

## Verification

- `coli doctor`: `model.shards` now passes; full suite of family_registry tests still green (35 passed).

### DIFF
--- a/c/family_registry.py
+++ b/c/family_registry.py
@@ -364,17 +364,17 @@ def _dsv4_geometry(config, context, _model_dir):
     index_hd = _required_int(config, "index_head_dim", "deepseek_v4")
 
     ratios = config.get("compress_ratios")
-    if not isinstance(ratios, list) or len(ratios) != layers:
+    if not isinstance(ratios, list) or len(ratios) < layers:
         raise ValueError(
-            "deepseek_v4: compress_ratios must be a list matching "
-            "num_hidden_layers")
+            "deepseek_v4: compress_ratios must be a list of at least "
+            "num_hidden_layers entries")
 
     # Fixed window ring: n_layers * window * head_dim fp32.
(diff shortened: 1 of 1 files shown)


JustVugg:
Merged. Verified the mismatch against both sides before landing it: `deepseek_v4.c` gates on `layer >= config->compress_ratio_count` (so any count at or above `num_hidden_layers` works and trailing entries are never read), while `family_registry.py` required exact equality. The engine is the source of truth here and the planner had drifted from it.

The detail worth calling out is your second change: validating `ratios[:layers]` rather than the whole list. Relaxing only the length check would have left `coli doctor` rejecting a healthy checkpoint over a malformed value in an entry the engine never reads. That is the careful version, and it is the reason this went in without a round of review.

Third planner-vs-engine disagreement this week (#1110, #1117, now this), all with the same shape: Python re-stating a rule the C already enforces, more strictly. Worth a look at whether the remaining geometry checks share it.

PR #1134, opened 2026-08-20
Fix RSS guard memory measurement and reclamation under Linux

tonnthuir:
## Summary

The rss_guard code measures the peak memory usage under Linux, so it will see excess memory usage forever once the peak exceeds the cap. This has changed to measure the current RAM usage.

The rss_guard also freed memory, but under Linux this memory was still being held by the program. Add a malloc_trim(1kB) to run once inside rss_guard to actually free the memory and return it to the system.

All changes are inside #ifdef __linux__ blocks so that they won't change behavior on other systems.

## Validation

Program compiles and runs. Produces the same output when run with the same inputs (temperature=0 and fixed seed).

Memory usage was tracked. When memory usage exceeded the RAM_GB environment setting, rss_guard reduced the size of the LRU cache and the reported memory usage dropped. The memory usage reduction persisted and further calls to rss_guard did not continue reducing the LRU cache.

## Compatibility

All changes are inside linux ifdef blocks. I do not have a non-linux test system to verify if rss_guard has problems on other platforms. I can see other platform-dependent code paths in the same area and I did not change them.

rss_gb() returns the peak memory usage under Linux. I added a function current_rss_gb (only defined for Linux systems) to return the current active memory usage. rss_gb() is called in other parts of the code and should possibly return current usage instead of peak usage everywhere, but I did not investigate which value is intended for other rss_gb() calls.


ZacharyZcR (commented):
Reviewed from the neighbouring change (#1121, the ecap-respect fix) — checked the interaction first, then the patch itself.

**Interaction with #1121: orthogonal and complementary, no conflict.** The two fixes close the two halves of the #1034 loop independently: #1121 stops the cache *regrowing* past a lowered cap; this PR stops the guard *misfiring forever* (peak-based measurement) and makes the freed bytes actually leave the process (malloc_trim). Your diff context already sits on top of the ecap change, and nothing here touches victim selection.

**The peak-vs-current diagnosis is verified in-tree**: `rss_gb()` on Linux reads `getrusage().ru_maxrss` (colibri.c:774-779), which is the high-water mark — so once the peak crosses the budget, every subsequent guard pass sees 'over budget' regardless of what was freed, walks the cap down, and evicts uselessly every 16 tokens. Exactly the '#403 keeps firing' behaviour from #1034's logs. Good catch.

Two notes, one worth changing:

1. **`current_rss_gb()` returning -1.0 on failure silently disables the guard**: -1 always passes the `rss <= lim*1.02+0.3` early-out, and the guard exists to prevent OOM kills — failing silent-open inverts its purpose. Suggest falling back to `rss_gb()` instead of returning -1: over-guarding on a peak figure is the lesser evil than never guarding, and it preserves today's behaviour in the (unlikely) /proc-unreadable case.

2. **macOS has the same peak bug and keeps it** — `ru_maxrss` there too (:776). Your linux-only scoping is honest and right given you can't test it, just flagging for the maintainer that the same one-liner shape (mach `task_info` `resident_size`) exists when someone with the hardware wants it.

Also worth saying: `malloc_trim` targeting the small allocations is the right call — the >128KB slabs are mmap'd and already return on free (the comment above rss_guard says so), so what you observed being retained was the small-block arena, which trim is precisely for.

JustVugg:
Both halves of this are real, and the diagnosis is better than the issue tracker had. Worth stating plainly since it is your first contribution here:

- `rss_gb()` uses `getrusage(RUSAGE_SELF).ru_maxrss`, which is the **high water mark**, not current usage. Once the process peaks above the cap — which happens naturally during load — the guard sees that peak on every subsequent call and keeps evicting forever. It cannot recover, and no amount of freeing helps.
- `free()` on glibc returns memory to the arena, not to the kernel, so the guard's evictions did not move RSS either. `malloc_trim` is the right instrument, and guarding it with `#ifdef __linux__` is the right scope.

**One thing before this can go in, and it is two lines.** The failure path opens the guard rather than closing it:

```c
double rss = current_rss_gb();       /* -1.0 when fopen fails */
if (rss <= lim*1.02+0.3) return;     /* -1.0 <= anything, so: return, every time */
```

If `/proc/self/status` is unreadable, `rss_guard` becomes a no-op for the rest of the run. This matters because of what the guard is for: #403 exists so the process does not get OOM-killed, and the current behaviour on that path is over-eviction (annoying, survivable) while the new one is no guard at all (OOM). Rare as an unreadable `/proc` is — hardened containers, some `hidepid` setups — a safety mechanism should degrade toward being conservative, not toward being absent.

The fix is to fall back rather than surrender:

```c
double rss = current_rss_gb();
if (rss < 0) rss = rss_gb();          /* peak-based, but a live guard beats none */
```

That makes the change strictly better than the status quo in every environment instead of most of them. While you are in there, the `[RSS] failed to open` line fires on every call (every 16 tokens), so it wants a `static int said;` latch like the other one-shot notices in that file, or a healthy run in an odd container floods stderr.

CI is green and the rest of the change is exactly right, so this is a small revision rather than a rethink. Ping when it is pushed and it goes in the same day.

Also: I retargeted this from `main` to `dev` — every PR here targets `dev`, and `main` only moves through a release PR. Nothing you did wrong, the contributing docs could say it louder.

tonnthuir:
I adjusted current_rss_gb to fall back to rss_gb if it can't measure the current memory usage. I think it is a little cleaner that way instead of having the calling function check the return value to see if it needs to call rss_gb. It also gives a negligible performance boost by removing a branch from the standard code path, but that isn't a reason to choose one way over the other.

It will also only print a failure message once. There are two print statements so that either failure (can't open proc or can't find the line in the file) will add a line to the log file stating that peak memory usage is being monitored. They use the same latch, so only one of them will be printed.

JustVugg:
Merged. Both revisions are in and both are right:

```c
return rss_gb();   /* peak-based, but a live guard beats none */
```

and the `announced_proc_failure` latch so the notice fires once rather than every sixteen tokens.

The underlying bug is the valuable part and worth restating for anyone who finds this later: `rss_gb()` reads `getrusage(RUSAGE_SELF).ru_maxrss`, which is the **high water mark**. Once the process peaked above the cap — which happens during load, by design — the guard saw that peak on every subsequent call and kept evicting forever, with no way to recover no matter how much it freed. And the freeing itself did not move RSS either, because glibc returns memory to its arena rather than to the kernel; `malloc_trim` is what actually gives it back.

So the guard was measuring the wrong number *and* its remedy was invisible to the number it was measuring. Two independent faults in one mechanism, both found from the outside.

Good first contribution, and thank you for taking the fail-open note rather than arguing it — the guard exists so the process is not OOM-killed (#403), and a safety mechanism that degrades to absent is worse than one that degrades to conservative.

PR #1142, opened 2026-08-20
attn: tile the prefill k/v reconstruction past a ceiling (#768 cause 2)

ZacharyZcR:
Addresses **cause 2** of #768 — the `kvb_all` materialisation — per the maintainer's triage there ("cause 2 removes one"; cause 1 belongs to #553's KV8 work).

## The problem, as the issue quantified it

`kvb_all` is `Tk × H × (qk_nope + v_head)` floats — **30.1 GB at ctx 262144** — allocated per attention call during prefill and freed after, but `cap_for_ram()` reserves the full amount **permanently** at every context length. @ThefloorMiner's refinement is the key fact: `kvb 0.000s` during decode — the buffer exists *only* in prefill, yet the reservation is resident forever, and it is what drives the expert cache to its cap=1 floor at long context (the 5.46 → 0.86 tok/s cliff).

## What this does

Above `KVB_FLASH_MB` (default 2048 MB), the reconstruction is tiled: rebuild `kv_b` for `KVB_TILE_MB` (default 512 MB) worth of tokens at a time, and fold scores/values through an online (flash-style) softmax with running (max, sum) per (row, head), accumulating into the existing `ctx` buffer.

- **Rebuild total unchanged** — still exactly one matmul pass over the context, just in tile-sized calls (`matmul_qt` is row-independent).
- **Peak transient**: `Tk*kvb_dim*4` → one tile + 2 floats per (row, head). At ctx 262144: 30.1 GB → ~0.6 GB.
- **`cap_for_ram` reserve mirrors the same trigger** (same env, same default), so the freed ~28 GB goes back to the expert cache — which is the throughput half of the issue.

## Numerics, stated plainly

Scores and values keep their exact t-order; the only change is that softmax normalisation is applied incrementally instead of after a full-row max/sum. So the tiled path is **not bit-identical** to one-shot — it is the same divergence class as the CUDA/Metal attention arms and the DSA gather path (#510): rounding-level, can flip a near-tie argmax. Three honest mitigations:

- It only engages past the 2 GB ceiling, where the one-shot path is currently either reserving tens of GB or OOMing — small contexts are byte-identical to today.
- `KVB_FLASH_MB=0` restores the old behaviour everywhere; `KVB_FLASH=0/1` is the per-run A/B switch.
- A one-line `[ATTN]` notice prints when tiling engages, so a changed output is attributable.

**DSA rows keep the one-shot path.** The top-keep list is emitted in two bands (`> thr` then `== thr`, each in position order — colibri.c's partial-select comment), so it is not tileable by t-range without reordering the accumulation. `dnsel` present → old path, explicitly. A gather-rebuild for the DSA case (rebuild only the selected rows — the same 30 GB is allocated there today for a top-k read) would be the follow-up.

## Verification

- `glm_tiny` TF oracle: **32/32 token-exact** under the tiled path — single-tile, and 8-tile with the row floor lowered locally so every causal-clipping boundary is crossed. (Exactness on tiny is expected — margins there are wide; the honest claim is "rounding-class divergence", verified structurally, not "bit-exact".)
- `make test-c` green, zero-warning build.
- What I could not do on this box: a real long-context prefill A/B on GLM-5.2 744B. @ThefloorMiner's 4× A6000 host with the ctx-32768/262144 protocol from the issue is exactly the machine — the prediction is: identical RSS behaviour below 2 GB of kvb, ~tile-sized transient above it, and `cap=` in the `[RAM_GB]` line rising at ctx 262144 by roughly `28 GB / row_bytes` slots.

Relation to the rest of #768: cause 1 (fp32 width) stays with #553; #795 (ragged KV pages) already covered the CUDA serving-growth slice. This PR is only the transient-materialisation half, which the triage ranked the better buy.


ZacharyZcR:
Second commit closes the follow-up flagged in the description: **the DSA case is now covered too, and unlike the dense tiling it is bit-identical.**

Under DSA the loops read only each row's top-keep list, but the one-shot matmul still rebuilt all Tk rows — the same 30 GB paid for a top-k read. Past the same `KVB_FLASH_MB` ceiling the rebuild now gathers the **union of selected rows** (compact `kvb_map`, rows with `ns==0` force their full causal range in). Per-row rebuild is independent and the read order along each tlist is unchanged, so output is byte-for-byte the one-shot path's — the two-band tlist order that blocked t-range tiling is irrelevant to a gather. Falls back to one-shot when the union doesn't at least halve the buffer, or on allocation failure.

Coverage status, stated precisely: the dense paths are regression-verified here (tiny TF oracle 32/32 in both modes, `make test-c` green); **the gather branch itself has not executed on this box** — it needs a DSA-active checkpoint, which this machine cannot run. The bit-exactness argument is structural (same rows, same order, same kernel), but a real GLM-5.2 long-context run on the 4×A6000 host would turn that argument into a measurement, ideally alongside the dense-path protocol from the issue thread.

Issue #1143, opened 2026-08-20
feat(kimi_k3): wire up tool calling — XTML format verified against Moonshot's reference renderer, design for review


ZacharyZcR:
Claiming the gap named in #1029: Kimi K3 is the only flagship-class engine still returning `400 Tool use is not wired up`, and it is the one whose upstream model is explicitly agentic. I've done the format archaeology and the code-path survey; this issue is the design write-up for review before the PR, because one wire-level decision deserves a maintainer call.

## The format (verified against Moonshot's reference renderer)

K3 has no `chat_template.jinja` — the normative renderer is `encoding_k3.py` in the checkpoint repo. Tool use is pure XTML over the existing four special tokens (`<|open|>`, `<|close|>`, `<|sep|>`, `<|end_of_msg|>`); every tag and attribute is ordinary text, which means `kimi_k3.c`'s existing `cb_open/cb_text/cb_close` primitives already cover the alphabet:

1. **Declaration** — one system message, JSON schema in a fenced block:
   `<|open|>message role="system" type="tool-declare"<|sep|># Tools\n…\n```json\n{compact}\n``` <|close|>message<|sep|><|end_of_msg|>`
2. **Assistant calls** — after the `response` channel closes:
   `<|open|>tools<|sep|>` then per call `<|open|>call tool="{name}" index="{n}"<|sep|>`, per argument `<|open|>argument key="{k}" type="{string|number|boolean|null|object|array}"<|sep|>{text}<|close|>argument<|sep|>`, closes in kind. (String args carry the decoded string; non-strings keep their exact JSON literal — the renderer's `_parse_arguments_object` is one-level-deep on purpose. A `json type="object"` block is the fallback for unparseable argument strings.)
3. **Results** — `<|open|>message role="tool" tool="{name}" index="{n}"<|sep|>{content}<|close|>message<|sep|><|end_of_msg|>`, with results re-sorted into `tool_calls` order by `tool_call_id` (the reference does this normalization; we should too).
4. **`tool_choice`** — `required` / `none` are internal system messages of `type="tool-choice"` with fixed English bodies; a forced function narrows the declared set (same shape `render_chat_v4` already implements).

## The three layers to change

[...]

Issue #1136, opened 2026-08-20
[Bug]: DeepSeek V4 greedy decode is not reproducible across runs — .coli_usage changes the generated tokens


Crowmen007:
### Commit or version

`4ef9a9920dd4290b4ac26f74e73913fd18379bc9`

### Environment

Ubuntu 24.04, Linux 6.8.0-111-generic, AMD Ryzen 9 5950X (16C/32T, Zen 3, AVX2, no AVX-512), 62.7 GiB RAM, Samsung SSD 980 1TB NVMe (PCIe 3.0 x4, DRAM-less), ext4 on LVM, gcc 13.3.0, Python 3.12.3. CPU-only `make deepseek-v4`, no GPU backend.

Model: `deepseek-ai/DeepSeek-V4-Flash-0731`, official checkpoint, all 74 files verified against the HF manifest by size and sha256.

### Reproduction steps

Run the same prompt twice through the same binary, with the engine left to manage `.coli_usage` as it normally does:

```bash
cd c
P="I had 17 apples. I gave a third to friends, then ate 2. How many are left? Think step by step."
for i in 1 2 3; do
  python ./coli run --model <model_dir> --ram 32 --ngen 256 "$P" 2>&1 \
    | sed -n "/^generated_text=/,/^timing /p" | head -n -1 > /tmp/run_$i.txt
done
md5sum /tmp/run_*.txt
```

To isolate `.coli_usage` as the variable, take a snapshot once and restore it before each run:

```bash
cp <model_dir>/.coli_usage /tmp/usage_snapshot
for i in 1 2 3; do
  cp /tmp/usage_snapshot <model_dir>/.coli_usage
  python ./coli run --model <model_dir> --ram 32 --ngen 256 "$P" 2>&1 \
    | sed -n "/^generated_text=/,/^timing /p" | head -n -1 > /tmp/fixed_$i.txt
done
md5sum /tmp/fixed_*.txt
```

### Expected behavior

Identical answers. Decode is greedy on this engine, and the expert cache is a performance mechanism — which experts happen to be resident should not change the tokens that come out.

`docs/ENVIRONMENT.md` states this for the I/O options that reorder work:

> \| `PIPE` \| `0` (off) \| Overlap expert disk-load with matmul via I/O worker threads. **Byte-identical output**; reorders I/O. `PIPE=1` opts in. \|

and `PILOT_REAL` is described as "Value-preserving real cross-layer prefetch loads".

### Actual behavior and logs

[...]

PR #1144, opened 2026-08-20
feat(kimi_k3): wire up tool calling — XTML per Moonshot's reference renderer (#1143)

### LINKED ISSUE #1029: Function Calling Does Not Yet Support?
Am i run with wrong model or Colibri doesn't yet support the function calling capabilities?

ZacharyZcR:
Implements #1143 (closes #1029's last flagship gap), option **(a)** from the design write-up — literal re-emission, GLM-precedent marker suppression. If (b) is preferred I'll rework the output side; everything else is independent of that choice.

## What's in

**Format** — verified against `encoding_k3.py` in the Kimi-K3 checkpoint repo (the normative renderer; K3 has no jinja template). All four surfaces: `tool-declare` system turn, `tools > call > argument` blocks with per-argument XTML types (non-strings keep their exact JSON literal — `1e2` stays `1e2`, one-level-deep parse mirrored from the reference), `role="tool"` result turns re-sorted into `tool_calls` order by `tool_call_id` (the reference's normalization), `tool_choice` required/none/forced.

**Gateway** — `K3CHAT1` grows typed records (`Y`/`O`/`B`/`F`/`V`/`J`) instead of pre-rendered text: K3's rank-BPE segment boundaries are the tokenizer contract, so tags keep being constructed engine-side. `parse_k3_tool_calls` + the stream-marker machinery handle the return path, with the GLM path's unclosed-block recovery posture.

**Engine** — `chat_build_wire` parses the records through the existing `ChatB` primitives plus `cb_attr` (attribute segmentation matches the reference exactly: `" key"`, `'="'`, escaped value, `'"'` each their own segment; `&`/`"` → `&amp;`/`&quot;`). The serve loop re-emits **tool-structural runs only** literally; every other XTML run stays suppressed exactly as before, so non-tool traffic is byte-identical to today.

## Tests (three layers)

- `test_k3_chat_tools.c` — golden: records rendered against the tiny Kimi tokenizer (which now carries the four XTML tokens) must reproduce the reference byte stream exactly, incl. attr escaping and malformed-record refusal. 7/7.
- Gateway units — declaration/choice modes, call history, argument typing, json fallback, id-based result re-sort, generated-call parsing, unclosed-tail recovery. 11/11 (kimi-filtered).
- `test_openai_tools_k3_e2e.py` — real gateway subprocess vs mock SERVE engine: both response shapes, chunk-straddling stream suppression (markers split across 25 DATA frames), `tool_choice=none` omits the declaration. 4/4.
- Full regressions: `test_openai*` 163/163, Anthropic surface 28/28 (kimi moved from the reject-list test to a render assertion), `make test-c` green, zero-warning builds.

## Stated limits

- **Not executed against the real 2.8T checkpoint** — the e2e layer is engine-mocked by design (same as the GLM suite). A real-model smoke on a K3-capable host is the honest final gate before this ships in a release; happy to coordinate.
- `tool_choice=none` follows the V4 gateway precedent (tools simply not offered) rather than the reference's MUST-NOT message — noted in #1143, trivially changeable if you'd rather mirror the reference.
- docs/api.md matrix updated; ENVIRONMENT.md untouched (no new env vars).

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1102,6 +1102,9 @@ tests/test_inkling_serve_framing$(EXE): tests/test_inkling_serve_framing.c inkli
 tests/test_kimi_serve_framing$(EXE): tests/test_kimi_serve_framing.c kimi_k3.c kv_prefix.h serve_codec.h st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h quant.h omp_tune.h route_trace.h
 	$(CC) $(NOCUDA_CFLAGS) $< $(VK_OBJ) -o $@ $(NOCUDA_LDFLAGS)
 
+tests/test_k3_chat_tools$(EXE): tests/test_k3_chat_tools.c kimi_k3.c kv_prefix.h serve_codec.h st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h quant.h omp_tune.h route_trace.h
+	$(CC) $(NOCUDA_CFLAGS) $< $(VK_OBJ) -o $@ $(NOCUDA_LDFLAGS)
+
 # olmoe's matmul_q, not colibri's: compares the IDOT path against FP32
 # ACTIVATIONS rather than an integer reference. NOCUDA_* for the same reason
 # the olmoe target uses it -- olmoe.c contains no COLI_CUDA code.
--- a/c/kimi_k3.c
+++ b/c/kimi_k3.c
@@ -2155,6 +2155,41 @@ static void chat_assistant(ChatB *b, const char *reasoning, const char *text){
     cb_close(b,"message"); cb_special(b,b->sp_eom);
 }
 
+/* ---- tool-call XTML (#1143) -----------------------------------------------
+ * Attribute segmentation mirrors the reference renderer (encoding_k3.py):
+ * " key", "=\"", the escaped value, and "\"" are each their OWN text segment.
+ * Segment boundaries are token boundaries under K3's rank-BPE, so the split
+ * is part of the format, not a style choice. Values escape & -> &amp; and
+ * " -> &quot;, exactly the reference's _escape_attr_value. */
+static void cb_attr(ChatB *b, const char *key, const char *val){
+    char kb[80]; snprintf(kb,sizeof kb," %s",key); cb_text(b,kb);
+    cb_text(b,"=\"");
+    size_t n=strlen(val), extra=0;
+    for(size_t i=0;i<n;i++) extra += val[i]=='&'?4u : val[i]=='"'?5u : 0u;
+    char *esc=malloc(n+extra+1);
+    if(!esc){ fprintf(stderr,"OOM chat attr\n"); exit(1); }
+    char *w=esc;
+    for(size_t i=0;i<n;i++){
+        if(val[i]=='&'){ memcpy(w,"&amp;",5); w+=5; }
+        else if(val[i]=='"'){ memcpy(w,"&quot;",6); w+=6; }
+        else *w++=val[i];
+    }
+    *w=0; cb_text(b,esc); free(esc);
+    cb_text(b,"\"");
+}
+/* open tag with caller-supplied attributes: cb_open_begin, cb_attr..., cb_end */
+static void cb_open_begin(ChatB *b, const char *tag){ cb_special(b,b->sp_open); cb_text(b,tag); }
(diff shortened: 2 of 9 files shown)


JustVugg:
Timing: you opened this at 13:08 and my (b) call on #1143 landed after, so you wrote (a) exactly as you said you would and this is not you ignoring a decision. The offer to rework the output side is appreciated, and yes — **I am holding to (b)**, with two caveats that matter given how today has gone.

**First caveat, and you should weigh it:** I was confidently wrong twice in the last hour on #1136, and one of those wrong calls sent you to rewrite correct documentation into incorrect documentation. So do not treat this call as authoritative because it is mine. If the reasoning below does not survive contact with the codec, say so and I will move.

**The reasoning, restated so you can attack it.** The engine sees `<|open|>` as a token id — it has ground truth about whether a structural marker was generated by the model or merely typed. Option (a) converts that to bytes and asks the gateway to re-derive it by pattern matching, which destroys the distinction at exactly the boundary where it matters most. For GLM the consequence of a spoofed marker is wrong text on a screen; for K3, whose whole point is agentic use, it is **an action that runs**. GLM carrying this exposure is a reason to fix GLM, not to extend the pattern to the engine where the blast radius is largest.

The cost objection you raised in the design write-up was real when you wrote it and is not anymore: `serve_codec.h` now owns both directions (`coli_serve_write_data` and siblings), every engine's migration landed behind a byte-exact wire-transcript freeze, and Inkling has already ridden an engine-specific structured payload on it (#1116, audio as an opaque extension). Adding a record is one edit in one file with an existing test that proves nothing else moved — not the five-way protocol surgery it would have been last week.

**And the part I am not asking you to carry:** GLM migrating to the same mechanism is ours. You are not fixing GLM's hole as the price of doing K3 correctly; you are declining to copy it, and we will make the two symmetric afterwards.

**On the rest of this PR, nothing changes and it is good work.** The format side is the part I cannot check without the checkpoint and the part most likely to rot silently, so going to `encoding_k3.py` and mirroring its behaviour — including `1e2` staying `1e2`, the one-level-deep argument parse, and the `tool_call_id` re-sort — is what makes this trustworthy. Typed `K3CHAT1` records rather than pre-rendered text is right for the rank-BPE boundary reason you gave. `tok_kimi_tiny.json` plus the mock-engine e2e is the correct testing shape for a 2.8T model nobody here can run.

Since you say the rest is independent of the output choice: if the rework turns out to be larger than a day, split it — land the format, gateway records and tests behind a flag that keeps the 400, then flip it on with the sideband. Better than one PR ageing while dev moves under it, which is how #790 and #601 got where they are.

Ping me when it is up and I will review the codec change the same day.

JustVugg:
**Merged — and I owe you an apology, because I blocked this on a claim I had not verified.**

I told you that with option (a) a spoofed marker on K3 "is an action that runs". That is false, and I should have checked before writing it. `parse_tool_calls` returns `(content, tool_calls)`; the gateway *reports* the call in the OpenAI response and never executes anything — the only `subprocess.Popen` in `openai_server.py` launches the engine. The client application decides what to run. The real exposure is "we tell a client the model requested something it did not", which is a genuine trust-boundary problem but not the category I invoked, and my whole argument rested on that category.

The second thing I got wrong is the one that actually flips the decision. **GLM has shipped this exact mechanism for months.** So (a) on K3 does not open a new class of exposure, it extends an existing one — and if K3 had gone to the sideband while GLM kept markers, the gateway would carry **two mechanisms** until GLM migrated. This repo has already paid for that shape once: byte framing duplicated five times is how Windows binary mode silently disappeared from sibling engines (#748). Doing the sideband once, for both engines together, is better than doing K3 specially and retrofitting GLM afterwards.

So the sideband work becomes ours and covers both engines. You are not reworking anything.

To be explicit about the sequence, since I sent you back and forth today: you wrote (a) as announced, I asked for (b) on a bad premise, you offered to rework rather than argue, and the rework turned out to be unnecessary — the right call was the one you made at the start. The 852 lines here are unaffected by any of it: the format work against `encoding_k3.py`, the typed `K3CHAT1` records, `tok_kimi_tiny.json` and the mock-engine e2e all stand on their own, and the format side in particular is the part nobody could have checked without going to the reference renderer the way you did.

#1029 closes with this: Kimi K3 was the last flagship engine returning `400 Tool use is not wired up`.

PR #1145, opened 2026-08-20
docs: promote COLI_V4_ROWS16 to an ENVIRONMENT.md table entry (#1136)

ZacharyZcR:
Item 1 of the two interim actions from your #1136 verdict: `COLI_V4_ROWS16=0` now has a proper table entry in ENVIRONMENT.md's V4 section — default, mechanism, the run-to-run consequence in bold, and the 'set =0 for any quality A/B; throughput A/Bs do not need it' instruction verbatim. The surrounding recipe paragraph (from #1138) is reworded to state it is a known defect under the post-IDOT house rule with a convergence fix planned, not a documented trade-off — so nobody reads it as working-as-intended again.

The convergence fix itself (reference path adopting rows16's accumulation order, K1-planar-tail shape) I'm starting on next per the same thread.

### DIFF
--- a/docs/ENVIRONMENT.md
+++ b/docs/ENVIRONMENT.md
@@ -367,15 +367,22 @@ the ones you are most likely to set: `DSV4_CUDA` (GPU tier on/off),
 `COLI_V4_SAVE_USAGE=0` is an engine-specific alias that disables only V4's
 usage rewrite; the shared `USAGE_SAVE=0` covers this engine too.
 
-**Reproducible greedy runs (#1136):** on this engine, greedy text legitimately
-varies with the expert-cache state — hot experts run the vectorized `rows16`
-kernel, cold ones run the reference matvec, the two accumulate in different
-orders, and which experts are hot follows the autopin history (`.coli_usage`,
-rewritten by every run). For byte-identical output across runs, either freeze
-the history (`USAGE_SAVE=0`, after seeding it once) or remove the variable
-entirely (`COLI_V4_ROWS16=0 COLI_V4_AUTOPIN=0 USAGE_SAVE=0`: reference kernels
-only, no history). Details in
-[deepseek-v4.md — CPU-only behaviour](deepseek-v4.md).
+| Variable | Default | Effect |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


JustVugg:
Right shape, right instinct, and the recipes at the bottom are correct — but hold it a moment: you wrote this against my first comment on #1136, and I have since corrected that comment. The mechanism in this text is the one I got wrong, and I would rather not enshrine it in `ENVIRONMENT.md`, where it is worse than in a code comment because users read it as settled.

The text says, three times, that the two paths "accumulate in different orders". They do not. Having now read both kernels:

```c
/* cold: coli_fp4_matvec_ref */
coli_fp8_activation_qdq_ref(activation, activation_scales, input, columns, 128);
matmul_mxfp4(output, activation, weight->data, weight->scales, 1, columns, rows);

/* hot: coli_fp4_matvec_rows16_v10 */
product = _mm512_mul_ps(_mm512_mul_ps(_mm512_set1_ps(activation[column]), values), scale);
```

The cold path runs an **fp8 quantize-dequantize over the activations** that the hot path skips. That is a precision difference, not a rounding-order one, and it changes what a reader should take away: **the hot path is the more accurate of the two**, so pinning more experts does not just change the output, it improves it. The `deepseek_v4.c:7650` comment we both trusted is wrong about its own bug — my fault for repeating it rather than opening the kernels, which is exactly the check this repo tells itself to do.

Two consequences for this PR:

1. **The explanation needs rewriting** — something like: *hot-pinned experts run the vectorized `rows16` kernel on raw fp32 activations, while cold experts go through the reference matvec, which first quantize-dequantizes the activations to fp8. The two therefore compute at different precision, and which experts take which path follows the expert-cache state.*
2. **"a convergence fix (reference path adopting rows16's accumulation order) is planned" over-commits.** The direction is genuinely undecided: the GPU tier and rows16 both use raw fp32 and only the CPU reference inserts the QDQ, so two of three already agree — but V4 uses fp8 activations natively, so that QDQ may be deliberate emulation of the upstream reference rather than an oversight. Better to say a convergence fix is planned and that which side the three paths converge *on* is pending a comparison against the HuggingFace reference implementation.

Everything else stands, including the part I would not have thought to write: naming it a defect rather than a trade-off, and citing #1044/#1080 as the house rule it violates. The `COLI_V4_ROWS16=0 COLI_V4_AUTOPIN=0 USAGE_SAVE=0` recipe is correct regardless of mechanism, since it forces every expert through one path.

Push the revision and it goes in immediately — this belongs in the env table and the sooner the better.

JustVugg:
**Disregard my review above — it was based on a wrong correction of mine, which I have now retracted on #1136.**

Both paths do run the fp8 activation qdq (`deepseek_v4.c:15411` in the rows16 matvec, :15530 in the dual), so the mechanism is exactly what you wrote and what the in-tree comment said: **accumulation order**. I misread the kernel through a grep filter that could not match the `qdq` lines, then asked you to rewrite correct text into incorrect text.

**This PR was right as originally written.** Nothing to change. Merging it now.

Issue #1140, opened 2026-08-20
[Discussion] Upstreaming the #161 CUDA stack: attention is now 80% of resident-host decode, and the 2.4→11.1 line deserves a continuation


ZacharyZcR:
### Context: why this line deserves to continue

@VincentMarquez's #161 stack — GPU MLA (D3), fused S=1 decode (`CUDA_FUSE`), the device expert tier under UNIFIED, CACHE_ROUTE, PLD multi-S — took a single DGX Spark from 2.4 to **11.1 tok/s at full top-8, hit 99.9%**, on GLM-5.2 744B. For calibration against the rest of the world: the NVFP4 vLLM fork for Spark needs **4 machines to reach 22–42 tok/s** single-stream (8 for 48), and on our own 8-GPU cluster GLM-5.2 single-stream decode sits around **~24 tok/s**. One $4k box at 11.1 against that background is a remarkable efficiency point, and it comes from exactly the levers this engine owns: residency, cache-aware routing, and speculation that only pays *because* residency made verify forwards cheap.

But as far as I can tell from #161's own framing, most of that stack is still **experimental / local, not merged** — upstream `COLI_CUDA_ATTN` covers S≤4 attention, and the fused-decode / GPU-MLA / multi-S verify work lives in Vincent's patches. Meanwhile #161's physics note is explicit about where the ceiling sits now:

> full-k8 S=1 is roughly a **~7 tok/s wall (MLA bandwidth + top-8 experts)**. 30 needs multi-S verify cost ≲ ~1.1× S=1 with sustained AL ≳ 4, or a much faster S=1 baseline.

And the 11.12 PROFILE says the same thing from the other side: `expert-disk 0.466s | expert-matmul 0.137s | attention 9.275s` — **~80% of timed decode is attention**. The expert side of this engine is essentially done on resident hosts; the attention side is the whole remaining game.

### Questions for the two of you

1. **Is upstreaming the #161 CUDA stack on the roadmap?** GPU MLA + `CUDA_FUSE` + the multi-S verify path (residual staying on-GPU) are the pieces that turn "resident host" into "attention-bound host with a path to 30". If the blocker is review bandwidth or generality (GB10-specific assumptions, managed-KV interactions), knowing which would let others help carve it into mergeable slices.

[...]

Issue #161, opened 2026-07-14
Benchmark: NVIDIA DGX Spark (GB10) 121GB — 11.1+ tok/s full-k8 CUDA (PLD); was 2.4 / 3.33 chat


VincentMarquez:
Machine:** NVIDIA DGX Spark (`NVIDIA_DGX_Spark`) · Ubuntu 24.04 · Linux 6.17.0-1008-nvidia aarch64
**CPU:** 10× Cortex-X925 @ 3.9 GHz + 10× Cortex-A725 @ 2.8 GHz (20 cores)
**Memory:** 121 GB unified (CPU/GPU shared)
**GPU:** NVIDIA GB10 (sm_121), driver 580.126.09
**Disk:** NVMe (`/dev/nvme0n1p2`), model on local ext4
**Model:** GLM-5.2 colibri int4 · **int8 MTP heads** (sizes match matey-0 clone: 3527131672 / 5366238584 / 1065950496)
**Engine base:** `JustVugg/colibri` @ `a78a06f` + local experimental patches (CACHE_ROUTE / TRACE_EXPERTS / CUDA unified path)

## Disk (README protocol)

`c/iobench` · 19 MB × 64 reads · 8 threads · shard `out-00099.safetensors`:

| mode | measured |
|------|----------|
| buffered | **4.25 GB/s** |
| **O_DIRECT** | **9.69 GB/s** |

## How to read these numbers

Community table rows are usually **stock routing (full top-8)** with **decode tok/s** from the coli stats line after warm-up. This report has **two tiers**:

1. **Tier A — full top-8, no routing change** (config `num_experts_per_tok=8`)
2. **Tier B — experimental `CACHE_ROUTE`** (opt-in, not upstream default; ~14% expert substitution vs true top-8; cache-aware routing, arXiv:2412.00099 max-rank)

Also:
- **Decode tok/s** = generation window only (coli `decode tok/s`). **Overall** includes prefill — do not use overall for ranking.
- **`MTP=0` / `DRAFT=0`** for these speed cells.
- Chat used **`TEMP=0.7`**. Output on the primes prompt was the correct first-20 list every turn.
- Short context via **`:reset` each turn** (clears KV only; pin/LRU survive).

---

## Tier A — full top-8, stock routing (`TOPK` unset, no CACHE_ROUTE)

[...]

PR #1151, opened 2026-08-20
CPU KV cache quantization: KV8 (fp8 e4m3) + KV_TQ — #553 rebased onto dev

ZacharyZcR:
The #553 rebase, carried per the go-ahead there. **All three commits are @NeuralNotwerk's, authorship preserved through the rebase**; every resolution I made is listed below for his (or your) veto, per the review plan.

Why this matters for the thread that revived it: this is the "single highest-value unblock" from #1140 — fp32 KV is bandwidth the S=1 MLA kernel must move every token, and on GLM-5.2 it is 3.2× llama.cpp's footprint (#768 cause 1). KV8 cuts the latent cache ~3.9× at the cost of fp8 rounding, opt-in.

## What landed

- `KV8=1`: latent Lc/Rc in fp8 e4m3 + per-row f32 scale. `KV_TQ=3|4`: rotated-int4 / PolarQuant tier. Both CPU-only in this PR (CUDA/Metal f32-row paths guard themselves off), both opt-in, mutually exclusive. Disk persistence versioned with upgrade/reject/self-heal.
- One commit of #553 was **dropped**: the PROF=1 disk-throughput metrics — unrelated to KV quantization, and dev's PROF accounting has since been rebuilt. It can return as its own PR if still wanted.

## The semantic resolutions (the three the maintainer flagged, plus one new since)

1. **KV producer loop** (`attention_rows`): dev parallelized this loop meanwhile (CPU-only pragma gate, shadow-shrink inside the loop); #553 had hoisted the CUDA shadow-shrink out of the loop to enable its own `if(S>8)` parallel. **Dev's structure wins** — same goal, already merged — and the KV8/KV_TQ producers ride it as write branches, with the shadow-shrink hoisted above all three branches (every branch rewrites the row). The KV8 motivation (≈576 libm encodes per row per layer wants the OMP pool awake) is preserved by dev's `if(S>1)` gate, which is broader than the PR's `if(S>8)`.
2. **Metal fused paths** (decode + full-layer): dev's conditions grew the fmt allowlists and `g_metal_prefill`; #553 adds `!g_kv8 && !g_tq`. Merged as dev's full condition **plus** the quant guard — the fused kernels read f32 Lc/Rc rows that are not even allocated under quantized KV.
3. **kvb reconstruction** (`attention_rows`, the #1142 overlap — I wrote both sides of this one): dev now has the tiled/flash arm and the DSA gather arm reading f32 `Lc` rows directly; #553 dequantizes Lc into an f32 staging buffer before the one-shot rebuild. Resolution: **quantized KV keeps #553's one-shot + staging design** (`kv_quant` gates the flash/gather arms off), so each side keeps its original semantics untouched. Making the flash arm KV8-aware (per-tile dequant staging) is a clean follow-up — the 30 GB kvb transient is independent of KV width, so long-context + KV8 would eventually want both.
4. **Makefile/TEST_BINS/.gitignore**: dev now derives TEST_BINS from build rules precisely to kill this conflict class — the PR's hand-written list is dropped, the three kv test rules alone make them gates. dev's explicit .gitignore list (the #368/#500 discipline) wins over the PR's glob; the two new binaries are added to the list.

## Verification

- **Default path is byte-identical**: glm_tiny TF oracle **32/32** with KV8/KV_TQ unset — the rebase does not touch anyone who doesn't opt in.
- `KV8=1`: oracle 30/32; `KV_TQ=4`: 23/32 — the expected lossy near-tie flips, and exactly why both are opt-in.
- All four kv suites pass (fp8 exhaustive roundtrip/RNE, TQ rotation/roundtrip/distortion, disk v1/v2 upgrade-reject-self-heal, alloc f32+fp8); full `make test-c` green (431 ok), zero-warning build.
- Not measured here: the actual long-context speed win. That needs a host where KV dominates — @ThefloorMiner's 4×A6000 protocol (#768) or the GB10 fleet (#161) are the natural datapoints once this merges.

KV_TQ is included because commits 2 and 3 harden it and splitting it out would have meant rewriting @NeuralNotwerk's base commit; if you'd rather ship KV8 alone first, say so and I'll carve it.


JustVugg:
Rebased onto current `dev` and pushed to your branch — #1107 (logprob channel) landed while this was open and touches `c/decode_batch.h`.

**One conflict, purely additive**: dev's `COLI_SUBMIT_TOPK_MAX` define and your `coli_kv_row8` helper landed at the same point in the file. Union, both kept, no judgement involved. Build clean, zero warnings, and @NeuralNotwerk's authorship survives both rebases.

**On the resolutions themselves — all four read right to me, and the third is the one that matters.** Keeping quantized KV on #553's one-shot + staging design and gating the flash/gather arms off with `kv_quant`, rather than trying to make one path serve both, is the correct call: each side keeps the semantics it was written and tested with. Making the flash arm KV8-aware is a genuine follow-up rather than something skipped, and you are right that the 30 GB kvb transient is independent of KV width, so long-context plus KV8 eventually wants both.

Resolution 1 also looks right for a reason beyond "dev's structure wins": dev's `if(S>1)` gate is strictly broader than the PR's `if(S>8)`, so the KV8 motivation (≈576 libm encodes per row per layer wanting the pool awake) is *better* served by dev's version than by the original.

Dropping the PROF disk-metrics commit was right too — unrelated, and dev's PROF accounting was rebuilt since.

**What this still needs**, and it is not mechanical: @NeuralNotwerk's look at the four resolutions. They are his commits and his design, and the review plan on #553 said he would see them. Given how fast `dev` is moving — this PR went conflicting within hours of being opened, twice now — that ask has a clock on it.

@NeuralNotwerk: no rework is being asked of you, only a veto if any of the four is wrong. The full list is in the PR body above.

JustVugg:
Merged.

**@NeuralNotwerk — your veto is still open, and merging did not consume it.** Both paths are off by default (`g_kv8=0`, `g_tq=0`, and both env reads default to 0, which I checked rather than took from the description), so a wrong resolution is wrong in a path nobody executes unless they ask for it. That bounded the risk enough to stop the clock on rebases — this PR went conflicting twice in a few hours — without spending your review. If any of the four is wrong, say so and we change it with nothing shipped in the meantime.

**Why this one was worth unblocking rather than parking.** It is roadmap 3.2 and the first real cut at the thing #1140 identifies as the actual ceiling: on a resident host, attention is ~80% of decode (`expert-disk 0.466s | expert-matmul 0.137s | attention 9.275s`), and fp32 KV is bandwidth the S=1 MLA kernel has to move every token. Everything the K-series did this week improved the 0.137 seconds. This is the first change aimed at the 9.275.

@ZacharyZcR — the rebase was careful in the way that matters: you resolved four semantic collisions and wrote down the reasoning for each instead of picking whichever side applied cleanly, and you dropped the one commit that did not belong. The third resolution in particular — keeping quantized KV on #553's one-shot staging and gating the flash arm off, rather than fusing two designs into a path neither had been tested on — is the call I would have wanted and would not have insisted on.

Next: measurement. Nobody has yet run KV8 on a real long-context prompt, and #768's numbers (~57 GB avoidable on GLM-5.2, 5.46 → 0.86 tok/s at 262k) are the claim this has to meet.

PR #1152, opened 2026-08-20
fix(inkling): pass qgs at the Metal MoE callsites (#1150)

### LINKED ISSUE #1150: macOS Metal build: Inkling MoE callsites miss qgs
macOS `METAL=1` builds fail because Inkling still calls the Metal MoE API with the pre-`qgs` signature.

At `33e67a9c004b6e608d1f19dfbdcc20793377f94f`, `backend_metal.h` declares `qgs` after `fmt` for `coli_metal_moe_block_begin(...)` and `coli_metal_moe_block(...)`, but `c/inkling.c` omits it at the Inkling MoE callsites:

- shared-expert async path: `fmt=5`
- routed-expert async/sync paths: `fmt=1` or `fmt=2`

Observed errors:

```text
inkling.c:1484:77: error: too few arguments to function call, expected 18, have 17
./backend_metal.h:183:5: note: 'coli_metal_moe_block' declared here

inkling.c:1388:70: error: too few arguments to function call, expected 16, have 15
./backend_metal.h:198:21: note: 'coli_metal_moe_block_begin' declared here
```

Minimal fix: pass `0` for `qgs` at those Inkling callsites. `qgs` is only meaningful for `fmt=4`, so it is unused for `fmt=1/2/5`.

JustVugg:
Fixes #1150 (reported by @jerome-benoit, with the exact error output and the two failing lines — that is what made this a five-minute fix).

`backend_metal.h` declares `qgs` after `fmt` for both `coli_metal_moe_block` and `coli_metal_moe_block_begin`; `inkling.c` had not been updated, so `METAL=1` builds fail with `too few arguments`.

**The value is documented rather than inferred.** The header says: *"qgs is the fmt=4 group size (ignored, pass 0, for fmt!=4)"*. Inkling uses fmt=5 for the shared expert and fmt=1/2 for the routed ones — all non-4, so `0` at every site, and the kernel ignores it there anyway.

**Three callsites, not two.** The report lists `:1386` (shared, fmt=5) and `:1483` (routed sync). The async `coli_metal_moe_block_begin` in the routed path at `:1468` has the same signature and the same omission — it was invisible in the report because the compiler stops before reaching it.

Verified the CPU build stays clean (zero errors, zero warnings). The Metal path itself I cannot build here; CI's macOS job builds every engine, which is the gate that would have caught this in the first place had Inkling's Metal path been in it.

@jerome-benoit — if you have the Mac handy, a `METAL=1` build on this branch would confirm it before merge.

### DIFF
--- a/c/inkling.c
+++ b/c/inkling.c
@@ -1383,7 +1383,7 @@ static void moe(Model *m, Layer *l, int layer, float *x, int S, float *out) {
             }
         }
         sxoff[ns] = ns*S;
-        sh_h = coli_metal_moe_block_begin(ns, D, I, 5, sgp, sup, sdp,
+        sh_h = coli_metal_moe_block_begin(ns, D, I, 5, 0, sgp, sup, sdp,
                                           sscale, sscale, sscale,
                                           sxg, sxoff, snr, srows, srw);
         free(srows);
@@ -1466,7 +1466,7 @@ static void moe(Model *m, Layer *l, int layer, float *x, int S, float *out) {
                  * the round; shared_done stays set either way. */
                 if (base + cap >= npair && !sh_h) {
                     ColiMetalMoeHandle *h = coli_metal_moe_block_begin(
-                        nb, D, I, q4 ? 2 : 1, mgp, mup, mdp, mgs, mus, mds,
+                        nb, D, I, q4 ? 2 : 1, 0, mgp, mup, mdp, mgs, mus, mds,
                         mxg, mxoff, mnr, mrows, mrw);
                     if (h) {
                         double ts = now_s();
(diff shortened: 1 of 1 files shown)


jerome-benoit:
Hello, I''ve already built with a similar local patch,, it's fine. and works.

JustVugg:
Confirmed by CI rather than by argument: #1160 added a `METAL=1` compile pass to the macOS job, and on `dev` — which contains this fix — the step reports **success**. Inkling builds again with Metal, and the three arguments were the complete set: no further error was hiding behind the ones the compiler had room to report.

@jerome-benoit, your report is what made this findable, and it produced more than its own fix. Chasing it turned up the reason it lasted a month: the macOS job ran `make <engine>` with Metal off, so every line behind `COLI_METAL` was invisible to CI, while `metal-test` compiled only `backend_metal.mm` and not the engines calling into it. The signature changed in v1.6.0 (`19d4004`, 2026-07-24) and CI stayed green for four weeks with the build broken.

That gap is closed now, for 15 seconds of runner time per macOS job. The next signature drift fails the same day.

PR #1155, opened 2026-08-21
attn: KV8/KV_TQ awareness for the flash and gather kvb arms (stacked on #1151)

ZacharyZcR:
**Stacked on #1151 — only the last commit (`attn: KV8/KV_TQ awareness…`) is new.** Draft until #1151 merges; will rebase to a one-commit diff then.

This is the follow-up you named in the #1151 review: *'long-context plus KV8 eventually wants both.'* The rebase gated the flash/gather arms off under quantized KV so each side kept its tested semantics — right for the merge, but it left the two features unable to compose: the 30 GB kvb transient is independent of KV width, so a KV8 run at ctx 262144 still paid the one-shot buffer (and its cap_for_ram reservation went back to being the binding constraint the moment you turned KV8 on).

Both arms now dequantize the latent rows they need through the same per-row codecs the one-shot path uses:
- **flash arm**: one tile's worth of Lc rows staged to f32 (~8 MB at the default tile) before the kv_b matmul, per tile;
- **gather arm**: exactly the union rows it rebuilds, one row at a time;
- **roped-key reads**: the one-shot path's three-way branch inlined verbatim, so each representation keeps its exact accumulation order (an earlier draft used a helper returning a partial sum — rejected here because it would have changed the f32/TQ float order relative to the shipped inline branches).

The f32 paths are byte-identical to before — every new branch sits on `g_kv8`/`g_tq` only, and the `kv_quant` gates are simply removed.

**Verified** (glm_tiny TF oracle): f32 default 32/32 unchanged; KV8 and KV_TQ under `KVB_FLASH=1` score 30/32 and 23/32 — **identical to their one-shot scores** — in both single-tile and 8-tile runs (row floor lowered locally to force multi-tile causal clipping). `make test-c` green (440 ok), zero-warning build.

Same caveat as #1142/#1151: the long-context RSS behaviour this composes needs a big-KV host to demonstrate end to end — the arithmetic says a KV8 + flash run at ctx 262144 now holds ~12 GB of KV plus a ~0.6 GB reconstruction transient where fp32 one-shot held 47.7 + 30.1.

### DIFF
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -4015,6 +4015,23 @@ static unsigned metal_fused_layer_fmt_miss(const Layer *l){
     return miss;
 }
 
+/* #1151 follow-up: the flash/gather kvb arms under quantized KV.
+ * kv_lc_rows_f32 dequantizes latent rows [t0, t0+n) into an f32 staging buffer
+ * (the same per-row codecs the one-shot path uses); the roped-key reads
+ * inline the one-shot path's three-way branch so each representation keeps
+ * its exact accumulation order. */
+static void kv_lc_rows_f32(Model *m, int layer, int64_t t0, int64_t n, float *dst, Cfg *c){
+    if(g_tq){
+        int lbb=coli_kvq_row_bytes(c->kv_lora,g_tq_bits,g_tq_codec);
+        for(int64_t t=t0;t<t0+n;t++)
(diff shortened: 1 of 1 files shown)


ZacharyZcR:
#1151 merged — rebased onto current dev, now a clean one-commit diff, out of draft. All verification claims re-run post-rebase: f32 default 32/32 unchanged, KV8/KV_TQ flash scores identical to one-shot, test-c green (440).

PR #1157, opened 2026-08-21
perf(dsv4): give the in-flight layer the whole slot pool during prefill

JustVugg:
**Draft until the A/B on the real checkpoint finishes** — it is running now and the numbers go here when it does. Opening early because the reasoning is reviewable without them.

## The problem

Batched prefill is layer-major: `target_batch` sweeps the **entire prompt** through layer L before L+1 starts. But the expert cache is partitioned statically per layer (`layer_slots() = state->slots + layer * slots_per_layer`). So while L runs, the other 42 partitions are dead weight — roughly 12 GB on a 25 GiB budget — and L thrashes its own 23 slots against a union of 150–256 experts.

Measured, 1606-token prompt, 1 token generated:
```
expert_requests=70184 hits=4599 misses=65585 bytes=876828426240
65,585 misses / 11,008 distinct experts = 5.96 reads each
876 GB read where <=138 GB suffices
```

Causal confirmation, same 188-token prompt, only `V4_PREFILL_CHUNK` varied — halving the chunk adds a third sweep and the cost tracks it:

| chunk | chunks | misses | bytes | TTFT |
|---|---|---|---|---|
| 128 (default) | 2 | 8,904 | 119.0 GB | 617 s |
| 64 | 3 | 11,063 | 147.9 GB | 1140 s |

**The engine already knew.** From the CUDA tier's route-aware refill: *"the persistent bank is a perfect per-layer expert cache … every expert is read from the store AT MOST once per layer — the CPU union's LRU cache re-reads the same experts several times per layer under prefill's expert-major sweeps (measured ~4-6x the layer's expert bytes)"*. That guarantee exists only on CUDA. This is the CPU equivalent.

## The change

`coli_v4_expert_store_prefill_pool(store, layer)` hands the whole slot array to one layer; `layer < 0` restores the partition. `layer_slots()` returns the base and a new `layer_slot_count()` returns the width, so pooling lives in two functions and every scan is bounded correctly.

**989 slots against at most 256 experts per layer** — each expert read exactly once per layer, the same guarantee the CUDA bank gives.

## The four things that could have gone wrong, and what stopped each

1. **Decode must not pool.** `target_batch` and `target_token` are adjacent and share the text `coli_v4_layer_free(engine, &layer); if (result) return -1;`. My first patch matched **both** — an assertion on the match count caught it. Hooking decode would have been the worst possible outcome: every layer is touched once per token, so one shared LRU smaller than a token's working set (43 × 6 = 258 references, no reuse) degenerates to a 0% hit rate. Verified after: 4 call sites in `target_batch`, **0** in `target_token`.
2. **Slots carry no layer tag** — the partition made it implicit. In a pool, layer 3's expert 5 would be returned as layer 7's. Changing owner clears identities; slabs stay allocated, so a layer switch costs a scan, not free/malloc.
3. **`calloc` zeroes the struct**, and 0 is a valid layer — the pool would have been born open and owned by layer 0. `pool_layer = -1` is written explicitly at init.
4. **One scan computes its base by hand** rather than through `layer_slots()`. Widening its bound would have read past the array for every layer > 0. Left at the partition width, with a comment saying why; it is a telemetry read-out, so approximate residency during prefill is acceptable.

Also: the pool is closed on **all three** exits of `target_batch` (load failure, cancel/error, normal), because a leaked pool would hand decode the degenerate cache.

## Constraints from the report, honoured

- Cache policy only — routing, precision and accumulation order untouched, so output is token-identical **by construction**, not by luck.
- Lease contract respected: leased slots are skipped by the reset; one `release()` per `lookup()` unchanged.
- `V4_PREFILL_CHUNK` untouched — the 128 clamp is a kernel contract, not tuning.
- Decode untouched.
- No new peak RSS: slabs are reused, never additionally allocated.
- CUDA path untouched — it has the bank.

## Verification in flight

```
./deepseek_v4 <model> "<188-token prompt>" --max-tokens 1 --memory-gb 22
```
Primary metric is `misses / distinct experts`, target 5.96 → ~1.0, equivalently `bytes=` 119 GB → ~20 GB on this prompt. Judged on the deterministic counters, not wall-clock: run-to-run noise on this box is ~35% on decode and worse on TTFT.

Closes the CPU half of the gap the CUDA bank comment names.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -6987,6 +6987,15 @@ typedef struct {
     double matmul_sec; /* cumulative expert-forward compute time (#890): the phase the
                         * dashboard needs alongside disk_sec so it stops folding
                         * everything into "other". One shared instance per store. */
+    /* Prefill-scoped slot pooling (#1157). Batched prefill is layer-major:
+     * the whole prompt sweeps layer L before L+1 starts, so while L runs, the
+     * other layers' partitions sit idle and L thrashes its own slice against a
+     * union of 150-256 experts. -1 = off (per-layer partition, the decode
+     * shape); >=0 = every slot belongs to that one layer. Slots carry no layer
+     * tag -- the partition made it implicit -- so switching owner MUST clear
+     * the pool, which is why the field holds the owning layer rather than a
+     * flag. */
+    int pool_layer;
     uint8_t *ehit;     /* layers*experts_per_layer: experts routed in the current turn */
     uint8_t *eheat;    /* layers*experts_per_layer: cumulative routing selections, capped 63 */
 } V4ExpertStoreState;
@@ -7073,7 +7082,17 @@ static V4ExpertRecord *get_record(V4ExpertStoreState *state, ColiExpertKey key) 
 }
 
 static V4ExpertSlot *layer_slots(V4ExpertStoreState *state, int layer) {
+    if (state->pool_layer >= 0) return state->slots;   /* prefill pool: one region */
     return state->slots + (size_t)layer * state->slots_per_layer;
+}
+
+/* Slots reachable through layer_slots(): the layer's own partition normally,
+ * the whole array while the prefill pool is open. Every scan over a
+ * layer_slots() result must bound itself with this rather than
+ * slots_per_layer, or the pool is allocated and never searched. */
+static int layer_slot_count(const V4ExpertStoreState *state) {
+    return state->pool_layer >= 0 ? state->layers * state->slots_per_layer
+                                  : state->slots_per_layer;
 }
 
 static void fill_tensor_view(ColiTensorView *view,
@@ -7111,7 +7130,7 @@ static int lookup(ColiExpertStore *store, ColiExpertKey key,
     state->stats.requests++;
     V4ExpertSlot *slots = layer_slots(state, key.layer);
     V4ExpertSlot *slot = NULL;
-    for (int i = 0; i < state->slots_per_layer; i++) {
+    for (int i = 0; i < layer_slot_count(state); i++) {
         if (slots[i].slab && slots[i].expert == key.expert) {
             slot = &slots[i];
             state->stats.hits++;
@@ -7119,7 +7138,7 @@ static int lookup(ColiExpertStore *store, ColiExpertKey key,
         }
     }
     if (!slot) {
-        for (int i = 0; i < state->slots_per_layer; i++) {
+        for (int i = 0; i < layer_slot_count(state); i++) {
             if (!slots[i].references && (!slot || !slots[i].slab ||
                                          (slot->slab && slots[i].used < slot->used)))
                 slot = &slots[i];
@@ -7225,7 +7244,7 @@ static int prefetch(ColiExpertStore *store, const ColiExpertKey *keys,
         if (!record) continue;
         int resident = 0;
         V4ExpertSlot *slots = layer_slots(state, keys[i].layer);
-        for (int slot = 0; slot < state->slots_per_layer; slot++)
+        for (int slot = 0; slot < layer_slot_count(state); slot++)
             if (slots[slot].slab && slots[slot].expert == keys[i].expert) {
                 resident = 1; break;
             }
@@ -7355,6 +7374,9 @@ int coli_deepseek_v4_expert_store_open(
     }
     if (state->slots_per_layer > state->experts_per_layer)
         state->slots_per_layer = state->experts_per_layer;
+    /* calloc would leave pool_layer at 0, which this code reads as "pool open,
+     * owned by layer 0". Off is -1 and must be written, not assumed. */
+    state->pool_layer = -1;
     state->slots = calloc((size_t)state->layers * state->slots_per_layer,
                           sizeof(*state->slots));
     if (!state->slots) {
@@ -7786,7 +7808,7 @@ static void hot_repin_locked(V4HotPolicy *policy, V4ExpertStoreState *state,
         pins[rank] = best;
     }
     V4ExpertSlot *slots = layer_slots(state, layer);
-    for (int i = 0; i < state->slots_per_layer; i++)
+    for (int i = 0; i < layer_slot_count(state); i++)
         if (slots[i].slab && slots[i].expert >= 0 &&
             hot_is_pinned(policy, layer, slots[i].expert))
             slots[i].used = ++state->clock;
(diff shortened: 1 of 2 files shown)


JustVugg:
## A/B on the real checkpoint — the mechanism works, the speedup is not yet demonstrated

DeepSeek-V4-Flash-0731, 193-token prompt (2 chunks at the default 128), 1 token generated, `--memory-gb 22`, `.coli_usage` restored from a snapshot before each arm.

| | baseline | pool | delta |
|---|---|---|---|
| expert_requests | 5,355 | 5,355 | identical |
| **misses** | **5,128** | **3,513** | **−31.5%** |
| hit_rate | 4.24% | **34.40%** | 8.1× |
| **bytes** | **68.56 GB** | **46.97 GB** | **−31.5%** |
| TTFT | 413.5 s | 557.9 s | see below |
| generated_text | `I` | `I` | identical |

**Reads per expert: 1.46 → 1.00.** If 3,513 is the count of distinct (layer, expert) pairs this prompt touches — and it must be, since a pool of 989 slots against at most 256 experts per layer cannot evict anything it will need again — then the pooled arm reaches the theoretical floor of exactly one read per expert per layer. That is the same guarantee the CUDA bank gives, which is what this was for.

The predicted gain for a 2-chunk prompt was ~1.4×. Measured 1.46×.

## What this does not show

**TTFT went the wrong way, and I am not going to explain it away.** 413 → 558 s is +35%. This box's run-to-run noise on TTFT is documented at 72–183 s on identical runs, a factor of 2.5, so 1.35× sits inside it and the number carries no information either way — but neither does it support the change. Reading 31.5% fewer bytes should cost less time, and on this pair of runs it did not.

So: the deterministic counters say the mechanism does exactly what it was designed to do, and the wall-clock says nothing. A performance change that has not been shown to make anything faster should not merge on counters alone, which is why this stays a draft.

**What would settle it:** the same A/B on a ~1600-token prompt — 13 chunks instead of 2, where the model predicts ~6× rather than ~1.4×, and where the effect should clear the noise floor by enough to be unambiguous. Roughly 1.5 h per arm here. Running next.

**Token-exactness is also under-tested:** one generated token matching is a match, not a proof. The long run will generate more, and the change is cache policy only — routing, precision and accumulation order are untouched — so any divergence would be a bug rather than a tradeoff.

JustVugg:
## Why the TTFT went backwards: the store holds its mutex across the disk read

Verified in `deepseek_v4.c`, and it is not this PR's doing — it changes what this PR is worth, though, so it belongs here.

```c
7110  pthread_mutex_lock(&state->mutex);
7118  for (i < layer_slot_count(state))     /* linear scan: hit  */
7126  for (i < layer_slot_count(state))     /* linear scan: LRU victim */
7133  slot->slab = malloc(record_bytes);
7147  coli_st_read_at_streaming_rep(...)    /* scales  — 13.37 MB pread */
7150  coli_st_read_at_streaming_rep(...)    /* weights — under the same lock */
7159  pthread_mutex_unlock(&state->mutex);  /* released only here */
```

The `pthread_mutex_unlock` calls at 7128 and 7135 are error branches. The success path holds the lock through both preads, so **the whole engine performs one disk read at a time**, whatever `V4_LOADER_LANES` says.

Three previously unexplained observations collapse into this:

- **3 vs 6 loader lanes measured +2%.** Extra lanes cannot overlap I/O if I/O is serialized.
- **The disk delivers 152 MB/s during prefill** (876.8 GB / 5754.9 s) on a drive that does ~1.0 GB/s at QD4. That is the QD1 regime, exactly.
- **Neither disk nor CPU is saturated.** The signature of a lock, not of a device limit.

**And it explains this PR's result.** The pool makes the two linear scans 43× longer — 23 slots to 989 — and those scans sit inside the critical section that serializes every read in the engine. So it cut redundant bytes by 31.5% while lengthening the section that determines throughput. On a fully serialized path, critical-section length is throughput. The counters and the wall-clock were both right.

## What this changes about the plan

This PR should not merge alone. The lock fix should land first or alongside:

- under the lock: find or reserve the slot, mark it loading, `references++` so it cannot be chosen as a victim;
- unlock; read; re-lock; publish; signal;
- a second request for the same expert waits on a condition variable rather than issuing a second read;
- a short read leaves the slot invalid (`expert = -1`), which the existing comment at the read site already anticipates.

While in there, the linear scans want an index. They are O(slots) today and this PR makes them 43× more expensive; behind a lock that no longer serializes I/O they stop mattering for throughput, but they are still the wrong shape.

The two multiply rather than add: this PR removes ~6× of redundant bytes at long context, the lock fix removes ~4× of serialization. `bytes=` and `misses` must be **identical** across the lock fix — it moves the same bytes, just concurrently — which makes it easy to validate against this PR's counters.

Credit for the diagnosis to @JustVugg, from a read of the engine rather than a profile.

**Long-prompt A/B (1600 tokens, 13 chunks) is running now** and will land here. It measures this PR against the same serialized lock, so it will show the byte reduction and probably not a proportional time reduction — which is the expected shape given the above, not a failure of the change.

JustVugg:
Closing this for now — a maintainer decision, and the reason is worth recording rather than leaving as silence.

**Not because the defect is wrong.** It was verified by reading the code, not inferred from a measurement, and nothing here contradicts it.

**Because we cannot measure it on the hardware we have.** The dev box's run-to-run spread on this workload is ±100%: the same prompt, the same binary, produced TTFT 413 s and 821 s hours apart. A change worth 30% is invisible against that, so every result we produced was either a deterministic counter (which showed the mechanism working) or a wall-clock number that could not be defended. Continuing to spend that machine on it was buying nothing.

The work is in the branch history and this issue keeps the analysis. If someone with a machine that can measure — a quiet box, or simply one whose noise floor is under 10% — wants to pick it up, everything needed is written down here.

Reopening is cheap. Leaving an unprovable claim open, and re-reading it every triage, is not.

**What was established, and it is the clearest result of the three:**

Batched prefill is layer-major, so while layer L sweeps the whole prompt the other 42 slot partitions sit idle while L thrashes its own. Handing the in-flight layer the whole pool took reads per expert from **1.46 to 1.00** on a 2-chunk prompt — the theoretical floor, and the same guarantee the CUDA tier's per-layer bank already gives. Predicted gain for 2 chunks was ~1.4x; measured 1.46x. Generated text identical.

It did not show a time improvement, and #1158 explains why: the two linear scans this widens by 43x live inside the critical section that serialises every read in the engine. It cut redundant bytes while lengthening the section that governs throughput. Both measurements were right.

**One real bug was found here and is worth carrying forward if anyone revives this:** `V4ExpertStoreState` has the same layout in both store units, but `V4ExpertSlot` does **not** — one carries `aligned_slab`, the other does not, so `sizeof` differs and a single shared definition indexing `state->slots[]` strides wrongly for one backend. The fix is for the public entry point to touch only integers on the state and leave the slot walk to a `static` helper compiled inside each unit. That trap is not specific to this change and will catch the next person too.

PR #1177, opened 2026-08-22
feat(glm52): optimize inference on single RTX 3090 (24GB) via Direct I/O and MTP

up1t3:
## Summary

This PR introduces an optimization recipe, benchmarking suite, and CLI interactive chat client for running **GLM-5.2 (744B MoE)** on a single consumer GPU (**NVIDIA RTX 3090 24GB VRAM + 64GB host RAM**) under Windows.

### Key Achievements & Benchmarks:
- **Direct I/O Streaming (`DIRECT=1`):** Bypasses Windows file cache, boosting NVMe expert read throughput from **0.28 GB/s to 1.16 GB/s** (4.1x faster).
- **Decode Throughput:** Increased from `0.09 tok/s` (11.6s/token) to **`0.25 - 0.35 tok/s`** (2.9 - 4.0s/token), yielding a **2.5x - 3.5x speedup**.
- **Prefill Latency:** Reduced from 60-120s down to **12-15s**.
- **MTP Speculative Decoding (`DRAFT=1`):** Achieves **80% acceptance rate** with GPU-accelerated verify layer.
- **Windows MinGW Stability:** Documented optimal memory flags (`REPIN=0`, `PIN=auto`, `CUDA_RELEASE_HOST=1`, `--ram 28 --cap 3`) preventing commit charge OOM and 500 runtime crashes.
- **Latent Drift Fix:** Verified that pure Top-4 routing (`CACHE_ROUTE=0`) prevents hidden state drift and repetitive loop attractors during long context generation.

### Changes:
1. Added comprehensive documentation and benchmark guide: `GLM52_OPTIMIZATION_GUIDE.md`.
2. Added streaming CLI terminal client: `c/cli_chat.py` with proper stop token filtering and sampling parameters.
3. Added Windows convenience batch scripts: `chat.bat`, `c/chat.bat`.


ZacharyZcR (commented):
Read with interest — a single-3090/64 GB datapoint with a 4.1× Direct I/O measurement is genuinely useful territory, and the three-tier placement numbers (522 VRAM + 484 RAM pinned) are worth having in the tracker. But as a PR this has structural and evidentiary problems worth naming before review effort goes into details:

**Structure**
- `GLM52_OPTIMIZATION_GUIDE.md` lands in the **repository root**, in Russian, while every other doc lives under `docs/` in English. A translated `docs/` page — or better, a datapoint issue following `docs/benchmarks.md`'s protocol — is the shape the repo curates.
- `chat.bat` appears **twice** (root and `c/`), and `cli_chat.py` duplicates what `coli chat` / the gateway already provide. If the batch launcher solves a real Windows pain point, one copy under `c/` with a note on what `coli` can't do would carry it.

**Evidential claims that need data or softer wording**
- `TOPK=4` truncates top-8 routing — that **changes model output**, and the guide asserts '100% intelligence preserved, no degradation' with no perplexity/oracle measurement attached. This repo's own bar for routing changes is high: CACHE_ROUTE ships opt-in with a measured ~14% expert-substitution figure and its results are reported in a separate tier (#161) precisely so stock numbers stay comparable. A 50%-routing-cut claim needs at least the `coli bench` three-task suite against stock before 'no degradation' can stand.
- The +150% headline mixes four levers (DIRECT, TOPK, MTP, placement) with no ablation, and the 80% MTP acceptance at hit 77% sits in the regime where #161/#455 measured speculation as net-negative on streaming hosts — worth an A/B with `DRAFT=0` on the same config to separate what MTP contributes from what it costs in verify-widened reads.

The measurements are worth salvaging into a datapoint issue with the standard protocol (n≥3, medians, per-lever A/B); the guide's tuning advice would then have numbers to point at. As-is, the doc placement and the unmeasured no-degradation claims are what a reviewer hits first.

JustVugg:
Thank you for writing this up, and for the profiling numbers: a 3090 with 64 GB of host RAM driving GLM-5.2 is a configuration people ask about.

I cannot merge it in this form, for three reasons that are about placement, not about the content:

- The guide is in Russian and lives in the repository root. Documentation is in English under `docs/`; the only per-language files are the README translations. A translated guide would need an English original next to it.
- The public texts of the repository do not use emoji (headings, release notes, docs), so the file would need to be plain.
- `cli_chat.py` and the two `chat.bat` files duplicate what `coli chat`, `coli serve` and `coli web` already do, and they would drift from them.

What would go in: a page `docs/glm52-single-gpu-recipe.md` in English with the same measurements (hardware, commands, before and after numbers), without the client and the batch files. If you open that, I will review it.

PR #1181, opened 2026-08-23
build(v4/cuda): add CUDA_ARCH=portable-pre-ampere for Pascal and Turing

8PotatoChip8:
## Summary

The generic DeepSeek-V4 CUDA kernels do not require Ampere, but nothing in the
build or the docs says so, and `CUDA_ARCH=portable` starts at `sm_80` — so an
owner of a GTX 10-series or RTX 20-series card has no reason to think the card
is usable at all. (`CUDA_ARCH=sm_61` already fell through to the `-arch=`
escape hatch, but it was undocumented for these targets and produced a binary
for exactly one card.)

Reading `backend_cuda_dsv4.cu`, neither of the two things that usually force a
minimum architecture actually does so here:

- **bf16 is storage only.** Every use goes through `__bfloat162float` /
  `__float2bfloat16` and the arithmetic is fp32. There is no `__hmul` /
  `__hadd` / bf16 MMA anywhere in the file, so no `sm_80` bf16 ALU is needed.
- **fp8 weights are decoded, not computed.** They go through the 256-entry
  constant lookup table the engine already carries (`e4m3()` / `e4m3_round()`),
  not fp8 tensor-core MMA.

One caveat I want to state rather than gloss over: the generic build's two
activation-quantise kernels, `fp8_sim64` and `fp8_sim64_dynamic`, do call
`__nv_cvt_float_to_fp8`. That is not an MMA op, and it is not a hard floor —
`cuda_fp8.h` guards the native `cvt.rn.satfinite.e4m3x2.f32` PTX behind
`__CUDA_ARCH__ >= 890` and otherwise falls through to its own software bit
manipulation — so it compiles and runs pre-Ampere. But it does mean those two
kernels take a *different code path* on Pascal/Turing than on Ada, which is the
one place in this change where behaviour is not identical by construction. That
made it the thing worth checking on real silicon rather than reasoning about,
so I did: both a 1080 Ti and a 2080 Ti reproduce the CPU reference to 2^-26.
Numbers in Validation below.

So this is a gencode list, not a kernel change — the file compiles unmodified.
`CUDA_ARCH=portable-pre-ampere` spans `sm_61` through `sm_90` and closes with a
`compute_90` PTX image, giving one binary that covers Pascal through Hopper
natively and newer cards through driver JIT.

It deliberately does **not** mirror `portable` all the way to `sm_120`.
`compute_120` requires CUDA >= 12.8 — the same toolkit floor the mandatory
`NO_TC=1` companion exists to stay below — so carrying it would make the preset
unbuildable by exactly the older toolkits its users are likeliest to have. (For
the same reason, stock `CUDA_ARCH=portable` does not build on CUDA 12.4.) The
trailing PTX keeps newer hardware covered without that dependency.

Two constraints are worth stating explicitly, and the diff records both in
`docs/deepseek-v4.md`:

- **`NO_TC=1` is required**, and for a reason that is easy to misread as an
  architecture limit: the opt-in `DSV4_CUDA_TC` path calls cuBLASLt
  block-scaling APIs (`CUBLASLT_MATMUL_DESC_A_SCALE_MODE` and friends) that
  need CUDA >= 12.8. That is a **toolkit** dependency, not an architecture one
  — which is exactly what `NO_TC=1` already exists for.
- **DeepGEMM stays `sm_90+` and cannot be lowered.** It calls
  `cuTensorMapEncodeTiled` (TMA), a hardware unit introduced in Hopper, so
  there is no software path to fall back to. Pre-Ampere cards get the generic
  tier and nothing more.

Two files, +56/-1, all of it the new branch and its documentation. No
kernel, engine, or CPU-path code is touched.

## Validation

- [x] `make -C c check`

**Compile coverage.** CUDA 12.4, clean (the only warning is the pre-existing
unused `g_pin_bytes`, once per arch):

```
$ make -f Makefile.deepseek-v4 CUDA=1 NO_TC=1 CUDA_ARCH=portable-pre-ampere \
       backend_cuda_dsv4.o
$ cuobjdump -lelf backend_cuda_dsv4.o
ELF file    1: backend_cuda_dsv4.1.sm_61.cubin
ELF file    2: backend_cuda_dsv4.2.sm_75.cubin
ELF file    3: backend_cuda_dsv4.3.sm_80.cubin
ELF file    4: backend_cuda_dsv4.4.sm_86.cubin
ELF file    5: backend_cuda_dsv4.5.sm_89.cubin
ELF file    6: backend_cuda_dsv4.6.sm_90.cubin
$ cuobjdump -lptx backend_cuda_dsv4.o
PTX file    1: backend_cuda_dsv4.1.sm_90.ptx
```

**Numerical validation on real pre-Ampere hardware.** Codegen is not
correctness, so this was run on actual Pascal and Turing silicon before
opening the PR. `tests/test_backend_cuda_dsv4.cu` drives matvec, the expert
path, QKV, the router and `dsv4_fp8_simulate`, comparing every result against
a CPU reference and failing above `1e-3`. Built per card with CUDA 12.8 /
MSVC 17.14 on Windows:

```
nvcc -O3 -std=c++20 --expt-relaxed-constexpr --expt-extended-lambda \
     -DCOLI_DSV4_NO_TC -gencode arch=compute_61,code=sm_61 \
     backend_cuda_dsv4.cu tests\test_backend_cuda_dsv4.cu \
     -o dsv4_test_sm_61.exe cudart.lib cublasLt.lib
```

| GPU | arch | result |
|---|---|---|
| GTX 1080 Ti | sm_61 | `dsv4-cuda worst_abs=1.49012e-08` — PASS |
| RTX 2080 Ti | sm_75 | `dsv4-cuda worst_abs=1.49012e-08` — PASS |

`1.49012e-08` is 2^-26, i.e. agreement at the limit of fp32 rounding, ~5
orders of magnitude inside the tolerance — and **identical on both
architectures**. That is the answer to the one open question in the Summary:
the `__nv_cvt_float_to_fp8` software fallback that Pascal/Turing take instead
of the native `cvt` instruction produces the same numbers as the reference.

Two notes on the checkboxes I left unticked, rather than ticking them loosely:

- `make -C c cuda-test` was **not** run, because it does not exercise this
  code — it targets `backend_cuda.cu` (the GLM backend), not
  `backend_cuda_dsv4.cu`. The DSV4 CUDA test above appears to have no make
  target at all; I compiled it directly. Happy to add a `cuda-test-dsv4`
  target in this PR or a follow-up if you would like one.
- No performance claims are made, so the third box does not apply. This
  change adds architectures to a gencode list; it does not touch a kernel.

## Compatibility

The new value is opt-in and additive: `native`, `portable`, `sm_XX`, and
`DEEPGEMM=1` all take the same branches they did before — the new `else ifeq`
sits inside the non-DeepGEMM `else`, so `DEEPGEMM=1` (which ignores
`CUDA_ARCH`) is untouched. Building the extra two architectures costs compile
time only when the new value is explicitly selected.

### DIFF
--- a/c/Makefile.deepseek-v4
+++ b/c/Makefile.deepseek-v4
@@ -111,7 +111,8 @@ endif
 # backend_loader_dsv4.o. Elsewhere there is no loader: the same
 # backend_cuda_dsv4.cu is compiled by nvcc and linked straight into the
 # engine, so the dsv4_cuda_* symbols are the real ones. Defaults to the
-# GENERIC kernels (any sm_80+; CUDA_ARCH=native|portable|sm_XX);
+# GENERIC kernels (CUDA_ARCH=native|portable|portable-pre-ampere|sm_XX;
+# portable covers sm_80+, portable-pre-ampere adds sm_61/sm_75 and needs NO_TC=1);
 # DEEPGEMM=1 adds the sm_120a DeepGEMM tensor-core paths and needs the
 # patched sm120 DeepGEMM checkout at DEEPGEMM_HOME (see
 # patches-deepgemm-sm120-msvc.patch and docs/deepseek-v4.md). Untested here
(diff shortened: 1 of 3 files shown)


8PotatoChip8:
Follow-up: the gencode change alone was not enough to actually run on these cards.

`dsv4_cuda_backend_arch_ok` returns `prop.major >= 8`, so the loader rejects an
sm_61 or sm_75 card at startup regardless of what the binary contains — the
kernels compile for those cards and then never run. Pushed a second commit
lowering the generic floor to `major >= 6`.

Justification, kept to what is actually verified:

- bf16 is storage only here — every use is a `__bfloat162float` /
  `__float2bfloat16` convert and the arithmetic is fp32.
- fp8 weights are decoded through a constant lookup table, not fp8 tensor cores.
- The warp primitives in this file (`__shfl_*_sync`, `__syncwarp`) are sm_30-era
  and impose no floor of their own.

sm_60 is the floor because Pascal is the oldest architecture this has been
validated on. Maxwell would very likely work and is deliberately not claimed,
because nobody has run it.

The check stays independent of what the binary contains, on purpose: a
`CUDA_ARCH=portable` build has no sm_61 cubin, and running it on such a card
fails at launch with "no kernel image is available" — loud and obvious — rather
than producing a wrong answer. Cards below sm_80 need
`CUDA_ARCH=portable-pre-ampere`, which is what that preset is for.

DeepGEMM is unaffected: still `compute 12.x` only.

Hardware validation is unchanged from the original post — GTX 1080 Ti (sm_61)
and RTX 2080 Ti (sm_75), `dsv4-cuda worst_abs=1.49012e-08` on both, i.e. 2^-26,
identical across the two architectures.

PR #1212, opened 2026-08-24
fix(qwen36): eliminate all compiler warnings (-Wall -Wextra)

Petsku01:
## Summary

Eliminates all compiler warnings from c/qwen36.c under gcc 13.3 -Wall -Wextra -fopenmp. No logic or behaviour changes.

### Commit 1: Initial warning fixes (5 fixes)

| # | Warning | Fix |
|---|---------|-----|
| 1 | -Wcomment: nested /* in block comment (line 354) | buf/*bn -> buf (*bn) |
| 2 | -Wformat-truncation: jb[160] too small (line 477) | -> jb[512] (widened per review) |
| 3 | -Wformat-truncation: jb[1200] too small vs esc[1024] (line 490) | -> jb[2048] |
| 4 | -Wformat-truncation: jb[400] too small vs esc[256] (line 525) | -> jb[768] |
| 5 | -Walloc-size-larger-than: calloc(c->n_layers, ...) without cast (5 sites) | Add (size_t) cast |

### Commit 2: Widen jb[256] to jb[512] per review feedback

Review noted jb[256] was too tight (8 bytes margin with default model). getenv("MODEL") can return arbitrary length strings.

### Commit 3: Early n_layers validation per cross-review feedback

- Move n_layers validation to load_cfg() before first calloc (is_attn)
- Add compiler-visible guard in model_init() after validate_cfg
- Clamp n_layers local copy before DN_rec/DN_conv calloc

## Verification

| Check | Result |
|-------|--------|
| Build (gcc -O3 -Wall -Wextra -fopenmp) | 0 warnings |
| C unit tests (test_qwen36_ctx + 12 others) | All pass |
| Full test suite (567 tests) | 567/567 pass (55 skipped = CUDA) |
| Oracle (token-exact) | Not run (torch unavailable on this host) |

## Cross-review

Independent cross-model review performed (GPT-5.6 Sol). Findings addressed:
- n_layers validation ordering fixed (was validated after first allocation)
- m->V calloc cast confirmed present in all call sites

## Notes

- (size_t) casts applied to all calloc(c->n_layers, ...) call sites for consistency
- MODEL environment variable length is unbounded (pre-existing, not in scope)
- Buffer sizes chosen to comfortably exceed worst-case snprintf output

## Hardware

- CPU: Intel i7 (WSL2, Linux 6.x)
- No GPU (CUDA tests skipped)
- gcc 13.3.0, OpenMP

### DIFF
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -920,6 +920,7 @@ static void load_cfg(Cfg *c, const char *snap) {
     c->n_experts = 256; c->topk = 8; c->inter = 512; c->shared_inter = 512;
     c->n_group = 1; c->topk_group = 1; c->norm_topk = 1; c->has_qk_norm = 1; c->has_bias = 0;
     c->attn_output_gate = 1; c->n_active = 0;
+    if (c->n_layers <= 0 || c->n_layers > 512) { fprintf(stderr, "load_cfg: n_layers=%d out of range 1..512\n", c->n_layers); exit(1); }
     c->is_attn = calloc((size_t)c->n_layers, sizeof(uint8_t));
     for (int i = 0; i < c->n_layers; i++) c->is_attn[i] = (i % 4 == 3) ? 1 : 0;
 }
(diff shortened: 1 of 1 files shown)


JustVugg:
Reviewed, and the substance is good; worth saying clearly that two of these are latent bug fixes rather than cosmetic silencing: the truncated `jb` buffers would have produced invalid JSON in the streaming chunks once a long model id or an escape-heavy token came through, so widening them fixes real output corruption, not just a warning.

One thing before this can land: the branch now conflicts with `dev`, and that is our doing, not yours: `qwen36.c` was touched by two merges since you opened this (the expert-slot index in #1215 and the segment adapters in #1227). A rebase onto current `dev` should be mostly mechanical since your changes are local to the streaming and init paths. If you would rather not fight the conflicts, say so and we can rebase it for you, keeping your authorship.

One optional cleanup while you are in there: in `model_init` the `if (n_layers <= 0) n_layers = 1;` clamp is dead code (the exit above it already guarantees the range); using the same loud-exit style as the line above would keep one convention. Not blocking.

Petsku01:
Rebased onto current dev. New PR with updated branch coming — all fixes preserved plus reviewer feedback addressed.

Changes in new PR:
- jb buffer widening (160->512, 1200->2048, 400->768)
- size_t casts on all 7 calloc call sites
- n_layers range guard (1..512) in load_cfg
- n_layers re-assert in model_init_range after validate_cfg
- buf/*bn -> buf (*bn) comment fix (-Wcomment)
- Optional cleanup: dead n_layers=1 clamp replaced with loud-exit

Build: 0 warnings in qwen36.c (-Wall -Wextra -O2)

PR #1218, opened 2026-08-24
Quantize dense weights incrementally during load (55.8% peak RSS reduction on Qwen3.6-35B-A3B gs64)

Blakeolson21:
Quantizes each dense f32 tensor to signed int8 immediately after load and releases the f32 source before the next tensor loads, instead of holding the full f32 model in memory before quantization begins.

Measured on a real Ornith/Qwen3.6-35B-A3B gs64 checkpoint (Linux, /proc measurements, pre- and post-change builds from identical make invocations):

- Peak RSS (VmHWM): 11,591,168 KiB before, 5,126,380 KiB after: a 55.8% reduction
- Per-process swap: 0 KiB throughout on both builds
- First generated token and expert-cache hit/miss counts identical pre/post (hit=328, miss=1592)
- Both real-model forward passes exit 0

Also included: COLI_DENSE_I8=0 retains and dispatches the exact f32 reference tensors (with executable coverage), COLI_KEEP_F32 preserved, packed-int4 gs64 routed-forward behavior covered by the focused regression (c/tests/test_qwen36_dense_batch), and docs for the runtime controls. Focused tests pass in optimized and ASan/UBSan builds.


Blakeolson21:
Closing to resubmit as a minimal, focused change; this branch accidentally carried unrelated history.

PR #1226, opened 2026-08-25
fix(serve): authenticate K3 tool-call sideband

### LINKED ISSUE #1147: serve: carry Kimi K3 tool-call structure on the shared codec instead of the text channel
serve: carry Kimi K3 tool-call structure on the shared codec instead of the text channel

**Corrected scope.** This issue originally proposed doing GLM and Kimi K3 together. That was wrong, and checking the two engines rather than assuming they were symmetric is what showed it:

- **Kimi K3**: `<|open|>`, `<|close|>`, `<|sep|>`, `<|end_of_msg|>` are real special tokens — `chat_special()` resolves them to ids held in `sp[0..3]` (`kimi_k3.c:2126`, `:2203`). The engine therefore *knows* whether a structural marker was generated by the model or merely looks like one.
- **GLM**: `<tool_call>` does not appear in `colibri.c` at all. The markers are ordinary text rendered into the prompt and recovered by `parse_tool_calls` in `openai_server.py`. **The engine has no ground truth to preserve**, so a sideband would carry nothing better than the text already does. GLM's exposure is inherent to its format and can only be narrowed in the parser, which `_unclosed_tail` already does for the lenient path.

**So: K3 only.**

**What it would fix.** #1144 ships K3 tool calling with the structural run re-emitted as literal text for the gateway to pattern-match (option (a) from #1143). That discards a distinction the engine holds: a user prompt that gets the model to echo `<|open|>call tool="…"<|sep|>` as ordinary characters is, downstream, indistinguishable from a real call. The gateway reports `tool_calls` and never executes — the client application decides what to run — so this is a trust-boundary weakness rather than code execution, but a client acting on our `tool_calls` field is acting on a statement of fact we would be getting wrong.

**Why the codec makes this cheap.** `c/serve_codec.h` owns both directions since the framing migration finished (#1087, #1090, #1096, #1116), each engine landed behind a byte-exact wire-transcript freeze, and Inkling already rides an engine-specific structured payload on it (audio, as an opaque extension). A tool-call record is the same shape: one file, one new record type, an existing test that proves nothing else moved.

**Not urgent, and not user-visible.** The wire between engine and gateway is internal; the OpenAI `tool_calls` contract clients see is identical either way, so this can land whenever without a migration for anyone.

Context: #1143 (design), #1144 (implementation), #1029 (original gap).

bherald:
## Summary

- add a byte-counted `TOOL <id> <n>` engine-to-gateway record for Kimi K3
- have a zero-byte TOOL frame declare the sideband authoritative for the request, keeping marker-looking DATA as ordinary text
- route only special-token K3 tool structures and enclosed payload through the trusted channel
- preserve compatibility with engines that do not emit TOOL, and cover OpenAI/Anthropic streaming and non-streaming responses

Closes #1147.

## Validation

- `git diff --check`
- `python3 -m py_compile c/openai_server.py c/tests/test_openai_server.py`
- `make -C c tests/test_serve_codec tests/test_kimi_serve_framing tests/test_k3_chat_tools`
- `./c/tests/test_serve_codec`
- `./c/tests/test_kimi_serve_framing`
- `./c/tests/test_k3_chat_tools`
- `cd c && PYTHONPATH=. python3 tests/test_openai_server.py` (155 tests)

### DIFF
--- a/c/kimi_k3.c
+++ b/c/kimi_k3.c
@@ -2770,6 +2770,11 @@ static void serve_data(const char *id, const char *p, int n){
     coli_serve_write_data(stdout,id,p,(size_t)n);
 }
 
+static void serve_tool(const char *id, const char *p, int n){
+    if(n<0) return;
+    coli_serve_write_tool(stdout,id,p,(size_t)n);
+}
+
 static int serve_one(Model *m, Tok *T, ServeReq *q){
     int cap=65536, *ids=malloc((size_t)cap*sizeof(int)), np=0;
     if(!ids){ coli_serve_write_error(stdout,q->id,"out of memory"); return 0; }
@@ -2802,6 +2807,10 @@ static int serve_one(Model *m, Tok *T, ServeReq *q){
         coli_serve_write_error(stdout,q->id,message); free(ids); return 0;
     }
     coli_serve_write_accept(stdout,q->id,np);
+    /* Declare the structured sideband before any generated DATA. Even an
+     * empty sideband is meaningful: it tells the gateway that literal K3
+     * marker lookalikes in DATA are content, not engine-proven structure. */
+    if(chat) serve_tool(q->id,NULL,0);
     /* KV PREFIX REUSE (#639 for GLM; this engine re-prefilled every turn).
      * A chat client resends the whole transcript each turn, so turn N used to
      * re-process turns 1..N-1 from scratch — the cost of a message grew with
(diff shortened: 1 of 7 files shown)


JustVugg:
Merged. This closes a real injection vector, and the design is the right one: the engine is the only party that knows whether a marker came from an actual special token or from text that merely resembles one, so authority has to originate there. The zero-byte TOOL frame declaring the sideband authoritative is the detail that makes the whole thing sound; without it, an old-style DATA parse would have remained reachable on every request.

What was verified in review beyond CI: the separate incremental UTF-8 decoder for the sideband (interleaved DATA/TOOL chunks must not share decoder state, and they do not), the backward path for pre-sideband engines, the wire-level test proving a lookalike in DATA stays content while the sideband call is parsed, and the golden-bytes codec coverage. All held up. Thanks for the careful scoping in #1147 too: checking that GLM's markers are plain text while K3's are real special tokens, instead of assuming symmetry, is what made this fix land clean.

PR #1225, opened 2026-08-25
docs: add Mintlify site and OpenAPI reference

hamedrabah:
## Summary

- add a Mintlify documentation configuration for the existing guides
- add a concise docs landing page and HTTP API overview
- add an OpenAPI 3.1 reference for the implemented inference, model, health, routing, and profiling endpoints
- document both bearer and `x-api-key` authentication plus the authenticated telemetry boundary

## Why

The HTTP API guide is detailed but lives in one long Markdown page. The new structure keeps that guide intact while adding navigable endpoint pages, request schemas, response shapes, and an interactive-reference foundation.

## Validation

- ran the official Mintlify CLI validator successfully (`mint validate`)
- rendered and reviewed the local Mintlify preview, including the OpenAPI reference
- checked the documented routes against `c/openai_server.py` and the web client types
- verified all 19 navigation targets resolve
- verified every internal OpenAPI reference resolves
- verified every operation has a summary, description, operation ID, and 200 response
- parsed `docs.json` and `openapi.json` with `jq`
- ran `git diff --check`

PR #1228, opened 2026-08-25
docs: restore Inkling literal tokens

hamedrabah:
## Summary

- restore the literal Inkling chat-template tokens in `docs/inkling.md`
- escape the pipe characters so the tokens remain valid inside the Markdown table

The rendered Mintlify page displays `<|content_text|>`, `<|message_model|>`, and `<|audio|>` without the source-level backslashes.

## Validation

- `mint validate --telemetry=false`
- local Mintlify preview render checked for the exact displayed token strings

Follow-up to #1225.

### DIFF
--- a/docs/inkling.md
+++ b/docs/inkling.md
@@ -84,9 +84,9 @@ help in proportion to the RAM you can give them.
 | Invocation | What it does |
 |---|---|
 | `-p "text" [-n N]` | streaming greedy generation (stops at eos or N tokens) |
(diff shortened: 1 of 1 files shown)


hamedrabah:
@JustVugg I tested the escaped form against Mintlify: escaping only the table pipes passes strict validation and renders the literal tokens exactly. This follow-up restores the three strings you flagged. Thanks again for catching it.

JustVugg:
Merged, thanks for the quick follow-up. Bonus discovered in review: per the GFM spec, unescaped pipes inside code spans still delimit table cells, so the pre-#1225 table was technically mis-rendered on GitHub as well; your escaping fixes both renderers at once.

PR #1232, opened 2026-08-25
fix(qwen36): widen jb buffers, add size_t casts, fix nested comment

Petsku01:
## Summary

Rebase of #1212 onto current dev (conflicts with #1215 and #1227 resolved).

## Changes

- jb[160]->512, jb[1200]->2048, jb[400]->768: prevent truncated JSON in streaming chunks with long model ids or escape-heavy tokens
- calloc(c->n_layers,...)->calloc((size_t)c->n_layers,...): silence -Walloc-size-larger-than on all 7 call sites
- n_layers range guard (1..512) in load_cfg before first calloc
- n_layers re-assert in model_init_range after validate_cfg (compiler-visible guard)
- buf/*bn -> buf (*bn) in comment to silence -Wcomment

## Reviewer feedback addressed

Per @JustVugg comment on #1212:
- Rebased onto current dev
- Optional cleanup: dead n_layers=1 clamp replaced with loud-exit style

## Verification

- Build: 0 warnings in qwen36.c (-Wall -Wextra -O2)
- No new warnings introduced
- Pre-existing -Wmisleading-indentation warnings are outside this PR scope

### DIFF
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -91,7 +91,7 @@ static int hexnib(char c){
  * pre-tokenize + per-piece ByteLevel map + BPE merges. */
 typedef struct { char **keys; int *vals; int *used; int cap; } SMap;
 static unsigned shash(const char *s){ unsigned h=2166136261u; while(*s){ h^=(unsigned char)*s++; h*=16777619u; } return h; }
-static void smap_init(SMap *m,int cap){ m->cap=cap; m->keys=calloc(cap,sizeof(char*)); m->vals=malloc(cap*sizeof(int)); m->used=calloc(cap,sizeof(int)); }
+static void smap_init(SMap *m,int cap){ m->cap=cap; m->keys=calloc((size_t)cap,sizeof(char*)); m->vals=malloc((size_t)cap*sizeof(int)); m->used=calloc((size_t)cap,sizeof(int)); }
 static void smap_put(SMap *m,const char *k,int v){ if(!k)return; unsigned h=shash(k)&(m->cap-1); while(m->used[h]){ if(m->keys[h]&&strcmp(m->keys[h],k)==0){m->vals[h]=v;return;} h=(h+1)&(m->cap-1);} m->used[h]=1; m->keys[h]=(char*)k; m->vals[h]=v; }
 static int smap_get(SMap *m,const char *k){ if(!m||!m->cap||!k)return -1; unsigned h=shash(k)&(m->cap-1); while(m->used[h]){ if(m->keys[h]&&strcmp(m->keys[h],k)==0)return m->vals[h]; h=(h+1)&(m->cap-1);} return -1; }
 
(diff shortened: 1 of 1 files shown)


c/qwen36.c
OPS-NeoRetro:
```suggestion
static void smap_init(SMap *m,int cap){ m->cap=cap; m->keys=calloc((size_t)cap,sizeof(char*)); m->vals=malloc(cap*sizeof(int)); m->used=calloc((size_t)cap,sizeof(int)); }
```

c/qwen36.c
```diff
@@ -1304,7 +1308,7 @@ static void model_init_range(Model *m, const char *snap, int cap, int bits,
             #undef LD4
         }
     }
-    m->cache = calloc(c->n_layers, sizeof(LCache));
+    m->cache = calloc((size_t)c->n_layers, sizeof(LCache));
     for (int i = layer_begin; i < layer_end; i++) {
         m->cache[i].cap = cap;
         m->cache[i].slots = calloc(cap, sizeof(Slot));
```
OPS-NeoRetro:
```suggestion
        m->cache[i].slots = calloc((size_t)cap, sizeof(Slot));
```

OPS-NeoRetro (changes requested):
You didn't apply `size_t` casts in `calloc` calls that you have overlooked.

OPS-NeoRetro:
@Petsku01

Petsku01:
Applied the requested review cleanups in 8a36782.

Changes:
- cast `cap` to `(size_t)` in `smap_init()` allocations
- cast `cap` to `(size_t)` in `m->cache[i].slots = calloc(...)`

Verification I ran locally on this branch:
- `make qwen36`
- `python3 tools/run_tests.py tests/test_qwen36_ctx tests/test_qwen36_cache_index tests/test_segment_runtime tests/test_segment_conformance`
- full `make test` completed successfully earlier on the same PR branch before this cleanup; this follow-up is a no-behavior-change allocation cast cleanup only

Thanks for the catch.

OPS-NeoRetro (approved):
APPROVED!!! 👍

JustVugg:
Merging. Verified rather than taken on trust: 0 warnings building `qwen36.c` with this branch, against 4 on pristine dev, and the four it removes are exactly the ones a contributor reported independently in #1253 after building v1.9.0 on a Sandy Bridge ThinkPad.

Two things here are worth more than the warning count.

**The `n_layers` range guard before the first calloc** is a correctness fix, not a cosmetic one, and it landed at a good moment: the sanitizer just found a heap overflow in `load_meta` on the same field, where `is_attn` was sized from config.json and then filled using the count from qwen36_meta.json (#1262, fixed in #1277). Yours guards the allocation, that one guards the write. They are complementary.

**The widened `jb` buffers** are also not cosmetic. `-Wformat-truncation` on the SSE chunk writers means a long model id or an escape-heavy token could truncate JSON mid-stream, and a truncated frame corrupts a streaming response rather than failing loudly.

This also closes the third of the three patches from #1253. The other two were merged as #1258 and #1259, and I had left that issue open specifically because this one was still outstanding; I will close it now pointing here.

Thanks for rebasing #1212 and for keeping the scope tight — the `size_t` casts and the nested-comment fix are the kind of change that is easy to review precisely because it does not wander.

PR #1248, opened 2026-08-26
feat(glm53): add GLM-5.3-Flash runtime

### LINKED ISSUE #1243: [Feature]: Add support for GLM-5.3-Flash
### Problem

Another MoE model that released today is GLM-5.3-Flash. It outperforms GLM 5.2 while being less than half the size at 320B 18B active.

Huggingface link: https://huggingface.co/zai-org/GLM-5.3-Flash

### Proposed solution

CPU only first then add GPU support.

### Alternatives considered

_No response_

### Scope and compatibility

_No response_

ZacharyZcR:
## Summary

- register the GLM-5.3-Flash family and official 62-shard checkpoint layout
- add the complete CPU text runtime with KDA, DSA, mHC, MTP, streaming routed experts, and prefix-cache reuse
- support official FP8 expert tensors on CPU and native FP8 dispatch through the shared CUDA and Metal backends
- add one-shot CLI and OpenAI-compatible completions serving
- extend install, doctor, resource planning, documentation, and release contracts
- add deterministic Transformers 5.16.1 tiny oracles, production-layout validation, segment conformance, malformed-input coverage, and mock GPU dispatch tests

## Validation

- token-exact F32 and FP8 full/cached tiny generation
- official checkpoint header/layout contract
- OpenAI `/v1/completions` end-to-end request
- CUDA Linux host link and Windows MSVC DLL builds
- Metal build on macOS
- full Python discovery and repository cross-platform CI

## Acceptance boundary

This PR implements the CPU-first support requested by #1243 and the subsequent CUDA/Metal tier. Full 320B checkpoint generation and real-device performance measurements remain hardware acceptance work; they are not inferred from tiny fixtures or toolchain-only CI.

Closes #1243

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -97,6 +97,22 @@ jobs:
       - name: Generate fixture and run target oracle
         run: make -C c deepseek-v4-tiny-check
 
+  glm53-tiny:
+    name: Linux (generated tiny GLM-5.3 oracle contract)
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+          cache: pip
+          cache-dependency-path: c/tools/requirements-glm53-tiny.txt
+      - name: Install pinned tiny-fixture dependencies
+        run: python -m pip install -r c/tools/requirements-glm53-tiny.txt
+      - name: Generate and validate oracle contract
+        run: make -C c glm53-tiny-check
+
   macos:
     # clang; libomp for the threaded path (Makefile falls back to
     # single-threaded automatically if it's ever missing).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -316,7 +316,7 @@ jobs:
           # step with "Version not available" before nvcc is ever reached.
           cuda: '12.6.2'
           method: network
-          sub-packages: '["nvcc"]'
+          sub-packages: '["nvcc", "cudart", "cublas"]'
       - name: Compile backend_cuda.cu (syntax only, no GPU)
         run: |
           cd c
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,7 +210,7 @@ jobs:
         run: |
           cd c
           rc=0
-          for t in colibri inkling kimi_k3; do
+          for t in colibri glm53_flash inkling kimi_k3; do
             echo "::group::$t METAL=1"
             make clean >/dev/null 2>&1 || true
             make $t METAL=1 || { echo "FAILED: $t METAL=1"; rc=1; }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,7 +190,7 @@ jobs:
         run: |
           cd c
           rc=0
-          ENGINES="colibri inkling kimi_k3 olmoe qwen36"
+          ENGINES="colibri glm53_flash inkling kimi_k3 olmoe qwen36"
           if [ "${{ matrix.v4 }}" = "1" ]; then ENGINES="$ENGINES deepseek-v4"; fi
           for t in $ENGINES; do
             echo "::group::$t"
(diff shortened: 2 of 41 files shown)


JustVugg:
Thank you for this, and for the honesty of the acceptance boundary in the description - saying plainly that full-checkpoint generation and device numbers are still ahead is exactly the right way to present work like this.

I read the whole thing rather than skimming it, and I want to start with what holds up, because it is most of it. I checked eight mechanisms line by line against `modeling_glm5_next.py`, deliberately picking the places where ports usually go wrong:

- the KDA recurrence: decay applied per-k **before** the accumulation, delta taken from the pre-update `kv_mem`, output read from the updated state. Correct.
- q/k L2-norm with eps **inside** the sqrt, matching FLA rather than `F.normalize` - the reference calls that out as a deliberate difference, and you got it.
- the low-rank output gate (`g_a` then `g_b`), which is where GLM-5.3 differs from Kimi K3's single full gate.
- `RMSNormGated` in the right order: norm, then weight, then `sigmoid(gate)`.
- both indexer scales in the right places (`n_heads**-0.5` on the weights, `head_dim**-0.5` on the scores), and the ReLU expressed as `if (dot > 0)` - mathematically identical, since ReLU commutes with a positive scale.
- the router: the correction bias affects **selection** but not the returned weight, then normalize, then scale by 2.5. That distinction is subtle and frequently missed.
- the final stream collapse as an unweighted mean, which the reference explicitly notes differs from DeepSeek V4.
- k-pooling: pool completeness, the `ape` bias, per-channel softmax pooling, tail width `topk + kpool - 1`. This is the only genuinely new mathematics in the model and it is faithful.

The oracle is real work too: token-exact teacher forcing and greedy, logits within 2e-4, cache-versus-full parity within 1e-5, plus a checkpoint validator run against all 62 real shard headers. And adding your family to the Segment conformance matrix is exactly what that gate exists to force.

**So the blocker is not your mathematics. It is that the engine has no expert cache.**

In `matrix_multiply`, every use of a routed expert is malloc, full read from the shard, compute, free. The same path serves prefill and decode. Per token that is 42 layers x 8 experts x 3 matrices of ~8.4 MB, so roughly **8.4 GB of disk reads per token with zero reuse**, plus about a thousand 8 MB allocations. With 175 GB of experts and a page cache that cannot hold them, that cost is paid on every token. There is also no batching: `mlp3` is called with `rows=1` inside the token loop, so prefill never groups tokens per expert.

That matters more here than it would in most projects, because expert streaming is what colibri *is*. Every other engine carries the machinery: `colibri.c` has a per-layer LRU plus a learned hot store plus PILOT lookahead and batch-union; DeepSeek V4 has the expert store with pinning and coalesced loads; Kimi K3 has its own LRU with a RAM budget. Whichever engine lands for GLM-5.3 becomes the example the next contributor copies, so it needs to teach that pattern rather than the opposite one.

Two consequences worth stating explicitly, both measured rather than estimated:

- **Resident set.** Non-FP8 tensors are loaded as float32, four bytes where the checkpoint uses two. The KDA projections alone are 4.56 B parameters, which is 18 GB; the total resident set lands near **28 GB** before any expert cache.
- **The planner does not know that.** `resource_plan.py` derives `dense_bytes` from on-disk tensor sizes, about 14.4 GB here, so `coli plan` would tell a 24 GB machine that the model fits and the engine would then run out of memory. That is the shape of a bug report we would rather not receive.

Given all that, we are going to build the GLM-5.3 engine on top of `kimi_k3.c`, because that inherits the streaming machinery - LRU, async loader pool, batching, RAM budget - instead of retrofitting it. We also want the int4-g64 container (measured cosine 0.994 against the FP8 source on real weights), which halves the bytes moved per token and brings the checkpoint to ~194 GB, and the vision tower, which this PR scopes out.

I am not going to merge this PR, and I would rather say so directly than leave it open indefinitely.

What I would like to do instead, with your name on it: **take your tiny-oracle generator and your checkpoint validator as they are.** `tools/make_glm53_tiny.py` and `tools/check_glm53_checkpoint.py` are infrastructure we need regardless of which engine ships, they are well built, and re-deriving them would be waste. Your k-pool implementation is also the reference I will be checking ours against. If you would rather land those yourself as a separate PR, that is the outcome I would prefer - tell me and I will review it the same day.

To be clear about where the line falls: the review found no error in what you wrote. The disagreement is architectural, about what an engine in this repo has to carry before it ships, and that is on us to have stated earlier in #1243 rather than after you had written 3,600 lines.

ZacharyZcR:
Superseded by #1260, which contains only the tiny oracle generator, checkpoint validator, pinned dependencies, and isolated CI requested in review. The rejected runtime and cache architecture remain excluded.

PR #1255, opened 2026-08-27
perf(dsv4): replace ldexpf with arithmetic

TechPizzaDev:
## Summary

clang/LLVM does not inline `ldexpf`. Replace with arithmetic which is inlined/const everywhere. 
Alternatively, `pow(2.0f, -9)` can be used and gets inlined, but a simple div should be nice and accurate everywhere.
GCC inlines `ldexpf` and `pow` all the same.

## Validation

- [x] `make -C c check`
  - https://c.godbolt.org/z/djW93nx8j shows the diff. I don't use colibri myself so I don't even know how to run it and don't feel like making a bench, but this is in a hot path so should be an easy win for anyone using clang/LLVM.
  The codegen replicates locally when disassembling the `verify_fp8_decode.c` tool. 

## Compatibility

### Note
FYI, `fp8_decode8` is duplicated in `c/tools/verify_fp8_decode.c`. I feel like that tool should call the actual `fp8_decode8` from `c/deepseek_v4.c` instead.


c/tools/verify_fp8_decode.c
```diff
@@ -25,7 +25,8 @@ static inline __m256 fp8_decode8(__m256i codes) {
         _mm256_slli_epi32(man, 20));
     __m256 nval = _mm256_castsi256_ps(nbits);
     /* subnormal magnitude (exp==0): (float)man * 2^-9  (exact) */
-    __m256 sval = _mm256_mul_ps(_mm256_cvtepi32_ps(man), _mm256_set1_ps(ldexpf(1.0f, -9)));
+    float man_factor = 1.0f / (float)(1 << 9);
```
OPS-NeoRetro, line 28:
```suggestion
    float man_factor = (float)((127 - 9) << 23);
```

c/deepseek_v4.c
```diff
@@ -16110,7 +16110,8 @@ static inline __m256 v4_fp8_decode8(__m256i codes) {
         _mm256_slli_epi32(_mm256_add_epi32(exp, _mm256_set1_epi32(120)), 23),
         _mm256_slli_epi32(man, 20));
     __m256 nval = _mm256_castsi256_ps(nbits);
-    __m256 sval = _mm256_mul_ps(_mm256_cvtepi32_ps(man), _mm256_set1_ps(ldexpf(1.0f, -9)));
+    float man_factor = 1.0f / (float)(1 << 9);
```
OPS-NeoRetro, line 16113:
```suggestion
    float man_factor = (float)((127 - 9) << 23);
```

OPS-NeoRetro (changes requested):
NOT APPROVED 👎

Using reciprocal division is not good at all because it would be optimized away with `-O0 -ffast-math`. Compilers when they get the `-ffast-math` arg causes the compiler to generate code that sets FPU MSRs to have DAZ (Denormals As Zeros) turned on. Instead, apply the recommendations that I gave you. It makes the code look worse but it is efficient. I don't know if compilers now do constant folding for reciprocal divisions with an integer constant casted to a float, but I prefer my recommendations because it shows the implementation in a clearer way. The comment above already said `(float)man * 2^-9`, so I did the exponent load as a full-on exponent shift casted to a float.

c/tools/verify_fp8_decode.c
TechPizzaDev: 1. No one should use the x87 FPU in 2026.
2. fastmath does not break this.
3. Your suggestion is full of magic constants, and produces a completely incorrect value. It only works if you bit-cast it.

PR #1261, opened 2026-08-28
feat(xdna): optional explicit AMD XDNA2 lane for the GLM shared expert

Kenneth-Javier:
## Summary

Adds an **optional, explicitly requested** compute lane that runs two GLM
operations — the shared expert's gate and up projections — on an AMD XDNA2
(Ryzen AI) NPU on Windows. Nothing else, on no other platform.

It is off by default, it is never selected by discovering hardware, and the
qualified operations compute in BF16, so output can differ from the normal path.
`--xdna` prints that warning before any work starts.

## What changes

| | |
|---|---|
| new engine files | `c/backend_xdna.{c,h}` — the host side. Links no XRT. |
| new helper | `c/backend_xdna_helper.cpp` — the only file that links XRT, built separately, shipped separately |
| engine seam | `c/colibri.c` **+145/−0**: the include, a side pointer on `QT`, its release hook, and two guarded lines in `moe()` |
| CLI | `c/coli` **+12/−0**: one `--xdna` flag, one line setting `COLI_XDNA=1` |
| build | `XDNA=1` (default `0`) compiles the lane in; the Windows release job uses it |
| release | an optional sidecar carrying the helper and four qualified artifacts |

The two dispatch lines use the same idiom as `vk_matmul_qt` — return 0 means
"not handled", and the current path runs. Like the GPU dispatches beside them,
they skip the lane inside an OpenMP team: `backend_xdna.c` keeps process-wide
staging state with no locking.

```c
#ifdef COLI_XDNA
    if(omp_in_parallel() || !coli_xdna_try_matmul(COLI_XDNA_FAMILY_MOE_SHARED_GATE_UP, &l->sh_gate.xdna, ...))
#endif
#ifdef COLI_VULKAN
    if(!vk_matmul_qt(&l->sh_gate, sg, x, S))
#endif
    matmul_qt(sg, x, &l->sh_gate, S);
```

## Default behaviour

Unchanged, and measured against this PR's base rather than asserted:

| | dev `692af87` | this branch, built `XDNA=1` |
|---|---|---|
| canonical scoring control | `-22.105853` | `-22.105853` |
| XDNA helper loaded | — | 0 |
| device opened | — | 0 |
| XRT imports in the host | — | 0 of 12 (all CRT + `KERNEL32`) |
| `[XDNA]` diagnostics | — | 0 |

The optional sidecar was installed and valid during that run. The default path
ignores it. An `XDNA=1` host has the same import table as a default one and
starts normally on a machine with no NPU.

One thing does reach every build. A host built without `XDNA=1` still reads
`COLI_XDNA` at startup, and when it is set to a non-zero value prints one stderr
line — `[XDNA] requested, but this build has no XDNA support: continuing on the
normal path` — so an explicit request is answered rather than silently ignored.
With the variable unset it prints nothing and nothing else changes.

## Turning it on

```bash
coli run --xdna --model <glm-model> "your prompt"
```

`--xdna` sets `COLI_XDNA=1` for the engine, and the engine enables the lane when
that variable parses as a non-zero integer. Exporting it yourself has the same
effect, which is how a harness driving the engine directly gets in; the flag wins
when both are present. All `[XDNA]` output is stderr, never stdout, so
machine-readable score output is unaffected.

## Qualified scope

```
platform      Windows, AMD XDNA2 (Ryzen AI)
family        MoE shared-expert gate / up   (sh_gate, sh_up)
K = 6144, N = 2048, fmt=4 grouped int4, group size 64, pair layout
M   1..64    -> M64 artifact       65..256 -> M256 artifact
    > 256    -> normal path
```

Everything else — the down projection, routed experts, attention, every other
model — uses the normal path. The semantic family is passed explicitly by the
call site and never inferred from a shape, so an unrelated 6144×2048 operation
cannot inherit this one's qualification.

## Failure and fallback

Every failure is fail-closed to the current path, with a diagnostic that names
the actual problem:

```
package not found · package incomplete (names the file) · artifact hash
mismatch (names the file) · helper not usable · registry invalid
```

The package integrity gate runs **before** the helper is loaded, so unknown or
corrupt bytes never reach execution. No partial NPU result is ever exposed.
Restoring the correct bytes restores the lane with no residue.

## Packaging and dependencies

The helper and the artifacts need the XRT SDK and the AIE toolchain, which the
release runners do not have, so CI cannot build them. They ship as an optional
sidecar produced on a qualified machine:

```
colibri-<tag>-windows-x86_64-xdna.{zip,manifest.txt,sha256}
```

The name keeps the core archive's stem, so the release job's existing
`sha256sum colibri-*` and `gh release upload artifacts/*` already cover it — no
workflow change is needed to publish it. `--verify-release` refuses if the
archive disagrees with its manifest, if the manifest names a different version,
or if any artifact digest disagrees with the SHA256 compiled into the engine. The
archive is byte-reproducible: same inputs, same sha256, regardless of staging
path or file timestamps.

**XRT and the MSVC redistributable are never bundled.** They are the user's to
install, and a test asserts they cannot appear in the archive.

## Correctness and quality

Device execution is qualified against a BF16 oracle. That is not a claim of
equivalence with `matmul_qt`, and the two legitimately differ: the current path
accumulates f32 activations against dequantised int4, the lane is BF16
throughout.

What was measured on the real model, both halves:

```
teacher-forced, 24-question internal benchmark, all 96 scoring calls exposed
  primary acc_norm          19/24 baseline -> 19/24 with the NPU
  predicted answers changed  0
  underlying per-option scores  CHANGED on every operation the NPU handled

free-running generation, 8 prompts
  5 of 8 produced different text
```

The first half is not evidence of equivalence and is not offered as such — at
n=24 the confidence interval is roughly 31 percentage points. Both halves are in
`docs/xdna.md` so a user can make the call themselves.

## Performance

**No speed claim is made.** The lane is not chosen on speed or cost and there is
no automatic selection. For context on one commit title: an earlier large cold
integration penalty was not reproduced after the weight-conversion refinement in
`perf(xdna): accelerate prepared weight conversion` (84.20 ms → 20.30 ms per
tensor, byte-identical output). That is a non-reproduction, not a speedup.

## Testing

```
dev 692af87        705 passed / 72 skipped
this branch        749 passed / 72 skipped        make check, exit 0
```

The +44 are all new and XDNA-owned: 19 packaging/integrity/determinism tests, 18
loader and ABI-rejection tests, 7 clean-ownership tests. No skip change, nothing
disabled.

Five C owners (registry, prepared state, QT state, execution, failure — 248
assertions in the failure owner alone) run in ordinary CI against synthetic
helper DLLs the tests build themselves. **No NPU or XRT is needed for any of
this.**

Two of them, execution and failure, are Windows-only: they load the synthetic
helper through `LoadLibraryExA`. The other three — registry, prepared state and
QT state — touch no loader and stay in the default `make check` everywhere, so
on Linux and macOS that run now compiles `backend_xdna.c`, and `colibri.c` with
`-DCOLI_XDNA` for the QT-state owner.

On real hardware, `tests/xdna_physical_probe` drives the production registry,
integrity check, loader, weight preparation, eligibility gates and candidate
function against the device. Both buckets: 6 dispatches, 6 completions, 1 device
open. Above the range: declined, 0 dispatches. Removing the helper, removing an
artifact and flipping one byte in an artifact each fall back to the exact default
result.

## Known limitations

- Windows only.
- One operation family, one shape; `M > 256` uses the normal path.
- Reduced precision; output can differ, and free generation has been observed to
  diverge.
- Explicit opt-in only. No automatic selection, no scheduler, no cost model.
- Needs an external AMD XRT / Ryzen AI runtime and the MSVC redistributable.
- The artifacts cannot be built in CI, so the sidecar is produced by a maintainer
  on a qualified machine.
- No performance claim.

## Why this shape

The lane sits behind an explicit flag rather than hardware discovery because the
consequence is reduced precision, and that is a decision a user should make
rather than inherit from owning a laptop. It is one operation family rather than
a backend because that is what has been qualified end to end — the artifacts,
the integrity chain, the fallback and the packaging — and a wider claim would
outrun the evidence.


Kenneth-Javier:
The remaining red check (`Qwen3.6 tiny oracle (token-exact + ASan/UBSan)`) is a
pre-existing defect that this PR exposes rather than introduces. Full write-up
and reproduction in #1262; the attribution in short:

**The relevant Qwen sources are unchanged by this PR.** `qwen36.c` is
`c7ace3609c259dc0f581b0ba0c54fa95dfd4043516bd52df70092d677ccbc628` on both `dev`
(692af87) and this branch, as are `qwen36_tier.{c,h}`, `st.h`, `json.h` and
`compat.h`. Nothing this PR touches is compiled into that binary.

**What did change is whether that job's sanitizer step rebuilds.** `clean.py`'s
hand-written engine list did not contain `qwen36`, and `qwen36$(EXE)`
(`c/Makefile:1036`) has no build-configuration dependency, so a different
`EXTRA_CFLAGS` alone does not invalidate the target. Measured at 692af87:

```
make qwen36                      -> sha256 a658efd4…   nm | grep -c __asan  ->  0
make clean                       -> "clean: removed 0 files/dirs"
qwen36 after clean               -> present, byte-identical
make qwen36 EXTRA_CFLAGS="-fsanitize=address,undefined …"
                                 -> "make: 'qwen36' is up to date."
                                 -> sha256 a658efd4…   nm | grep -c __asan  ->  0
```

So the step named "Same run under ASan + UBSan" has been re-running the
un-instrumented binary from the step before it. On this branch the same two
commands give `clean: removed 1 files/dirs` and a binary with 33 ASan symbols.

**Forcing a genuinely fresh sanitized build at 692af87 reproduces the defect**,
with the fixture built from `c/tools/oracle-requirements.txt` (same reference
token sequence CI produces):

```
==1169==ERROR: AddressSanitizer: stack-use-after-return on address 0x74ea81707898
READ of size 8 at 0x74ea81707898 thread T32

Matching tokens: 16/16
```

It is intermittent — roughly 3–5 reports per 20 runs, on the base and on this
branch alike — and the run still exits 0 with all 16 tokens matching, which is
why the token-exactness steps could never have caught it.

One caveat for whoever picks it up: **no source frame is available.** The ASan
report consistently stops after the `READ` line even in a local `log_path`
capture with `symbolize=1`, `fast_unwind_on_fatal=0` and `sleep_before_dying=5`,
so the missing frames are not a GitHub log-truncation artifact. I am not
guessing at a faulting line.

I have not touched `qwen36.c`, and I have not weakened the gate, reverted the
`clean.py` ownership or suppressed the diagnostic — any of those would hide a
real defect and put the sanitizer step back to asserting nothing. Happy to
follow whatever order you prefer: if a green board is required before this
merges, #1262 would need fixing first.

JustVugg:
Reviewed. For a default build this is essentially free: not compiled, not linked, and the new field on QT is inside the ifdef so the struct layout does not change. The seam expands correctly in all four ifdef combinations, with no dangling else and no double dispatch. The allocation care is real, both products guarded against overflow and the old staging buffer freed only after the new one succeeds.

`test_xdna_failure` is the right test for a fallback lane: it drives each failure class and asserts the result is bit-identical to `matmul_qt` by memcmp. That is the only assertion that matters for a path whose promise is that it never changes the answer, and it runs without hardware against synthetic helper libraries the tests build.

One change I do want before merge. The two dispatch lines in `moe()` are missing the `!omp_in_parallel()` guard that the Vulkan line immediately above them has. `backend_xdna.c` holds process-wide mutable state, realloc'd staging buffers and a wrapped-identity pair, with no locking anywhere. Tracing the call sites this is latent rather than live today, because the shared-expert tail runs after the parallel region closes. But it is one pragma away from a heap race, and matching the sibling line is two tokens.

Two surface changes not in the body, worth correcting there. The else arm adds a `COLI_XDNA` probe and a stderr line to every build, so it is not quite all-guarded. And three XDNA test owners are not in TEST_EXCLUDE, so a default `make check` on Linux and macOS now compiles the backend and the engine with the define.

One question. `release.yml` builds the shipped Windows host with `XDNA=1`, so every released Windows binary carries the lane and the startup probe. Was shipping it compiled in intended for this first slice, rather than leaving `XDNA=1` to a maintainer build, given the artifacts it needs cannot be produced in CI?

Kenneth-Javier:
Thanks for the careful review.

**The guard** is in `e802919`: both dispatch lines now skip the lane under `omp_in_parallel()`, same as the GPU dispatches beside them, so inside a team `matmul_qt` runs. **The two surface points** are corrected in the PR body: the stderr line a default build prints when `COLI_XDNA` is set, and the three portable XDNA owners that now compile on Linux and macOS.

**On `XDNA=1` in the release build: yes, that was intended.** The sidecar is designed to sit beside the ordinary Windows download. Without the switch, a user who installs both and runs `coli --xdna` gets "this build has no XDNA support", and the sidecar has nothing to plug into. `EXTRA_CFLAGS` cannot stand in for it either, because the `colibri` rule never compiled `backend_xdna.c`.

What it costs the shipped binary, as measured while qualifying this lane:

```
size                       +30,809 bytes
import table               unchanged: same 12 entries, no XRT, no coli_xdna.dll
startup without --xdna     one getenv, nothing printed
default output             bit-identical on real GLM-5.2 (both buckets), with a
                           valid sidecar installed; the helper was never mapped
```

Your underlying point stands, though. Compiling the lane into every release only helps if a sidecar is actually published for that release, and CI cannot build one. As things stand, that is a manual step on a machine with XRT and the AIE toolchain, and it doesn't scale.

It doesn't need to be per release, because almost nothing in the sidecar changes between releases:

- the four artifacts are pinned by SHA256 in `backend_xdna.c`, so new bytes require a code change
- the helper is bound by ABI generation, not by release version, and its source has not changed since it was written
- only the archive and manifest carry the version, and `build_xdna_package.py --release` is pure Python: no XRT, builds nothing

So I'd like to propose, as a **separate follow-up PR after this one merges**:

1. **Build the inputs once.** The helper, the four artifacts and a machine-generated provenance file (source hashes of the helper, ABI version, registry hashes, toolchain versions) go into one inputs zip, built from the bytes already qualified here, with nothing rebuilt.
2. **Pin them in the release workflow.** The Windows release job downloads that zip, checks a pinned SHA256, refuses if the provenance no longer matches the checked-out source, and stamps the sidecar for the tag. It writes outside `dist/`, so it can never end up inside the core archive, and the upload step is extended to carry the manifest and checksum.
3. **Keep releases safe without inputs.** If inputs aren't configured, the step skips with a warning, so an optional sidecar can never block a release. Configured but wrong fails the build.

Manual work then happens only when the helper or the kernels change, as a reviewed PR that updates the pin.

**The one decision that is yours: where should the inputs live?** My suggestion is a pre-release in this repository (a tag like `xdna-inputs-1` doesn't match `v*`, so it never triggers the release build), rather than anything the workflow would fetch from a contributor fork. It would mean accepting bytes built on my machine, which is why the provenance file and the pin are part of the design. If you'd rather keep `XDNA=1` out of the release build until that is in place, that works too. It's one line, plus the two doc paragraphs that describe it.

JustVugg:
The code is ready as I said; holding it one release for a reason that is about the release, not about the PR. It flips the Windows release build to `XDNA=1`, and CI never compiles that configuration, so the first build of it would be the tag's release job. Please add `XDNA=1` to the Windows CI build (the ordinary one, no XRT needed, as you say) so the configuration is proven before it ships; with that green it goes in right after 1.12.1.

Kenneth-Javier:
Done in `6368f28`, and green on the current head `844dfca` ([run](https://github.com/JustVugg/colibri/actions/runs/35906968092/job/107336943607)). The Windows leg of the all-engines job in `ci.yml` now passes the same `make_args: "XDNA=1"` to the same engine loop as `release.yml`. A new step checks the binary rather than the command line: `colibri.exe` must carry the XDNA lane, must not be the default build, and must import neither XRT nor `coli_xdna.dll`. On the runner it reports the same 12 imports as locally. The loader contract tests later in that job now run against this host too (100 run, 0 skipped). I kept it out of `make check`, because `XDNA=1` adds `-DCOLI_XDNA` to the global `CFLAGS`, so `check.yml` keeps covering the default Windows build. Thanks.

JustVugg:
Checked again on `04775f8`. `6368f28` does what I asked: the Windows engine build in CI now uses the same `XDNA=1` as `release.yml`, and the new step verifies the binary itself, lane present and no XRT or helper import. All 29 checks are green and the OpenMP guard from my review is on both dispatch lines.

This merges right after the `v1.12.1` tag, as agreed on the 23rd, so the release keeps the build it was validated on. Nothing else is needed from you. Thanks for the careful follow-through.

Issue #1249, opened 2026-08-27
[Performance]: Dual-NVMe expert streaming on Threadripper PRO 7965WX — a second drive is worth +37.5%, and DIRECT=1 more than doubles the effect


T-Rex-Git-Hub:
### Commit

dd7df2c (v1.8.0)

### Hardware and storage

**CPU:** AMD Ryzen Threadripper PRO 7965WX — 24 cores / 48 threads, Zen 4
**ISA:** avx512_vnni, avx512_bf16, avx512f/bw/cd/dq/vl, vbmi, vbmi2, ifma, bitalg, vpopcntdq, avx2
**Platform:** TRX50, 4-channel DDR5
**RAM:** 123 GB — swap 8 GB, 97 MiB used across the whole campaign (no swap pressure)
**GPU:** RTX 5090, 32 GB — **present but not engaged in any arm below.** All results below are CPU-only.

**Storage — two NVMe on independent controllers, asymmetric bandwidth:**

| | Drive A | Drive B |
|---|---|---|
| Model | Samsung 990 PRO 4TB | Samsung 9100 PRO 2TB |
| Device / mount | `nvme0n1` → `/models` | `nvme1n1` → `/` |
| PCIe | Gen4 x4, 16.0 GT/s (= its own `max_link_speed`) | Gen5 x4, 32.0 GT/s (= its own `max_link_speed`) |
| Filesystem | ext4, `rw,relatime` | ext4, `rw,relatime` |
| Fill after model copy | 12% | 48% |
| Other load | none — dedicated | OS, swap, `/var/lib/docker` |
| iobench O_DIRECT | 6.70–7.06 GB/s | 8.05–8.48 GB/s |

Neither drive is link-downshifted; both negotiate their own ceiling.

**Comparability audit between the two drives:** filesystem and mount options match
exactly. Two asymmetries could not be eliminated and are disclosed — fill level
(12% vs 48%) and competing I/O (drive B carries the OS and swap). Docker
containers were stopped for all arms.

**Container:** `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp` — 142 shards,
**429.3 GB** as reported by `coli doctor` (the README model table says 372 GB).
MTP heads

### Software environment

**OS:** Ubuntu 26.04 LTS, kernel 7.0.0-30-generic, glibc 2.43
**Compiler:** gcc 15.2.0 (Ubuntu 15.2.0-16ubuntu1)
**Python:** 3.14.4
[...]

PR #1270, opened 2026-08-29
fix(mirror): size stripe chunks by measured bandwidth, not 1/nsf

Unknown-Findout:
`mir_pread_striped` splits a coalesced expert read into one chunk per replica and `pthread_join`s all of them, so the read costs whatever the **slowest** leg costs. The chunk size was `len/nsf` regardless of how fast each drive is.

The engine already knows how fast each drive is. It prints it:

```
[MIRROR] probe: primary 5.15 GB/s | mirror 4.92 GB/s | mirror 0.54 GB/s
[MIRROR] 3 drives | read split 48% / 46% / 5% (measured)
```

That split decides **which** drive holds an expert, and was then ignored for **how much** of one each drive reads. A drive correctly weighted down to 5% still got an equal third of every 19 MB stripe.

## Measured

2x NVMe + 1x SATA (990 Pro 5.69, SN850P 5.03, MX500 0.44 GB/s), 19 MB expert, O_DIRECT, one thread per leg as the caller does:

```
2 legs, equal      9.5 / 9.5 MB          join waits  2.0 ms
3 legs, equal      6.33 / 6.33 / 6.33    join waits 12.6 ms
3 legs, weighted   9.69 / 8.56 / 0.75    join waits  1.9 ms
```

End to end on GLM-5.2 int4, n=5 per arm, interleaved on an idle machine:

```
2 legs stock       0.900 tok/s   sd 0.025
3 legs stock       0.666 tok/s   sd 0.010     -26%
3 legs weighted    0.888 tok/s   sd 0.029     +33% over stock
```

Adding a drive the engine had already measured at 10x slower cost **26%**. The same drive weighted is a wash instead of a penalty.

The mechanism shows up in the byte counters independently of the timing - same read **count** either way, 6x fewer bytes to the slow leg:

```
3 legs stock       mirror2 18.30 GB (3353 reads)
3 legs weighted    mirror2  3.00 GB (3196 reads)
```

18.18 / 3.01 = 6.04x, which is the weight ratio.

## Matched drives are unaffected

Equal weights still produce an equal split, and the test pins that. This only changes behaviour where the drives actually differ - it cannot make a matched pair worse.

## Scope

The split is factored into `mir_stripe_plan()` so the arithmetic is testable without fds or threads. It reads only its arguments plus `g_mir_cut`.

`tests/test_mirror_stripe_split.c` checks five properties. Two are negative controls: **equal weights must still split evenly** (without it the suite passes on a plan that always favours replica 0), and **chunk 0 must still land on the routed replica** so the hash keeps spreading first-chunk load. Sizes must also sum to exactly `len` with contiguous offsets for every rotation of `rep` - a gap there is silent corruption, not a slow read.

One thing worth stating about the first assertion. A plain `> equal` on the fast leg **passes on the old equal-split code**, by 2,731 bytes of 4K rounding. I found that by reverting the plan and watching that assertion stay green while its sibling went red, so both bounds are now deliberately far from `equal`. An assertion that survives the bug is not testing for the bug.

Negative control run rather than assumed: with the plan reverted to `len/nsf`, 3 of the 5 fail; restored, all pass.

## Verification

```
gcc + clang       0 warnings in the changed function under -Wall -Wextra
                  -Wconversion -Wsign-conversion -Wshadow -Wcast-align -pedantic
ASan + UBSan      clean over the full C suite (Linux)
python suite      721 tests
all five engines  build
CUDA_DLL=1        links
```

## What is not covered

No macOS or ARM hardware here, so those paths are CI-only from my side. The function has no platform guards and macOS reaches it through `F_NOCACHE` rather than `O_DIRECT`, so the runtime behaviour there is unexercised by anything I ran - clang compiling clean is not the same as it having run.

Tested on Windows 11, i9-14900K, RTX 3090, GLM-5.2 int4 g64.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1561,6 +1561,8 @@ tests/test_corpus_draft$(EXE): tests/test_corpus_draft.c colibri.c st.h uring.h 
 
 tests/test_cap_precedence$(EXE): tests/test_cap_precedence.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
+tests/test_mirror_stripe_split$(EXE): tests/test_mirror_stripe_split.c colibri.c st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h tier.h quant.h sample.h kv_persist.h telemetry.h
+	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 tests/test_v4_hybrid_policy$(EXE): tests/test_v4_hybrid_policy.c deepseek_v4_hybrid.h hybrid_split.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/colibri.c
+++ b/c/colibri.c
@@ -2588,6 +2588,51 @@ static void *mir_stripe_worker(void *a){
     return NULL;
 }
 /* Returns total bytes read (== len) on success, -1 to make the caller fall back. */
+/* Split `len` across `nsf` replicas PROPORTIONALLY to their measured bandwidth,
+ * rather than len/nsf. Stripe i reads sz[i] bytes at off[i] from replica
+ * srep[(rep+i)%nsf]; offsets are contiguous and the sizes sum to exactly len.
+ *
+ * mir_pread_striped joins every stripe, so an equal split makes the whole read
+ * as slow as the slowest drive - and the weights needed to avoid that are
+ * already computed and already in scope. g_mir_cut[] is the same bandwidth
+ * ranking expert_route() uses for whole-expert placement, from
+ * COLI_DISK_WEIGHTS or the startup probe. Before this it decided WHICH drive
+ * holds an expert and was ignored for how much of one each drive reads, so a
+ * correctly down-weighted slow drive still got an equal share of every stripe.
+ *
+ * Measured on 2x NVMe + 1x SATA (990 Pro 5.69, SN850P 5.03, MX500 0.44 GB/s),
+ * 19 MB expert, O_DIRECT, one thread per leg as in the caller:
+ *
+ *   2 legs, equal      9.5 / 9.5 MB          -> join waits  2.0 ms
+ *   3 legs, equal      6.33 / 6.33 / 6.33    -> join waits 12.6 ms
+ *   3 legs, weighted   9.69 / 8.56 / 0.75    -> join waits  1.9 ms
+ *
+ * So adding a drive the engine already knows is 10x slower cost 6.3x on every
+ * striped read, and the same drive weighted is a small net win.
+ *
+ * Split out of the caller so the arithmetic is testable without fds or threads
+ * (tests/test_mirror_stripe_split.c). Everything it reads is an argument except
+ * g_mir_cut, so a test can drive it by setting that alone. */
+static void mir_stripe_plan(int64_t len, int nsf, const int *srep, int rep,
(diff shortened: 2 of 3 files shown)


JustVugg:
Merging. I checked your claims rather than reading them, and they hold.

**The negative control reproduces.** I put the bug back (`len/nsf` in `mir_stripe_plan`) and reran: 3 of 5 fail, exactly as you said, and the right 3. `equal weights still split ~evenly` and the sum/contiguity checks stay green, which is what makes the other three meaningful rather than a suite that fails on any change.

**The default path is untouched.** `mir_pread_striped` returns -1 before reaching any of this unless there are at least two replicas and the read is 4 MB or more, so a single-drive host never enters the changed code. And `g_mir_cut` is cumulative cuts of 256, so `len*w` peaks around 5e9 with a 19 MB expert: no overflow room worth worrying about.

**The telemetry fix is worth more than its line count.** Once a share can round away and get skipped, the stripe index stops being the leg index, so the old `(rep+i)%nsf` would have billed the wrong drive in exactly the configuration this change creates. Catching that while writing the feature that causes it is the harder half.

The result itself is the part I want on the record: adding a drive the engine had already measured at 10x slower **cost 26%**. That is not an obvious failure mode, and it means anyone who added a spare SATA disk to a mirror was quietly paying for it. Your byte counters confirming the timing independently at 6.04x against the weight ratio is what makes it a diagnosis rather than a benchmark.

Two notes on where this lands. It is the experimental half of #1249, which measured a second NVMe on independent controllers at +37.5% and is now in `docs/benchmarks.md`: that report showed symmetric drives paying off, yours explains why asymmetric ones did not. Together they move the README's multi-SSD row from hypothesis to measured. Worth a line in that table once someone reruns the three-drive arm on current dev.

On your own scoping of what is not covered: agreed, and thank you for stating it. `mir_stripe_plan` is pure arithmetic with no platform guards, so macOS differs only in reaching it through `F_NOCACHE`, and the sizes are contract-tested independently of the read path. That is a low-risk gap, but it is a real one and it should be someone with the hardware who closes it, not a compiler.

The observation about the assertion surviving the bug by 2,731 bytes of 4K rounding is the most useful sentence in the PR. I shipped the same class of mistake in #1267: a release test that checked the engine was copied and never that it was built, which is how v1.9.0 went out with no archives.

Unknown-Findout:
Reran this on current dev for the benchmarks row. Rebuilt both the engine and the CUDA dll from dev today so it's not my branch talking.

```
                    tok/s (n=5)        bytes to the SATA leg
2 legs  D+H         1.030  sd 0.041    -
3 legs  D+H+C       1.006  sd 0.034    5.63 GB / 6002 reads
```

So about 2% apart, which is inside the run to run spread. Same box and same prompt that gave 0.666 and 18.30 GB to that leg before the fix.

Honestly the timing is the less interesting half. The SATA share came out 5.61, 5.63, 5.64, 5.64, 6.08 across the five runs. It barely moves. That's the split doing what the probe says it should, and it doesn't care whether the clock was having a good day.

Box: RTX 3090, i9-14900K, 128 GB. Samsung 990 Pro 2TB and a WD_BLACK SN850P 1TB for the two fast legs, third is a Crucial MX500 that probes at 0.54 GB/s next to 5.91 and 5.27. GLM-5.2 int4 g64, 490 token prompt, --ngen 8 --temp 0 --ctx 4096, arms interleaved.

One thing I should mention. My first pass had three runs at 0.68 to 0.80 and I nearly posted them, since 0.68 is close enough to the old 0.666 that it looked like the fix wasn't holding. It was me. I had a recursive grep crawling the same two drives the benchmark was streaming off. The byte split was normal in all three, which is the only reason I caught it instead of filing it. On an idle machine they come back 1.04 to 1.06.

Issue #939, opened 2026-08-11
Very slow cache warmup


biemster:
This might be just a documentation issue, but the docs grew very large and I can't find why I'm running into the following issue:
I've contributed the benchmark numbers for the fully RAM resident Xeon machine #389, improved the numbers ~10% by pulling in the latest version, but after reinstalling colibri and redownloading the exact same model I can't get above 0.2 tok/s anymore. Even using the exact same commands as before.
Also when trying to reproduce #634 I get the same issue.

What I noticed when profiling the machine (either the Xeon or the AWS graviton4) while running colibri, is that the RAM usage climbs rather slow, around 100MB/s. Also the hit rate starts very low on every subsequent invocation, so it seems it has to warm up the cache every time using this slow 100MB/s rate which takes hours.

Since I managed initially to get proper performance and 100% hit rate, what step am I missing? Or did something change in the code in the meantime which I have to take into account in the run parameters?
The setup instructions for full RAM resident machines is hard to find, maybe the documentation can be improved on this.

Secondly, there is the handle `PIN_GB=all`, why wouldn't that load all shards in RAM on startup and immediately have a full cache ready?

JustVugg:
You are almost certainly running the v1.5.0 regression, and your symptoms match it point for point. #885 measured exactly this on v1.5.0: **~60 GB of RAM left unused and 13 points of expert hit rate lost** — the pin budget subtraction shrank the cache, so RAM climbs slowly, the resident set never builds, and every invocation warms up from cold. "Reinstalled and can't get above 0.2 tok/s with the same commands" is the regression's signature, not a step you're missing.

It's fixed on `dev` (#869 + #914 — the reporter of #885 confirmed RSS and hit rate restored), and the release carrying the fix is imminent.

To confirm: post the commit you're on (`coli --version` / `git rev-parse HEAD`). If it's the v1.5.0 tag, either build from `dev` now or wait for the release — your #389 numbers should come back without any parameter changes.

One unrelated note for the fully-resident use case: warmup persistence across invocations depends on the OS page cache, so the first run after a reinstall (which re-downloads the model) is genuinely cold once — but at Xeon RAM sizes that first pass should run at disk speed, not 100 MB/s. The 100 MB/s ceiling is the regression throttling the fill, not your disk.

JustVugg:
**v1.6.0 is out and it is your fix**: https://github.com/JustVugg/colibri/releases/tag/v1.6.0

[...]

PR #1271, opened 2026-08-29
qwen36: vectorise the int4 expert unpack (2.10x CPU decode, bit-exact)

lornecodes:
Every expert cache miss in `qwen36.c` unpacks a packed-int4 expert to int8 in the slot, and the loop doing it was indexed by **element**:

```c
for (i = 0; i < want_w; i++) {
    uint8_t byte = raw[i >> 1];
    int8_t v = (i & 1) ? ((byte >> 4) & 0xF) : (byte & 0xF);
    if (v & 8) v -= 16;
    s->g[i] = v;
}
```

`want_w` is `3 * inter * hidden` = 6,291,456 for Qwen3.6-35B-A3B — 6.29M iterations per miss, each reloading `raw[i>>1]`, doing an `i&1` select, and taking a branch.

Two changes, in order. Walking **bytes** and sign-extending by shifting removes the branch. Vectorising then needs an **interleaving** store, because the two nibble streams land consecutively in the output — which is why no compiler reaches it from the scalar form. Checking the disassembly after the branchless rewrite confirmed **zero vector registers** in the loop. Written explicitly it emits `vst2q` on NEON and `vpunpcklbw`/`vpunpckhbw` on AVX2, with the scalar form kept as the tail and the portable fallback.

### Bit-exactness

Integer throughout, so unlike the float reductions in #442 there is no reassociation question — this is exact by construction rather than by tolerance. In vectors the signed value of a nibble `n` is `(n ^ 8) - 8`; the scalar tail gets the same result by casting the nibble into the top four bits and arithmetic-shifting back down.

Verified identical to the **original branching form** (not merely to the intermediate one): over all 256 byte values, at every length around a vector boundary, and on a full-size 6.29M-value random expert, on both AVX2 and NEON.

### Measurements

Standalone kernel, single-threaded: **11.84 → 0.18 ms** on an i9-12900K, **3.19 → 0.08 ms** on an Apple M4 Pro.

End to end — Qwen3.6-35B-A3B int4-gs64 (`Kreuzzelg/qwen36-35b-a3b-colibri-i4-gs64`), i9-12900K (avx-vnni, no AVX-512), Debian trixie, gcc 14.2, Samsung 990 Pro ext4, container fully in page cache, `cap 32`, `N_NEW=48`, `MemoryMax=40G`, three runs per rung, median:

```
SNAP=$M TOK=$M/tokenizer.json N_NEW=48 ./qwen36 32 4 prompt.txt
```

| | tok/s | decode | TTFT |
|---|---|---|---|
| original (branchy scalar) | 1.66 | 28.7 s | 7.84 s |
| branchless scalar | 2.60 | 18.5 s | 5.21 s |
| **vectorised** | **3.49** | **13.8 s** | **3.93 s** |

**2.10x.** The expert miss count is **6824 in every one of those runs, to the unit**, and the hit rate never moves — placement behaviour is untouched, only what a miss costs. Generated text is character-identical on the real model, and the tiny fixture scores the same as before the change.

### How it was located rather than guessed

Decode time on this model is a straight line in miss count — `time = 11.1s + 2.62ms * misses` predicts LRU, PILOT and pinned configurations within 1.1% — which put ~60% of decode inside the per-miss path and pointed here.

After the change the same fit reads `10.9s fixed + 0.43 ms/miss` — fixed compute unchanged, as it must be for a patch touching only the unpack, and the per-miss term down 6.1x. The unpack is now **21% of decode instead of 62%**, so the CPU bottleneck for this engine has moved to the matmul and attention path.

That the per-miss cost was CPU rather than I/O is confirmed most simply by this patch working: vectorising an integer unpack cannot produce 2.10x on a workload bound by storage. Independently, bracketing the runs against `/proc/diskstats` shows **9.49 GB of device reads per run** — about 1.6 s at this drive's measured 5.78 GB/s, so ~6% of the original 28.4 s decode, and essentially accounted for by the one-time dense-weight load (~9.25 GB) rather than by expert misses.

Possibly of interest for #939 (very slow cache warmup), since warmup is all misses; and as a datapoint for #442, where the same class of win is blocked on float reassociation that does not apply here.

### Checks

- `make check` on the 12900K: **704 tests, OK (skipped=35)**
- Build warnings unchanged: 4 on this branch, the same 4 on pristine `dev` (all pre-existing, in the JSON emission path)
- Disassembly confirms the intended instructions on both gcc/AVX2 and clang/NEON

### DIFF
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -1375,6 +1375,64 @@ static void slot_ensure_allocated(Model *m, Slot *s) {
     s->g4 = s->u4 = s->d4 = NULL;   /* packed int4 (allocated on int4 load if GPU int4 active) */
 }
 
+
+/* Unpack packed signed-int4 experts to int8. Two nibbles per byte; LOW nibble is
+ * element 2k, HIGH is 2k+1, each signed 4-bit two's complement. Must stay in step
+ * with pack_int4 in c/tools/convert_qwen36.py.
+ *
+ * This is the hottest thing on this engine's CPU decode path. Every expert cache
+ * miss unpacks a whole expert -- 3*inter*hidden = 6.29M values on
+ * Qwen3.6-35B-A3B -- and a fit of decode time against miss count put ~60% of
+ * decode inside it.
+ *
+ * The loop this replaces was indexed by ELEMENT, so it reloaded raw[i>>1], did an
+ * i&1 select and took a branch for every value. Walking BYTES and sign-extending
+ * by shifting removes the branch. Vectorising then needs an INTERLEAVING store,
+ * which is why no compiler gets there from the scalar form: the two nibble
+ * streams are consecutive in the output, so it needs vst2q on NEON or
+ * unpacklo/unpackhi on AVX2 rather than a strided store. Checking the
+ * disassembly after the branchless rewrite confirmed zero vector registers.
+ *
+ * Sign extension is branchless in both paths. In vectors the signed value of a
+ * nibble n is (n ^ 8) - 8; the scalar tail gets the same result more cheaply by
+ * casting the nibble into the top four bits and arithmetic-shifting back down.
+ *
+ * Integer throughout, so unlike a float reduction this is bit-exact by
+ * construction rather than by tolerance -- verified identical to the original
+ * branching form over all 256 byte values, at every length around a vector
+ * boundary, and on a full-size 6.29M-value random expert, on AVX2 and NEON. */
+static void unpack_int4_to_int8(int8_t *out, const uint8_t *raw, int64_t n)
+{
+    int64_t nb = n / 2, b = 0;                  /* n is 3*inter*hidden, always even */
+#if defined(__AVX2__)
+    const __m128i m4 = _mm_set1_epi8(0x0F), e8 = _mm_set1_epi8(8);
+    for (; b + 16 <= nb; b += 16) {
+        __m128i by = _mm_loadu_si128((const __m128i *)(raw + b));
+        __m128i lo = _mm_and_si128(by, m4);
+        __m128i hi = _mm_and_si128(_mm_srli_epi16(by, 4), m4);   /* 16-bit shift: mask after */
+        lo = _mm_sub_epi8(_mm_xor_si128(lo, e8), e8);
+        hi = _mm_sub_epi8(_mm_xor_si128(hi, e8), e8);
+        _mm_storeu_si128((__m128i *)(out + 2 * b),      _mm_unpacklo_epi8(lo, hi));
+        _mm_storeu_si128((__m128i *)(out + 2 * b + 16), _mm_unpackhi_epi8(lo, hi));
+    }
+#elif defined(__ARM_NEON)
+    const uint8x16_t m4 = vdupq_n_u8(0x0F);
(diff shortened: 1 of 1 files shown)


c/qwen36.c
```diff
@@ -1388,6 +1388,64 @@ static void slot_ensure_allocated(Model *m, Slot *s) {
...
+    const __m128i m4 = _mm_set1_epi8(0x0F), e8 = _mm_set1_epi8(8);
+    for (; b + 16 <= nb; b += 16) {
+        __m128i by = _mm_loadu_si128((const __m128i *)(raw + b));
+        __m128i lo = _mm_and_si128(by, m4);
+        __m128i hi = _mm_and_si128(_mm_srli_epi16(by, 4), m4);   /* 16-bit shift: mask after */
+        lo = _mm_sub_epi8(_mm_xor_si128(lo, e8), e8);
+        hi = _mm_sub_epi8(_mm_xor_si128(hi, e8), e8);
+        _mm_storeu_si128((__m128i *)(out + 2 * b),      _mm_unpacklo_epi8(lo, hi));
+        _mm_storeu_si128((__m128i *)(out + 2 * b + 16), _mm_unpackhi_epi8(lo, hi));
+    }
```
OPS-NeoRetro, line 1430: 🤔 Why you didn't take advantage of 256 bits for AVX2 though? We need an SSE4.1/SSE4.2/SSSE3 path here.

OPS-NeoRetro (changes requested):
@lornecodes, I felt something weird because I only see SSE4.1, SSE3 or SSE2 intrinsics in the AVX2 path.

OPS-NeoRetro:
@JustVugg, merge #1232 first then take a look at this and #1239

JustVugg:
Merging. I verified the bit-exactness claim independently rather than reading it: a standalone harness comparing your `unpack_int4_to_int8` against the **original branching form** (not the intermediate branchless one) over all 256 byte values, every even length from 0 to 1024 so the vector boundary is crossed in both directions, and a full-size 6.49M-value random expert. Identical byte for byte, rerun on the current head after the rebase.

Kernel timing on my box, gcc -O2 -mavx2: 11.01 ms to 1.00 ms. Your machine's before-figure (11.84 ms) lines up with mine closely enough that the comparison is the same one.

Also checked: only `qwen36.c` is touched, so the blast radius is exactly one engine; token-exact at caps 1, 8 and 16; clean under a genuinely instrumented ASan build; the four qwen36 C tests pass; and the malformed-container refusal still fires.

The part I want on the record is how you found it. A fit of decode time against miss count, `11.1s + 2.62ms * misses` predicting LRU, PILOT and pinned configurations within 1.1%, is a measurement that tells you where to look before you look. And re-fitting afterwards to `10.9s + 0.43ms/miss` — fixed term unchanged, as it must be for a patch touching only the unpack — is the check most people skip.

The 9.49 GB of device reads from `/proc/diskstats`, accounted for by the one-time dense load rather than by expert misses, is what makes "this was CPU, not I/O" a demonstration instead of an assertion.

@OPS-NeoRetro's sequencing was right and has been followed: #1232 is merged, and #1239 has been reviewed — it is red on two defects in its own test scaffolding (a memcmp assertion that only holds when the compiler contracts FMA, and `_sse41` targets that pass x86 flags on arm64), both reported there with the isolation.

PR #1280, opened 2026-08-30
feat(qwen38): image preprocessing, pinned against the official processor

JustVugg:
**This is half of vision, and the half that decides whether it is correct rather than nearly correct.** The tower is not implemented and images are still refused; this is the piece the tower would otherwise be built on top of blind.

I would rather ship it separately than bundle it into a tower PR where it becomes invisible, because it is independently verifiable *today* and the tower is not.

## Why preprocessing is worth its own gate

It is where vision goes wrong silently. A canvas one patch too wide, a different patch order, the other model's normalisation constants — none of that raises. The model answers anyway, and answers worse. So the reference is the real `Qwen2VLImageProcessor`, not a re-reading of the paper.

**It does not need the weights.** The reference builds from `preprocessor_config.json` alone, which is 390 bytes. All of this was developed and verified without downloading the 185 GB checkpoint, which is also why the same approach will work for whoever writes the tower.

## Measured

Eight shapes, chosen to hit every branch of `smart_resize` — below the pixel window so the canvas grows, inside it, far above so it shrinks, and non-square in both directions:

| | |
|---|---|
| geometry and patch order | **identical on all eight** |
| pixels, no resampling (256x256, 640x480) | **0.0000** |
| pixels, with resampling | 0.0157 worst case |

That last figure is Pillow's bicubic against torchvision's, and it is measured and printed rather than declared zero. What makes the tolerance meaningful instead of permissive: **a wrong patch order shows up near 1, not 0.01.** The two exact rows are the real check.

An extreme aspect ratio is refused rather than turned into a degenerate canvas, matching the reference. A 1x5000 image has no sensible representation in square patches, and failing here beats failing further down.

## Two differences from GLM-5.3, and why this is its own file

**Resolution is dynamic.** GLM-5.3 fits everything onto a 448 canvas and pads. Qwen keeps the aspect ratio and picks a canvas whose *area* lands inside `[shortest_edge, longest_edge]`, so there is no padding but the token count depends on the image.

That has a consequence worth stating plainly: **a 1080p photo becomes 2040 tokens.** On an engine that streams experts from disk that is a prefill nobody will sit through, which is exactly why `GLM53_MAX_IMAGE_TOKENS` exists. `preprocess(max_tokens=...)` is the same lever, and it shrinks rather than crops — what is lost is detail, not pieces of the image.

**Normalisation is 0.5/0.5**, not the CLIP constants.

## For whoever picks up the tower

27 blocks, hidden 1152, 16 heads, patch 16, spatial merge 2, projecting to 2560, `gelu_pytorch_tanh`, 2304 learned position embeddings. `deepstack_visual_indexes` is empty, so there is no deepstack merging to reproduce. The prompt side is already settled by the template, which splices images as `<|vision_start|><|image_pad|><|vision_end|>`.

Stacked conceptually after #1279 (tool calling) but independent of it: this branch is cut from dev and shares no code with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,3 +145,4 @@ c/*.obj
 c/kimi_k3_tiny/.vendor_kimi_k3/
 c/kimi_k3_tiny/.coli_usage
 c/qwen38_vision_tiny/
+c/qwen38_mm_tiny/
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,4 @@ c/*.obj
 # vendor modeling cache for the tiny K3 oracle (maintainer tool)
 c/kimi_k3_tiny/.vendor_kimi_k3/
 c/kimi_k3_tiny/.coli_usage
+c/qwen38_vision_tiny/
--- a/c/Makefile
+++ b/c/Makefile
@@ -1151,6 +1151,15 @@ qwen38-vision-check: tests/test_qwen38_vision$(EXE)
 	$(PYTHON) tools/make_qwen38_vision_tiny.py --out ./qwen38_vision_tiny
 	./tests/test_qwen38_vision$(EXE) ./qwen38_vision_tiny
 
+# Il percorso completo: frame IMAGE, torre, sostituzione degli embedding,
+# generazione. Prova che due immagini diverse danno risposte diverse, che e'
+# l'unica domanda a cui una fixture con pesi casuali sa rispondere onestamente.
+.PHONY: qwen38-vision-serve-check
+qwen38-vision-serve-check: qwen38$(EXE)
+	$(PYTHON) tools/make_qwen38_multimodal_tiny.py --out ./qwen38_mm_tiny
+	$(PYTHON) tools/make_edge_tiny_tokenizer.py --vocab-size 64 ./qwen38_mm_tiny
+	$(PYTHON) tests/test_qwen38_vision_serve.py --binary ./qwen38$(EXE) --fixture ./qwen38_mm_tiny
+
 tests/test_qwen38_prefix$(EXE): tests/test_qwen38_prefix.c qwen38.c qwen38_core.h qwen38_nfc.h qwen38_nfc_tables.h st.h json.h compat.h quant.h route_trace.h serve_codec.h tok_unicode.h tok_unicode_o200k.h
 	$(CC) $(NOCUDA_CFLAGS) $< -o $@ $(NOCUDA_LDFLAGS)
 
--- a/c/Makefile
+++ b/c/Makefile
@@ -1141,6 +1141,16 @@ tests/test_qwen38_config$(EXE): tests/test_qwen38_config.c qwen38.c qwen38_core.
 tests/test_qwen38_serve_framing$(EXE): tests/test_qwen38_serve_framing.c qwen38.c qwen38_core.h qwen38_nfc.h qwen38_nfc_tables.h st.h json.h compat.h quant.h route_trace.h serve_codec.h tok_unicode.h tok_unicode_o200k.h
 	$(CC) $(NOCUDA_CFLAGS) $< -o $@ $(NOCUDA_LDFLAGS)
 
+tests/test_qwen38_vision$(EXE): tests/test_qwen38_vision.c qwen38_vision.h st.h json.h compat.h
+	$(CC) $(CFLAGS) -Wno-unused-function $< -o $@ $(LDFLAGS)
+
+# Genera la fixture della torre e la confronta con l'oracolo upstream. Non serve
+# il checkpoint: 240k parametri casuali, meno di un megabyte.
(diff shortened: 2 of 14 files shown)


JustVugg:
Pushed the tower. This PR now has both halves of vision that can be verified without the checkpoint.

## The tower

`qwen38_vision.h`: 27 blocks, hidden 1152, 16 heads, patch 16, spatial merge 2, projecting to 2560. Transcribed from `Qwen4ExpVisionModel`, not deduced from the paper.

`tools/make_qwen38_vision_tiny.py` builds a **240 kB** tower with random weights from the upstream class and records its forward; the C reads the same fixture. `make qwen38-vision-check`. No weights, again.

## Two findings, both invisible until the oracle existed

**The merger uses a different GELU from the blocks.** The blocks take `ACT2FN[hidden_act]` — `gelu_pytorch_tanh` for this checkpoint — while the merger instantiates `nn.GELU()`, the exact erf one. Using one for both matched to **4.6e-3**: close enough to look right, far enough to move the image tokens.

How it was localised rather than guessed: I fed the merger the **reference's own** `last_hidden`. It still disagreed, which separates "the merger is wrong" from "the merger is amplifying an earlier error" — and those two are indistinguishable from the final number alone. That is why the oracle records the output before the merger as well as after.

**A fixture can be too weak to test what it claims to.** My first version scaled the random weights to 0.05. That makes the q·k products so small that softmax comes out essentially uniform: attention degenerates into the mean of the values and stops depending on the scores at all. **A tower with no RoPE whatsoever passed that fixture** — verified by disabling RoPE and watching the test stay green, not feared in the abstract.

At 0.6 the scores have a real range, and all four negative controls now fail as they should: rope off, wrong GELU, raster patch order, flat position interpolation.

## The tolerance is measured, not chosen

The reference in float32 differs from **itself in float64** by 1.83e-4 on these activations. So a gap of that order is the arithmetic of the thing we compare against, not a defect of ours. The threshold sits between that and the 4.6e-3 the real bug produced — deliberately, so it can tell them apart.

Final numbers: pooler 2.5e-4, last_hidden 4.8e-4, merger in isolation 4.6e-5.

## What is left

Wiring the tower into the engine's sequence and accepting images at the gateway. Images stay refused until that exists — an engine that accepts a picture and ignores it is worse than one that says no.

Two things for whoever does it. The prompt side is already settled: the template splices images as `<|vision_start|><|image_pad|><|vision_end|>`. And attention in the tower is full and quadratic in patches, with a 1080p image producing 8160 of them, so the token ceiling is a requirement rather than a convenience.

JustVugg:
Closed the last mile. Images work end to end now.

## The path

The gateway preprocesses the image, replaces the content part with `<|vision_start|>` + N x `<|image_pad|>` + `<|vision_end|>`, and sends the patches in an `IMAGE` frame ahead of the `SUBMIT` they belong to. The engine runs the tower once and substitutes its output for the embedding of each placeholder.

**N is not a constant**, because the resolution is dynamic. The engine **refuses** a request whose prompt and grid disagree rather than guessing which vectors go where: a mismatch means the template and the preprocessing saw two different images, and proceeding would put the right vectors in the wrong positions.

## Three decisions worth stating

**`serve_codec.h` gains an IMAGE command, and five engines share that file.** The change is additive, and the payload is consumed even by an engine that cannot use it — leaving it in the stream would shift the next frame by megabytes and the gateway would read pixels as a header. So the refusal lives in the caller, not the codec.

**The position map is by absolute position, not by offset within a chunk.** Prefill arrives in pieces, and a relative index would hand the second piece the wrong image *while matching silently*.

**The map is dropped at the end of every turn**, for the same reason: a stale one would answer the next request about the previous photo, and would match rather than fail.

## Verified end to end, still without the checkpoint

`tools/make_qwen38_multimodal_tiny.py` merges the text tiny model and a tiny tower into one loadable fixture. `make qwen38-vision-serve-check`.

The gate deliberately does **not** check that the model answers. With random weights it answers regardless, and answers identically while ignoring the picture entirely. It checks that **two different images produce two different answers** — the only question a random fixture answers honestly, and the one that catches the likeliest defect: patches loaded, tower run, result dropped somewhere between the merger and the embeddings. Both refusals are covered too.

    ok   two different images give two different answers
    ok   a byte count that disagrees with the grid is refused
    ok   a prompt with too few placeholders is refused
    ok   a text-only request still works

## Regression, since a shared file was touched

All eight engines build. qwen38 stays token-exact across all eight `Q38_PREFILL_BATCH`/`Q38_NATIVE_BF16`/cap combinations, so the text path is unchanged. The serve-framing tests for qwen38 and DeepSeek V4 pass, and the Python suites are green.

Remote URLs stay refused rather than fetched. One image per request for now: the engine holds a single pending image, and a second arriving before its `SUBMIT` drops the first and says so.

PR #1283, opened 2026-08-30
fix(glm53): expose supported KV slots

donovan-yohan:
GLM-5.3 already implements up to 16 KV slots, but the family registry limits the CLI to one. Align the registry with the engine and pin the limit in the registry test.


JustVugg:
Correct fix: `glm53.c` defines `GLM53_MAX_SLOTS 16` and clamps `KV_SLOTS` to it, and the registry said 1, so `serve` could never use what the engine has.

The branch had gone `CONFLICTING` because #1302 raised the interactive output cap on the same `FamilyLimits` line. I resolved it on your branch as a merge of `dev` (no rebase, nothing of yours rewritten): dev's arguments plus your `max_kv_slots=16`. 48 registry tests green with yours included. Approved the CI runs; if green, this goes into the 1.10.2 patch release.

PR #1284, opened 2026-08-30
build: list the headers each engine includes as Makefile prerequisites

Avicennasis:
### The problem

`make` tracks file timestamps, not `#include` graphs, and `c/Makefile` has no `-MMD`/`-include` dependency generation — each engine rule names its prerequisites by hand. Those lists have drifted behind what the sources actually include.

The failure mode is the quiet one: `make` reports success, changes nothing, and leaves a **stale binary that looks freshly built**.

### Reproduce

```
cd c
make colibri                          # builds
touch edge_runtime.h && make colibri
make: 'colibri' is up to date.        # ...but colibri.c includes edge_runtime.h
```

Touching a header the rule *does* list relinks normally:

```
touch st.h && make colibri
gcc -O3 -march=native ... -o colibri  # relinked
```

So this is specific to the omitted headers, not a general staleness issue.

### What's missing

Every one of these is a header the engine source `#include`s and that exists in `c/`:

| engine | missing |
|---|---|
| `colibri` | 13 — `abl.h` `backend_cuda.h` `backend_metal.h` `backend_vulkan.h` `decode_batch.h` `edge_adapters.h` `edge_runtime.h` `edge_tok_internal.h` `schema_gbnf.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` `tier.h` |
| `inkling` | 8 — `backend_cuda_ink.h` `backend_metal.h` `edge_adapters.h` `edge_runtime.h` `edge_tok_internal.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` |
| `kimi_k3` | 9 — `backend_metal.h` `backend_vulkan.h` `edge_adapters.h` `edge_runtime.h` `edge_tok_internal.h` `hybrid_split.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` |
| `olmoe` | 7 — `edge_adapters.h` `edge_runtime.h` `edge_tok_internal.h` `fused_simd.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` |
| `qwen36` | 6 — `edge_adapter_internal.h` `edge_adapters.h` `edge_runtime.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` |

### The fix

Adds the missing headers to the five engine rules, plus `c/tests/test_makefile_deps.py` so the two lists cannot silently drift apart again. The test compares each engine rule's prerequisites against the headers its source includes, counting only headers that exist in `c/` (system headers and generated files are not prerequisites).

It was checked in both directions rather than just observed to pass — **red** against the unfixed `Makefile`, **green** with the fix — so it discriminates rather than passing vacuously. The existing `unittest discover -s tests -p 'test_*.py'` picks it up with no Makefile change.

`make -C c test-python`: **705 tests, OK (skipped=35)**.

### Notes

Found while merging this into a downstream fork, where an equivalent test was already in place and went red on contact. No behaviour change — prerequisites and one new test only.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -716,7 +716,7 @@ $(file >.build-config,$(BUILD_CONFIG))
 endif
 .build-config: ;
 
-colibri$(EXE): colibri.c cli_args.h st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h quant.h sample.h kv_persist.h telemetry.h route_trace.h omp_tune.h kv_fp8.h kv_tq.h $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) $(VK_SPV) .build-config
+colibri$(EXE): colibri.c cli_args.h st.h uring.h json.h tok.h tok_unicode.h compat.h grammar.h quant.h sample.h kv_persist.h telemetry.h route_trace.h omp_tune.h kv_fp8.h kv_tq.h abl.h backend_cuda.h backend_metal.h backend_vulkan.h decode_batch.h edge_adapters.h edge_runtime.h edge_tok_internal.h schema_gbnf.h segment_adapter_internal.h segment_adapters.h segment_runtime.h tier.h $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) $(VK_SPV) .build-config
 	$(CC) $(CFLAGS) colibri.c $(CUDA_OBJ) $(METAL_OBJ) $(VK_OBJ) -o colibri$(EXE) $(LDFLAGS)
 
 # Vulkan backend object (plain C + vulkan headers) and its SPIR-V shaders.
@@ -997,7 +997,7 @@ fp8-bench: fp8_bench$(EXE)
 # the GPU sits idle. Same guard #783 put on kimi_k3, which since gaining an
 # MXFP4 expert path no longer needs it. tests/test_makefile_cuda_scope.py
 # asserts this shape.
-olmoe$(EXE): olmoe.c cli_args.h st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h omp_tune.h route_trace.h serve_codec.h
+olmoe$(EXE): olmoe.c cli_args.h st.h json.h compat.h sample.h tok.h tok_unicode.h tok_unicode_o200k.h omp_tune.h route_trace.h serve_codec.h edge_adapters.h edge_runtime.h edge_tok_internal.h fused_simd.h segment_adapter_internal.h segment_adapters.h segment_runtime.h
 	$(CC) $(NOCUDA_CFLAGS) olmoe.c -o olmoe$(EXE) $(NOCUDA_LDFLAGS)
 
 # Qwen3.6-35B-A3B engine (hybrid Gated Attention + Gated DeltaNet + streaming
@@ -1019,7 +1019,7 @@ QWEN36_TIER_SRC =
 QWEN36_CFLAGS   = $(NOCUDA_CFLAGS)
 QWEN36_LDFLAGS  = $(NOCUDA_LDFLAGS)
 endif
-qwen36$(EXE): qwen36.c cli_args.h qwen36_tier.h st.h json.h compat.h $(QWEN36_TIER_SRC) $(CUDA_OBJ)
+qwen36$(EXE): qwen36.c cli_args.h qwen36_tier.h st.h json.h compat.h edge_adapter_internal.h edge_adapters.h edge_runtime.h segment_adapter_internal.h segment_adapters.h segment_runtime.h $(QWEN36_TIER_SRC) $(CUDA_OBJ)
 	$(CC) $(QWEN36_CFLAGS) qwen36.c $(QWEN36_TIER_SRC) $(CUDA_OBJ) -o qwen36$(EXE) $(QWEN36_LDFLAGS)
 
 # Qwen3.8-Flash-Next text-only sibling. This target is intentionally CPU-only;
@@ -1071,7 +1071,7 @@ tests/test_qwen36_json_escape$(EXE): tests/test_qwen36_json_escape.c qwen36.c qw
 tests/bench_qwen36_dense_batch$(EXE): tests/bench_qwen36_dense_batch.c qwen36.c qwen36_tier.h st.h json.h compat.h $(QWEN36_TIER_SRC) $(CUDA_OBJ)
 	$(CC) $(QWEN36_CFLAGS) $< $(QWEN36_TIER_SRC) $(CUDA_OBJ) -o $@ $(QWEN36_LDFLAGS)
 
-inkling$(EXE): inkling.c cli_args.h st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h omp_tune.h route_trace.h kv_prefix.h serve_codec.h $(INK_CUDA_OBJ) $(METAL_OBJ)
+inkling$(EXE): inkling.c cli_args.h st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h omp_tune.h route_trace.h kv_prefix.h serve_codec.h backend_cuda_ink.h backend_metal.h edge_adapters.h edge_runtime.h edge_tok_internal.h segment_adapter_internal.h segment_adapters.h segment_runtime.h $(INK_CUDA_OBJ) $(METAL_OBJ)
 	$(CC) $(CFLAGS) inkling.c $(INK_CUDA_OBJ) $(METAL_OBJ) -o inkling$(EXE) $(LDFLAGS)
 
 # ENGINES WITHOUT A CUDA BACKEND (#783).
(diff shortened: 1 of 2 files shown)


JustVugg:
Thanks for this. The defect is real and I reproduced it on `dev` before reading your diff:

```
make colibri                          -> builds
touch edge_runtime.h && make colibri  -> make: 'colibri' is up to date.   (wrong)
touch st.h          && make colibri   -> gcc -O3 -march=native ... relinks (correct)
```

That is the worst shape a build bug can take: `make` reports success, exits 0, and leaves a stale binary that looks freshly built. On your branch the same `touch edge_runtime.h` relinks. Good catch, and thank you for adding a test rather than just fixing the lines.

I have retargeted the PR to `dev`. All PRs in this repo target `dev`; only the release PR goes `dev` -> `main`. That retarget matters here for a second reason, see below.

Two things before I merge.

**1. Two engines are still missing.** Your branch is cut from `main`, which has five engines. `dev` has seven: `qwen38` landed after v1.9.0. Checking every rule against the includes of its own source on `dev`:

| engine | headers included but not listed |
|---|---|
| colibri, inkling, kimi_k3, qwen36, olmoe | none, fixed by this PR |
| glm53 | `backend_vulkan.h` `edge_adapter_internal.h` `edge_adapters.h` `edge_runtime.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` |
| qwen38 | `edge_adapter_internal.h` `edge_adapters.h` `edge_runtime.h` `qwen38_vision.h` `segment_adapter_internal.h` `segment_adapters.h` `segment_runtime.h` |

Behaviourally, on your branch:

```
touch edge_runtime.h && make glm53  -> make: 'glm53' is up to date.
```

so `glm53` still has exactly the bug you are fixing for the other five. (`qwen38_vision.h` missing from `qwen38` is my own doing, I added that header and never added it as a prerequisite. Your test is what would have caught me.)

**2. `ENGINE_RULES` is the same class of bug the PR fixes.** The test keeps the engine list by hand:

```python
ENGINE_RULES = {
    "colibri$(EXE)": "colibri.c",
    ...
}
```

and `_prereqs_by_target()` builds its result by filtering on that dict, so the `assertEqual(set(prereqs), set(ENGINE_RULES))` is only load-bearing in one direction. It catches a rule that was renamed or removed. It cannot catch a rule that was never added, which is why it passes green today while `glm53` is broken. The next engine we add will be invisible to it in the same way.

Could you derive the list from the Makefile instead? Every rule matching `^NAME$(EXE):` that has a `NAME.c` next to it is an engine, and then there is no list to keep in sync. That single change is worth more to us than the header lines, because it keeps working without anyone remembering it.

If you would rather stop at what you have, say so and I will do both follow-ups myself and credit you. But the PR is yours and the test is the good part of it, so you should get the chance first.

CI is green so far on the current head (23 of 24 checks passed, one still running).

Avicennasis:
Both done, and rebased onto `dev` so the two engines that only exist there are in scope. Thanks for reproducing it independently before reading the diff — and for retargeting, which is what surfaced that my branch was a five-engine view of a seven-engine tree.

### 1. glm53 and qwen38

Confirmed your table exactly before changing anything, and it turns out this needs no compiler at all — `make -q` answers the timestamp question without running the recipe, so the whole thing reproduces on a tree that has never been built:

```
touch glm53 && sleep 1 && touch edge_runtime.h
make -q glm53; echo $?
0        # "up to date", though glm53.c includes it
```

After the fix, `make -q` exits 1 — would relink — for every engine against a header that was previously absent from its rule: glm53 (`edge_runtime.h`, `backend_vulkan.h`), qwen38 (`edge_runtime.h`, `qwen38_vision.h`), plus colibri, inkling, kimi_k3, olmoe and qwen36.

### 2. `ENGINE_RULES` is gone

You're right that it was the same class of bug as the one being fixed, and it's worth stating exactly how, because the measurement is unambiguous:

| scenario | old (hand-list) | new (derived) |
|---|---|---|
| fixed Makefile (baseline) | PASS | PASS |
| **glm53+qwen38 unfixed — the real bug** | **PASS** | **FAIL** |
| **a new engine rule omitting `st.h`** | **PASS** | **FAIL** |
| rule syntax changed, parse matches nothing | FAIL | FAIL |

Two things I'd like your eye on, because deriving is not free:

**`iobench` is now covered too.** Your rule as stated — every `NAME$(EXE):` with a `NAME.c` beside it — picks up `iobench`, which is a benchmark tool rather than an engine. I left it in: it's built from one translation unit by the same kind of rule, so it has the identical failure mode and gets the guarantee for free, and its prerequisite list is already complete so it changes no Makefile line. `fp8_bench` is correctly *not* picked up, since it's built from `backend_cuda.cu` and there is no `fp8_bench.c`. Say the word if you'd rather the derivation excluded non-engines and I'll narrow it.

**Deriving can go vacuous in a way listing cannot.** If the rule syntax ever changes, the parse matches nothing, and a header check over zero targets passes green — which reads exactly like a clean tree. That's a worse failure than the one we're removing, so the parse is now asserted non-empty and cross-checked against `FAMILIES`. That reuses the authority you already established in `test_family_registry` rather than keeping a second list here; `deepseek_v4` is excluded by name, since it's built by `Makefile.deepseek-v4` through a phony target and has no prerequisite list here to check.

Two smaller things, both in the silent-failure direction:

- Line continuations are folded before parsing, with the same `re.sub(r"\\\n[ \t]*", " ", ...)` idiom `test_family_registry` uses. No engine rule wraps today; without it a wrapped rule would report headers as missing that are listed on the next line.
- The include regex now tolerates leading whitespace. No engine has an indented `#include "..."` today (0 across all eight sources), but one inside an `#ifdef` is exactly where a new header would land, and missing it fails silently rather than loudly.

`make -C c test-python` is green on the rebased branch.

bherald:
I independently tested head 2f645832840db04ffde8893e3ded94fc17538afb on x86_64 Linux. With `PYTHONPATH=c`, `python3 c/tests/test_makefile_deps.py` passed 2 tests, and a clean `make -C c -j2 kimi_k3` passed. After touching `c/edge_runtime.h`, `make -C c -q kimi_k3` correctly returned 1; a normal make rebuilt `kimi_k3`, after which `make -C c -q kimi_k3` returned 0. I did not load a model or test Windows or the other engines; this is model-free build-dependency evidence only.

Avicennasis:
Thanks — and thank you especially for stating the scope rather than letting it be inferred. One clarification worth making explicit, because the evidence is narrower than it first reads:

`kimi_k3` is one of the five engines fixed by the **first** commit. The two engines this round of review actually asked for, `glm53` and `qwen38`, arrived in the second commit, and you note you did not test other engines — so your run independently confirms the *mechanism* (a listed header relinks, an omitted one silently does not) on an engine that was already covered, rather than the new coverage. That is still the useful half to have from a second pair of hands, since the mechanism is the part I could have been wrong about. The glm53/qwen38 evidence is currently mine alone:

```
touch glm53 && sleep 1 && touch edge_runtime.h
make -q glm53   ->  0  before,  1  after
```

Your `make -q` → rebuild → `make -q` sequence is a better check than mine, incidentally: I only measured the 1, and 1-then-0 also rules out the rule being permanently dirty for an unrelated reason.

**The head you tested has moved.** `dev` gained 9 commits (through v1.10.0) and `c/Makefile` picked up `cli_args.h` on every engine rule, which conflicted with these lists. Rebased onto `68821ec` and resolved as a union rather than by taking a side — upstream's `cli_args.h` plus the headers added here. Re-verified after the rebase:

- derived test: 2 passed, 8 targets discovered
- `make -q` after `touch edge_runtime.h`: relinks for all seven engines
- `make -q` after `touch cli_args.h`: relinks for colibri and glm53 — i.e. the resolution kept upstream's side, not just mine

New head is `a40384d`, so the `2f64583` result is superseded. Sorry for invalidating your run.

**On `PYTHONPATH=c`** — that is a real wart this PR introduced and you worked around it silently. The previous version of the test imported only stdlib and ran standalone; it now imports `family_registry` for the non-vacuity cross-check, so `python3 c/tests/test_makefile_deps.py` gives `ModuleNotFoundError` while the file still carries a `__main__` block advertising exactly that invocation.

It is at least consistent: 7 of the 9 test files in `c/tests` that import from `c/` have the same limitation, `test_family_registry.py` among them, and `make -C c test-python` works for all of them because discovery runs from `c/`. So I have left it alone rather than deviate from the majority. Two lines fix it if @JustVugg would rather it stayed directly runnable:

```python
sys.path.insert(0, str(C_DIR))
```

Happy either way — say the word.

bherald:
I re-ran the model-free checks on current head `a40384de19b1b79c315820d6ecf3db42e345d173` on x86_64 Linux, replacing my superseded `2f64583` result. With `PYTHONPATH=c`, `python3 c/tests/test_makefile_deps.py` passed 2 tests and `python3 c/tests/test_family_registry.py` passed 43 tests. A clean `make -C c kimi_k3` passed; after a clean `make -q` baseline, touching `c/edge_runtime.h` made `make -C c -q kimi_k3` return 1 as required.

I also checked the two newly covered engines without compiling them: after placing newer dummy `glm53` and `qwen38` targets, touching `c/edge_runtime.h` made `make -C c -q glm53` return 1, and touching `c/qwen38_vision.h` made `make -C c -q qwen38` return 1. Cleanup returned the worktree to a clean state.

I did not load a model or test this PR head on Windows. This is current-head Linux build-dependency evidence only.

JustVugg (approved):
You did everything asked, and I verified it rather than taking the diff's word for it.

**Completeness, computed across all seven engines** — every header each `.c` includes, checked against its rule's prerequisites:

```
colibri  glm53  inkling  kimi_k3  olmoe  qwen36  qwen38
missing: none on any of them (0 total)
```

That includes `cli_args.h`, which landed after your first revision — so the derived list picked up a header neither of us listed by hand. That is the point of deriving it.

**Behaviourally**, on your branch: `touch edge_runtime.h && make glm53` now relinks, where on `dev` it answered *"'glm53' is up to date"* and left a stale binary. That silent staleness was the whole bug.

`test_makefile_deps.py` green, 25/25 in CI.

Thanks for coming back to it and for taking the "derive the list, do not hand-maintain it" note seriously — that part outlives the header lines.

PR #1307, opened 2026-08-31
fix(coli): discover serve processes where there is no /proc

BrianHeeseIs:
`coli stop` scans `/proc` to find the serve wrapper and its engine. The scan is guarded against `/proc` being absent, so it does not crash on macOS or the BSDs — it iterates an empty list and prints

```
```

while the serve and its multi-GB engine are both still running.

That is a worse failure than the crash it replaced. This command exists because the engine re-execs itself for OpenMP tuning and no longer carries the name it was started with — which, per its own docstring, is how two ghost engines once OOM'd a box. A silent success is an invitation to walk away from exactly that situation.

### What changed

Discovery is now **ported** rather than merely guarded:

- `/proc` is still used wherever it exists — the Linux path is unchanged.
- Elsewhere, `ps` is asked **once** for the whole process table, and its space-separated output is normalised into the NUL-separated byte form that `_serve_cmdline_matches_port` and `_serve_environ_matches_port` already parse. **Those matchers are untouched**, so port-aware behaviour is identical on both paths.
- The environment is read only for processes whose name already matches a known engine, so the common case adds no work.

An earlier draft called `ps` once per PID and took 38 s; the single-enumeration version runs in 3.6 s.

### Evidence (macOS 15, Apple M3 Max)

Before, with a live `coli serve` running:
```
```

After:
```
would stop 26211: coli serve (cmdline, port 8000)
```

And with nothing running it still correctly reports `nothing running` — no false positives.

### Test

`c/tests/test_coli_stop_portable.py` (stdlib `unittest`, no new deps) plants a decoy `coli serve` and asserts `cmd_stop` discovers it. On the unported code it fails for the right reason — *"Discovery must be PORTED, not just guarded"* — with 3 of its 4 cases already passing, which is precisely the shape of this bug: the crash was guarded, the discovery never ported.

```
Ran 4 tests in 3.576s
OK
```

### Note

`c/tests/test_openai_server.py` already fails on this branch's parent commit (`12a5c46`) and is unrelated to this change — I left it alone rather than fold an unrelated fix into this PR.

---

### Notes for review

**Windows.** Discovery here is POSIX-only by design. `cmd_stop` confirms an engine target by reading another process's environment (`SERVE=1` + `COLI_SERVE_PORT`); neither `tasklist` nor `wmic` can read another process's environment block, and doing it properly needs `NtQueryInformationProcess`/`ReadProcessMemory` or a third-party dependency, while `coli` is stdlib-only. A wrapper-only scan would let `cmd_stop` kill the `coli serve` parent while its engine survived, which is the failure this command exists to prevent. `cmd_stop` reads the pidfile before scanning and that path is `/proc`-independent, so Windows `coli stop` still works for the supported flow. `test_discovers_a_coli_serve_process` is therefore skipped on win32 with that reason inline; the file's other three tests still run there.

**Environment-probe frequency.** The target loop is `if comm in process_names and _serve_environ_matches_port(_stop_environ(pid), a.port)`. Python short-circuits, so `_stop_environ` is invoked once per process whose `comm` is already in `process_names` — the same population `dev` read the environment for — not once per scanned process. What changes is the cost of each of those reads on the no-`/proc` path: `dev` only ever did one `open()` of `/proc/<pid>/environ`, whereas the fallback spawns one `ps eww -o command= -p <pid>` per name-matching process. Per `coli stop` invocation on that path: one `ps -Ao` for the whole table, plus M × `ps eww`, with M the number of name-matching processes. On `/proc` it remains M file reads and no subprocesses. `coli stop` is a one-shot interactive command rather than a hot path, and the full table is requested once rather than per pid. Batching the probes into a single `ps eww -p pid1,pid2,...` is an easy follow-up if the cost matters.

**Over-match test.** `test_does_not_match_the_word_server` runs on every platform, but on win32 it passes vacuously — it asserts the pid is *absent*, and discovery finds nothing there regardless. Its protective value is POSIX-only today.

### DIFF
--- a/c/coli
+++ b/c/coli
@@ -1818,6 +1818,59 @@ def cmd_cluster_worker(a):
     finally:
         stop_heartbeat.set()
 
+def _stop_proc_table(port):
+    """(pid, cmdline, comm) for every process, in /proc's NUL-separated byte format.
+
+    macOS and the BSDs have no /proc, so cmd_stop's scan used to iterate an empty list and
+    report "nothing running" while a live serve and its multi-GB engine kept the box. That is
+    worse than a crash: this command exists to stop the ghost engines that once OOM'd a
+    machine, and a silent success invites the operator to walk away.
+
+    `ps` is asked for everything ONCE rather than per pid, and its space-separated output is
+    normalised to NUL-separated bytes so _serve_cmdline_matches_port keeps working unchanged."""
+    try:
+        pids=[int(d) for d in os.listdir("/proc") if d.isdigit()]
+    except OSError:
+        pids=None
+    if pids is not None:
+        table=[]
+        for pid in pids:
+            try:
+                with open(f"/proc/{pid}/cmdline","rb") as f: cmd=f.read()
+                with open(f"/proc/{pid}/comm") as f: comm=f.read().strip()
+            except (OSError,PermissionError): continue
+            table.append((pid,cmd,comm))
+        return table
+    try:
+        out=subprocess.run(["ps","-Ao","pid=,comm=,command="],capture_output=True,text=True,
+                           timeout=10).stdout
+    except (OSError,subprocess.SubprocessError):
+        return []
+    table=[]
+    for line in out.splitlines():
+        parts=line.strip().split(None,2)
+        if len(parts)<3 or not parts[0].isdigit(): continue
+        pid=int(parts[0]); comm=os.path.basename(parts[1])
+        try: argv=shlex.split(parts[2])
+        except ValueError: argv=parts[2].split()
+        table.append((pid,b"\0".join(a.encode("utf-8","replace") for a in argv)+b"\0",comm))
+    return table
+
+def _stop_environ(pid):
+    """A process's environment as NUL-separated KEY=VAL bytes, /proc or `ps eww`."""
+    try:
+        with open(f"/proc/{pid}/environ","rb") as f: return f.read()
+    except OSError:
+        pass
+    try:
+        out=subprocess.run(["ps","eww","-o","command=","-p",str(pid)],capture_output=True,
+                           text=True,timeout=5).stdout
+    except (OSError,subprocess.SubprocessError):
+        return b""
+    # `ps eww` appends the environment after the command line; keep only KEY=VAL tokens.
+    pairs=[t for t in out.split() if "=" in t and t.split("=",1)[0].replace("_","").isalnum()]
+    return b"\0".join(t.encode("utf-8","replace") for t in pairs)+b"\0"
+
 def cmd_stop(a):
     """Shut down a running `coli serve` AND its engine — one command, no pkill.
     The engine re-execs itself for OMP tuning, so its process is named `exe`,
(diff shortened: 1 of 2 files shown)


JustVugg:
Status: the Windows UCRT64 job is the one red check here, which is also the platform this change is about (discovering serve processes without /proc). Please look at that log and push the fix; the Linux and macOS legs are green.

JustVugg:
Real bug, right fix, and one job away from green.

Twenty-four jobs pass including Linux, macOS and the V4 platform gate. The single failure is Windows UCRT64, in your own new test `test_discovers_a_coli_serve_process`, asserting that discovery must be ported and not merely guarded.

The cause is that the port is for macOS and BSD, which is what the description promises, but the test asserts on every platform. That job installs a native Windows Python, so there is genuinely no `/proc` and `_stop_proc_table` takes the ps branch; the only ps on PATH there is the MSYS2 one, which does not take `-Ao pid=,comm=,command=` and in any case never sees a native Windows process.

Either add a Windows path, tasklist or wmic, or skip that one test on win32 with a comment saying Windows discovery is still pidfile-only. Then this merges.

Worth keeping: `test_does_not_match_the_word_server` documents a real over-match in the existing `_serve_cmdline_matches_port`, and that is a find on its own.

One behavioural note for the body rather than a change. The refactor moved the name check so `_stop_environ(pid)` now runs for every matching process name on every scan, where dev only read the environment after the name matched. On the ps path that is one subprocess per candidate per invocation.

BrianHeeseIs:
First, an apology for the delay — I was on vacation and left this sitting after you'd already done the work of diagnosing it twice. That wasn't fair on you, and thank you for the second nudge rather than just closing it.

Pushed: rebased onto current `dev`, and skipped that one assertion on win32.

### Why the skip rather than a tasklist/wmic path

`cmd_stop` has two kinds of target. The `coli serve` wrapper is confirmed from its command line, which `wmic process get CommandLine` could supply. The **engine** is confirmed by reading *another process's environment* for `SERVE=1` and `COLI_SERVE_PORT` — and neither `tasklist` nor `wmic` can read another process's environment block. Doing it properly on Windows needs `NtQueryInformationProcess` + `ReadProcessMemory`, or a third-party dependency; `coli` is stdlib-only today and I didn't want to change that in this PR.

So a Windows scan could find the wrapper but never port-confirm the engine behind it. That failure mode is specifically bad here: discovery would report the wrapper, `cmd_stop` would kill it, and the multi-GB engine would survive — which is the ghost-engine situation the command exists to prevent. A partial port would also make `test_discovers_a_coli_serve_process` pass while the property its docstring asserts ("discovery must be PORTED, not just guarded") was still false on that platform.

`cmd_stop` reads the pidfile before it scans, and that path doesn't touch `/proc`, so Windows `coli stop` keeps working for the supported flow. Happy to open a separate issue for real Windows engine discovery if you want it tracked.

### On the `_stop_environ` call frequency

Your read of the cost is right, and I've put a note in the PR body. One correction in your favour though, because I don't want to agree to something worse than what's there: the guard is `if comm in process_names and _serve_environ_matches_port(_stop_environ(pid), a.port)` — Python short-circuits, so `_stop_environ` runs once per process whose `comm` is already in `process_names`, not for every process scanned. That's the same population `dev` read the environment for.

What genuinely changed is the **cost of each of those reads on the no-`/proc` path**. On `dev` it was only ever one `open()` of `/proc/<pid>/environ`. The fallback is a `ps eww` subprocess per name-matching process, so per `coli stop` invocation it's one `ps -Ao` for the table plus M × `ps eww`, where M is the number of processes whose name matches a known engine. On `/proc` it stays M file reads and no subprocesses.

If you'd like that tightened I'd rather do it as a follow-up than widen this PR — batching the probes into a single `ps eww -p pid1,pid2,...` is the obvious one.

### On the over-match test

Glad that one earned its keep. Worth flagging its limit honestly: on win32 it passes *vacuously*, because it asserts the pid is absent from the output and discovery finds nothing there anyway. Its protective value is POSIX-only. I left it running on all platforms rather than skipping it too, since it costs nothing and would start genuinely protecting Windows the day discovery lands there.

### Verification

I can't run the Windows job locally, so I reproduced the mechanism instead: stripping `ps` from `PATH` on macOS makes `_stop_proc_table` return an empty table — the same condition native Windows hits — and that reproduces your assertion exactly, `'<pid>' not found in '...nothing running...'`. With the skip applied, the POSIX suite still runs all four tests with none skipped, a simulated win32 load reports `OK (skipped=1)`, and the PATH-stripped repro **still fails** on macOS, which is what proves the skip didn't quietly widen to POSIX.

CI on `51bc55593dedcfc25fd662056b3e570fb354c7c8`: **28/28 green**, including `Windows UCRT64` — `Ran 1042 tests … OK (skipped=146)`, so the test was collected and skipped rather than missed.

PR #1308, opened 2026-08-31
fix(dsv4): report real CPU and memory in HWINFO on macOS

BrianHeeseIs:
`v4_hwinfo_emit()` reads `/proc/cpuinfo` for the CPU name and `/proc/meminfo` for memory. Neither exists on macOS, both `fopen()` calls return NULL, and the telemetry line goes out with zeroes and no error.

The gateway forwards that faithfully, so the dashboard's **Runtime panel renders `unknown` with `0 GB RAM / 0 GB free`** on every Mac — while correctly showing the core count, which is what makes it look like a minor display quirk rather than absent detection.

### Evidence — measured on the engine itself, macOS 15 / Apple M3 Max

Before (pristine `12a5c46`):
```
HWINFO 16 0.0 0.0 0 0.0 unknown|v4-cpu
```

After:
```
HWINFO 16 137.4 105.5 0 0.0 Apple M3 Max|v4-cpu
```

Which matches the machine — `sysctl machdep.cpu.brand_string` → `Apple M3 Max`, `hw.memsize` → 137.4 GB.

### What changed

- Darwin branch using `machdep.cpu.brand_string` and `hw.memsize`.
- Availability delegates to the existing `coli_v4_os_available_memory()` (`deepseek_v4.c:1053`), which already carries a reviewed Darwin branch — free + inactive + purgeable pages, the `MemAvailable` equivalent — rather than duplicating the detection.
- It fills **only what `/proc` could not supply**, so the Linux path is byte-identical and untouched.

### Two details worth flagging for review

**Units.** The Linux branch reports `kB/1e6`, i.e. *decimal* GB, and that is the contract the web UI was built against. This branch therefore uses `bytes/1e9`, not `bytes/2^30`. Using GiB here would have silently disagreed with the Linux path.

**Include placement.** `#include <sys/sysctl.h>` sits directly above its only caller rather than with the other platform includes near the top. This file is an amalgamation compiled once per `-DCOLI_V4_UNIT_*`, and that upper include region is not part of the unit containing `v4_hwinfo_emit` — placing it there produces `call to undeclared function 'sysctlbyname'`. Happy to move it if there is a preferred convention for this.

Builds clean (`make -f Makefile.deepseek-v4 deepseek-v4`, rc=0). No behaviour change on Linux.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -13500,6 +13500,24 @@ extern void coli_v4_expert_store_emit_hits(ColiExpertStore *store);
 extern double coli_v4_expert_store_disk_sec(ColiExpertStore *store);
 extern double coli_v4_expert_store_matmul_sec(ColiExpertStore *store);   /* #890 */
 
+#ifdef __APPLE__
+/* #macos-port: needed by the Darwin branch inside v4_hwinfo_emit below. It sits HERE, next to
+ * its only caller, rather than with the platform includes near the top: this file is an
+ * amalgamation compiled once per -DCOLI_V4_UNIT_*, and that upper include region is not part
+ * of the unit that compiles this function, so an include placed there yields
+ * "call to undeclared function 'sysctlbyname'". */
+#include <sys/sysctl.h>
+#endif
+
+#ifdef __APPLE__
+/* #macos-port: needed by the Darwin branch inside v4_hwinfo_emit below. It sits HERE, next to
+ * its only caller, rather than with the platform includes near the top: this file is an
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Reviewed line by line: the amalgamation include quirk is handled correctly (the duplicated guarded include is one per compilation unit), the Linux path stays byte-identical, availability reuses the existing Darwin branch instead of re-detecting, and the decimal-GB contract of the web UI is respected and documented. CI green including the macOS jobs, which are the ones that execute this.

Issue #1254, opened 2026-08-27
[Feature]: Add DeepSeek-V4-Flash-0731-reap-150b


rssned:
### Problem

While not an official model, it has the same safetensors format as base DeepSeek-V4-Flash-0731 with half the size (experts reduced from 256 to 132 per layer).
The interesting part is (based on author claims) this model scores mostly the same/outperforms base model in certain benchmarks.

**Claimed benchmarks** (author has 4 deepseek variants, REAP-150b is the interesting one)
| Model | Experts | Size | GSM8K | MATH-500 | HumanEval+ | MBPP+ | mean |
| --- | --- | --- | --- | --- | --- | --- | --- |
| base 284b | 256 | 156 GiB | 0.9484 | 0.7060 | 0.8720 | 0.7407 | 0.8168 |
| REAP 200b | 178 | 104 GiB | 0.9401 | 0.6880 | 0.8720 | 0.7407 | 0.8102 |
| REAM 200b | 178 | 104 GiB | 0.8620 | 0.6080 | 0.8841 | 0.7698 | 0.7810 |
| **REAP** **150b** | **132** | **79 GiB** | **0.9295** | **0.7140** | **0.8963** | **0.7593** | **0.8248** |
| REAM 150b | 132 | 79 GiB | 0.6922 | 0.5020 | 0.8537 | 0.7328 | 0.6952 |

**HuggingFace**: https://huggingface.co/puwaer/DeepSeek-V4-Flash-0731-reap-150b

I tried to run it with colibri and `coli` breaks on
```
expert is not two contiguous ranges: layer=0 expert=20
```

I did a quick search in `deepseek_v4.c` and found it fails in `contiguous_group` check.
Using a `stderr fprintf` to see what breaks (after the `qsort` line, before the `if` that fails)
```c
[...]

PR #1310, opened 2026-09-01
Enable support for quantized DeepSeek V4 Flash

zameo94:
Enable support for quantized DeepSeek V4 Flash in c/deepseek_v4.c

## Problem

At the moment, Colibrì's deepseek_v4 file can't run the `DeepSeek-V4-Flash-0731-reap-150b` version. This is a quick explanation:

`
Original problem: the quantized REAP DeepSeek-V4-Flash model (85 GB, FP4 experts) was failing with expert is not two contiguous ranges: layer=0 expert=20 because, in the REAP version's shards, the experts' scales (w1/w2/w3) are not always contiguous and can end up in different shards from the weights.
`

## Summary

`This code has been generated with the huge support of LLMs (Kimi K3 and DeepSeek V4 Flash) and human support`

Updating the `deepseek_v4.c` file, this software is now able to run the quantized version of DeepSeek V4 flash 150B (you can download it [here](https://huggingface.co/puwaer/DeepSeek-V4-Flash-0731-reap-150b)) and the standard version (download [here](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-0731)).

This update was not created for the `puwaer` version only:

| Aspect | Before (standard DeepSeek V4 Flash) | After this change |
|---|---|---|
| 3 expert weights (w1/w2/w3) | contiguous **in the same shard** | contiguous, or read per-matrix (each tensor with its own shard/offset) |
| 3 expert scales (w1/w2/w3) | contiguous **in the same shard** | contiguous, or read per-matrix |
| scales and weights | forced **same shard** | different shards allowed |
| FP4 expert format (I8 packed e2m1 + UE8M0 scale / 32 cols) | same | **unchanged** |

`Why it works with the standard model`: when an expert's tensors are contiguous in one shard, the code takes the identical path as before (per_matrix=false, same shards, same fast direct-I/O). No behavior change → zero regression risk.

`What other quantized variants it unlocks`: any DeepSeek V4 Flash quantized checkpoint that keeps the official FP4 tensor format (expert_dtype=fp4; weights as packed I8 e2m1 nibbles, scales as UE8M0 with one scale per 32 columns; dense/attention weights as FP8 E4M3 + UE8M0) but packs the tensors so that an expert's weights/scales end up non-contiguous or in different shards. This is typical of repacked/sharded uploads (e.g. REAP, FLOCK-style re-packs) and now loads with no other change.

`What quantizations must be for this to apply`:
- Routed experts: FP4 in the official packed-nibble layout (I8 container, e2m1, UE8M0 scale, block size 32 columns). Not a different FP4 encoding, not a different block size.
- Dense/attention weights: FP8 E4M3 + UE8M0 scales (unchanged, already supported).
- Anything else (e.g. FP8 experts, GPTQ/AWQ-style formats, different scale grouping) still fails the existing validators (config.json / validate_matrix) by design — this change does not add new quantization support, only layout flexibility.

## Validation

- [x] `make -C c check`

## Compatibility

##  Benchmark

Final result:

----------------------------------------------------------------------
Ran 704 tests in 93.004s

OK (skipped=42)

Tested on:

M1 Pro, 16 Gb of RAM

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -7255,12 +7255,23 @@ enum { V4_W1 = 0, V4_W2 = 1, V4_W3 = 2, V4_MATRIX_COUNT = 3 };
 typedef struct {
     const ColiSafetensorsTensor *weight[V4_MATRIX_COUNT];
     const ColiSafetensorsTensor *scale[V4_MATRIX_COUNT];
-    int shard;
+    int scale_shard;
+    int weight_shard;
     uint64_t scale_offset;
     uint64_t scale_bytes;
     uint64_t weight_offset;
     uint64_t weight_bytes;
     uint64_t record_bytes;
+    /* Per-matrix fallback when scale (or weight) tensors are split across
+     * shards (REAP-style packed checkpoints). per_matrix=1 selects m_off/
+     * m_len/m_shard directly; the group fields above are ignored then. */
+    int per_matrix;
+    int m_scale_shard[V4_MATRIX_COUNT];
+    int m_weight_shard[V4_MATRIX_COUNT];
+    uint64_t m_scale_offset[V4_MATRIX_COUNT];
+    uint64_t m_scale_bytes[V4_MATRIX_COUNT];
+    uint64_t m_weight_offset[V4_MATRIX_COUNT];
+    uint64_t m_weight_bytes[V4_MATRIX_COUNT];
 } V4ExpertRecord;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


JustVugg:
This is the right shape — the contiguous fast path untouched, per-matrix strictly as fallback — and it answers a real request (#1254). Being upfront about the LLM support is appreciated; the review treats the code on its merits either way.

One thing stands between this and merging: **the new path has no test, and CI cannot see it.** The 25 green checks exercise the contiguous layout only — the per-matrix fallback is reached by no fixture in the repo, so today its only evidence is your machine. That is exactly the kind of green that has burned us before.

Two asks, the first one small:

1. Paste the actual run against the REAP checkpoint: the startup banner, a few generated tokens, and whether the expert-miss counters look sane. That anchors the claim while the test gets built.
2. A fixture test: `tools/make_deepseek_v4_tiny.py` builds the tiny container CI uses. If you extend it with an option that splits an expert's scales into a different shard from its weights (the REAP layout), the existing `deepseek-v4-tiny-check` oracle would then exercise your fallback token-exactly on every CI run. If you would rather not dig into the fixture generator, say so and we will add that part — the feature is worth landing properly.

One more reviewer note: the per-matrix path multiplies reads per expert (2 → up to 6). As a fallback that is the correct trade, but worth a comment in the code saying it is a fallback cost, so nobody later "optimizes" the contiguous path away.

rssned:
@JustVugg, thanks for considering this and @zameo94 for finding the solution.

**I can confirm this model loads and runs.**

Please consider extending this to other engines as this isn't a bug in safetensors format and would allow experimenting with other pruned models.

We can trade *some* quality for *some* speed - in my case loading most of used weights in memory (similar topic, seeing constant 0 reads) didn't helped because my problem seems to be related to matmul being slow (zen2 avx2)

Here's 1. to save @zameo94 some time (and in the case he doesn't have the model on disk)

## Run
```bash
env DIRECT=1 DRAFT=0 URING=1 ./coli run --model ~/models/puwaer/DeepSeek-V4-Flash-0731-reap-150b/ "write djb hash in c, code only"
```

## Result
```
     ▄▀▀▀▄  ▄        colibri v1.10.1
  ▄▄▄▄▀▀▀▀▄▀▀        tiny engine, immense model
      ▀▀▀▀▀▀▀        DeepSeek V4 Flash · 284B MoE · 85 GB on disk
        ▀▀▀▀         run
          ▀          
  ──────────────────────────────────────────────────────────
[OMP] deepseek-v4: 13 compute threads (16 logical CPUs minus 3 expert-loader workers); OMP_NUM_THREADS=<n> overrides, COLI_NO_OMP_TUNE=1 disables
ram_tiers available=42.77GiB dense=resident(6.23GiB) target_slots=57 target_cache=30.52GiB head=resident-bf16 projected=42.62GiB
v4_open index=0.09s plan=0.00s (index built once, reused by plan/inventory/head)
v4_ssd_io mode=direct-aligned fallback=buffered-pread
v4_hot_policy pin_slots_per_layer=16 repin_interval=6 mode=resident-ram rows16=hot-pins
v4_cli mode=chat memory=auto target_only=1
v4_dense_resident layers=43 bytes=6.226GiB
```
```c
#include <stddef.h>
#include <stdint.h>

uint32_t djb_hash(const char *str) {
    uint32_t hash = 5381;
    int c;

    while ((c = *str++)) {
        hash = ((hash << 5) + hash) + c;
    }

    return hash;
}
```
```
<｜end▁of▁sentence｜>

v4_tokens prompt=13 generated=75 total=88 expert_requests=20619 hits=16906 misses=3713 hit_rate=81.992 bytes=49640374272 target_only=1
TUNE decode: 75 tokens in 174.664s
generated_text=[...]
timing time_to_first_token=33.764s after_first=140.897s
[STATS] 20619 selections across 3302 distinct experts -> /home/colibri/models/puwaer/DeepSeek-V4-Flash-0731-reap-150b/.coli_usage
v4_rows16 packed_slots=1055
v4_direct reads=2728 flock_reads=0 fallbacks=985 payload_bytes=34326183936

JustVugg (approved):
Merging on the strength of two things: the code review (contiguous fast path untouched, the fallback packs the slab in the same scales-then-weights layout downstream consumers already expect, 27 per_matrix sites covering index and demand-read both) and rssned's independent run on the real REAP-150b checkpoint -- coherent generated code, internally consistent counters (16906+3713=20619, 82% hit rate), and the read-heavy profile this fallback predicts. That was ask 1. Ask 2, the split-layout fixture test so CI exercises this path forever, becomes our follow-up as promised -- it stays on me, not on you. Thanks for unblocking #1254, and thanks @rssned for running it.

zameo94:
Hey guys, sorry i was busy today. I wanted to dig into and write the test myself. About the run, i had no doubt (I tried standard DeepSeek V4 Flash and the quantized one). Sorry that the test will be in charge of the Colibrì teams. I will be focus on the "Issue" tab, I will be glad to help this project (I have free time now, so glad to help). That been said, thanks for merging my PR @JustVugg

PR #1313, opened 2026-09-01
perf(quant): compute four output rows per pass in matmul_fp8

BrianHeeseIs:
`matmul_fp8` computes one output row per pass. Computing **four**, each with its own accumulator, is **bit-exact** and substantially faster.

### Why it is bit-exact
Each output row's sequence of additions is unchanged — same operands, same column order, same intermediate roundings. Only the *interleaving* of four **independent** accumulator chains differs, and interleaving independent chains cannot alter any row's result. No reassociation, no partial sums, no column blocking, no change to dequantisation or scale application.

The speedup comes from four independent floating-point dependency chains hiding latency, plus one load of `xs[i]` feeding four rows. A tail of fewer than four rows clamps the spare indices onto the last valid row and guards the stores, avoiding a separate remainder loop.

### Measured
Apple M3 Max, prompt of 256 tokens, 40 generated tokens, N=2, both arms deterministic, within-arm spread 0.2%:

| arm | TTFT | decode | tok/s |
|---|---|---|---|
| `12a5c46` | 203.9 s | 49.97 s | 0.7805 |
| this patch | 169.2 s | 41.18 s | **0.9471 (+21.34%)** |

**Bit-exactness is measured, not argued:** the patched arm reproduces the unmodified engine's generated-text md5 `74bb489b74731bc2ce2595c7ede07309` exactly.

Found by attaching `sample` to the decode window: `matmul_fp8.omp_outlined` was 71,430 self-samples here versus 35,291 in a faster fork; after this patch it is 31,223.

### Caveat, stated explicitly
`c/quant.h` is shared with other engines, so this compiles into `colibri` and `kimi_k3` as well. **I have not verified those builds or benchmarked them** — only the DeepSeek-V4 engine is measured. Treat the cross-engine effect as unverified.

Companion change for the FP4 path is in a separate PR; together they measure +36.80% tok/s and -18.0% TTFT.

---

### Update — rebased, and the bit-exactness claim corrected

The original claim above ("bit-exact, measured not argued") was **wrong as this project builds**, and CI was right. The four-accumulator shape is algebraically identical to the one-row loop, but that is a statement about the C abstract machine, not about codegen: under the default `-O3 -march=native`, GCC contracts the multiply-adds differently in the four-chain form and the result drifts by roughly one unit in the last place. The measurement that produced the original figure was taken on an M3 Max with clang and NEON — not the path where the contraction differs.

**Fix:** contraction is now constrained for this kernel only, scoped with `push_options`/`pop_options` so nothing else in the header changes. On x86-64 with AVX2+FMA, `tests/test_qwen38_native_weights` goes from 179 of 258 rows differing to a clean pass.

**Rejected alternative, recorded because it is the obvious one:** rounding each product into a named `float` before accumulating does *not* work — GCC contracts across statements, and it still fails 179 of 258.

**clang is deliberately not constrained.** There the reference and this kernel already contract identically and the test passes; forcing contraction off under clang makes them disagree, and the same test then fails on arm64.

### Performance, scoped honestly

- **arm64** — the guard is invisible to clang (`__clang__` is defined alongside `__GNUC__`), so the compiled binary is byte-identical with and without it. The previously measured +21.34% tok/s therefore still applies unmodified to that path: it is literally the same binary.
- **x86-64** — **not measured.** The guard applies there, and forbidding the fused multiply-add in that loop may return some of the gain. I have no x86-64 host; my only such environment is emulated, so any timing from it would be meaningless.

`tests/test_qwen38_native_weights` passes locally on arm64/clang and on x86-64/gcc.

---

### Update 2 — the fp-contract guard was a no-op; the fix is now compiler-selected

CI was right again, and my previous correction was itself wrong. `Engine (Linux, CPU)` stayed red on `8355ef5`, and the reason is that **the guard I added does nothing**.

**What is actually true about GCC.** `#pragma GCC optimize ("fp-contract=off")` is silently ignored: contraction is resolved as a language/dialect option at parse time, not through the per-function optimize attribute. Measured on gcc 13.4, clean build per cell, against `tests/test_qwen38_native_weights`:

| form | effect |
|---|---|
| `#pragma GCC optimize ("fp-contract=off")` | ignored (no-op) |
| `#pragma GCC optimize ("-ffp-contract=off")` | ignored (no-op) |
| `__attribute__((optimize("-ffp-contract=off")))` | ignored (no-op) |
| `#pragma STDC FP_CONTRACT OFF` | unimplemented — GCC warns `ignoring '#pragma STDC FP_CONTRACT'` |
| command-line `-ffp-contract=off` | works |

Only the command-line flag has any effect, and that is a global numerics decision this kernel should not be making for the whole project.

**Why the earlier evidence misled me.** The `179 of 258` probe was real, but it used the *command-line* flag; translating that result into a source pragma was the invalid step. And the four-row kernel's pass/fail is arch-dependent: with no guard at all it is exact at `-march=znver3` and `-march=x86-64-v3`, but **not** at `-march=haswell`. I had only ever checked arches that pass. Compounding it, `ARCH=` alone does not invalidate object files, so one of my sweeps linked objects built under mixed `-march` and reported the opposite result on identical source. Both of those are my error, not CI's.

**The fix.** The four-row kernel is now selected under `__clang__` only; GCC keeps upstream's one-row kernel verbatim. Under GCC this header preprocesses **byte-for-byte identically to `dev`** (1935 lines, zero diff), so the Linux CI path is provably unchanged rather than merely retested. The diff against `dev` contains **zero deleted lines** — upstream's kernel is untouched.

**Verification**

| gate | result |
|---|---|
| GCC-preprocessed header vs `dev` | identical, 1935 lines |
| gcc x86, clean build: haswell / znver3 / x86-64-v3 | pass |
| clang x86, same three arches | pass |
| arm64/clang full `make test-c` | `rc=0`, all pass |
| arm64 test binary md5 | `4964e2e290eab04d39d0ba326fdd4406` — unchanged |
| pragma/contract warnings | 0 |

The arm64 binary being byte-identical is the point: the measured win is preserved exactly, and nothing on the clang path moved.

**Consequence for the scope of this PR, stated plainly.** As it now stands the speedup lands on clang only; GCC builds get upstream's kernel and no gain. Extending it to GCC would mean compiling the fp8 kernel with `-ffp-contract=off`, which changes numerics beyond this function — your call, not mine. Happy to pursue that separately if you want it, or to leave this as the clang-only win.

### DIFF
--- a/c/quant.h
+++ b/c/quant.h
@@ -523,21 +523,44 @@ static inline int64_t fp8_nblk(int n){ return ((int64_t)n + FP8_BLOCK - 1) / FP8
 static void matmul_fp8(float *y, const float *x, const uint8_t *q8, const float *bscale,
                        int S, int I, int O){
     int64_t nblkI = fp8_nblk(I);
+    /* Four output rows per pass, each with its own accumulator.  Every row's
+       addition sequence is identical to the one-row form - same operands, same
+       column order - so results are bit-identical; only the interleaving of four
+       independent dependency chains differs.  The gain is latency hiding plus one
+       load of xs[i] feeding four rows.  A tail of fewer than four rows clamps the
+       spare indices onto the last valid row and the stores are guarded, which
+       avoids a separate remainder loop. */
     #pragma omp parallel for schedule(static)
-    for(int o=0;o<O;o++){
-        const uint8_t *w = q8 + (int64_t)o*I;
-        int64_t blkO = o / FP8_BLOCK;
-        const float *scl = bscale + blkO*nblkI;
+    for(int o=0;o<O;o+=4){
+        int o1=o+1<O ? o+1 : o, o2=o+2<O ? o+2 : o, o3=o+3<O ? o+3 : o;
+        const uint8_t *w0=q8+(int64_t)o*I,  *w1=q8+(int64_t)o1*I;
+        const uint8_t *w2=q8+(int64_t)o2*I, *w3=q8+(int64_t)o3*I;
+        const float *scl0=bscale+((int64_t)o /FP8_BLOCK)*nblkI;
+        const float *scl1=bscale+((int64_t)o1/FP8_BLOCK)*nblkI;
+        const float *scl2=bscale+((int64_t)o2/FP8_BLOCK)*nblkI;
+        const float *scl3=bscale+((int64_t)o3/FP8_BLOCK)*nblkI;
         for(int s=0;s<S;s++){
             const float *xs = x + (int64_t)s*I;
-            double a=0;
+            double a0=0,a1=0,a2=0,a3=0;
             for(int64_t bi=0; bi*FP8_BLOCK<I; bi++){
                 int base=(int)(bi*FP8_BLOCK); int blen=FP8_BLOCK; if(base+blen>I) blen=I-base;
-                float sc=scl[bi]; float acc=0;
-                for(int i=base;i<base+blen;i++) acc += e4m3_decode(w[i])*xs[i];
-                a += (double)acc*sc;
+                float acc0=0,acc1=0,acc2=0,acc3=0;
+                for(int i=base;i<base+blen;i++){
+                    float xv=xs[i];
+                    acc0 += e4m3_decode(w0[i])*xv;
+                    acc1 += e4m3_decode(w1[i])*xv;
+                    acc2 += e4m3_decode(w2[i])*xv;
+                    acc3 += e4m3_decode(w3[i])*xv;
+                }
+                a0 += (double)acc0*scl0[bi];
+                a1 += (double)acc1*scl1[bi];
+                a2 += (double)acc2*scl2[bi];
+                a3 += (double)acc3*scl3[bi];
             }
-            y[(int64_t)s*O+o]=(float)a;
+            y[(int64_t)s*O+o]=(float)a0;
+            if(o1!=o) y[(int64_t)s*O+o1]=(float)a1;
+            if(o2!=o) y[(int64_t)s*O+o2]=(float)a2;
+            if(o3!=o) y[(int64_t)s*O+o3]=(float)a3;
         }
     }
 }


JustVugg:
The structure of this is exactly right, and the speedup is plausible — but I reproduced the CI failure locally and measured it, and the bit-exact claim does not hold on the standard build.

`tests/test_qwen38_native_weights` (S=2, I=257, O=129, memcmp against a scalar reference) on your branch, gcc -O3 -march=native, x86-64 AVX2+FMA:

```
DIFF s=0 o=0  want=316.075348   got=316.075409
DIFF s=0 o=2  want=106.774582   got=106.774628
...
tot diff: 179 of 258
```

Last-ulp differences, spread across most rows — so not a tail-clamping bug. Your reasoning about per-row addition order is correct **at the C level**; what changes is what the optimizer does with four independent accumulators in one inner loop (FMA contraction / SLP packing differs from the single-accumulator loop the reference compiles to). The C abstract machine agrees with you; the codegen does not.

Why this matters here more than in most projects: the engines' CI gates are token-exact greedy oracles, and a one-ulp logit difference can flip an argmax. Bit-exactness is the invariant that keeps those gates meaningful, so a tolerance-based comparison is not an option for this path.

Two ways forward, either of which we would take:

1. Find a formulation whose codegen stays per-row identical to the scalar loop (for instance, verify with `-ffp-contract=off` locally to confirm FMA contraction is the mechanism — if the diff disappears there, the fix is constraining contraction in this kernel rather than restructuring).
2. Keep the interleave, and bring evidence at the token level instead: run the qwen38 tiny oracle (`make qwen38-tiny-check`) and the Qwen3.6/GLM gates across caps on your branch. If every token-exact gate is green with the new rounding, the maintainer can consider re-pinning the scalar reference — but that is a contract change, and it needs that evidence attached, not the bit-exact claim.

The +21% is worth chasing. Happy to re-run the measurement on anything you push.

BrianHeeseIs:
You're right, and thank you for the precise diagnosis — that saved me a lot of flailing.

My error is narrower than the claim I made: I proved the per-row argument for the **C abstract machine**, then verified it empirically on **arm64/clang only** and wrote it into the PR title as an unqualified property. It isn't one. "Bit-exact by construction" is a statement about codegen, not about source semantics, and I didn't have the evidence for it on the target that matters.

What I can confirm from this side, which is consistent with your reading:

- The arithmetic structure is identical to `reference_fp8_matmul` — f32 accumulation inside each 128-column block, f64 after applying the block scale, same operand casts. So the difference cannot be in the formulation.
- `e4m3_decode` is a plain 256-entry LUT returning `float`, so there's no intermediate-precision subtlety hiding in the dequant.
- On arm64 the cited test passes at `-ffp-contract=off`, `=on` (default) **and** `=fast`. So arm64 codegen agrees regardless, which is why the ARM and macOS jobs are green and only `Engine (Linux, CPU)` fails.

That leaves exactly what you said: the optimizer treats four independent accumulators differently from one, and the C-level reasoning doesn't reach it.

I don't have an x86 AVX2 machine to hand, so I'm setting up an emulated amd64 environment to reproduce the 179-row diff and run your `-ffp-contract=off` check myself. I'd rather not spend your offer to re-run on a guess — I'll come back either with a formulation that's verified byte-identical on x86, or with the token-level evidence for path 2 framed honestly as the contract change it would be.

Also noted, and appreciated: your point that `-march=native` on every x86 job means those green checks never execute the fallback loop in #1314. That's a sharper statement about what the CI can and cannot witness than anything I had, and I'll carry it into how I evidence kernel changes here.

Feel free to close this in the meantime if a stale PR is noise — I'll reopen or resubmit with evidence attached.

JustVugg:
The shape of this is attractive and four rows per pass is the right idea, but CI is contradicting the bit-exactness claim, so it cannot go in as it stands.

`Engine (Linux, CPU)`, `Linux` and `Windows UCRT64` all fail on the same assertion:

```
tests/test_qwen38_native_weights.c:587: check failed: !memcmp(want,got,sizeof want)
```

That test is the gate for exactly what the PR title promises. A four-row kernel is bit-exact only if each output row folds its groups in the same order and with the same roundings as the one-row path: holding a decoded weight tile across four rows is free, but changing the accumulation order or fusing a multiply-add that used to round twice is not, and either shows up as a memcmp failure like this one.

Two things would make this reviewable:

1. Say which of the two it is. If the difference is the accumulation order, the fix is to keep the per-row fold identical and only hoist the decode. If it is intentional, then the claim in the title has to become "close, not exact", and the PR needs a numerical argument instead of a memcmp.
2. Rebase. The run is from 2026-09-02 and `dev` has moved a long way since, so the numbers would want re-measuring anyway.

The +21.34% is worth chasing, this is not a rejection of the idea. Please push a revision and I will look again.

JustVugg:
The optimisation is reasonable and the algebra in your description is correct as written. The headline claim is not true as this project builds, and that is what the CI is telling us.

`tests/test_qwen38_native_weights` fails on Linux and on Windows UCRT64. The failing assertion is `c/tests/test_qwen38_native_weights.c:587`, a memcmp of `q38_weight_matmul` against an independent reference in the same translation unit.

Reproduced and isolated: on current dev the test passes; with this patch applied it fails; rebuilding the same patched tree with `-ffp-contract=off` makes it pass again. So the four-accumulator shape is algebraically identical but under the project's default `-O3 -march=native` GCC contracts the multiply-adds differently in the four-chain form than in the one-chain form, and the result drifts by about one unit in the last place.

That matters beyond the fixture, because `c/qwen38_core.h`, `c/colibri.c` and four call sites in `c/deepseek_v4.c` all route through this function. And the +21.34% was measured on an M3 Max with clang and NEON, which is not the path where the contraction differs.

Two ways forward and I am happy with either. Force the contraction off for this function so the bit-exact claim holds as we build it. Or drop the claim and treat it as a new numeric contract, in which case the contract string the test pins needs to change deliberately rather than by accident. Rerunning CI will not help either way.

BrianHeeseIs:
First, an apology for the delay — I was on vacation, and you'd already diagnosed this twice before I got back to it. Thank you for the persistence rather than just closing it.

Rebased onto current `dev`, and you were right about the mechanism — I've constrained contraction rather than restructuring.

### The fix

`#pragma GCC optimize ("fp-contract=off")`, scoped around this kernel only with `push_options`/`pop_options` so nothing else in the header changes. On x86-64 with AVX2+FMA the cited test goes from 179 of 258 rows differing to a clean pass.

I set up an emulated x86-64 environment to check this rather than guess, and reproduced your numbers exactly first — `want=316.075348 got=316.075409`, 179 of 258 — then confirmed `-ffp-contract=off` clears it, which is the experiment you suggested.

**One thing worth recording, because it is the first fix a reviewer would reach for:** rounding each product into a named `float` before accumulating does *not* work. GCC contracts across statements, so it still fails 179 of 258. Only constraining contraction does it.

### clang is deliberately left alone

The guard is `#if defined(__GNUC__) && !defined(__clang__)`. Under clang the reference and this kernel already contract identically and the test passes; forcing contraction off there makes them disagree and the same test fails on arm64. I tried the portable version first and it broke the platform that was already correct.

### What that means for the number in the title

I've dropped the bare percentage from it, because you were right that the measurement was taken where the bug does not manifest.

- **arm64** — the guard is invisible to clang, so the binary is *byte-identical* with and without it (same md5, and clang's preprocessor emits zero occurrences of `fp-contract`). The +21.34% I measured on an M3 Max therefore still applies unmodified to that path. It is the same binary.
- **x86-64** — **unknown, and I can't measure it.** The guard applies there, and forbidding FMA in that loop may give back some of the gain. My only x86 environment is emulated, so any timing from it would be meaningless.

You offered to re-run the measurement on anything I push. If that offer still stands I'd gratefully take it for the x86 number, since you have the hardware and I don't. If it turns out the constraint costs most of the gain there, this may be worth landing for arm64 alone or not at all — your call, and I'd rather find that out than quote a figure I can't defend.

`c/tests/test_qwen38_native_weights` passes on both arm64/clang and x86-64/gcc locally.

BrianHeeseIs:
Still running the last tests before pushing an update, bear with me, it's coming either this evening or tomorrow morning :)

BrianHeeseIs:
Thank you for actually standing up an x86-64 environment and reproducing it rather than taking my word for it — and yes, that is the experiment, and your result matches mine exactly.

I have to flag two things before you spend any more time on this, because one of them is a mistake I already made and the other is a trap in the method we are both now using.

**1. The flag works. The pragma does not — and that is what kept CI red.**

The `179 of 258` probe used the *command-line* `-ffp-contract=off`, which is sound. My error was translating that into a source pragma in `quant.h`. GCC silently ignores it. Measured on gcc 13.4, clean build per cell, against `tests/test_qwen38_native_weights`:

| form | effect |
|---|---|
| `#pragma GCC optimize ("fp-contract=off")` | ignored (no-op) |
| `#pragma GCC optimize ("-ffp-contract=off")` | ignored (no-op) |
| `__attribute__((optimize("-ffp-contract=off")))` | ignored (no-op) |
| `#pragma STDC FP_CONTRACT OFF` | unimplemented — GCC warns `ignoring '#pragma STDC FP_CONTRACT'` |
| command-line `-ffp-contract=off` | works |

Contraction is resolved as a language/dialect option at parse time, not through the per-function optimize attribute. So "constrain contraction for this kernel only, from source" is not achievable on GCC. My previous update to this PR claimed otherwise; it was wrong.

**2. An emulated x86-64 host can give you a false green here. It gave me one.**

In my emulated container `-march=native` resolves to `znver3`, and the four-row kernel passes there **even with no guard at all**. The divergence only appears at `-march=haswell`:

| | haswell | znver3 | x86-64-v3 |
|---|---|---|---|
| upstream one-row | pass | pass | pass |
| four-row, unguarded | **fail** | pass | pass |
| four-row + command-line `-ffp-contract=off` | pass | pass | pass |

Reproducing the probe does not imply reproducing the test failure — different paths, different data. If you are checking a candidate fix, please pass `-march=haswell` explicitly rather than relying on `native`.

One more trap worth having in writing: `ARCH=` alone does not invalidate object files, so switching arch without `make clean` links objects built under mixed `-march`. That produced *opposite results on identical source* for me twice before I noticed.

**Where that leaves the PR.** As pushed, the four-row kernel is now selected under `__clang__` only and GCC keeps your one-row kernel verbatim. Under GCC the header preprocesses byte-for-byte identically to `dev` (1935 lines, zero deletions), so the Linux path is provably unchanged, and `make test-c` at `-march=haswell` is green end to end. The cost is that the win is clang-only.

If you would rather have it on GCC too, the only mechanism that works is `-ffp-contract=off` at the build level, which changes numerics beyond this one function. That is your call, not mine — I did not want to make a global numerics decision inside a kernel patch. Happy to prepare either shape.

PR #1314, opened 2026-09-01
perf(dsv4): four output rows per pass in coli_fp4_matvec_rows16_order, preserving 13b8f16's accumulation order

BrianHeeseIs:
`13b8f16` deliberately traded throughput for determinism so the reference matvec accumulates in the same order as the optimized rows16 path, recording 155 against 235 GFLOP/s and roughly 6-9% end-to-end.

**That ordering guarantee can be kept while recovering most of the speed.**

### Why it is bit-exact
Processing four output rows per pass, each with its own accumulator, leaves every row's addition sequence untouched — same operands, same column order, same intermediate roundings. Only the interleaving of four independent chains differs. No reassociation, no column grouping, no change to scale placement. The gain is latency hiding plus one activation load feeding four rows.

### Measured
Apple M3 Max, 256-token prompt, 40 generated tokens, N=2, deterministic. With the companion `matmul_fp8` change:

| arm | TTFT | decode | tok/s |
|---|---|---|---|
| `12a5c46` | 203.9 s | 49.97 s | 0.7805 |
| both changes | 166.9 s | 36.51 s | **1.0683 (+36.80%)** |

Output md5 identical to the unmodified engine (`74bb489b...`). `c/tests/test_deepseek_v4` passes, including `rows16 convergence: ok (hot and cold bits identical)`.

The quoted figure is for **both** changes together; applied alone this recovers part of it.

### Relevant to your determinism decision
After this widening, the measured cost of the `13b8f16` guarantee on this hardware drops from the 6-9% you recorded to about **3.4%** — I built and measured a variant that reverses the routing to check. **The property you paid for became substantially cheaper, which argues for keeping it.** Details in the linked issue.

### DIFF
--- a/c/deepseek_v4.c
+++ b/c/deepseek_v4.c
@@ -15965,19 +15965,52 @@ void coli_fp4_matvec_rows16_order(float *y, const uint8_t *q4,
     }
 #else
     #pragma omp parallel for schedule(static)
-    for (int o = 0; o < O; o++) {
-        const uint8_t *w = q4 + (int64_t)o * rb;
-        const uint8_t *scl = e8s + (int64_t)o * ng;
-        float sum = 0.0f;
+    for (int o = 0; o < O; o += 4) {
+        int o1 = o + 1 < O ? o + 1 : o;
+        int o2 = o + 2 < O ? o + 2 : o;
+        int o3 = o + 3 < O ? o + 3 : o;
+        const uint8_t *w0 = q4 + (int64_t)o * rb;
+        const uint8_t *w1 = q4 + (int64_t)o1 * rb;
+        const uint8_t *w2 = q4 + (int64_t)o2 * rb;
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Verified beyond CI, because CI cannot see this change: the modified loop is the non-AVX2 fallback, and every x86 CI job compiles with -march=native, so the green x86 checks never execute it. I rebuilt deepseek_v4 with EXTRA_CFLAGS=-mno-avx2 on dev and on this branch and ran the token-exact tiny oracle on both: PASS on both, including the KV prefix-reuse checks. Together with the green macOS job (arm64, which does run this path), that is the evidence the sibling PR #1313 is still missing. The accumulation-order reasoning in the comment is honest about what it preserves, and the tail clamping with guarded stores is correct for O not divisible by 4.

PR #1290, opened 2026-08-31
Add MLX affine qpack reader, inspector, and packed affine Metal GEMV

Avicennasis:
## What this adds

Support for reading and executing MLX affine-quantized qpack containers on the Metal backend, in four commits:

1. **`c/affine_quant.h`** — the MLX affine (unsigned `scale*q + bias`) Q4/Q8 quantization contract with a CPU reference oracle, distinct from the existing signed-int4 multiplicative fmt=4 path. 214-line test.
2. **`mm_affine_q4` / `mm_affine_q8` Metal kernels** plus a validated host dispatch API in `c/backend_metal.{h,mm}`, behind explicit capability checks. Pipeline failures fail hard; there is no silent fallback.
3. **Strict qpack v1 reader** (`c/qpack.{c,h}`) with hard limits on layers/experts/sections and a 412-line test.
4. **`qpack-inspect`** utility for validating containers offline.

## Why

qpack is the packed distribution format used by [Swiftlet](https://github.com/leonickson1/Swiftlet) for MLX-affine Qwen MoE weights. This lets colibri consume those containers and run packed affine Q4/Q8 directly on Apple GPUs without int8 expansion — complementary to the CPU packed-int4 expansion work merged in #1271.

The affine GEMV kernels are adapted from Swiftlet commit `b3a0467` (Apache-2.0, same as colibri); `THIRD_PARTY_NOTICES.md` is extended accordingly.

## Validation

On an Apple M4 Max (macOS 26.6.2, native arm64), `make metal-test`:

- affine Q4 and Q8, each with F32/F16/BF16 scale/bias tensors: zero error vs the CPU oracle;
- handle reuse, descriptor-mismatch rejection, and resident-byte accounting asserted per case;
- the existing kernel battery (rmsnorm, add, silu_mul, int8/int4 GEMV, fmt=4 grouped-int4, fp8 LUT) unaffected and passing.

Cross-implementation check: a container generated by Swiftlet's own `swiftlet-repack` from its tiny fixture is accepted by `qpack-inspect` and reports the expected v1 layout, affine metadata, expert layout, and file sizes. CPU-side tests (JSON, qpack reader, affine oracle, inspector) also pass on Linux x86_64.

## Scope note

This is deliberately infrastructure-only: no engine consumes the qpack reader or affine kernels yet. Qwen3.6 orchestration wired to this path is in development and will follow as a separate PR; keeping the format/kernel contract reviewable on its own seemed preferable to one large change. Happy to restructure if you'd rather see them together.

PR #1315, opened 2026-09-01
Add resumable, hash-verified qpack installers for Hugging Face and static mirrors

Avicennasis:
## What this adds

Two stdlib-only tools under `c/tools/` that download a Swiftlet-format qpack container straight into its final on-disk layout, with resume and hash verification — the "download half" of qpack support. Consuming the installed container needs the reader in #1290; this PR is code-independent of it (the only shared token is the `qpack.hashes.v1` schema string) and can land in either order.

1. **`qpack_install_policy.py`** — the durability contract: a resumable install journal bound to an immutable source revision (cross-revision resume is refused), exclusive install sessions via POSIX locking, exact-range resume of partial files, manifest-last publication with directory `fsync`, and path / symlink / Windows-reserved-name validation. qpack v1 carries sizes but no digests, so integrity comes from transport validators plus a locally persisted full-file SHA-256 — the docstrings say exactly that.
2. **`qpack_http_install.py`** + **`qpack_mirror_install.py`** — one engine, two thin front ends: a commit-pinned Hugging Face adapter (resolves a revision, streams ranges into place, `If-Match` restart, no-ETag hash-authoritative resume) and a static HTTPS/R2 mirror installer keyed on a `hashes.json` index (strict legacy-Swiftlet and `qpack.hashes.v1` parsing, optional out-of-band SHA-256 pin of the index itself, credentials only via an explicitly named env var, stripped on cross-origin redirects).

Follows the existing per-format downloader pattern in `c/tools/` (`download_glm52.py`, `download_fp8.py`); no new dependencies; `THIRD_PARTY_NOTICES.md` gets its own section for the Swiftlet-derived format details. Neither commit rewrites what the other introduced.

## Validation

- 53 deterministic tests (18 policy, 21 HTTP, 14 mirror), green on Linux and on macOS (Darwin arm64) — the macOS run covers the real POSIX-lock and directory-fsync paths. Full `make test-python` on current `dev`: 778 tests, OK.
- Real transfer: the 16.9 GB Qwen3.6-35B qpack from Swiftlet's public R2 mirror, with the `hashes.json` digest pinned out-of-band (verified from two network vantages). Deliberately `SIGTERM`ed at 2.6 GB; the restart resumed the in-flight file exactly (`resumed_files: 1`) and completed hash-verified. Re-running against the finished install is a no-op (`already_complete: true, downloaded_bytes: 0`) and leaves every file's size and mtime untouched.
- Not claimed: Windows crash durability (validated by design and tests only, not on a Windows host).


Avicennasis:
Pushed `d76e7be` — `fix(tools): accept a qpack manifest that does not declare layout.json`.

**Defect.** Run against the production Qwen3-Next-80B container on Swiftlet's public mirror, both installers refused before writing a byte: `qpack mirror install failed: qpack manifest does not declare packed_experts/layout.json`. That container's `manifest.json` sizes 49 files (`model.safetensors` + 48 layer blobs) and nothing else; `layout.json`, `config.json` and the tokenizer files are indexed only by `hashes.json` (61 entries). The 35B container from the same writer declares `layout.json` and its aux files, which is the only shape the installer had been tested against. Swiftlet's own reader never consults the files table for `layout.json`, and the C reader takes the same position (verify when declared, don't require).

**Fix.** `parse_manifest` no longer requires a `layout.json` entry; a declared entry is still size-checked at commit and before publication. Because the reader opens `packed_experts/layout.json` directly, relaxing the policy alone would have produced an install that "completes" without it — so the HTTP engine now fetches it as a required sidecar whenever the manifest omits it: on a mirror it must be covered by `hashes.json` (checked before any transfer, like `config.json`) and is hash-verified; on Hugging Face a 404 refuses. It must be a non-empty JSON object, its name and `.part` name are reserved, and it is re-verified offline on an already-complete install. Every runtime blob stays required.

**Tests (53 → 64).** Fixtures are the exact published bytes of both containers' `manifest.json` + `hashes.json` (pins `2133857f…` / `1605766290…`), with a fixtures-scoped `.gitattributes` so a CRLF checkout can't break the pins. Covered: the real 80B index is accepted and the session opens (was refused after two requests); the real 35B index still is; the undeclared sidecar is fetched, recorded and re-verified, and refused when absent from the index or the mirror; declared `layout.json` with a wrong size refuses on `Content-Length` and on a sized index; a missing layer blob refuses on coverage and on 404; every request on both paths (revision lookup, index, manifest, sidecars, ranged resume) carries `User-Agent: colibri-qpack-install/1.0`, which the mirror requires (it answers 403 to `Python-urllib/*`). `make test-python`: 789 OK on Linux, 728 OK on macOS; ruff clean.

**Live check** (macOS, 80B mirror, index pinned, stopped after the first weight to avoid the 42 GB transfer): the previously refused command logged `installed model.safetensors (1357687320 bytes)` with the local SHA-256 matching `hashes.json`, had `packed_experts/layout.json` (2130 bytes) and all sidecars on disk, and was streaming `packed_experts/layer_00.bin` when killed at 34 s. Probe directory deleted afterwards.

Not addressed here: the Windows UCRT64 lane was already red on `41ef7b9` with 21 `OSError: [Errno 9] Bad file descriptor` errors from `commit_file` (`os.fsync` on an `O_RDONLY` descriptor, which Windows rejects) — a separate, pre-existing defect not addressed by this commit.

Avicennasis:
Follow-up on the Windows lane: the UCRT64 job was already red on this PR's first head (`41ef7b9`) — all 21 `commit_file` paths failed with `OSError: [Errno 9] Bad file descriptor`. Root cause: `os.fsync` maps to `_commit` on Windows, which rejects a read-only handle; POSIX flushes a read-only descriptor fine. `f62549c` opens the partial `O_RDWR` on `nt` only (directory fsync was already skipped there).

Verified on real Windows CI rather than reasoned about: I ran this exact head (`f62549c`) through the full `CI` + `check` workflows on my fork — **25/25 jobs green**, including `Windows UCRT64` and `All engines (windows)` — https://github.com/Avicennasis/colibri-contrib/actions/runs/33673457835 and https://github.com/Avicennasis/colibri-contrib/actions/runs/33673458024. Fork-side draft used only to trigger CI; closed without merging.

Net for this PR since filing: `layout.json` is verify-when-declared and fetched as a required sidecar when the manifest omits it (exposed by the production Qwen3-Next-80B container), the User-Agent is asserted on every request path, and the Windows durability path now actually runs. 64 tests in the qpack suite; Linux `make test-python` 789 OK; macOS 728 OK.

Avicennasis:
Rebased onto `dev` at `f58a2679` (the #1336 merge): `feat/qpack-installers` `f62549ca` -> `6b6106a7`. The four commits replayed with no conflicts (`git range-diff` shows all four as `=`; `dev` had not touched any of the 13 files), so the diff against `dev` is unchanged at 13 files, +3737.

Local gates on the rebased head (Linux x86-64, gcc 13.3.0, Python 3.12.3):

- `make -C c check` exit 0: `test-c` ran 107 gate binaries through `tools/run_tests.py` with no `FAILED:` line; `test-python` `Ran 797 tests`, `OK (skipped=35)`.
- From `c/`: `python3 -m pytest -q -p no:cacheprovider tests/test_qpack_install_policy.py tests/test_qpack_http_install.py tests/test_qpack_mirror_install.py` -> 64 passed, 76 subtests passed. The same command under `python3 -O` -> 64 passed, 76 subtests passed.

The earlier `check` workflow run on `f62549ca` (33649671092) had been cancelled while the `CI` run (33649671138) passed, which is why the PR showed as unstable. This push queued fresh runs for `6b6106a7` (`CI` 33908776419, `check` 33908776432); both currently show `action_required`, so they need a maintainer approval to start.

Avicennasis:
Rebased onto `dev` `a90bed95` (the #1326 merge): `feat/qpack-installers` `6b6106a7` -> `3ea116fd`. The four commits replayed with no conflicts (`git range-diff f58a2679..6b6106a7 origin/dev..HEAD`: all four `=`); `git diff --stat origin/dev..HEAD` is still the PR's 13 files, +3737/-0, and none of the 354 `dev` commits since `f58a2679` touch them.
Gates at `3ea116fd` on Linux x86-64 (Python 3.12.3, pytest 9.0.2), from `c/`:
- `python3 -m unittest discover -s tests -p 'test_qpack_*install*.py'`: `Ran 64 tests`, OK.
- `python3 -m pytest -q -p no:cacheprovider tests/test_qpack_install_policy.py tests/test_qpack_http_install.py tests/test_qpack_mirror_install.py`: 64 passed, 76 subtests passed.
- `make -C c test-python` (the full Python suite): `Ran 1074 tests in 271.183s`, `OK (skipped=92)`.
- `python3 -m unittest tests.test_makefile_deps`: 2 tests OK; `git diff --check` clean.
No C sources change, so no `test-asan` or Apple/Metal targets were run for this branch.

JustVugg:
Thank you for the care in this: the resume policy, the manifest-last completion and the hash verification are the right shape, and the fixtures being byte-exact production files is exactly how it should be pinned.

I am holding it rather than merging it, for one reason only: it is the download half of a format colibri cannot consume yet. Merged today it would put two installers in `c/tools/` whose output no engine can open, and the tools README would advertise a path that stops at the download. The right moment for this is together with the consumer, in the same release, so that a person who runs the installer can also run the model.

If you want to take the consuming half too, open with the container reader and we review the pair; otherwise this stays open and I will come back to it when qpack support lands on the engine side.

PR #1316, opened 2026-09-01
fix(planner): honor cgroup memory limits

Avicennasis:
## Summary

`resource_plan.memory_available()` read host `/proc/meminfo` `MemAvailable` only. Inside a container with a cgroup memory limit that is the host's number, so the planner admitted a model the cgroup could not hold and the process was OOM-killed at the limit instead of being refused up front. (`c/tools/bench_*` and the CUDA path already reason about container limits elsewhere; the planner did not.)

The change makes the planner's memory figure `min(host MemAvailable, remaining finite cgroup budget)`:

- cgroup v2 `memory.max` / `memory.current` walking `/proc/self/cgroup` ancestors (`max` = unlimited), with a cgroup v1 `memory.limit_in_bytes` / `memory.usage_in_bytes` fallback and its unlimited sentinel; the memory-controller mount is resolved from `/proc/self/mountinfo`, and when several views are visible the tightest ancestor limit wins.
- `inactive_file` (v2) / `total_inactive_file` (v1) from `memory.stat` is subtracted from usage, matching the working-set figure Docker/cAdvisor report, so clean page cache charged by a previous run does not make the next run look exhausted.
- `memory_available()` is a tri-state: an int (including an authoritative 0), `None` when nothing could measure it (non-Linux, no procfs — the planner keeps its historical 8 GB fallback), or a typed `CgroupError` when a present cgroup/procfs input is malformed, oversized or unreadable — surfaced by `coli doctor` as `memory.ram: fail <reason>` instead of being mistaken for "measured 0".
- `build_plan` owns the single probe and refuses a known over-budget model before any allocation; `doctor` reads the plan's figure back.

All reads are bounded, ASCII-only, and there is no `assert` on the decision path. Mount-namespace hardening (descriptor pinning, overmount detection) is deliberately not part of this change.

## Validation

- [x] `make -C c check` — rc=0 (C unit tests OK; Python suite 798 OK, 35 skipped, with the built engine present)
- [ ] CUDA changes were tested with `make -C c cuda-test` — not applicable, no CUDA code touched
- [ ] Performance claims include hardware, commands, and repeatable measurements — no performance claim; one extra bounded procfs read at plan time

From `c/` on Ubuntu 24.04 (Linux 7.0.0, x86_64, cgroup v2, Docker 29.1.3, Python 3.12.3):

```
python3 -m unittest -v tests.test_cgroup_memory      # 60 OK   (also under python3 -O)
python3 -m unittest -v tests.test_resource_plan      # 68 OK   (also under python3 -O)
python3 -m unittest discover -s tests -p 'test_*.py' # 791 OK, skipped=34 before the engine is built (origin/dev: 725 OK, skipped=34);
                                                     # 798 OK, skipped=35 once `make check` has built ./colibri (tests/test_cuda_env.py collects only with the binary)
ruff check  # only the pre-existing F401 `os` in tests/test_resource_plan.py
git diff --check origin/dev..HEAD
```

Tests drive the resolver against fixture proc/cgroup trees: unlimited, finite, nested ancestor-min, usage above limit, `inactive_file` subtraction (v2 and v1), a hidden ancestor limit across two visible mounts (the tighter one wins), malformed/oversized/unreadable inputs (typed refusal), absent procfs (unknown → fallback), the hybrid `1:net_cls:/` + `0::/…` membership layout systemd and Docker both produce, and `doctor`'s three outcomes (measured / unknown / refused).

Live cgroup v2 smoke (`python:3.12-slim`, checkout mounted read-only; values in bytes):

| container | `memory_available()` | `limit − (current − inactive_file)` |
|---|---|---|
| `--memory=2g` | 2 130 833 408 | 2 130 833 408 |
| `--memory=6g` | 6 425 481 216 | 6 425 481 216 |
| no limit | 36 207 153 152 | n/a (= host `MemAvailable`, read 1.4 MB apart) |

In the 2 GiB container `coli doctor --ram 16` reports `planned RAM budget exceeds available memory`; the automatic budget resolves to 1.87 GB rather than the 8 GB fallback.

Not exercised: cgroup v1 on a real kernel (fixtures only), macOS/Windows beyond the unchanged `None` path, an actual OOM-boundary run.

## Compatibility

- [x] The default CPU build remains dependency-free — `resource_plan.py` uses the standard library only

Prepared with AI assistance; I reviewed the change and ran the validation above.


JustVugg:
Heads-up: this went `CONFLICTING` a few minutes ago, and **you do not need to do anything** — #1322 landed on `dev` and touched the same statement in `doctor.py`.

The two changes are complementary rather than competing:

- yours reads the memory probe out of the plan (`available_memory = plan["memory"]["available_bytes"]`)
- #1322 gates the GPU arguments on `linkage.linked`, so a CPU-only binary stops being handed VRAM it cannot spend

I resolved it keeping both, and pushed the resolution to `merge/1316-doctor` in this repo so nothing lands on your branch without you. `tests/test_doctor` + `tests/test_cgroup_memory`: 96 tests, green together.

Two ways forward, your call: pull that branch into yours, or leave it and I will merge from it and mark this PR merged with your commits intact. Either way the rebase is not yours to do — a conflict created by our merge order is our problem, not the contributor's.

Reviewing the substance of the cgroup work next; the 917 lines of tests are noted and appreciated.

Avicennasis:
Apologies for the slow answer — this sat behind other work on my side.

Second route, please: merge from `merge/1316-doctor` and mark this PR merged with the commits intact. Your resolution is the right one — reading the memory probe out of the plan and #1322's `linkage.linked` gate on the GPU arguments are complementary, and I have nothing to add to it.

One heads-up since you'll be doing the merge. `dev` has moved about 210 commits since `745ee0e7`, and a `git merge-tree` of `merge/1316-doctor` against today's `dev` (`b5968eb`) conflicts again, in the same seam:

- `c/resource_plan.py`, two hunks: #1377's `memory_available()` rewrite (`windows_available_bytes`, the commit limit) lands where this PR feeds the cgroup ceiling into `available_memory`, and #1456's `placement_unified` rename lands on the 8 GB fallback that this PR narrows to "only when every host-memory probe failed".
- `c/tests/test_resource_plan.py`, one hunk: #1456's two new tests and this PR's `test_auto_ram_keeps_reserve_below_small_finite_headroom` were inserted at the same point.

Complementary again rather than competing: the cgroup probe is a Linux-only tightening of the same number #1377 now measures on every platform, so the resolution is "keep both" in all three places.

If you would rather not resolve that a second time, say so and I will rebase the two commits onto current `dev`, re-run `tests/test_doctor` + `tests/test_cgroup_memory` + `tests/test_resource_plan`, and force-push a fresh head. Otherwise I will leave the branch exactly as it is so what you merge is what you reviewed.

JustVugg:
Reopening right away: the queued CI runs for this branch had expired while they waited for approval, and closing and reopening is the only way to make GitHub create them again. Nothing about the pull request itself is being judged here.

JustVugg:
Sorry for the two notifications just now, and for the wait before them.

**Why this never got a CI result.** The runs were created when you pushed, then sat in `action_required` waiting for a maintainer to approve them, and expired before anyone did. Today I approved every pending run that still existed, and twenty pull requests are building now. This one was past the expiry, so there was nothing to approve; closing and reopening did not make GitHub create new ones, which is what those two notifications were.

**Where the branch stands.** It is behind `dev` and currently conflicting. The four files you touch have moved this much since your `745ee0e`:

- `c/resource_plan.py`: 4 commits
- `c/tests/test_resource_plan.py`: 6 commits
- `c/doctor.py`: 3 commits
- `c/tests/test_cgroup_memory.py`: untouched, it is yours

That is a small surface for a change of this size, and the planner is somewhere a cgroup limit genuinely matters: today the budget comes from `MemAvailable`, which inside a container is the host's number and not the one the process will be killed for.

**When you push, a fresh run appears and I approve it the same day.**

PR #1319, opened 2026-09-02
fix(qwen36): read tokenizer.json merges in string and pair spellings

Avicennasis:
## What this fixes

`load_tokenizer` in `qwen36.c` indexes the BPE merge table from `tokenizer.json` by splitting each entry on a space — the legacy `"a b"` string spelling. `tokenizers >= 0.20` (transformers 4.45+) writes each merge as a two-element array `["a", "b"]`, which is what current Qwen3.6 checkpoints ship. Every entry failed the string test and was skipped, so the reader finished with **zero merges and no message**; nothing downstream treats an empty merge table as an error, so `encode_text` silently degraded to one token per byte-symbol.

The fix accepts both spellings. It was found running the engine against a real Qwen3.6 tokenizer: the prompt "The capital of France is" tokenized to 24 byte-level tokens instead of 5, and the model still produced text, which is why it went unnoticed.

## Validation

- New `c/tests/test_qwen36_tok_merges.c` (auto-discovered by the Makefile): builds a tokenizer with each spelling and asserts the same merge table and the same encoding; fails on the pre-fix code.
- `make -C c qwen36` and the new test pass on current `dev` (Linux x86_64) and macOS arm64.
- Also adds `c/tests/tmp_*` to `.gitignore` for test scratch dirs.

Standalone; cherry-picked from a larger qpack series so it can land on its own.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,8 @@ c/tests/test_*
 !c/tests/test_*.cu
 !c/tests/test_*.mm
 !c/tests/test_*.py
+# scratch fixture directories the test binaries write at run time
+c/tests/tmp_*
 # bench_* and fuzz_* build to extensionless binaries in the same directory and
 # were not covered here, which is how c/tests/bench_omp_grain reached the tree.
 c/tests/bench_*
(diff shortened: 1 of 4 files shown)


JustVugg (approved):
The failure this fixes is the kind worth hunting: the pair spelling indexed **zero** merges and `bpe_piece` reads an empty table as "nothing to merge" rather than as an error, so tokenisation silently degraded to one token per byte — 24 tokens for a 24-character prompt, with nothing in the logs.

Verified on your branch:

- the test covers **both** spellings, so the legacy string form is protected from becoming a regression while fixing the new one;
- negative control, which is what convinced me: disabling the pair branch makes the test report `expected [3, 4], got 5 ids: 0 1 2 0 1` — the byte-level degradation itself, not a generic mismatch. A test that names the real symptom is worth more than one that just goes red.

The type checks on the array form (`J_ARR`, `len==2`, both kids `J_STR` and non-NULL) mean a malformed tokenizer.json falls through to `continue` instead of dereferencing something. Right call.

25/25 in CI, test green locally.

PR #1320, opened 2026-09-02
fix(build): fall back to the Homebrew prefixes when brew is off the PATH

Avicennasis:
## What this fixes

The Darwin arm of `c/Makefile` locates libomp with `$(shell brew --prefix libomp)`. On a Command Line Tools-only Mac driven non-interactively (`make` from a script, over SSH, in CI) Homebrew is routinely installed but not on `PATH`, so the probe prints nothing, the OpenMP arm never engages, and the build is a perfectly valid **single-threaded** engine with libomp sitting right there on disk. The only signal is the existing "libomp not found: building single-threaded" line, which is easy to read as "not installed".

This falls back to the two standard Homebrew prefixes (`/opt/homebrew/opt/libomp`, `/usr/local/opt/libomp`) before concluding libomp is absent. The existing `omp.h` + `libomp.*` artifact check still gates the flags, so a prospective-but-uninstalled prefix is still rejected.

## Measured

M4 Max, CLT-only, Qwen3.6-35B: 0.87 tok/s from the silent single-threaded build → 4.44 tok/s once OpenMP engaged (TTFT 3.9 s → ~0.7 s), output byte-identical — engine code is unchanged by this commit; every `omp` site is row-partitioned and deterministic by construction. `make -n` from a shell without `brew` on `PATH` now shows `-fopenmp -I/opt/homebrew/opt/libomp/include … -lomp` engaging.

Makefile-only; standalone.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -52,6 +52,13 @@ ifneq (,$(DARWIN))
 # Niente -march: su arm64 NEON e' baseline (i kernel __ARM_NEON si attivano da soli).
 CC      = clang
 OMPDIR := $(shell brew --prefix libomp 2>/dev/null)
+# CLT-only Macs routinely have Homebrew installed but off the non-interactive
+# PATH (`make` from scripts, SSH, CI), which turned this probe into a silent
+# single-threaded build with libomp sitting right there on disk. Fall back to
+# the two standard Homebrew prefixes before concluding it is absent.
+ifeq ($(OMPDIR),)
+OMPDIR := $(firstword $(wildcard /opt/homebrew/opt/libomp /usr/local/opt/libomp))
+endif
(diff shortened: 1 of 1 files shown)


JustVugg (approved):
Safe by construction, which is what a build-probe fix should be:

- it only fires when `OMPDIR` is already empty — `brew` on PATH keeps winning, unchanged;
- `wildcard` returns only paths that exist, so a machine with neither prefix ends up exactly where it started;
- line 64 still gates on `omp.h` and `libomp.*` actually being there, so a stale prefix cannot produce broken flags.

The failure it fixes is the kind worth catching: a silent single-threaded build on a machine that had libomp sitting on disk the whole time. Silent is the operative word.

Verified the Linux build is untouched (those paths do not exist there). 25/25 green.

PR #1321, opened 2026-09-02
glm53: size the expert cache around the model, not around MemAvailable

ronaldcklomp:
### The problem

`expert_cache_init` sizes the LRU expert-slot cache as `MemAvailable - 3 GB`.

`MemAvailable` counts reclaimable page cache as free. So on a machine where the model already fits in RAM, the engine allocates almost all of it as *anonymous* memory — and in doing so evicts the very model pages that the slot cache then has to re-read from disk. Both caches end up holding the same bytes, and the cheap copy is the one that loses.

### The fix

Leave the model room to stay in page cache and take only what is left:

```
budget = MemTotal - model - 8% margin
```

still clamped by `MemAvailable`, still overridable with `GLM53_EXPERT_GB`. When the model can't fit in RAM at all, page cache can't help either, so the previous take-what-is-available behaviour is kept.

### Numbers

GLM-5.3-Flash int4-g64, 182 GB model, 247 GB RAM, 8-core EPYC 7F32, model on NVMe, `--ngen 60 --temp 0 --no-think`:

| `GLM53_EXPERT_GB` | tok/s | slot misses | avg `expert_read` |
|---|---|---|---|
| default (~238) | 1.387 | 9,936 | 8.45 ms |
| 24 | 2.013 | 15,978 | 3.52 ms |
| 40 | 2.061 | 13,625 | 4.46 ms |
| 44 | 2.060 | 13,167 | 4.79 ms |
| 56 | 1.971 | 11,931 | 5.69 ms |

Note the shape, which is what makes this counter-intuitive: the smaller cache takes **~60% more slot misses and is still 48% faster**, because each miss is served from page cache instead of the drive. Measured disk reads per run drop from ~47 GB to ~16 GB (`/sys/block/nvme0n1/stat`).

On this machine the new formula picks 44.3 GB, which is the measured optimum.

### Does it generalise?

The sweep above tunes on one prompt at `--ngen 60`, so I re-checked on an unrelated prompt at `--ngen 120` (old default emulated with `GLM53_EXPERT_GB=230`, runs interleaved):

| | run 1 | run 2 | mean |
|---|---|---|---|
| old default (230 GB) | 1.431 | 1.607 | 1.519 |
| new auto (44.3 GB) | 1.743 | 1.940 | **1.842** |

Same direction, **+21%** rather than +48%. So the size of the win is clearly workload-dependent — I'd not want the 48% read as typical. What reproduces is the direction and the mechanism; the magnitude depends on how much of the routing misses the resident tier.

Run-to-run spread is also wider at this length (first run of each config is the slower one), so treat these as two samples, not a tight interval.

### Verification

The `teacher_forcing` line (argmax at every prompt position) is byte-identical before and after, so this is purely a memory-placement change with no effect on output.

Builds and runs against a clean `main` checkout with no other changes.

### Caveat

All of this is one machine, one model. Specifically untested:

- **Model much smaller than RAM.** The formula hands back a large budget, which should be harmless, but I haven't run it.
- **Model larger than RAM.** This takes the `model_gb >= total - margin` fallback to the old behaviour. I reasoned that path through and wrote it, but had no way to exercise it here.
- The 8% margin is a judgement call tuned at a model-to-RAM ratio of ~0.74, and the model size is estimated as `sparse * n_experts * e_slot * 1.07` rather than read from the files on disk.

Happy to adjust the margin, derive the size from actual file sizes, or land this behind an opt-in flag first if you'd rather not move a default on one machine's evidence.


rcklomp:
Rebased under #1325 as asked: the branch is now the single glm53 commit (`cfac6de2`, `c/glm53.c` +37/−6) on top of current `dev` (`d0f1791`, v1.10.2). The `st.h`/`qwen38_core.h` mapping hunks that used to ride along are gone from the diff since they landed as #1325 (`eabeb9a`).

The formula is unchanged: `budget = MemTotal − model − 8 % margin`, still clamped by `MemAvailable`, still overridable with `GLM53_EXPERT_GB`, and the take-what-is-available behaviour is kept when the model cannot fit in RAM. It applies cleanly and builds on `dev`.

Two things from the fork since, for context rather than for this PR: the rig has been running this budget under `COLI_MAP_EXPERTS`-style mapping for four days (`budget esperti: 44.3 GB (247.3 totali, 183.2 modello, 19.8 margine)`), and @monotophic's scarcity datapoint on #1324 says the copy path stays the right default below ~1:1 model:RAM — which is exactly where this budget still matters.

Authored with Claude Fable 5.1 in Claude Code.

JustVugg:
The observation behind this is right, and it is one I hit independently on another engine this week: the expert LRU and the page cache hold the same bytes, so sizing the LRU from `MemAvailable` means allocating anonymous memory that evicts exactly the pages the next miss would have been served from. A machine can end up reading from disk what it already had in RAM twice over.

What is holding it is that this changes the default budget on every machine, and the right budget is an empirical question, not an analytical one. The patch decides it with `model_gb = sparse * n_experts * e_slot * 1.07` against an `8%`-or-4-GB margin, and those constants are the whole behaviour. If the 1.07 is off, or if `MemAvailable` on the target box already excludes most reclaimable cache, the new budget is worse than the old one and nobody finds out except as slower tokens.

What would make this mergeable is a cold A/B, on the same machine, same prompt, same seed, caches dropped before every run:

- one case where the model fits within `total - margin` (the new branch), and
- one where it does not (the old branch, to show the fallback is untouched),

reporting tok/s and the expert hit rate for both old and new. If the claim is that the page cache does the work the LRU was doing, the hit rate can fall while tok/s rises, and that is the interesting shape to see.

Two smaller notes for the revision:

- `memory_total_gb` opens and scans `/proc/meminfo` a second time; one pass returning both fields would do.
- #1456 has just changed `memory_available_gb` in this same function so that it is not Linux-only (on macOS it returned 0, collapsing the budget to 1 GB). Rebase on top of that, or the two will disagree about what "available" means on a Mac.

Keeping it open, this is worth finishing.

rcklomp:
Thanks — that is the right thing to ask for, and running it changed the patch.

Rebased onto current `dev` (so on top of #1456), and both of your smaller notes are fixed. The measurements below are on the rebased version, not the original one.

## First: two unit bugs, one of them mine

While rebasing I found the original patch was comparing different units:

* `memory_total_gb` computed `kb / 1048576.0`, which is **GiB**, and weighed it against a `model_gb` built from byte counts (`/1e9`, true **GB**).
* `compat`'s existing `kb / 1e6` is also not GB — `/proc/meminfo`'s "kB" is KiB, so it understates true GB by 2.3%. That was harmless while "available" was only ever compared against itself, and stops being harmless the moment a model size in real GB is subtracted from it.

Both now return true GB. This matters for your review: it is why the original PR reported the formula picking 44.3 GB, and why on the same machine it now picks **61.1 GB**. Same box, same model — the old number was a unit error.

## The cold A/B

GLM-5.3-Flash int4-g64, 265.6 GB RAM, 8-core EPYC 7F32, model on NVMe. Model shards evicted with `posix_fadvise(DONTNEED)` before **every** run. Same prompt, `--greedy 60`, so both binaries emit the same 60 tokens and no seed is involved. Two reps each, `dev` and the patch interleaved:

| | tok/s | hit rate | slots/layer | anonymous resident | budget chosen |
|---|---|---|---|---|---|
| `dev` | 1.120 / 1.106 | 72.5% | 288 | 171.2 GB | 241.3 GB |
| patch | **1.290 / 1.292** | 65.3% | 102 | 60.6 GB | 61.1 GB |

**+15.6%**, and it is the shape you predicted: the hit rate *falls* while tok/s *rises*. Hits and misses are bit-identical between reps (greedy decode, same routing), so the difference is placement and nothing else. On the old path the engine takes 171 GB of *anonymous* memory on a 265 GB box holding a 183 GB model, which is the double-hold this PR is about.

## Is the 8% margin the right constant?

This is your real question, so here is the curve rather than an argument. Same conditions, `GLM53_EXPERT_GB` pinned (it is a direct GB budget, so these points measure the machine, not the formula):

| `GLM53_EXPERT_GB` | slots/layer | hit rate | tok/s |
|---|---|---|---|
| 24 | 40 | 45.4% | 1.288 |
| **40** | 67 | 56.4% | **1.312** |
| 48 | 80 | 60.2% | 1.296 |
| 61 (what the formula picks) | 102 | 65.3% | 1.287 |
| 100 | 168 | 71.5% | 1.136 |
| 241 (old default) | 288 | 72.5% | 1.113 |

The hit rate rises monotonically with the budget across the whole range while throughput peaks at 40 GB and then falls. More hits, slower machine.

For the constants specifically: the curve is **flat within 2% from 24 GB to 61 GB**, a 2.5x range of budgets. The formula lands at 61, on the plateau, 1.9% off the peak. So the 8% margin is not tuned, but the function it is choosing on is broad enough that it does not need to be — being off by half costs ~2%, while being off by 4x (the old default) costs 15%. If you would rather have the peak than the plateau, a larger margin biases downward; I did not tune it, because tuning it on one machine and one prompt is how the first version of this PR got a number that did not generalise.

## Your two smaller notes

Both have the same answer: `compat.h` gains `compat_meminfo_gb(total, avail)` — one call, one pass, Linux/macOS/Windows — and `compat_mem_available_gb()` now delegates to it, so there is still exactly one definition of "available".

* The second `/proc/meminfo` open and scan is gone.
* It settles the #1456 interaction rather than merely rebasing past it: `total` is now measured the same cross-platform way as `available` (`sysctl hw.memsize` on macOS, `GlobalMemoryStatusEx` on Windows), so they cannot disagree about what they mean. Worth noting the bug you would otherwise have inherited — the old `memory_total_gb` returned **0 on macOS and Windows**, which took the `total <= 0` fallback and silently disabled this entire fix on both platforms. Only Linux ever ran the new branch.

`tests/test_mem_available.c` said in its own comment that the "available <= physical RAM" bound had to be skipped on macOS for want of a total. It no longer has to be, and it now also pins the wrapper against the one-pass helper.

## What I did not test

The `model_gb >= total - margin` fallback is **not exercised on hardware**. No model on this box exceeds RAM, and unprivileged user namespaces are disabled here, so I could not fake `/proc/meminfo` to force it. The branch is unchanged in behaviour from the code it replaces and `total <= 0` now routes into it as well, but it is reasoned, not measured, and should be read that way.

## One thing worth flagging

Once `COLI_MAP_EXPERTS` flips on by default (#1350 describes itself as preparation for exactly that), `expert_read` serves experts straight from the shard mapping, and an engine that sizes its LRU at one slot per expert bypasses this budget entirely. On my own tree, where that flip has already happened, this patch is a no-op. It is still right for `dev` as it stands today — `COLI_MAP_EXPERTS` is opt-in and the LRU governs — but if the flip lands, the blast radius of this code shrinks to checkpoints whose experts are not all mappable, and that may change how much you care about the constants.

I have left the four other hand-rolled total-RAM queries alone (`telemetry.h` via sysctl, `olmoe.c` via `sysconf(_SC_PHYS_PAGES)`, `kimi_k3.c`, and the test), each with its own `#ifdef` ladder and its own definition. Consolidating them onto `compat_meminfo_gb` is a separate change and I did not want to grow this one.

rcklomp:
Heads-up on the red CI at `ee8642e`, since it will have pinged you: **it was not the patch, and it was my fault for letting the branch rot.**

Both `CI` and `check` failed having executed **zero jobs** — they never started, which `gh` reports as a workflow-file problem rather than a build failure. The same workflows passed at `cfac6de`. The branch was **43 commits behind `dev`**, and `dev` had since added a step to `.github/workflows/ci.yml` (the multi-SSD mirror / `O_DIRECT` equivalence checks from the mirror-io work), so the stale head could not produce a usable run.

Rebased onto current `dev` as `cf7288a`. Not a no-op merge — two commits had touched `c/glm53.c` since my base, `fix(omp): size the OpenMP team in the four engines that never did` and `fix(glm53): refuse an over-long serve prompt with CONTEXT_EXCEEDED`. Both auto-merged cleanly against `expert_cache_init`; the change itself is unaltered.

Re-verified before pushing rather than assumed: `glm53` builds clean against current `dev`, and `tests/test_mem_available` passes with both fields populated (`disponibile: 249.80 GB, totale: 265.56 GB`), which is the thing the `compat_meminfo_gb` consolidation could plausibly have broken in a merge.

CI is running on `cf7288a` now — it started and queued jobs this time, instead of dying instantly, which is itself the evidence that staleness was the cause. I will follow up here with the result either way rather than leave you to check.

Nothing needed from you on this; flagging it so the failure mail is not mistaken for a problem with the change.

rcklomp:
Following up as promised: **both workflows are green on `cf7288a`** — `CI: success`, `check: success` — and the PR now reads `MERGEABLE/CLEAN` rather than `UNSTABLE`.

So the red at `ee8642e` was staleness and nothing else. Nothing about the change moved to fix it; only its base did.

Note on your CI as a nicety, take it or leave it: a branch that falls far enough behind fails with **zero jobs executed**, which reads as a broken patch in the notification mail rather than as a stale base. If `.github/workflows` gains a step the PR head predates, that is indistinguishable from a real failure to the contributor until they dig into the run. A base-freshness check that says so explicitly would have saved this round trip. Entirely your call, and not something I would send a PR for uninvited.

JustVugg:
Reviewed. The formula is safe in the direction that matters: the new budget is the minimum of the old one and the model-aware figure, so it can only shrink and cannot reintroduce the OOM class. And the flagship case, model much larger than RAM, takes the untouched fallback branch. The body is unusually honest about the 48 percent being workload-specific.

One thing the body does not advertise and should. The Linux arm also fixes a unit bug, `kb / 1e6` becoming `kb * 1024.0 / 1e9`, which raises the reported available RAM by about 2.4 percent for every caller of `compat_mem_available_gb`, not only for glm53.

Two things before merge.

The consolidation is half done, and your own comment is the argument against stopping here. It says every caller hand-rolls its own ifdef ladder and that the definitions do not agree. But the diff routes only `glm53.c` through the new helper: `olmoe.c` still reads /proc/meminfo directly with the old scaling, and `telemetry.h` and `kimi_k3.c` keep their own paths. After this lands, HWINFO will report an available-RAM figure 2.4 percent away from the one glm53 budgets against, which is exactly the inconsistency the comment condemns. Follow-up is fine, but I would like it named.

And no test touches the new budget arithmetic. `test_mem_available.c` gains a wrapper-agreement check, which is good, but the three branches that decide the cache size are unpinned. Could you factor the total/model/margin/free computation into a testable helper and pin the three cases: model much larger than RAM, model much smaller, and total reported as zero?

Small trap worth renaming: `compat_meminfo_gb` is one letter from the existing Windows-only `compat_meminfo`.

PR #1322, opened 2026-09-02
fix(doctor): omit GPU plan for CPU-only engine

configurowebmax:
## Summary

- make `coli doctor` build its placement report from the installed engine's GPU capability, not just host GPU inventory
- clear GPU selectors and requested VRAM when the selected engine is CPU-only or its backend is unavailable
- cover both CPU-only and GPU-enabled binaries with regression tests

## Reproduction

On Windows with the v1.10.1 Qwen3.6 release binary and an NVIDIA RTX A2000, `coli doctor` correctly warned:

```text
[warn] accelerator.gpu    GPU detected but the engine is CPU-only
```

but the same report then advertised a VRAM hot tier, described a mixed CPU/GPU bottleneck, and recommended `COLI_CUDA_PIPE=1`. The plan therefore described hardware the installed engine could not use.

After this change the GPU remains visible in the diagnostic warning, while the executable placement plan reports the CPU path, zero VRAM budget, and no CUDA-only tuning.

## Tests

- `python -m unittest discover -s tests -p test_doctor.py -v` - 36 passed
- validated the modified diagnostic against the real 41-shard Qwen3.6 container: CPU-only warning retained; VRAM tier removed
- full Python discovery ran 723 tests; 9 environment-only errors require Unix `rm` or `make`, neither available in this Windows shell
- `make check` could not run because GNU Make is not installed in this environment

### DIFF
--- a/c/doctor.py
+++ b/c/doctor.py
@@ -540,9 +540,17 @@ def run_doctor(model, ram_gb=0, context=4096, gpu_indices=None, vram_gb=0, *,
     try:
         if resolved is None:
             raise ValueError("placement requires a registered model family")
-        plan = build_plan(model, ram_gb, context, gpu_indices, vram_gb,
+        # The placement report describes this installed engine, not merely the
+        # hardware visible on the host. A CPU-only binary cannot spend VRAM;
+        # passing discovered GPUs through here made doctor contradict its own
+        # accelerator check and emit CUDA-only tuning advice. Keep inventory
+        # details in accelerator.gpu above, but build an executable plan below.
+        plan_gpus = detected_gpus if linkage.get("linked") else []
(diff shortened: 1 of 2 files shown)


JustVugg (approved):
The separation is the right one: hardware inventory stays in `accelerator.gpu`, while the *plan* is gated on `linkage.linked` — what this binary can actually spend, not what the host happens to have. That is what made doctor contradict its own accelerator check.

What earns the approval is the second test. `test_gpu_engine_with_detected_gpu_keeps_vram_plan` pins the case that must keep working, so the fix cannot quietly become a regression for people who do have a linked CUDA build. A one-sided test here would have been the easy version and the dangerous one.

Ran `tests/test_doctor` on your branch locally: 36 tests, green. 25/25 in CI.

PR #1324, opened 2026-09-03
glm53: serve expert weights from a file mapping instead of copying them in

ronaldcklomp:
### The problem

An expert is 14.16 MB in six pieces (gate/up/down, each weights + scales). On a slot miss `expert_read` malloc's a 14.16 MB slot and `pread`s the six pieces into it — copying bytes the kernel is already holding in page cache. Those private copies also make the slot cache compete with the page cache for the same RAM.

### The fix

Measured first: **0 of 12096 experts** have their six pieces consecutive in the file, so a slot cannot just point at one contiguous run. But nothing requires the pieces to be adjacent *in memory* either — `expert_mats` only builds three read-only `Mat` views over them. So a slot becomes six pointers rather than one buffer, and each piece points straight into its file's mapping.

On GLM-5.3-Flash int4-g64 (182 GB, 45 layers, 12096 routed experts) all of them qualify and the copy disappears entirely:

```
[MAP] 59 file mappati, esperti mappabili 12096/12096 (tutti: nessuna copia)
```

### Numbers

8-core EPYC 7F32, 247 GB RAM, model on NVMe, `--ngen 60 --temp 0 --no-think`, two runs each:

| | tok/s |
|---|---|
| `GLM53_NO_MMAP=1` (current behaviour) | 1.319 / 1.322 |
| mapped | **1.965 / 1.971** |

**+49%.** `teacher_forcing` — the argmax at every prompt position — is identical either way, so this is purely a memory-placement change.

### Details worth knowing

- Mapping is per **file**, not per expert: piece offsets are not page-aligned, but mapping the whole file lets the kernel deal with that. A piece is used only if its file mapped, it fits, and its offset is 4-byte aligned (the scale pieces are dereferenced as `float *`).
- Slots are reused and their buffer was never freed, so a recycled slot must not be `pread` into while it still points at a read-only mapping. Hence the split between `piece[]` (what `expert_mats` reads) and `own` (the malloc'd buffer, allocated only for experts that cannot be mapped). Teardown frees `own`, never a mapping.
- When every expert is mappable a slot costs no memory of ours, so eviction has nothing to reclaim and `cap` becomes `n_experts`.
- Guarded by the file's existing `#ifndef _WIN32`; on Windows `GLM53_HAVE_MMAP` is undefined, `g_map_active` stays 0 and every read takes the original `pread` path. **Compiled and verified both ways.**
- `GLM53_NO_MMAP=1` forces the copy path at runtime.

### Interaction with #1321

When everything maps, the slot cache no longer holds private copies, so the expert-cache budget that #1321 corrects stops mattering (`cap` is overridden to `n_experts`). #1321 still matters where mapping is unavailable — Windows, or a checkpoint whose pieces are not 4-byte aligned — so the two are complementary rather than duplicates, but **this one supersedes it on the common path**. Happy to rebase either onto the other, in whichever order suits you, or to close #1321 if you would rather have just this.

### Caveats

One machine, one model. The 4-byte alignment and whole-file mapping assumptions hold for this checkpoint's layout; a checkpoint that fails them falls back to the existing copy path per piece rather than breaking. I have not tested on Windows or macOS beyond confirming the non-mmap path compiles.

### DIFF
--- a/c/glm53.c
+++ b/c/glm53.c
@@ -78,6 +78,10 @@ static int g_vk_ready = 0;
 #include <time.h>
 #ifndef _WIN32
 #include <sys/resource.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#define GLM53_HAVE_MMAP 1
 #endif
 #include "hyper_connections.h"   /* mHC, condiviso con deepseek_v4.c */
 
@@ -1145,7 +1149,17 @@ typedef struct ERef {
     int contig;                           /* i sei pezzi sono consecutivi in un file */
 } ERef;
 
-typedef struct { int eid; uint8_t *base; uint64_t used; } Slot;
+/* I sei pezzi non sono adiacenti nel file, ma non devono esserlo nemmeno in
+ * memoria: expert_mats ci costruisce sopra solo tre viste in sola lettura. */
+typedef struct { int eid; uint8_t *piece[GLM53_EXPERT_PIECES]; uint8_t *own; uint64_t used; } Slot;
+
+/* Una mappatura per file, non per esperto: gli offset dei pezzi non sono
+ * allineati alla pagina, ma mappando il file intero ci pensa il kernel. */
+#define GLM53_MAXFD 4096
+static uint8_t *g_fmap[GLM53_MAXFD];
+static size_t   g_fmaplen[GLM53_MAXFD];
+static long     g_map_serve, g_map_copy;
+static int      g_map_active, g_map_all;
 typedef struct LCache { Slot *s; int n, cap; } LCache;
 
 /* Lunghezze e posizioni dei sei pezzi dentro allo slot. Gate e up sono
@@ -1225,8 +1239,64 @@ static double memory_available_gb(void) {
     return gb;
 }
 
+/* Un pezzo e' mappabile se il suo file e' mappato, ci sta dentro, ed e'
+ * allineato a 4 byte (le scale si leggono come float *). */
+static int piece_mappable(const GModel *m, const ERef *ref, int q) {
+    const int fd = ref->fd[q];
+    if (fd <= 0 || fd >= GLM53_MAXFD || !g_fmap[fd]) return 0;
+    if ((size_t)ref->off[q] + (size_t)m->e_len[q] > g_fmaplen[fd]) return 0;
+    return (ref->off[q] & 3) == 0;
+}
+
+/* Gli esperti si leggevano copiando 14 MB per slot da byte che il kernel teneva
+ * gia' in page cache. I sei pezzi si possono puntare direttamente nella
+ * mappatura del file: nessuna copia, e la residenza la gestisce il kernel.
+ * Su Windows non si mappa e resta il percorso originale. */
+static void expert_map_init(GModel *m) {
+#ifdef GLM53_HAVE_MMAP
+    if (getenv("GLM53_NO_MMAP") || !m->eref) return;
+    const Cfg *c = &m->c;
+    int files = 0;
+    for (int i = 0; i < c->n_layers; i++)
+        for (int e = 0; e < c->n_experts; e++) {
+            const ERef *ref = &m->eref[(size_t)i * c->n_experts + e];
+            if (ref->fd[0] <= 0) continue;
+            for (int q = 0; q < GLM53_EXPERT_PIECES; q++) {
+                const int fd = ref->fd[q];
+                if (fd <= 0 || fd >= GLM53_MAXFD || g_fmap[fd]) continue;
+                struct stat stt;
+                if (fstat(fd, &stt) != 0 || stt.st_size <= 0) continue;
+                void *pm = mmap(NULL, (size_t)stt.st_size, PROT_READ, MAP_SHARED, fd, 0);
+                if (pm == MAP_FAILED) continue;
+                g_fmap[fd] = (uint8_t *)pm;
+                g_fmaplen[fd] = (size_t)stt.st_size;
+                files++;
+            }
+        }
+    if (!files) return;
+    g_map_active = 1;
+    long full = 0, partial = 0;
+    for (int i = 0; i < c->n_layers; i++)
+        for (int e = 0; e < c->n_experts; e++) {
+            const ERef *ref = &m->eref[(size_t)i * c->n_experts + e];
+            if (ref->fd[0] <= 0) continue;
+            int ok = 1;
+            for (int q = 0; q < GLM53_EXPERT_PIECES; q++)
+                if (!piece_mappable(m, ref, q)) { ok = 0; break; }
+            if (ok) full++; else partial++;
+        }
+    g_map_all = (partial == 0 && full > 0);
+    if (getenv("GLM53_VERBOSE"))
+        fprintf(stderr, "[MAP] %d file mappati, esperti mappabili %ld/%ld%s\n",
+                files, full, full + partial, g_map_all ? " (tutti: nessuna copia)" : "");
+#else
+    (void)m;
+#endif
+}
+
 static void expert_cache_init(GModel *m) {
     const Cfg *c = &m->c;
+    expert_map_init(m);
     const char *setting = getenv("GLM53_EXPERT_GB");
     /* Il default si misura: quello che resta libero dopo i pesi residenti,
      * meno un margine per lo stato della conversazione, i temporanei del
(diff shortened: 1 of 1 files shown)


ronaldcklomp:
Two things I should have found before opening this, both arguing for changes. Flagging them myself rather than letting review spend time on them.

**1. I reinvented `st_map_raw`.** `st.h` already has `st_map_raw`/`st_unmap_raw` over `compat_map_readonly`, which handles Windows via `CreateFileMappingA`/`MapViewOfFile` and carries the aligned view/handle needed for cleanup. `kimi_k3.c` already uses it. My patch instead calls POSIX `mmap` directly and disables the whole feature on Windows with an `#ifndef _WIN32`, which is strictly worse: it duplicates the mechanism, leaks the mapping (I never unmap — the process exits, but that is not a reason), and gives Windows nothing.

Reworking on top of `compat_map_readonly` would make the fast path portable rather than POSIX-only. The one wrinkle is that `st_map_raw` maps per *tensor*, while I map per *file* deliberately — expert piece offsets are not page-aligned, and one mapping per expert piece would mean 12096 × 6 mappings. So it likely wants a per-shard variant of the same helper rather than `st_map_raw` as-is.

**2. The no-mmap choice in `st.h` looks deliberate, and I did not engage with it.** The header says, in as many words:

> legge con pread (niente mmap) + posix_fadvise(DONTNEED) -> le pagine NON restano residenti nel processo. E' la correzione del bug di RSS [...] (vedi memoria mmap-rss-bug)

So pread + `DONTNEED` is there to keep pages out of the process, and my change reverses that for expert weights. Mapped file pages are reclaimable page cache rather than anonymous memory, so I do not *think* this reintroduces the original bug — but I measured on a machine with 247 GB of RAM and a 182 GB model, which is precisely the case where residency is free. The README advertises running a 744B model in 15–26 GB, and that is the configuration where explicit `DONTNEED` discipline could well beat the kernel's LRU, and where my +49% could turn into a regression. I have not tested it and should not claim otherwise.

Given both points, I would suggest **not merging this as it stands**. I am happy to:

- rework it onto a per-shard `compat_map_readonly` helper so Windows benefits too, and
- gate it behind an opt-in (env or a plan flag) defaulting to the current pread path until someone can measure the low-RAM case,

or to just close this if you would rather the mapping question be settled at the `st.h` layer for every engine at once, which is arguably where it belongs — `glm53.c` is not the only engine paying for these copies.

The measurement itself stands: 0 of 12096 experts are file-contiguous, and pointing the six pieces at a mapping instead of copying 14.16 MB per miss is worth +49% on this hardware, with identical output. It is the mechanism I got wrong, not the win.

ronaldcklomp:
Follow-up on the low-RAM worry I raised above — I have now measured it, on a different engine.

I ported the same idea to `qwen38` (Qwen3.8-Flash-Next-FP8, 176B, 186 GB, 512 experts/layer), this time built on `compat_map_readonly` rather than raw `mmap`, so it is portable and Windows gets the fast path too. That let me A/B the exact configuration I said I had not tested: a cache too small to hold the working set, so experts are constantly evicted and re-read.

**Cache holds every expert** (no misses left to serve):

| | tok/s | expert_read | anonymous RAM |
|---|---|---|---|
| copy | 1.467 | 0.00s (0%) | 65 GB |
| mapped | 1.466 | 0.00s (0%) | **17 GB** |

**Cache too small — experts must be evicted** (the configuration the streaming engine exists for):

| | tok/s | expert_read | anonymous RAM |
|---|---|---|---|
| copy | 0.925 | 18.92s (**29%**) | 30 GB |
| mapped | **1.288 (+39%)** | 0.01s (**0%**) | 16 GB |

Identical generated text in all four runs.

So the concern inverts: under memory pressure the mapping is **faster**, not slower, and it is *also* the configuration where it saves the most memory. Where the cache is large enough that misses have stopped, it is throughput-neutral and simply removes 48 GB of anonymous memory. I could not construct a case where it lost.

That is consistent with what `compat.h` already says about this primitive:

> pages are faulted when the caller reads them and remain reclaimable file-backed cache pages

which is the distinction that matters versus the original RSS bug: these are reclaimable file pages, not the anonymous working set that `pread`-into-`malloc` creates.

This does not change my earlier recommendation for **this** PR — it still open-codes POSIX `mmap` and gives Windows nothing, and should be reworked onto `compat_map_readonly` or closed in favour of doing it at the `st.h` layer for every engine. But the underlying idea now has evidence from two engines and both memory regimes, so if you would like it in some form I am happy to do the rework in whichever shape you prefer.

JustVugg:
This is the most impressive contribution this project has received in months, and not mainly for the number. You found a +49% win, then reviewed yourself harder than any maintainer would have, then went and measured the exact regime you had flagged as your own blind spot — on a second engine, with the portable primitive, with identical output. That sequence is the whole review methodology of this repo, executed unprompted. Thank you.

**Decision: option (b).** Do it at the `st.h`/`compat_map_readonly` layer, for every engine at once. You said it yourself — that is where it belongs — and your qwen38 port already proves you can make it portable. Every engine pays the same copy today; none of them should keep paying it individually.

Shape we would like:

1. A per-shard mapping helper next to `st_map_raw` (your per-file reasoning is right: piece offsets are not page-aligned, and 12096×6 mappings would be absurd). Windows comes along via `compat_map_readonly`, as in your qwen38 version.
2. **Opt-in first**: `COLI_MAP_EXPERTS=1`, default stays pread until the low-RAM case is measured. Not because we doubt the direction — because the flip should be a measurement, not a hope.
3. The RSS guards are ours to fix in the same series: they read `VmRSS`, which mapped pages inflate, so they would misfire under mapping. Counting `RssAnon` instead is the correction, and your anonymous-memory columns are exactly the argument for it.

**The missing datapoint is ours to provide, and we are uniquely equipped for it**: a 25.4 GB box with a 180+ GB model is this project's home turf, and a ~430 GB GLM-5.3 conversion finishes here in about a day. That is the one regime your 247 GB machine cannot see: global page-cache scarcity, where eviction stops being our LFRU/PIN policy and becomes the kernel's LRU. If the mapping holds there too, the default flips and this headlines the next release. If the kernel's LRU turns out to need help, the same mapping plus `madvise` hints driven by our existing hotness data is the obvious second act — your design does not preclude it, which is another thing it gets right.

On #1321: keep it open and rebase it under this — as you said, it still matters wherever mapping is unavailable, and "the correct budget where copies still exist" is worth having even if the common path stops copying.

Send the st.h-layer version whenever ready; it gets reviewed with priority. And the measurement discipline in this thread — the self-review, the adversarial follow-up on your own claim — please keep bringing exactly that.

monotophic:
The page-cache-scarcity datapoint you asked for above, from two boxes whose containers are 2.5× and 3.2× their RAM. Engine: GLM-5.2 (`colibri.c`), which has its own mapped path (`COLI_MMAP=1`, same per-shard design as `st_map_shard_range`); trunk at `12b0fa13`, so #1325 and #1350 are in. Instrument: OpenAI-server decode, 5 prompts × 64 tokens, temperature 0, usage history frozen per arm, copy/mapped/copy/mapped order.

| box | RAM | container | copy | mapped | engine hit rate |
|---|---|---|---|---|---|
| GB10, 20 cores | 121.6 GiB | GLM-5.2 int4, 384 GB | 0.491 / 0.455 tok/s | 0.189 / 0.192 | 76.6% → 56.2% |
| Ryzen AI Max+ 395 | 123.5 GiB | GLM-5.2 int3-g64, 305 GB | 0.410 / 0.394 | 0.432 / 0.288 | 74.8% → 64.0% |

Output identical in every arm. The direct-run decode tells the same story more cleanly (GB10 0.73 → 0.29 tok/s; Strix 0.77 → 0.33 / 0.54), and the profile shows why: bytes fetched per token go from 2.4 GB to 15.7 GB on GB10, and the expert-matmul share of decode goes from 19% to about 50%, which is page-fault time landing inside the kernel call.

Reading: at these ratios there are two eviction policies over the same bytes. The engine's slot LRU counts an expert resident while the kernel's page LRU has already reclaimed part of it, so a "hit" re-faults the expert one page at a time in the middle of the matmul, and the pre-touch on every load re-reads pages that the previous load had just brought in. Where the model fits in RAM (your 247 GB case) the two policies never disagree and mapping only removes the copy; below about 1:1 they disagree constantly. So the flip-the-default question has a measured answer for the scarcity regime: keep `pread` as the default there, or make the mapped slot own its pages (`mlock` at fill, unlock at evict) so the engine's policy is the only one.

What did help on these boxes, same instrument: `DIRECT=1 PIPE=1` (+12% replicated on GB10, +10 to 20% on Strix, hit rate unchanged), which is the opposite move, taking the page cache out entirely. We are now testing handing the page cache's share of RAM to the slot cache with `RAM_GB` on top of that. Raw per-arm records available if useful.

Authored by Claude Fable 5.1 in Claude Code, analysis in partnership with @Monotophic.

rcklomp:
Closing this as superseded. Per the decision above, the mapping went in at the `st.h` layer as #1325 (`eabeb9a`, merged 2026-09-05) with the RSS accounting fixed in #1350; `dev`'s `glm53.c` now serves experts through `st_map_shard_range`, which is why this branch conflicts — it is the older, engine-local version of the same mechanism, and rebasing it would only put a second mapping path next to the shared one.

@monotophic's scarcity datapoint above is the measurement the default flip was waiting for, and it says the copy path stays the default below ~1:1 model:RAM (and that `mlock`-at-fill or `DIRECT=1 PIPE=1` are the levers there). That belongs with #1325's follow-up, not with this branch. The "correct budget where copies still exist" part lives on in #1321, now rebased on `dev`.

Authored with Claude Fable 5.1 in Claude Code.

PR #1325, opened 2026-09-03
st.h: shared per-shard mapping primitive, opt-in via COLI_MAP_EXPERTS=1

ronaldcklomp:
Per your review on #1324 — this puts the mapping at the `st.h` layer, shared by every engine, opt-in, exactly as scoped.

### 1. Per-shard mapping helper next to `st_map_raw`

`st_map_shard_range(fd, off, nbytes)` maps a whole **file** once (not per tensor/range — piece offsets aren't page-aligned, and mapping each of a model's ~12000+ expert pieces individually would mean thousands of small mappings for no benefit over one per shard), keeps it for the process lifetime, and serves any byte range inside it with no allocation or copy. Portable via the existing `compat_map_readonly` (same one `st_map_raw` already uses), so Windows gets the same path as POSIX — not the raw-`mmap`, POSIX-only version from #1324.

A failed or disabled mapping returns `NULL`. Every call site here falls back to its unmodified `pread` path on `NULL` — the primitive never changes behavior on its own.

### 2. Opt-in first

`COLI_MAP_EXPERTS=1`, unset/0 by default. `st_map_experts_enabled()` short-circuits to `NULL` immediately when off, so every engine takes the exact `pread` path it always has. Agreed this should be a measurement, not a hope — didn't flip the default.

### 3. RSS guards

Left alone, per your note that this is yours to fix in the same series.

### Wired into the two engines that had their own version of this

- `glm53.c` — replaces the `g_fmap` table from #1324 with a call into the shared primitive. Same fallback structure (per-piece `own` buffer when a piece can't map), same teardown safety (frees `own`, never a mapping).
- `qwen38_core.h` — replaces the qwen38-port's `q38_shard_map` table the same way.

Neither engine's default behavior changes; this only removes duplicate per-engine tables in favor of one implementation. `#1324`'s underlying idea is superseded by this — recommend closing it in favor of this PR.

### Verified

Same prompt, `--temp 0`, `pread` vs `COLI_MAP_EXPERTS=1`, on a clean `origin/main` checkout (not my working branch):

| engine | pread | mapped | match |
|---|---|---|---|
| GLM-5.3-Flash int4-g64 | 1.103 tok/s | 1.983 tok/s | `teacher_forcing` md5 identical |
| Qwen3.8-Flash-Next-FP8 | — | — | generated-text md5 identical |

Also confirmed unaffected: `qwen38` (default CPU-only target) still builds and links with **no** Vulkan dependency either with or without `VK=1`, and `kimi_k3`/`colibri` (both already `#include st.h`) build clean — this addition is purely additive.

### On the low-RAM regime

Understood that's the missing datapoint and you're better equipped to produce it than I am. If it turns up something the mapping needs to account for (the `madvise`-hints-from-hotness-data idea you mentioned, or anything else), happy to take a pass at it once you have the number.

### On #1321

Ready to rebase it onto this branch whenever you'd like — holding off until this lands so I'm not rebasing twice.

### DIFF
--- a/c/glm53.c
+++ b/c/glm53.c
@@ -1145,7 +1145,9 @@ typedef struct ERef {
     int contig;                           /* i sei pezzi sono consecutivi in un file */
 } ERef;
 
-typedef struct { int eid; uint8_t *base; uint64_t used; } Slot;
+/* I sei pezzi non sono adiacenti nel file, ma non devono esserlo nemmeno in
+ * memoria: expert_mats ci costruisce sopra solo tre viste in sola lettura. */
+typedef struct { int eid; uint8_t *piece[GLM53_EXPERT_PIECES]; uint8_t *own; uint64_t used; } Slot;
 typedef struct LCache { Slot *s; int n, cap; } LCache;
 
 /* Lunghezze e posizioni dei sei pezzi dentro allo slot. Gate e up sono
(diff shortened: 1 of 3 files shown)


ronaldcklomp:
Also handled the two housekeeping items from your review:

- **#1321 rebased** onto this branch (`d9d38c5` → `7993e5c`) — clean rebase, no conflicts (the two PRs touch different functions in `glm53.c`), builds clean. It now applies on top of this one, as requested.
- Left **#1324 open** rather than closing it myself, since you're the one who should make that call — but this PR does supersede it, for the reasons in the description above.

JustVugg (approved):
This is the rework I asked for, done exactly as scoped, and it is smaller than the version it replaces because it deletes the per-engine tables instead of adding a third.

**Verified here rather than taken on trust:**

- gate genuinely off by default — `st_map_experts_enabled()` returns 0 without the variable and every call site falls back to its unmodified `pread`;
- `glm53`, `qwen38`, `colibri`, `kimi_k3` all build (everything that includes `st.h`);
- `test_st_pread` green;
- Kimi token-exact oracle passes **identically** with the gate off and on — so where it is used, it is a reordering of how bytes are fetched, not a change to what is computed.

**And the datapoint I owed you.** GLM-5.3, 391 GB int4-g64 container, 25.4 GB of RAM, 8 experts/layer — the low-memory regime your 247 GB box cannot show:

| | pread | `COLI_MAP_EXPERTS=1` |
|---|---|---|
| decode, 24 tokens | 7824 s | **6546 s (−16.3%)** |
| expert-disk wait | 4370 s | 3961 s |
| expert-matmul | 1003 s | 697 s |
| experts read per token | 575.0 | 575.0 — identical |

Same experts fetched, 16% less time: it copies less, exactly the mechanism you described. **It wins under memory pressure too**, which was the open question.

Two honesties about that number: single runs, and I measured 14% run-to-run variance on this box between two identical `pread` runs — so treat −16% as a direction, not a figure. A deterministic pair (`TEMP=0`, fixed seed) is running now and I will post it.

**One thing for the follow-up series**, observed while measuring: with mapping on, the planner reports an 18.2 GB budget and warns of a 0.8 GB overshoot that is not real — it counts reclaimable file pages as anonymous. That is the RSS-guard work I said was mine, and now I have evidence it is needed rather than a hunch.

Merging opt-in. The default flip waits for the deterministic run and the guard fix.

Issue #1045, opened 2026-08-16
[Feature]: Can you please consider adding qwen 3.8


blenderman94:
### Problem

Hello im really enjoying using kilibri so far and im wondered about if you maybe can add this model into the support list. https://huggingface.co/Qwen/Qwen3.8-27B   i know we can run theese models with quantizations everywhere but it would be curious to run the full precision modell. thx in advance if you consider doing this. have a nice day and keep up the good work with kolibri.

### Proposed solution

-

rofl0r:
qwen-3.8 27b is not a MoE model, so it's not really a target for this engine. however the 2.4T checkpoint is one, and should further MoE models from the family be released they should already be supported once #712 is merged.

JustVugg:
Thanks for the kind words — and this is a genuinely interesting request, but the answer needs a distinction that decides everything.

I checked the model's config: **Qwen3.8-27B is a *dense* model, not a Mixture-of-Experts one.**

```
model_type:        qwen3_5
architectures:     Qwen3_5ForConditionalGeneration
num_hidden_layers: 64      hidden_size: 5120
MoE fields:        none (no num_experts / n_routed_experts / num_experts_per_tok)
```

That matters because colibrì's entire mechanism is **streaming the experts a token actually routes to**. In an MoE model ~5% of the weights are active per token, so we can keep the dense path resident and stream the rest on demand. In a dense model **every weight is active on every token** — there is nothing to stream selectively, and colibrì's core advantage evaporates. For dense models, llama.cpp and Ollama are already excellent and I'd genuinely point you there rather than pretend we'd be better.

**But your actual goal — running it at full precision — is a real open thread here.** Qwen3.8-27B in bf16 is roughly 54 GB, which is exactly the "RAM is the hard blocker" case. There is an open proposal for a **streamable dense trunk** (#826) precisely so colibrì can run where the *dense* part doesn't fit. If that lands, an unquantized 27B on a small machine becomes plausible. That's the issue to follow — I've linked this one there.

Two more things worth knowing:

- Its hybrid **linear + full attention** layer alternation is architecturally close to what the Qwen3.6 engine in review (#712) already implements, so that machinery would partly transfer.
- If you want a **Qwen MoE** on colibrì today, #712 (Qwen3.6-35B-A3B: 35B total / 3B active) is the one under active review.

[...]

PR #1326, opened 2026-09-03
feat(qwen36): name a checkpoint by its geometry, and a converter that refuses what it cannot place

kreuzzelg:
Fixes the banner defect confirmed in #1045 (comment 5458211750), and lands the pieces that make the Qwen3.8-2.4T-A95B shape a verified one rather than a guessed one — without a single weight of it.

## The defect

`Qwen/Qwen3.8-2.4T-A95B` declares the same `model_type` as Qwen3.6-35B-A3B (`qwen3_5_moe_text`) and resolves to the qwen36 engine today, where the banner announces it as **"Qwen3.6-35B-A3B · 35B"**. A loader that misreports what it loaded costs someone a day.

## Registry: display by geometry

A `model_type` resolves to exactly one family (`_BY_TYPE`), so a second descriptor is not available. Instead the descriptor carries `display_variants`: a `DisplayVariant` pairs a geometry (config keys that must match exactly) with a name, and `display_for(resolved)` picks the first match.

| geometry | banner |
|---|---|
| 40 layers / 256 experts / hidden 2048 | `Qwen3.6-35B-A3B · 35B MoE · 21 GB on disk` |
| 92 / 512 / 8192 | `Qwen3.8-2.4T-A95B · 2.4T MoE` |
| anything else (a tiny fixture, a third sibling) | `qwen3_5_moe_text · 8L x 8E MoE` — its own model_type and measured numbers, never a sibling's parameter count |

The registry refuses a variant that matches everything, and a family whose static `display_name` is not among its variants (the READMEs are held to that name by the existing contract test). `coli`'s banner reads the family config, so the VL-wrapped 35B repo (`text_config`) and the flat 2.4T config both display right. `public_metadata` carries the variants.

## Converter: every tensor name is placed, or it stops

While checking the 2.4T index I found the class of defect you described for GLM-5.3, live in this converter. It selected `<prefix>layers.<i>.` plus four globals and let the rest fall through unmentioned. That skipped the `mtp.*` head and the `visual.*` tower **by accident**, and would have copied an unfamiliar tensor inside a layer as f16 without a word.

`tools/qwen36_tensor_kinds.py` now classifies every name before the first shard is read: layer tensors by an exact suffix list, `mtp.*` / `visual.*` skipped **on purpose** with a count and a reason (`skipping 19 `mtp.*` tensor(s): multi-token-prediction head …, config mtp_num_hidden_layers=1`), anything else refused with the names. Torch-free, so `tests/test_qwen36_tensor_kinds.py` runs wherever the registry tests run; it pins the contract to the safetensors indexes of both real checkpoints (1609 and 1045 tensors) and to the per-expert layout transformers saves.

One correction on the way: a comment claimed the real 35B ships per-expert tensors. **Both real checkpoints ship the fused layout** (`experts.gate_up_proj` / `down_proj`, one tensor per layer) and both carry the mtp head; only the tiny fixture is per-expert. So the "five-minute index check" I proposed in #1045 was aimed at the wrong regex — `expert_inventory` describes the *container*, not the checkpoint — and the real thing to verify was the fused split and the deliberate skip. Both are now under test.

## Fixture: the 2.4T shape at toy widths, in the real layout

`make_qwen36_tiny.py --geometry qwen38-2p4t`: 92 layers / interval 4 / 512 experts top-10 / 16:1 attention heads / 8:1 DeltaNet heads, and the saved shard rewritten into what the checkpoints ship — fused experts plus a one-layer `mtp.*` head with `mtp_num_hidden_layers: 1`. The converter has to split the one and skip the other; the engine has to match the transformers reference token for token. The default preset reproduces the shipped fixture's reference exactly, so the existing gate is unchanged.

CI (`qwen36-tiny-check`) gains a step: build the 2.4T fixture, assert the converter *said* it skipped 19 mtp tensors and placed 1587 in 92 layers, assert no `mtp.*` survives in the container, token-exact at cap 1 / 2 / 512; the sanitizer step now runs both fixtures.

## Verified

- old fixture: reference identical, converted shard **bit-identical** (`cmp`), 16/16 at cap 1/2/8
- 2.4T geometry: 19 mtp tensors skipped and said so, 1587 placed in 92 layers, **16/16 at cap 1/2/512**, 0.35 s per run; generator 17 s, converter 9 s
- ASan/UBSan (`nm` confirms 31 `__asan` symbols — the lesson of #1277): no diagnostic on either fixture under `PILOT=1 WIDE=2`
- registry planner on the real 2.4T config: KV **1.44 GiB @ 8k, 46.0 GiB @ 256k**, DeltaNet state 0.55 GiB context-free — the numbers quoted in #1045. The engine's KV cache is F32 (`ensure_kv` → `falloc`), so they stand; the F16 caveat I raised there is closed
- python suite: 743 tests, OK (skipped=35); the new CI step run locally word for word
- `--seed 3` for the 2.4T rung: the default seed collapses this geometry's reference to one repeated token, which a shape error could still reproduce; seed 3 gives twelve distinct tokens over sixteen

## Not in this PR, and said in `docs/qwen36.md`

Running the 2.4T. The warmstart keeps every expert in RAM by design, ~1.4 TB of int4 for this checkpoint; that needs the disk-streaming design, not this one. What this PR guarantees is that the shape is right before anyone spends the bandwidth.

Thanks to claude code!


JustVugg:
Reviewed and merging. The banner fix is what the title says, and the converter half is the substantial part and is understated rather than oversold: it also adds a missing-layers refusal and a note for a config that declares mtp where the checkpoint carries none, neither of which the body mentions.

The CI step is a real gate rather than a smoke test. It greps the converter's own output for the skip counts and the tensor-kind totals, asserts no mtp tensor survives in the container, and then runs token-exact at cap 1, 2 and 512.

One behaviour change nobody flagged, and I am fine with it but it should be in the body: `resolve_prefix` used to return an empty string when neither known prefix matched and now returns `model.` unconditionally, so a checkpoint with bare `layers.0.` keys is refused rather than silently mis-parsed. Fail-safe, and consistent with the rest of the PR.

One request, for auditability rather than correctness. The name lists in `test_qwen36_tensor_kinds.py` are transcribed by hand from two real index files, and the whole contract rests on them, in particular that the mtp tensors are bare `mtp.*` and not `model.mtp.*`. Could you add the sha256 of the two `model.safetensors.index.json` files, or a one-line command that regenerates the lists? Then a reviewer can check the transcription instead of trusting it.

kreuzzelg:
Thanks for the merge and the read. The provenance you asked for is in #1613: the sha256 of both `model.safetensors.index.json` files in the test header, the one line that regenerates a collapsed list from an index, and an `IndexBackedTest` that checks the lists against the real files when `QWEN36_INDEX_DIR` points at them (skipped in CI). For the 2.4T index the regenerated list is `QWEN38_2P4T` exactly; for the 35B it is the same 46 names under `model.language_model.` plus 21 `model.visual.*` kinds. You are right about `resolve_prefix` too -- refusing bare `layers.0.` keys instead of mis-parsing them was intended, and belonged in the body.

PR #1337, opened 2026-09-04
docs: report GLM-5.3-Flash PRO 6000 speed and quality findings

lEWFkRAD:
## Summary

Add an external llama.cpp field report for GLM-5.3-Flash on one RTX PRO 6000 Blackwell 96 GB, linked from the GLM-5.3 documentation. The final unprofiled v8/Q8-KV comparison measured 34.3605159041 generated tokens/s per user without speculation and 42.9185939037 with Q8 DFlash2 depth 3, for four users with 89,400-token prefixes, one sampled seed each and 512 measured output tokens each. Each arm has one measured run; these are not repeated-run medians.

Q8 was 24.91% faster, but all four outputs spent 512 tokens in reasoning and produced no final answer. No-spec produced two unanswered and two incorrect responses. Original source/prompt/token hashes match, but slot 3's sampled seed differs, so this is not an exact four-slot continuation comparison. GLM optimization was stopped because answer quality failed; neither the 60 tokens/s/user target nor useful-answer acceptance was reached.

The report preserves the clearly labeled historical speed table, sparse-padding and FP32 attention findings, finite Low/High quality matrix, speculative EOS accounting and two invalid profiles. These measure a separately patched llama.cpp runtime, not Colibri's engine. Hardware, normalized commands, warm-up policy, run counts, provenance and reproducibility gaps are explicit. The linked immutable companion evidence snapshot covers earlier results only and is explicitly not evidence for the final 42.9186 result.

## Validation

- Reconciled final rates, raw-event windows, reasoning/final-token counts and seed-matching limits against retained receipts and an independent local audit; reviewed the report for raw prompts, outputs, identities and internal operations.
- Checked Markdown links and `git diff --check`; verified final receipt and benchmark hashes locally. No raw benchmark artifacts are added to Colibri.
- [ ] `make -C c check` — not run for this documentation-only change.
- [ ] CUDA changes were tested with `make -C c cuda-test` — not applicable; no engine changes.
- [ ] Performance claims include hardware, commands, and repeatable measurements — hardware and command details are included, but these are single observations; repeated-run medians and a complete public reproduction package remain unavailable.

## Compatibility

- [x] The default CPU build remains dependency-free.
- [x] No model files, generated binaries, or benchmark artifacts are included; this PR adds Markdown only and links the minimized evidence separately.

### DIFF
--- a/docs/experiments/glm53-flash-rtx-pro-6000-2026-09-04.md
+++ b/docs/experiments/glm53-flash-rtx-pro-6000-2026-09-04.md
@@ -1,0 +1,130 @@
+# GLM-5.3-Flash on one RTX PRO 6000: progress through September 4, 2026
+
+**External implementation report:** these measurements use a separately patched llama.cpp runtime and do not benchmark or validate Colibri's `c/glm53.c` engine. They are shared for review of numerical correctness, long-context measurement and profiling methodology.
+
+The best measured result is **36.1237 generated tokens/s per user for four concurrent users**, each with an 89,400-token prefix, one sampled seed, and 512 measured output tokens. The target remains **60 tokens/s per user**. Correctness and profiling work has narrowed the remaining questions, but neither performance nor useful-answer acceptance has been reached.
+
+## Configuration and measurement
+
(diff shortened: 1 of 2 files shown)


JustVugg:
Thank you for the care in this write-up, but it does not belong in this repository: as the first paragraph says, the measurements are of a separately patched llama.cpp runtime and do not exercise colibri's glm53 engine. docs/experiments holds experiments on this engine; a report on another runtime, however rigorous, would mislead a reader looking for what colibri does on that card. If you run GLM-5.3-Flash through coli serve on the PRO 6000, that datapoint (with the correctness line the protocol now asks for) is welcome as a benchmarks.md row. Closing.

PR #1342, opened 2026-09-04
perf: O(1) LRU victim selection via intrusive recency lists (#1050)

Petsku01:
## Summary

Replaces the remaining O(cap) victim-selection scans in the three MoE engines (OLMoE, Qwen3.6, GLM) with intrusive doubly-linked recency lists that mirror the authoritative `used` stamps. Victim selection drops from O(cap) per miss to O(1); hit paths stay O(1) (`victim_touch`); LRU semantics are exactly preserved.

Follow-up to #1215 (per-expert hash index): that PR removed the O(cap) lookup scan; this removes the O(cap) victim scan. Addresses the scan-cost half of #1050 (the tok/s non-monotonicity datapoint showed eviction churn dominates small-cap throughput; the scan itself is only ~1-2% of per-miss cost, so this PR is a complexity/structure change, not an end-to-end speedup — see Benchmark below; the victim-pick cycle itself is measured at 2.9-6.8 ns (list) vs 457-462 ns (legacy scan) at cap=219, see the cycle-level microbenchmark below).

## Changes

- **olmoe.c**: two lists per cache (ev-list = resident && !pinned, pin-list = resident && pinned). `victim_pick` = ev-head else pin-head else -1 — same contract as the legacy 3-phase scan. Pin flips re-file via `victim_refile`; the pin-list is maintained in `used` order by an O(k) insert-scan (k bounded by the pin budget) so the all-pinned fallback matches the legacy scan even for delayed pin flips. Pilot speculation restricted to unpinned victims (`victim_pick_ex(allow_pinned=0)`), matching the legacy pilot scan exactly. Kill-switch: `COLI_VICTIM_SCAN=1` restores the legacy linear scan verbatim.
- **qwen36.c**: same derived change. `apply_resident` now re-files newly pinned slots ev->pin (previously they stayed ev-listable); pilot path uses `victim_pick_ex` as above.
- **colibri.c (GLM)**: single ev-list per layer (members: resident with slab and a fresh `used` stamp). `eslot_victim_pick` = #1034 free-with-slab reuse (per-layer counter, O(free)) else ev-head else legacy scan fallback. Membership is written only after the fresh `used` stamp (refile at all three pilot sites: SPMC, URING finalize, CUDA promo); per-layer heads/tails are initialized to -1 (not calloc-zero, which would read as head=0).
- **tests/test_olmoe_victim_index.c** (new): 10 categories — exact-LRU pick, pinned exclusion, all-pinned fallback, in-flight exclusion/republish, kill-switch with scan counters, churn with structural list audit, cap scaling, delayed pin-flip differential vs the legacy scan, speculation pin-exclusion, partial-pin state.
- **Makefile**: test target for the new file.

## Invariants

- `used` remains the authoritative recency stamp; the lists only mirror it.
- List maintenance happens at exactly the same lock sites the scans used (`g_pilot_mx`); no new synchronization.
- Links are slot indices, so realloc/moves cannot invalidate them.
- In-flight slots (eid < 0) are off-list; they re-enter via publish/refile.

## Verification

- gcc -O2 -Wall -Wextra: warning counts identical to base in all three engines (olmoe 26, qwen36 17, colibri 23); zero errors.
- test_olmoe_victim_index: 10/10 categories pass.
- bench_olmoe_victim_index (build on demand, not a test gate): 2.9-6.8 ns/cycle list pick vs 457-462 ns legacy scan at cap=219.
- Existing gates: test_olmoe_cache_index, test_qwen36_cache_index, test_glm_cache_index: all pass.
- make test: 672 tests OK (skipped=59), 175.7 s.

## Review history (round 1, two independent models)

Pre-registered review with Sol (gpt-5.6-sol, correctness lens) and Grok (grok-4.5, adversarial lens), dispatched blind in parallel. 11 findings, 7 structural, all fixed and covered by the new test assertions:

1. GLM list heads/tails calloc-zeroed (read as head=0) -> per-layer -1 init.
2. GLM publish ran before the fresh `used` stamp -> pilot residents could stay permanently off-list -> membership moved to refile-after-stamp.
3. GLM free-with-slab reuse priority was reversed (ev-head won over reuse candidates) -> counter + reuse loop first, matching the legacy scan.
4. Pin-list tail-append broke `used` order on delayed pin flips -> O(k) insert-scan (differential test added).
5. qwen36 `apply_resident` pinned without re-file -> refile on 0->1 transitions.
6. Pilot speculation could displace pinned slots (contract drift vs the legacy 2-phase pilot scan) -> `victim_pick_ex(allow_pinned=0)`.
7. Stale struct comment on the rlist encoding -> corrected (0 = unlinked, 1 = ev, 2 = pin, calloc-safe).

No LRU divergence, no dangling-link paths, no new synchronization — confirmed by both reviewers (Q1/Q3/Q4).

## Paired benchmark (OLMoE-1B-7B int8, cap=32, PPL=1, 6 threads, WSL2 14GB)

Correctness parity is exact: **TF-NLL 3.9657 and hit rate 79.4% in all 20 runs** (base 23b8bda vs this branch; cache behavior unchanged, 16152 hits + 4200 misses, RSS 1.79 GB after load on both sides).

### Round 1 (pairs 1-5, interleaved start order)

| pair | base tok/s | list tok/s |
|---|---|---|
| 1 | 0.61 | 0.64 |
| 2 | 0.78 | 0.70 |
| 3 | 0.73 | 0.76 |
| 4 | 0.83 | 0.80 |
| 5 | 0.82 | 0.67 |

### Round 2 (pairs 6-10, same protocol)

| pair | base tok/s | list tok/s |
|---|---|---|
| 6 | 0.80 | 1.95 |
| 7 | 0.77 | 0.76 |
| 8 | 0.97 | 0.91 |
| 9 | 0.96 | 0.72 |
| 10 | 0.98 | 0.70 |

**n=10 combined:** base median 0.81 (sd 0.12), list median 0.74 (sd 0.39 — pair 6 list = 1.95 is a 2.4x outlier whose load phase ran in 3.1 s vs 15.8 s for the paired base run in the same minute, i.e. it started on an idle machine before background load returned; excluding it, list median 0.72, sd 0.08). Paired sign test: base wins 7/10 pairs, two-sided p ~ 0.34 — **not statistically significant**. Host load average rose 0.2 -> 1.8-2.0 across the session; the slowest list runs (0.67/0.72/0.70) co-occur with the loaded window.

**Conclusion: no reliable throughput difference in either direction on this host/workload.** The O(cap)->O(1) change removes ~32 list-node visits per miss; the miss itself is a ~6 MiB expert read, so the end-to-end effect is below measurement noise, as the #1050 datapoint predicted. The PR makes no throughput claim: the deliverables are the O(1) structural guarantee (asserted by the victim-index test categories), exact LRU-order equivalence, and the legacy scan available verbatim behind `COLI_VICTIM_SCAN=1` (verified on the new binary: reproduces 3.9657/79.4%).

## Cycle-level microbenchmark (added 10.9, d30d6ea)

`tests/bench_olmoe_victim_index.c` (new, build on demand like bench_dsa_select: `make tests/bench_olmoe_victim_index`) measures the victim-pick cycle in isolation at cap=219 on the same state shape (publish-stamped `used`, every miss at full cap, all under `g_pilot_mx`): 2M pick+hide+publish cycles.

- list pick: **2.9-6.8 ns/cycle** (three runs: 2.9, 6.8, 6.8)
- legacy O(cap) scan: **457-462 ns/cycle** (~438M scan steps total)
- **speedup on the contended cycle: ~50-150x** at cap=219

This is the cycle the routing loop pays under `g_pilot_mx` on every miss at full cap — the end-to-end tok/s effect still stays below noise (6 MiB expert read dominates), which the paired benchmark above measures honestly.

## Notes / caveats

- The kill-switch env var is parsed in qwen36 for parity but qwen36's victim path is list-only; the fallback there is structural (ev-head -> pin-head), which is why the differential tests assert scan/list agreement in olmoe.
- Possible follow-up (not in this PR): a model-free GLM-finalizer test that exercises the URING/SPMC publish paths in isolation; current GLM coverage is via test_glm_cache_index + the in-engine paths exercised by make test.


Petsku01:
Rebased onto upstream dev (f58a267) to keep the diff clean (5 files: c/Makefile, c/colibri.c, c/olmoe.c, c/qwen36.c, c/tests/test_olmoe_victim_index.c).

Local verification on the rebased head (4acde78): gcc -O2 -Wall -Wextra warning counts identical to dev base (olmoe 26, qwen36 17, colibri 27 incl. one pre-existing -Wunused-function in dev), test_olmoe_victim_index 10/10 categories, and all three cache-index gates (olmoe/qwen36/glm) pass. n=10 paired benchmark data is in the description; no throughput claim.

Petsku01:
Ready for review — taken out of draft. CI is green (full check suite on this head). Happy to address any feedback.

Local verification recap: victim-index test 10/10, cache-index gates pass, warning counts match base. No throughput claim; structural O(1) + exact LRU parity only.

JustVugg:
Reviewed, and holding this one rather than taking it into the release. Not a rejection: the work is careful and the microbenchmark is honest.

The reason is the trade you describe yourself. The victim-pick cycle goes from 457-462 ns to 2.9-6.8 ns, but you also say the scan is only 1-2% of per-miss cost and that there is no end-to-end speedup. Against that, the change rewrites the eviction bookkeeping of three engines (`colibri.c`, `olmoe.c`, `qwen36.c`), and the failure mode of a list that drifts out of agreement with the `used` stamps is not a crash, it is the wrong expert being served: a quality loss nobody would trace back here.

What would move it: an end-to-end A/B on a real model, cold, showing tok/s and expert hit rate unchanged or better on at least one engine. That turns "the structure is better" into "the structure is better and costs nothing", which is the bar for touching a hot path in three engines at once.

One shape that would also make this much easier to take: split it per engine. `olmoe.c` alone, with its test, is reviewable on its own terms and would land quickly; all three at once means a single review has to hold three sets of list invariants in its head.

Leaving it open. Ping me when you have the A/B or a split, and I will pick it up.

Petsku01:
Thanks for the careful review, and the hold is fair — I ran the A/B you asked for and the honest headline is not the one I was hoping for.

**Setup.** OLMoE-1B-7B (i8-quantized, 7 GB) on a 4-physical-core CPU box, cold-start, deterministic teacher-forced run (PPL=1 env): 152 tokens, cap=219, bits=8. The PR's own `COLI_VICTIM_SCAN=1` kill-switch lets both sides run on the *identical binary*, so this is a pure code-path A/B. Four runs, alternating order.

**Quality — unchanged, exactly.** TF-NLL 3.9657 nats/token (ppl 52.76) and expert hit rate 95.3% (hit=19398, miss=954) in **all four runs, zero deviation**. The list cannot serve a wrong expert on this run: it mirrors the `used` stamps and the victim_index test asserts list/stamp agreement. So the quality-risk half of your hold now has a receipt.

**Throughput — no meaningful end-to-end win.** Best-pair comparison: 1.30 vs 0.82 tok/s, but the conservative pair (same disk-warmth position) is 0.83 vs 0.82 — and the run-to-run spread (0.69–1.30) tracks page-cache temperature and OMP variance, not the victim pick. The arithmetic says why: 954 misses × the 457 ns legacy scan ≈ **0.4 ms** out of a ~180 s run. The scan you called 1–2% of per-miss cost is, end-to-end with a disk-resident expert pool, closer to 0.0002% of total runtime.

**So I'm updating my own claim:** the microbenchmark is honest but it measures the wrong thing at this workload. When misses are backed by disk I/O and matmul, the victim-pick mechanism is nowhere near the critical path. The O(1) structure is real (the 2.9–6.8 ns microbench stands) but it buys nothing at cap=219 with this miss profile.

Given that, I don't think this PR should land as-is, and I don't plan to push it. Two honest options:

1. **Close it** — the A/B shows the class of change (mechanism swap, same policy) doesn't clear the hot-path bar in the regime that matters (partial residency). The failure-mode concern you raised stands even with quality receipts, since the win it buys is unmeasurable.
2. **Split olmoe.c + test out alone** — I've rebased the branch onto current `dev` (the PR's base): the two conflicts (Makefile deps, qwen36.c Slot struct merged with the new `pw` field) are resolved, all three engines compile, and CI is green on the rebased head — 28/28 checks pass, mergeable CLEAN. If you'd rather have the smaller review surface available for when cap sizes grow into the range where the scan matters, it's ready to go.

My recommendation is option 1. The right place for this work is probably a cap-policy PR where miss handling itself changes — the measurement setup (deterministic TF-NLL + hit rate + tok/s on identical binaries) is now in place to support that properly, and I'm glad the push for it happened either way.

JustVugg:
Closing as you recommend, and thank you for running the A/B and writing down the result you did not want: 954 misses times a 457 ns scan is 0.4 ms in a 180 s run, so the mechanism is real and the win is unmeasurable in the regime that matters. The measurement setup you built (deterministic TF-NLL plus hit rate plus tok/s on one binary) is the thing to keep; if a cap-policy change ever moves miss handling itself, this branch is the reference for how to prove it.

PR #1323, opened 2026-09-02
feat(qwen36): routed experts from a qpack container through bounded Metal slots

Avicennasis:
## What this adds

The Metal slot-pool half of the original series: `qwen36` runs a Swiftlet-format qpack container's routed experts through #1290's affine Metal path from a bounded, refillable slot pool. Dense weights and norms come from the loader PR this is stacked on (#1343).

Four commits of our own:

1. `feat(metal)`: whole-slot Metal buffers with offset-addressed affine dispatch (`coli_metal_slot_register` / `coli_metal_matmul_affine_slot`): a slot is wrapped zero-copy once (page-aligned only; copy-wrapping would silently detach the GPU from refills), projections addressed by `setBuffer:offset:` with bounds/alignment checks and a stale-descriptor tripwire. Host-pointer path and fmt=4 untouched. New cases in `tests/test_backend_metal.mm`.
2. `feat(qwen36)`: routed experts through bounded slots (`qwen36_qpack.{c,h}`): per-expert blobs read with one `pread` into an LRU pool (default 96, `QWEN36_QPACK_SLOTS=<n>`), per-slot generation counters checked at dispatch so a stale reference is refused rather than served; `moe()` gains the qpack branch ahead of the CUDA tier; `QWEN36_QPACK=<dir>` in `main` with container-vs-config refusal. CPU gate `tests/test_qwen36_qpack.c` (portable affine reference on a synthesized container; runs on every platform, no GPU).
3. `test(qwen36)`: Apple parity gates (`make qwen36-metal-test`, fixture-driven, loud skip marker when no fixture): one-layer MoE and prefill/decode logits vs the CPU reference under forced eviction, stale-reference refusal, resident-buffer accounting.
4. `feat(qwen36)`: end-of-run `[qpack] projections: metal=… cpu=… | slots=… fills=… evictions=…` counters, so a CPU fallback cannot masquerade as GPU parity.

Own diff: 9 files, +1775/-41 on top of the loader PR.

## Stacked on

#1343 (dense weights, norms, snap-view tool), which is itself stacked on #1290, #1319 and #1320. All of their commits show in this diff until they land: the first 10 commits here are the loader PR's branch verbatim.

Rebased onto `dev` at `f58a2679` on 2026-09-04. Two conflicts, both in commit 2:

[...]

PR #1343, opened 2026-09-04
feat(qwen36): load MLX-affine dense weights and norms from a qpack container

Avicennasis:
## What this adds

The loader half of what was #1323: `qwen36` reads a Swiftlet-format qpack container's dense weights and norms from the container's own `model.safetensors`, through #1290's affine contract. Nothing here touches Metal; the routed experts are unchanged, and the Metal slot pool that consumes them follows in #1323, which is now stacked on this PR.

Four commits of our own:

1. `feat(affine)`: a loader-side dequant oracle in the affine contract (`coli_affine_dequant_ref`, same bit conventions as the matmul reference), with `tests/test_affine_dequant.c`.
2. `feat(qwen36)`: dense weights from the container. `st.h` gains U32 as an additive dtype; `load_t_affine` expands the MLX-affine U32 triples at load with geometry derived from the shapes and re-validated against the config (a mismatch is a refusal, never a plausible out-of-bounds read). `tests/test_qwen36_dense_affine.c`.
3. `feat(qwen36)`: norm dialect declared, not guessed. mlx-lm materialises the `+1` in the RMSNorm gammas the engine applies itself; a `zero_centered_norms` meta flag (default = the HF convention, so converted snapshots are unchanged) unshifts once at load.
4. `tools(qwen36)`: `make_qwen36_qpack_snap.py` (stdlib-only) builds the flat `config.json` + shape-derived `qwen36_meta.json` snap view with symlinks into the read-only container, following the `convert_qwen36.py` contract ("engine reads flat config + meta; tooling stays smart"); `tests/test_make_qwen36_qpack_snap.py`.

Own diff: 8 files, +1221/-16 on top of the prerequisites.

## Stacked on

- #1290 (four commits: affine contract, packed affine Metal GEMV, strict qpack reader, inspector)
- #1319 (tokenizer merges in both spellings) and #1320 (Homebrew prefix fallback)

Their commits show in this diff until they land (16 of the 21 files, +2437/-35). The branch is a replay of their current heads (`7f3fe6c9`, `cfd51066`, `f01621ff`) onto `dev` at `f58a2679`; each replayed commit has the same patch-id as its source (`git range-diff` flags #1319/#1320 on hunk context only).

## Validation

Linux (x86-64, gcc), this branch at `05868e5b`:

- `make -C c check` (clean, portable build, `test-c` + `test-python`): exit 0. `test-c` ran 112 gate binaries, none failed; `test-python` 739 tests OK, 35 skipped.
- `make -C c test-asan`: exit 0, the same 112 binaries clean under ASan + UBSan.

Nothing in these four commits is Metal-specific, so CI's macOS `make check` lane compiles and runs the same gates as Linux. Not re-run in this pass (no Apple machine): the round-trip of the production container's dense tensors against an independent pure-Python dequant, reported as `max |diff| 0` in the original #1323 description on the pre-rebase series.

## Run recipe

```sh
python3 c/tools/make_qwen36_qpack_snap.py --container /path/to/model.qpack --out /path/to/snap
```

builds the snap view the engine reads through `SNAP=<dir>`. Running a qpack container end to end also needs the routed-expert path (`QWEN36_QPACK=<dir>`, `METAL=1`), which is #1323.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,8 @@ c/tests/test_*
 !c/tests/test_*.cu
 !c/tests/test_*.mm
 !c/tests/test_*.py
+# scratch fixture directories the test binaries write at run time
+c/tests/tmp_*
 # bench_* and fuzz_* build to extensionless binaries in the same directory and
 # were not covered here, which is how c/tests/bench_omp_grain reached the tree.
 c/tests/bench_*
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -32,6 +32,17 @@ executes an unmodified vLLM checkout (commit
 `ffd46bfab2128bb84146050e98b51a617c6575ab`) as a behavioural oracle for the
 native port; no vLLM code is vendored.
 
+## Swiftlet MLX affine Metal GEMV
+
+The MLX affine Q4/Q8 kernels in `c/backend_metal.mm` are adapted from
+`gemv_affine_fast` and `gemv_affine_fast8` in Swiftlet:
+https://github.com/leonickson1/Swiftlet at commit
+`b3a04676748c7597800c5bcc8b80a32508f9f43d`.
+
+Swiftlet is licensed under Apache License 2.0. The kernels were modified for
+Colibri's checked descriptor, batched dispatch, buffer ownership, and fallback
+contract. This repository's `LICENSE` contains the applicable Apache 2.0 text.
+
 ## DeepGEMM sm120 headers (fetched, not vendored: `c/third_party/deepgemm/`)
 
 The DeepSeek V4 CUDA tier's DeepGEMM flavour (`make cuda-dsv4-dg-dll`,
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -32,7 +32,12 @@ executes an unmodified vLLM checkout (commit
 `ffd46bfab2128bb84146050e98b51a617c6575ab`) as a behavioural oracle for the
 native port; no vLLM code is vendored.
 
-## Swiftlet MLX affine Metal GEMV
+## Swiftlet qpack and MLX affine Metal GEMV
+
+`c/qpack.c` and `c/qpack.h` implement the Swiftlet qpack v1 container schema
+documented by `Sources/SwiftletCore/Qpack.swift` at the commit linked below.
+The reader is a new portable implementation for Colibri and does not copy
+Swift source code.
 
 The MLX affine Q4/Q8 kernels in `c/backend_metal.mm` are adapted from
 `gemv_affine_fast` and `gemv_affine_fast8` in Swiftlet:
--- a/c/Makefile
+++ b/c/Makefile
@@ -52,6 +52,13 @@ ifneq (,$(DARWIN))
 # Niente -march: su arm64 NEON e' baseline (i kernel __ARM_NEON si attivano da soli).
 CC      = clang
 OMPDIR := $(shell brew --prefix libomp 2>/dev/null)
+# CLT-only Macs routinely have Homebrew installed but off the non-interactive
+# PATH (`make` from scripts, SSH, CI), which turned this probe into a silent
+# single-threaded build with libomp sitting right there on disk. Fall back to
+# the two standard Homebrew prefixes before concluding it is absent.
+ifeq ($(OMPDIR),)
+OMPDIR := $(firstword $(wildcard /opt/homebrew/opt/libomp /usr/local/opt/libomp))
+endif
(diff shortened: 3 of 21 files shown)


Avicennasis:
Apple/Metal gates at `05868e5b` from a fresh clone on an M4 Max (macOS 26.6.2, Command Line Tools, Apple clang 21.0.0): `make -C c metal-test` rc 0 (99 ok / 0 mismatch), `make -C c gemm-test` rc 0 (12/12, `GEMM large-batch: all clean`), `make -C c {qwen36,colibri,inkling,kimi_k3} METAL=1` rc 0. On an M1 Max the same ref gives 98 ok / 1 mismatch in `metal-test` — the single `fp8 LUT exactness (256/256 codes via GPU kernel) 96/256 mismatches` line, which also fails at `dev` `f58a2679` without this PR and passes on the M4 Max, so it is pre-existing and platform-specific, not from this PR. Per-target table in #1323.

JustVugg:
This is the base of the qpack stack and the one to rebase first. `c/qwen36.c` has moved 18 commits and is the busiest engine file in the repo, `c/Makefile` 39, `c/st.h` 3, `c/backend_metal.mm` 3. Your `qwen36.c` hunk at +174/-18 is the real work; the rest of what you touch is new files that cannot conflict.

Nothing has overtaken it: `c/qpack.c` and `c/affine_quant.h` are still absent from dev, and #1319 landing from this same stack is the sign the work is being taken.

I have closed #1290, since its 13 files are a strict subset of this PR's 21 and you should not be rebasing three branches for one body of work. Do this one, then #1323 on top of it.

Avicennasis:
Rebased onto `dev` `a90bed95` (the #1326 merge): `feat/qwen36-qpack-loader` `05868e5b` -> `f0733efa`, nine commits, zero behind. `git range-diff`: 7/9 patches identical; `0a92b5c0` and `03497454` differ only by the conflict integration below, and the one-line `affine_quant.h` prerequisite was re-applied to `dev`'s current `qwen36$(EXE)` line. Conflicts, each resolved as the union of both sides:
- `c/backend_metal.mm`: pipeline-state declaration and the `P(...)` creation list -- `dev`'s `g_moe_silu_clamped` (`a2148241`) kept beside `g_affine_q4`/`g_affine_q8`/`g_affine_width_ok`.
- `c/tests/test_backend_metal.mm`: `dev`'s `run_f32_noscale` harness and its four `main()` cases (`90ec4f67`) kept; the `affine_f16`/`affine_store_scalar`/`run_affine` helpers and the six affine cases follow them.
- `c/qwen36.c`: include block -- `dev`'s `expert_ffn.h` plus `affine_quant.h`.
- `c/Makefile`: `qwen36$(EXE)` prerequisites -- `dev`'s `expert_ffn.h omp_tune.h kv_prefix.h` plus `affine_quant.h`; the test recipes auto-merged this time.
Gates at `f0733efa`. Linux x86-64 (gcc 13.3.0, Python 3.12.3): `make -C c colibri qwen36 qpack-inspect` and the four new test binaries build; `tools/run_tests.py` 4/4 ok; `tests.test_makefile_deps` 2 OK; `tests.test_make_qwen36_qpack_snap` 6 OK; `git diff --check` clean.
Apple M4 Max (macOS 26.6.2, Apple clang 21.0.0, GNU Make 3.81), fresh clone, checkout by SHA and `git clean -fdxq` before every target: `make -C c qwen36 METAL=1` rc 0 (arm64 Mach-O); `make -C c metal-test` rc 0, 103 ok / 0 mismatch, the six `affine Q4|Q8 / f32|f16|bf16 batched` cases at nerr 0. `dev` `a90bed95` itself: `metal-test` 97 ok / 0 mismatch on the same machine.
#1323 is re-stacked on this head.

JustVugg:
Checked against today's dev for 1.12.1, and it is not a mechanical rebase this time, so it moves to the next release rather than being forced in.

The blocker is `c/json.h`: 16 conflicting hunks. Your branch adds a `strict` flag to the parser; dev merged #1605 two days ago, which added its own strict path (`json_parse_checked`, with `p->error` set on non-JSON whitespace, control characters, bad \u escapes, a NUL in a key and an unterminated string). Two strict modes in the one parser every engine uses is the thing to avoid, so the rebase should drop the `strict` field and route the qpack reader through `json_parse_checked`; if your reader needs a rule #1605 does not enforce, add it there with a test in `tests/test_json.c`. The other three conflicts (`c/Makefile`, one hunk in `c/qwen36.c`, `tests/test_json.c`) are prerequisite lines and the include block.

Also note that `c/qwen36.c` has moved again today (integer-dot trunk, #1674 and #1677 CUDA batching, #1559 mixed layout), so please re-run the qwen36 tiny oracle and the Metal gates on the rebased head. Once #1343 is in, #1323 restacks on it and #1315 goes in with them, as agreed: installers together with the consumer.

Issue #1040, opened 2026-08-15
qwen36 GPU detection: 'none compute-capable' on NVIDIA (GTX 1660 Ti) → CPU fallback


sehHeiden:
## Environment

- GPU: NVIDIA GeForce GTX 1660 Ti (discrete, Vulkan 1.4.315 loader, driver 580.178)
- Branch: `qwen36` (PR #602), built from source (gcc, no Vulkan SDK; headers from vulkan-headers pkg)
- Runs: `SNAP=… PORT=8888 ./qwen36_serve 16 4`

## Symptom

Server starts CPU-only; GPU never used:

```
[gpu] 2 device(s) found but none compute-capable -> CPU backend
[serve] model resident in 9.8s | RSS 9.34 GB | listening on http://0.0.0.0:8888
```

Result: ~0.3 tok/s (CPU-only) for a 35B-A3B model.

## GPU is fine — evidence

```
$ vulkaninfo --summary
GPU0: deviceType = PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
      deviceName = NVIDIA GeForce GTX 1660 Ti
      driverID   = DRIVER_ID_NVIDIA_PROPRIETARY
```

- `/usr/share/vulkan/icd.d/nvidia_icd.json` present, `libvulkan.so.1` loads fine.
- Two devices enumerated (1660 Ti + integrated) — both rejected as 'not compute-capable'.

## Suspected root cause (c/qwen36.c, ~L1846)

Queue-family flags are read via raw buffer arithmetic with hardcoded offsets:

```c
#define VK_QF_OFF_FLAGS 0    /* queueFlags is the FIRST field of VkQueueFamilyProperties */
#define VK_QF_STRIDE 32      /* sizeof(VkQueueFamilyProperties) = 4+4+12+4+8 = 32 */
...
uint32_t flags = *(uint32_t*)(qf + (size_t)j*VK_QF_STRIDE + VK_QF_OFF_FLAGS);
if (flags & VK_QUEUE_COMPUTE_BIT) { compute_devs++; break; }
```

If the returned `VkQueueFamilyProperties` layout/padding differs from the assumed 32-byte stride on this driver, `VK_QUEUE_COMPUTE_BIT` never tests true and every device is rejected.

## Suggestion

Read queue families type-safely instead of char-arithmetic:

```c
VkQueueFamilyProperties *qf = calloc(nf, sizeof(VkQueueFamilyProperties));
[...]

PR #1344, opened 2026-09-04
fix(qwen36): three CUDA expert-tier bugs (#1339, #1340, #1341)

### LINKED ISSUE #1339: qwen36 CUDA tier: qt_issue overruns G.is_x with two or more GPUs (stride 8*D into a 32*D buffer)
Found while working on #1338; present on `dev` @ f58a267, CUDA path only. I will send a fix PR.

## Summary

`qt_issue()` in `c/qwen36_tier.c` writes each device's input replicas into `G.is_x` at stride `8*D` floats per device, but `G.is_x` is allocated as `32*D` floats and a single device can receive up to 32 rows. With two or more GPUs the write can run past the end of the buffer.

## Where

- `c/qwen36_tier.c:201` — `G.is_x=malloc((size_t)32*D*sizeof(float));`
- `c/qwen36_tier.c:429-430` (in `qt_issue`) — `float *xr=G.is_x + (size_t)di*8*G.D;` then `memcpy(xr+(size_t)j*G.D, x, ...)` for `j < c`, where `c = G.is_cnt[di]` can be up to 32 (`is_k[QT_MAX_DEV][32]`, `K>32` rejected at `:405`).

## Effect

Device `di` owns the float range `[8*di*D, 8*di*D + 32*D)`. Anything past `32*D` is out of bounds:

| `ndev` | device 1 overruns when | device 4+ |
|---|---|---|
| 1 | never (`di=0`, 32 rows fit exactly) | — |
| 2 | more than 24 of the K routed experts land on device 1 (`eid % 2 == 1`) | — |
| 5+ | any row on device 4 (`8*4*D == 32*D`) | always |

The overrun is a heap write of up to `8*D` floats (64 KB at D=2048) past `is_x`; with two cards and top-k 32 it needs a token whose routed experts are mostly odd-numbered, so it is intermittent rather than immediate. Reachable with `COLI_GPUS=0,1` (or the auto-selected two devices) and a wide top-k; single-GPU runs and the Vulkan path in #1338 (single device) are unaffected.

## Fix I intend to send

Size the buffer per device and stride by the same amount: allocate `G.ndev*32*D` floats (or `QT_MAX_DEV*32*D`) and use `G.is_x + (size_t)di*32*G.D`. One line each.

### LINKED ISSUE #1340: qwen36 CUDA tier: qt_shutdown never signals cv_take, so pthread_join can hang if a group is open
Found while working on #1338; present on `dev` @ f58a267, CUDA path only. I will send a fix PR.

## Summary

`qt_shutdown()` in `c/qwen36_tier.c` sets `th_stop` and signals only `G.cv`, then `pthread_join`s the uploader. Every other wait in the file loops on `!G.th_stop` but sleeps on `G.cv_take`, which shutdown never signals. If the uploader is parked on the LFRU victim wait when shutdown runs, the join never returns.

## Where

- `c/qwen36_tier.c:496` — `G.th_stop=1; pthread_cond_signal(&G.cv);` (no `cv_take`).
- Waiters on `cv_take` that check `th_stop` and are never woken by shutdown: uploader victim wait `:98` (`while(G.issue_open && !G.th_stop) pthread_cond_wait(&G.cv_take,&G.mx);`), `qt_note_block` `:279`, `qt_note_planned` `:360`, `qt_fill_wait` `:373`.

## When it bites

`issue_open` is set by `qt_issue` and cleared by `qt_take` (`:457`). If the process shuts down between the two — an error exit or `atexit` handler running while a group is in flight, which is exactly when `qt_shutdown` is registered to run (`c/qwen36.c` registers it with `atexit`) — and the uploader has an LFRU swap queued, the uploader sits on `:98`, shutdown signals the wrong condvar, and `pthread_join` at `:497` blocks forever. The `HEAT_FILE` is written before the join, so the visible symptom is a process that never exits after printing the tier stats.

## Fix I intend to send

Broadcast `G.cv_take` alongside `G.cv` when setting `th_stop` (one line), and have the victim path skip the free-and-upload when it wakes because of `th_stop` rather than because the group closed.

### LINKED ISSUE #1341: qwen36 CUDA tier: int8 containers free the weights the tier still points at (regression from #1334)
Found while rebasing #1338 onto #1334; present on `dev` @ f58a267, CUDA path, int8 containers only. I will send a fix PR. This is a regression of the #1331 fix, not a fault in what it set out to do.

## Summary

For int8 containers the warmstart in `c/qwen36.c` hands the tier the live int8 weight pointers, then frees them on the next line. The tier keeps those pointers, and the engine's CPU fallback has nothing to rematerialize from. Any planned int8 expert that is later computed on the CPU dereferences NULL; any later re-staging of such an expert in the tier reads freed memory.

## Where

- `c/qwen36.c:2809-2818` — `wg = expert_is_int4 ? e->g4 : (const uint8_t *)e->g;` … `qt_note_planned(l, eidw, wg, wu, wd, ...)` … `if (!keep8 && e->g) { free(e->g); e->g = e->u = e->d = NULL; }`. The comment above says eviction "rematerializes from g4", which is true for int4 and false for int8 (`e->g4` is NULL on an int8 container; the weights *were* `e->g`).
- `c/qwen36_tier.c:359` — `qt_note_planned` stores the pointer: `if(!s->g4){ s->g4=g4; ... }`. It is used again by `enqueue_locked` → `stage()` at `:251`, and it qualifies the expert as an LFRU promotion candidate at `:393` (`s->g4` non-NULL).
- `c/qwen36.c:1532-1533` — `slot_ensure_int8` returns immediately when `s->g4` is NULL, so for an int8 slot with freed `e->g` it leaves `e->g == NULL`; `:1940-1941` then calls `matmul_qe(g, xs, e->g, ...)` with a NULL weight pointer.

## When it bites

Any path that computes a planned int8 expert on the CPU after warmstart:
- the upload failed (`c/qwen36_tier.c:130-132` leaves `resident=0`, budget marked full),
- `qt_issue` handed the expert back to the CPU (`coli_cuda_expert_group_issue` returned 0),
- the LFRU swap evicted it (`qt_lfru_tick_locked`, every 16 ticks once heat crosses the hysteresis).
The first two are a NULL dereference on the decode thread; the third is a use-after-free in the uploader (`stage()` reads `s->g4`), then the same NULL dereference when the evicted expert is next routed.

int4 containers are unaffected: there the source of truth is the packed `g4/u4/d4`, which is never freed, and `slot_ensure_int8` rematerializes from it.

## Fix I intend to send

Keep the int8 weights resident when the container is int8: make the early free conditional on `expert_is_int4` (the RSS saving it buys came from int4's second copy, which int8 does not have). Alternatively rematerialize from the container on eviction, which is more code and a disk read on the decode path; I would start with the first.

crichalchemist:
## Summary

Three pre-existing bugs in the Qwen3.6 CUDA expert tier, found while porting the tier to Vulkan (#1338) and filed as #1339, #1340, #1341. One commit per issue, each with a test that fails on the tree before its fix. The tests link `qwen36_tier.c` against a fake CUDA backend (`tests/qwen36_fake_cuda.h`, factored out of the existing `test_qwen36_tier_int8.c`), so they run in the plain CPU build and in CI.

- **#1339** `qt_issue` strides each device's input block by `8*D` floats into a buffer of `32*D` floats. With two or more GPUs, a full 32-row issue on device 1 writes past the end of the allocation. Fix: allocate `ndev*32*D` floats and stride by `32*D`. The new struct field `is_x_floats` exists so the test can assert against the real allocation rather than restate the arithmetic.
- **#1340** `qt_shutdown` signals `cv` but never `cv_take`, so the uploader parked on the LFRU victim wait (and `qt_note_*`/`qt_fill_wait` waiters) never notice `th_stop` and `pthread_join` hangs when a group is still open. Fix: broadcast `cv_take` too; the uploader abandons a swap it wakes into during shutdown instead of freeing a victim tensor an in-flight group may still reference.
- **#1341** After #1334, the warmstart hands the tier the live int8 weights and then frees them, assuming eviction rematerialises from `g4`. That only holds for int4; int8 has no second copy, so the tier's pointer dangles and the CPU fallback dereferences NULL. Fix: free the int8 copy only for int4 containers. The warmstart is extracted into `tier_warmstart()` so the engine-level test can drive it without `main`.

Notes for review: with int8 weights kept, the tier's slot pointer aliases a live engine slot; that is safe because `qt_init` requires `cap == n_experts`, so a planned slot is never recycled. The `cv_take` broadcast makes a benign post-shutdown enqueue reachable if a caller races `qt_note_*` against `qt_shutdown` (previously that caller blocked forever); left as is to keep the diff surgical. `qt_shutdown` still frees none of `G`'s allocations (pre-existing, untouched).

## Validation

- [x] `make -C c check` (OK, skipped=35) and `make -C c test-asan` (clean under ASan+UBSan)
- [x] CUDA changes were tested with `make -C c cuda-test` (if applicable) — on a Colab Tesla T4 (driver 580.82.07, CUDA 12.8), commit d937ad2: `make -C c cuda-test CUDA_ARCH=native` passes; the four tier tests pass; `make qwen36 CUDA=1 CUDA_ARCH=native` builds with 0 warnings; the CI tiny fixture (`--ebits 8`, int8 experts) decodes 16/16 tokens against the torch reference with the tier off, on (64/64 experts resident, 100 % VRAM hits), and on with a starved budget (`CUDA_EXPERT_GB=0.00002`: 1/64 resident, 317 CPU misses through the int8 fallback that #1341 fixes), tier-on output identical to tier-off. One warning appeared in that run from Colab's older gcc in the engine-test build, `qwen36.c:1450 -Waggressive-loop-optimizations`, in the pre-existing nibble-unpack tail loop that this PR does not touch; gcc 13 (CI) is clean.
- [x] Performance claims include hardware, commands, and repeatable measurements (no performance claims)

RED evidence, each against the tree without its fix:
- #1339: `test_qwen36_tier_multidev` under ASan with the old stride:
  ```
  =================================================================
  ==1336655==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x525000002100 at pc 0x7b42016fb303 bp 0x7ffeb9de4710 sp 0x7ffeb9de3eb8
  WRITE of size 256 at 0x525000002100 thread T0
      #2 0x5d36d942a1fc in qt_issue tests/../qwen36_tier.c:450
      #3 0x5d36d942c61f in main tests/test_qwen36_tier_multidev.c:76
  0x525000002100 is located 0 bytes after 8192-byte region [0x525000000100,0x525000002100)
  SUMMARY: AddressSanitizer: heap-buffer-overflow ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115 in memcpy
  ```
- #1340: `test_qwen36_tier_shutdown` hangs in `qt_shutdown`; the test's 10 s watchdog fires (`FAIL: qt_shutdown hung`). With the broadcast but without the drop path, the post-shutdown state checks fail (victim tensor freed, incoming expert uploaded after shutdown began).
- #1341: `test_qwen36_tier_int8_engine` against the tree with the extraction but the old free condition:
  ```
  int8 container
    ok   every planned int8 expert reached VRAM
    ok   three uploads per planned expert (gate, up, down)
    FAIL a planned int8 expert still has weights for the CPU fallback
    FAIL those weights are the bytes the loader wrote
    FAIL the pointer the tier kept still targets the live int8 block
  int4 container
    ok   every planned int4 expert reached VRAM
    ok   int4 still drops the int8 copy right after staging (peak RSS)
    ok   slot_ensure_int8 rebuilds it from the packed g4/u4/d4
  FAILED 3
  ```

Build: 0 warnings under the Makefile's flags for the tests and `make qwen36`.

## Compatibility

Fixes #1339, fixes #1340, fixes #1341.

https://claude.ai/code/session_019BacNGNxAJ1M57UdYE2M3N

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1650,6 +1650,12 @@ tests/test_qwen36_tier_int8$(EXE): tests/test_qwen36_tier_int8.c tests/qwen36_fa
 tests/test_qwen36_tier_multidev$(EXE): tests/test_qwen36_tier_multidev.c tests/qwen36_fake_cuda.h qwen36_tier.c qwen36_tier.h backend_cuda.h
 	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
 
+# Same fake backend, single device: proves qt_shutdown returns (under an
+# alarm(10) watchdog) instead of hanging forever when a group is still open
+# and an LFRU swap is parked waiting for cv_take (#1340).
+tests/test_qwen36_tier_shutdown$(EXE): tests/test_qwen36_tier_shutdown.c tests/qwen36_fake_cuda.h qwen36_tier.c qwen36_tier.h backend_cuda.h
+	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
+
 tests/test_serve_poll$(EXE): tests/test_serve_poll.c serve_poll.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/Makefile
+++ b/c/Makefile
@@ -1656,6 +1656,15 @@ tests/test_qwen36_tier_multidev$(EXE): tests/test_qwen36_tier_multidev.c tests/q
 tests/test_qwen36_tier_shutdown$(EXE): tests/test_qwen36_tier_shutdown.c tests/qwen36_fake_cuda.h qwen36_tier.c qwen36_tier.h backend_cuda.h
 	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
 
+# Same fake backend, but driving the ENGINE: the warmstart in qwen36.c hands the
+# tier raw pointers into the RAM expert slots, so only a test that includes
+# qwen36.c can prove the weights behind them are still there afterwards -- on an
+# int8 container there is no packed copy to rebuild them from (#1341). Includes
+# qwen36_tier.c in its own TU like the other tier tests, so nothing else is
+# compiled in and no model file is needed.
+tests/test_qwen36_tier_int8_engine$(EXE): tests/test_qwen36_tier_int8_engine.c tests/qwen36_fake_cuda.h qwen36.c qwen36_tier.c qwen36_tier.h backend_cuda.h cli_args.h st.h json.h compat.h
+	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
+
 tests/test_serve_poll$(EXE): tests/test_serve_poll.c serve_poll.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/Makefile
+++ b/c/Makefile
@@ -1641,7 +1641,13 @@ tests/test_cli_args$(EXE): tests/test_cli_args.c cli_args.h
 # coli_cuda_* e registra cosa riceve. E' il solo modo di provare in CI che un
 # esperto arriva davvero in VRAM e nel formato giusto (#1331) -- un test che si
 # fermasse a "qt_init ritorna 1" sarebbe passato anche col difetto.
-tests/test_qwen36_tier_int8$(EXE): tests/test_qwen36_tier_int8.c qwen36_tier.c qwen36_tier.h backend_cuda.h
+tests/test_qwen36_tier_int8$(EXE): tests/test_qwen36_tier_int8.c tests/qwen36_fake_cuda.h qwen36_tier.c qwen36_tier.h backend_cuda.h
+	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
+
+# Same fake backend (tests/qwen36_fake_cuda.h), two devices: proves the tier's
+# per-device input-replica block (G.is_x) is sized and strided so device 1's
+# block cannot run past the end of the buffer or into device 0's (#1339).
+tests/test_qwen36_tier_multidev$(EXE): tests/test_qwen36_tier_multidev.c tests/qwen36_fake_cuda.h qwen36_tier.c qwen36_tier.h backend_cuda.h
 	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
 
 tests/test_serve_poll$(EXE): tests/test_serve_poll.c serve_poll.h
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -2645,6 +2645,59 @@ static void serve_loop(Model *m){
     }
 }
 
+/* Warmstart body, lifted out of main so a test can drive it against an
+ * in-memory model without a container (the QT_NO_WARMSTART check stays at the
+ * call site). expert_is_int4 is what main probed from the on-disk expert size.
+ */
+static void tier_warmstart(Model *m, int expert_is_int4) {
+    /* Plan the set (heat order), then load+stage IN PARALLEL. The
+     * load path is thread-safe: expert_get locks the layer cache
+     * (g_pilot_mx), st_read_raw uses pread; entries are unique. */
+    double t0 = now_s();
+    int cap_total = m->c.n_layers * m->c.n_experts;
+    int *wpl = malloc((size_t)cap_total*sizeof(int));
+    int *wpe = malloc((size_t)cap_total*sizeof(int));
+    int wn = qt_plan_fill(wpl, wpe, cap_total);
+    /* Load ALL experts into RAM, not just the planned (VRAM) set:
+     * otherwise the first touch of a CPU-fallback expert triggers a
+     * ~12 ms container read in the middle of decode (measured: 139
+     * ms/token on a single-GPU run). Planned ones also go to VRAM. */
+    uint8_t *planned = calloc((size_t)cap_total, 1);
+    for (int i = 0; i < wn; i++) planned[wpl[i]*m->c.n_experts + wpe[i]] = 1;
+    int keep8 = getenv("COLI_KEEP_INT8") != NULL;
+    #pragma omp parallel for schedule(dynamic, 16)
+    for (int gi = 0; gi < cap_total; gi++) {
+        int l = gi / m->c.n_experts, eidw = gi % m->c.n_experts;
+        Slot *e; expert_get(m, l, eidw, &e);
+        /* int4: i puntatori impacchettati; int8: i pesi stessi. Prima
+         * qui si esigeva e->g4, che su un container int8 e' NULL: la
+         * promozione non partiva mai e il budget restava riservato a
+         * vuoto (#1331). */
+        const uint8_t *wg = expert_is_int4 ? e->g4 : (const uint8_t *)e->g;
+        const uint8_t *wu = expert_is_int4 ? e->u4 : (const uint8_t *)e->u;
+        const uint8_t *wd = expert_is_int4 ? e->d4 : (const uint8_t *)e->d;
+        if (planned[gi] && wg) {
+            qt_note_planned(l, eidw, wg, wu, wd, e->gs, e->us, e->ds);
+            /* The staging copy is done. On an int4 container the int8 copy
+             * can go RIGHT AWAY so it never shows up in peak RSS: g4/u4/d4
+             * stay as the source of truth, and slot_ensure_int8() rebuilds
+             * the int8 block on an LFRU eviction with no container access.
+             * An int8 container has no second copy -- e->g4 is NULL -- so
+             * freeing e->g there both dangles the pointer qt_note_planned
+             * just stored (any later stage() reads freed memory) and leaves
+             * slot_ensure_int8() unable to bring the expert back: it returns
+             * early without g4, and the CPU fallback in the decode loop then
+             * dereferences NULL. Keep it (#1341). COLI_KEEP_INT8 keeps its
+             * meaning for int4. */
+            if (!keep8 && expert_is_int4 && e->g) { free(e->g); e->g = e->u = e->d = NULL; }
+        }
+    }
+    qt_fill_wait();
+    free(wpl); free(wpe); free(planned);
+    fprintf(stderr, "[qtier] warmstart (parallel): all %d experts in RAM (int8 only for non-residents), %d in VRAM -- %.1f s\n",
+            cap_total, wn, now_s()-t0);
+}
+
 int main(int argc, char **argv) {
     const char *snap = getenv("SNAP");
     if (!snap) { coli_print_launcher_help("Qwen3.6"); return 1; }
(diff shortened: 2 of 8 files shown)


JustVugg:
Reviewed, and this supersedes the two PRs I had opened — I closed them. Yours came 16 hours first, covers all three, and does the one thing I had written off as impossible.

**On that last point specifically.** I claimed in my own PR that the caller-side violation of #1341 could not be caught without a GPU, because it lives in the warmstart rather than in the tier. `test_qwen36_tier_int8_engine.c` catches it: include `qwen36.c`, build an in-memory int8 model with no container behind it so `expert_get` takes its hit path, drive `tier_warmstart()` against the fake backend. That is the right answer and I had stopped one step short of it. Factoring `qwen36_fake_cuda.h` out so all three tier tests share it is the move that made it possible.

Three commits, one per issue, each with a test that fails before its fix: that is the shape we want.

**One suggestion, on #1341's condition.** You have:

```c
if (!keep8 && expert_is_int4 && e->g) { free(e->g); ... }
```

Correct today. What it encodes, though, is the *current* correspondence between weight format and memory ownership. Whoever adds a third format that also aliases `e->g` — an fp8 container, say — has no reason to read `expert_is_int4` as a statement about ownership, and would reintroduce this exact bug without noticing.

The invariant itself does not mention formats:

```c
if (!keep8 && e->g && wg != (const uint8_t *)e->g) { free(e->g); ... }
```

**Do not free what you just handed over.** A future aliasing format is then correct without anyone remembering; a future format with its own separate copy still frees as before. I verified it across the current two plus two hypothetical formats before suggesting it.

Take it or leave it — the PR is good either way and I am not blocking on a one-line preference. If you would rather not touch it, say so and I will send it as a follow-up with your name on the original.

CI approved and running. Reviewing the rest of the diff now.

JustVugg:
CI is red on Windows UCRT64, and it is the test rather than the fix:

```
tests/test_qwen36_tier_shutdown.c:115: error: 'SIGALRM' undeclared
tests/test_qwen36_tier_shutdown.c:116: error: implicit declaration of function 'alarm'
```

`alarm`/`SIGALRM` are POSIX and MinGW has neither. Everything else is green, and the tier code itself builds there.

I hit exactly this two days ago on #1336 with `pipe`/`dup2`, so the shape is familiar. A watchdog thread is the portable equivalent and keeps the test meaningful on Windows instead of skipping it there:

```c
static void *deadline(void *unused) {
    (void)unused;
    struct timespec limit = {5, 0};
    nanosleep(&limit, NULL);
    fprintf(stderr, "FAIL: qt_shutdown did not return within 5s (#1340)\n");
    _exit(1);
}
/* start it detached before qt_shutdown(), let the process exit normally if we get there */
```

Worth keeping it running on Windows rather than `#ifndef _WIN32`: on #1336 the Windows job — once the test could actually build there — found a real semantic difference in my own code that Linux could not see. A skipped test on the platform you are least able to check locally is the one you most want running.

Say if you would rather I push that as a commit on a branch here and you pull it, same as I did for #1316. The three fixes read well and I would like them landed.

kreuzzelg:
#1339 and #1340 are mine — the tier landed in #713 with both. Thank you for finding them, and more for the shape of the fix: a fake backend that lets the tier run in the CPU build is exactly what my own tests lacked, which is why these two survived. I read the three commits against current `dev` rather than against the description; notes below, two confirmations and three things I would still change.

## Confirmations

**#1340, the abandon path holds up against the tier's accounting.** An LFRU swap is enqueued with `v_eid >= 0`, so `enqueue_locked` never reserves budget for it (`!reserved && v_eid<0` is false) — abandoning it needs no `G.used` correction, and the patch correctly makes none. The two flags it restores are the two that `qt_lfru_tick_locked` and `enqueue_locked` changed: the victim's `resident` (cleared before enqueue) and the newcomer's `queued`. The victim still holds its tensors, so `resident=1` is the truthful state. And the race the test cannot pin — whether the uploader has reached the `cv_take` wait when shutdown fires — is handled in both orders: if it is parked, the broadcast wakes it into the abandon branch; if it has not arrived yet, `!G.th_stop` fails and it skips the wait into the same branch. Nothing depends on the poll in the test having caught it parked.

**The other three `cv_take` waiters exit cleanly too.** `qt_note_planned` and `qt_note_block` wake into `enqueue_locked`, which returns 0 on a full queue — `qt_note_planned` then returns its reservation via the existing `s->planned` branch. `qt_fill_wait` simply returns with `qn > 0`, which at shutdown is what we want.

**#1339 sizing:** `G.ndev` is final at the allocation (the `have < ndev` clamp is thirty lines above it), and `home()` never returns a device index at or beyond `ndev`, so `ndev * 32 * D` is the whole address space `qt_issue` can touch.

## Three things I would still change

**1. Name the row limit once.** After this patch the literal `32` stands in seven places that must agree: `is_k[QT_MAX_DEV][32]`, `K>32` in `qt_issue`, `tg/tu/td[QT_MAX_DEV][32]`, `rows[32]`, the `is_x_floats` computation and the stride. #1339 was three of those disagreeing. A `QT_MAX_ROWS` next to `QT_MAX_DEV`, used in all seven, plus one line in the multidev test that ties the stride to the array the rows are indexed by —

```c
check(sizeof G.is_k[0] / sizeof G.is_k[0][0] == 32 &&
      G.is_x_floats == (size_t)G.ndev * (sizeof G.is_k[0] / sizeof G.is_k[0][0]) * G.D,
      "replica block stride equals the per-device row capacity");
```

— makes the next drift fail here instead of in a driver.

**2. The warmstart line and the docs now say something false on int8.** `"[qtier] warmstart (parallel): all %d experts in RAM (int8 only for non-residents)"` was true because the int8 copy of every resident was freed. With #1341 that free is int4-only, so on an int8 container every resident keeps its weights and the parenthesis is wrong — and so is the RSS saving `docs/qwen36-cuda-tier.md` quotes (40 → 29 GB), which is an int4 number. Both should say so: the tier's RSS advantage is a property of packed containers, an int8 container pays full residency and gets VRAM speed only. Better a line that says "int8 container: all experts stay in RAM" than one that promises a saving the reader will not see.

**3. One cheap assertion for the mixed batch.** The multidev test proves the two blocks are disjoint and inside the buffer. It could also pin *where* they are — device 0's at `G.is_x`, device 1's at `G.is_x + 32*D` — which is the property the stride literal encodes and the one assertion that would have failed on the old code even with a single row on device 1.

## On the test infrastructure

Because `TEST_RULES` collects every `tests/test_*` rule, all four land in `make test-c` without a list edit, and `test-asan` rebuilds that suite under ASan/UBSan in the Sanitizers job. So the multidev test is also an ASan detector for #1339 from the day it merges, and the tier finally has a memory-safety gate that does not need a GPU. I will build the placement work I proposed in #1040 on `qwen36_fake_cuda.h` rather than on a second harness — thank you for factoring it out.

With the watchdog in place of `alarm` this is good to go from where I sit.

JustVugg:
Pushed the MinGW fix to your branch as one commit on top of yours, so nothing of yours moved: `test_qwen36_tier_shutdown.c` now bounds `qt_shutdown()` with a detached watchdog thread instead of `alarm()`/`SIGALRM`, which MinGW does not have. The test already used pthread and nanosleep, so this adds no dependency. Same 10 s, same failure line, same `_exit(1)`.

Verified here: the test passes on Linux, and a probe that never sets the flag exits 1 after exactly 10 s, so the watchdog still catches a hang. CI will say whether Windows agrees.

Reason for the hurry, stated plainly: #1334 is on `dev` since 1.10.1 and this PR is the fix for the use-after-free it introduced (#1341). The next release must not ship one without the other, and this test was the only thing standing between them. If you would rather carry the change yourself, revert my commit and push yours; either is fine.

mfethe1:
All three changes are in, plus the watchdog. I don't have push access to
`qwen36-tier-fixes`, so they're in a cross-fork PR against that branch rather than
pushed directly: **crichalchemist/colibri#1 (`ef3ac1e`)**. @crichalchemist — merge,
cherry-pick, or apply by hand, whichever suits. Two of your requests turned out to be
measurably complementary rather than overlapping, which is worth writing down.

## 1. `QT_MAX_ROWS`, and why your check #1 and your check #3 are both needed

`QT_MAX_ROWS` now stands in all seven places (`is_k`, the `K>` clamp, `tg/tu/td`,
`rows[]`, the topk clamp message, `is_x_floats`, the `qt_issue` stride), and the
multidev test carries your `sizeof` assertion plus a placement pin.

I built both and then deliberately broke the tree in each of the two drift directions,
three repeats each, to check neither assertion was decoration:

| injected drift | sizeof check | placement pin | in-bounds | disjoint |
|---|---|---|---|---|
| stride reverted to `8*D` (the #1339 shape) | fires | **fires** | fires | fires |
| `is_k[..][64]`, stride still 32 | **fires** | passes | passes | passes |

The second row is the one I did not expect. If a later edit widens the row array
without touching the stride, the placement pin is silent — device blocks really are
at `is_x + di*32*D` and really are disjoint; the buffer is just too small for a group
the array now permits. Your `sizeof` check is the only thing that catches it. And in
the first row the placement pin fires on a single-row device-1 issue where in-bounds
and disjointness both pass. So they detect different things and I kept both, rather
than folding one into the other as I had first intended.

Related: the PR's own in-bounds assertion passes under a too-*small* stride in the
mixed batch, because undersized blocks stay inside the allocation while overlapping
each other. In-bounds is not a sufficient invariant here — that is the argument for
the placement pin being an assertion rather than a comment.

## 2. The warmstart line and the docs

You are right, and the docs were worse than the log line. Both now say which container
they are talking about:

- the log line branches on `expert_is_int4` — `"int8 copy dropped for residents, kept
  for non-residents"` vs `"int8 container: all experts keep their weights in RAM"`;
- `docs/qwen36-cuda-tier.md` says the RSS saving is a property of packed containers,
  that an int8 container frees nothing since #1341 and pays full residency for VRAM
  speed, and that the 40 → 29 GB row is an int4 measurement with no int8 analogue.

## 3. The watchdog

`alarm`/`SIGALRM` are gone; a detached thread sleeps 10 s and `_exit(1)`s with the same
`FAIL: qt_shutdown hung` message. Same contract, and the intent is that the test **runs**
on UCRT64 rather than being compiled out — but see the caveat below: I have not built it
there, so that is the goal of the change, not something I have measured.

I checked it is a real oracle before proposing it. Fixed and unfixed trees, alarm and
watchdog harnesses, five runs per cell: identical verdicts in all four cells, 20/20, no
within-cell splits — passing cells 16–150 ms, hanging cells 10046–10270 ms, about two
and a half orders of magnitude apart and non-overlapping. And reverting just the
`pthread_cond_broadcast(&G.cv_take)` in `qt_shutdown` on the current tree reproduces the
hang 3/3 under the watchdog at 10.13–10.22 s, so the deadline is load-bearing and not
merely green.

**Caveat I want stated plainly:** all of the above is macOS/arm64 + Apple clang against
the fake backend. I have **not** compiled or run this on MinGW/UCRT64 — I removed the
non-portable API, I did not measure the Windows build. Until someone builds it there (I
intend to), the CI job is the only thing that will actually prove the port. Nothing here
is a performance claim either; these are liveness and memory-safety gates only.

Thanks for the review — the `#1040` placement work landing on `qwen36_fake_cuda.h` is
exactly why I pulled it out into a shared header.

mfethe1:
Verified on merged `dev` (`e9505363`): all four commits landed and all three fixes are intact in the tree.

Built and ran the tier suite against the fake-CUDA harness on darwin/arm64:

```
test_qwen36_tier_int8          ok
test_qwen36_tier_multidev      ok   (2 devices, per-device budgets)
test_qwen36_tier_shutdown      ok   (thread watchdog, no hang)
test_qwen36_tier_int8_engine   OK   (3/3: planned int4 reach VRAM,
                                     int8 copy dropped after staging,
                                     slot_ensure_int8 rebuilds from packed g4/u4/d4)
```

@JustVugg your SIGALRM→thread swap in `fa8142d7` is the right call — the watchdog fires correctly here and the test is now portable to MinGW without the signal dependency. Thanks for pushing it onto the branch rather than bouncing the PR back.

@kreuzzelg the #1340 shutdown fix reads well in final form — `pthread_cond_broadcast(&G.cv_take)` alongside the existing `signal(&G.cv)` at `qwen36_tier.c:514`, so the `qt_note_planned`/`qt_fill_wait` waiters on the shared condvar all observe `th_stop` and `pthread_join` can't hang.

One note on what this does and doesn't establish: this is the **fake-CUDA harness on CPU**, so it validates the tier's control flow — per-device replica sizing, shutdown wakeup, int8 slot retention — but not real CUDA behaviour. The multidev test proves the tier keeps per-device state separate; it doesn't prove correctness against actual multi-GPU hardware, which I can't test here. Worth someone running it on a real 2-GPU box before relying on #1339 being fully closed.

Nothing outstanding from my side. Thanks both — clean collaboration.

Issue #1345, opened 2026-09-04
[Bug]: metal-test fp8 LUT exactness fails on M1 Max (96/256): fast-math exp2() in mm_gemv fmt=8 decode is 1-2 ulp low for e4m3 exponents 1..6


Avicennasis:
### Version

`dev` @ f58a2679 (post-v1.10.0; merge of #1336). Also reproduces at the two PR heads in #1343 / #1323, and the offending line was introduced in 6ca61a1b (2026-07-22, "fp8: Metal mm_gemv kernel for fmt=7 native FP8-e4m3 passthrough"), so every `dev` since then should be affected on this hardware.

### Environment

- Fails: MacBook Pro, **Apple M1 Max**, 64 GB, macOS 26.6.2 (25G83), Command Line Tools only, Apple clang 21.0.0 (clang-2100.1.1.101), GNU Make 3.81.
- Passes: MacBook Pro, **Apple M4 Max**, 64 GB, same macOS / CLT / clang / make.
- Both runs from fresh `/tmp` clones of the same bundle, `git checkout --detach f58a2679`, `git clean -fdxq`, default `make -C c metal-test` (no flags).

### Steps to reproduce

```
git clone https://github.com/JustVugg/colibri && cd colibri && git checkout --detach f58a2679
make -C c metal-test
```

### Expected

```
  fp8 LUT exactness (256/256 codes via GPU kernel) 0/256 mismatches  ok
...
metal backend tests: ok
```

(That is what the M4 Max prints. Every other case in the suite, including all the other fmt=8 cases, passes on both machines with identical case sets; only the per-hardware `nerr` figures differ.)

### Actual (M1 Max)

```
Metal fmt=8 native FP8-e4m3 passthrough tests:
  fp8 LUT exactness (256/256 codes via GPU kernel) 96/256 mismatches  *** MISMATCH
  fp8 gate/up-shaped O=2048 I=6144 (spec example) S=1 worst_rel=3.01e-08 ...  ok
  ...
metal backend tests: FAILED
```

`make` exits 2, so the whole gate is red on M1 even though nothing else fails.

### Diagnosis

I patched `run_fp8_lut()` to print each mismatching code. All 96 are exactly the codes with e4m3 exponent field 1..6, both signs, all 8 mantissas (0x08–0x37 and 0x88–0xB7). Every "got" value is 1–2 ulp **below** the reference, e.g.:

```
[...]

PR #1346, opened 2026-09-05
fix(metal): bit-exact fp8-e4m3 decode via ieee754 exponent bitcast

SyedYousufFaizan:
# Title

fix(metal): bit-exact fp8-e4m3 decode via ieee754 exponent bitcast

## Overview

This pull request resolves a test failure in `make -C c metal-test` on Apple M1 hardware (Apple7 GPU architecture), where the FP8 LUT exactness check reported 96/256 mismatches against the CPU reference decoder.

## Problem

In `c/backend_metal.mm`, the `mm_gemv` kernel decodes normalized FP8 E4M3 values using:

```metal
float mag = (exp == 0) ? (float(mant) * 0.001953125f)
                        : (1.0f + float(mant)*0.125f) * exp2(float(int(exp) - 7));
```

Because the Metal shader library compiles with fast-math enabled by default, `exp2()` uses hardware approximation instructions.

On Apple M1 (Apple7 architecture), `exp2()` on negative arguments (`exp` in `1..6`, representing powers `2^-6` through `2^-1`) evaluates 1 to 2 ULPs below the exact IEEE-754 value. For example, `exp2(-6.0f)` returns `0.0156249981` (`0x3c7ffffe`) instead of `0.015625` (`0x3c800000`).

While Apple M4 (Apple9 architecture) evaluates these specific powers accurately enough to match the reference, Apple M1 causes 96 codes (`6 exponent values x 2 signs x 8 mantissas`) to fail the strict bit-level comparison (`yg[b] != ref`) in `c/tests/test_backend_metal.mm:run_fp8_lut()`.

## Solution

Construct the power of two directly in IEEE-754 binary32 format instead of calling `exp2()`:

- In single-precision float representation, the exponent field has a bias of 127 and occupies bits 23 to 30.
- For `2^(exp - 7)`, the biased exponent is `(exp - 7) + 127 = exp + 120`.
- For all valid normalized codes (`exp` in `1..15`), `exp + 120` ranges from 121 to 135, which falls safely within normal float boundaries.
- Shifting this value by 23 bits and bit-casting using `as_type<float>((uint(exp) + 120u) << 23)` forms the exact power of two with 0 fractional bits.

Multiplying `(1.0f + float(mant) * 0.125f)` by an exact power of two is exact under IEEE-754 rules. This guarantees bitwise parity with the CPU reference across all GPU architectures (M1 through M4) while executing faster than a transcendental call and preserving fast-math optimizations for the rest of the shader library.

## Changes

`c/backend_metal.mm`

```diff
@@ -137,7 +137,7 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
         wv = as_type<float>(0x7fc00000u);           // qNaN -- matches quant.h's e4m3_decode
       } else {
         float mag = (exp == 0) ? (float(mant) * 0.001953125f)                 // subnormal: mant*2^-9
-                                : (1.0f + float(mant)*0.125f) * exp2(float(int(exp) - 7));
+                                : (1.0f + float(mant)*0.125f) * as_type<float>((uint(exp) + 120u) << 23);
         wv = sign ? -mag : mag;
       }
       acc += wv * xr[i] * scl[i/128];
```

## Verification

1. `make -C c metal-test`
   - Before: `fp8 LUT exactness (256/256 codes via GPU kernel) 96/256 mismatches *** MISMATCH`
   - After: `fp8 LUT exactness (256/256 codes via GPU kernel) 0/256 mismatches ok`
   - Result: All test cases pass; exit code 0.

2. `make check`
   - Ran 670 tests in 66.468s.
   - Result: OK (skipped=65); exit code 0.

### DIFF
--- a/c/backend_metal.mm
+++ b/c/backend_metal.mm
@@ -137,7 +137,7 @@ kernel void mm_gemv(device const uchar* w      [[buffer(0)]],   // raw weight by
         wv = as_type<float>(0x7fc00000u);           // qNaN -- matches quant.h's e4m3_decode
       } else {
         float mag = (exp == 0) ? (float(mant) * 0.001953125f)                 // subnormal: mant*2^-9
-                                : (1.0f + float(mant)*0.125f) * exp2(float(int(exp) - 7));
+                                : (1.0f + float(mant)*0.125f) * as_type<float>((uint(exp) + 120u) << 23);
         wv = sign ? -mag : mag;
       }
       acc += wv * xr[i] * scl[i/128];


JustVugg (approved):
Verified at the root rather than on the reported cases: 127-7 = 120, so `(e+120)<<23` is the IEEE754 encoding of `2^(e-7)`. I checked it against `exp2f` across all 15 normal e4m3 exponents — identical on every one, not merely closer. The subnormal branch (e=0) is on the other side of the ternary and untouched.

This is exact by construction rather than more accurate: it removes the fast-math `exp2()` from the path entirely, which is why it fixes #1345 rather than narrowing it. 25/25 green.

PR #1347, opened 2026-09-05
fix(qwen36): do not free the int8 weights the tier is still pointing at

### LINKED ISSUE #1341: qwen36 CUDA tier: int8 containers free the weights the tier still points at (regression from #1334)

JustVugg:
Fixes #1341, reported by @crichalchemist. **This is my regression from #1334.**

## The chain

```c
qt_note_planned(l, eidw, wg, wu, wd, ...);   /* the tier RETAINS wg */
if (!keep8 && e->g) { free(e->g); ... }      /* and the next line frees it */
```

On int4 that free is safe: what is freed is the int8 copy, while the tier holds `e->g4`, a separate allocation, and `slot_ensure_int8()` can rebuild from it. On int8 the pointer handed over **is** `e->g` — dangling pointer for re-staging, and `slot_ensure_int8()` returns immediately when `g4` is NULL, so the CPU fallback dereferences NULL.

The comment directly under my change already said *"on eviction slot_ensure_int8() rematerializes from g4"* — true for int4, false for int8. The code was stating the precondition my new branch broke, and I read it as describing the surrounding code rather than constraining what I was adding.

## Fixed so it stays fixed

My first version said `expert_is_int4`. That works today and encodes the *current* correspondence between format and memory ownership — whoever adds a third format that aliases `e->g` has no reason to suspect that check is about ownership.

The condition now names no format:

```c
if (!keep8 && e->g && wg != (const uint8_t *)e->g)
```

**Do not free what you just handed over.** A future aliasing format is correct without anyone remembering; a future format with a separate copy still frees as before. Verified across five cases including two formats that do not exist yet.

## The honest cost

An int8 container has no half-size packed copy to keep instead of the full one, so the RAM saving int4 gets here is not available to int8. Pretending otherwise is what created the bug.

## What the test does and does not do

The tier test gains the missing contract: the tier **retains** the pointer, so the caller must not free it.

Stated plainly: **that check does not catch the caller-side violation.** Restore the unconditional free and the test stays green, because the defect lives in the warmstart, which needs a real CUDA tier to execute. This is precisely the class of bug a software backend would make testable in CI — a concrete example rather than an argument.

### DIFF
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -2811,11 +2811,27 @@ int main(int argc, char **argv) {
                 const uint8_t *wd = expert_is_int4 ? e->d4 : (const uint8_t *)e->d;
                 if (planned[gi] && wg) {
                     qt_note_planned(l, eidw, wg, wu, wd, e->gs, e->us, e->ds);
-                    /* The staging copy is done; free the int8 copy RIGHT AWAY
-                     * so it never shows up in peak RSS. On LFRU eviction
-                     * slot_ensure_int8() rematerializes from g4 (no container
-                     * access). */
-                    if (!keep8 && e->g) { free(e->g); e->g = e->u = e->d = NULL; }
+                    /* Liberare la copia int8 subito tiene basso il picco di RSS,
+                     * e su un container int4 e' sicuro: quel che si libera e'
+                     * `e->g`, mentre il tier tiene `e->g4`, un'allocazione
+                     * distinta, e slot_ensure_int8() sa rimaterializzare da li'
+                     * dopo uno sfratto LFRU.
+                     *
+                     * Su un container int8 NIENTE di tutto cio' e' vero, ed e' la
+                     * regressione di #1341: il puntatore appena consegnato al
+                     * tier E' `e->g`. Liberarlo lascia il tier con un puntatore
+                     * a memoria morta per il re-staging, e slot_ensure_int8()
+                     * -- che esce subito quando g4 e' NULL -- non ha nulla da
+                     * cui ricostruire, quindi il fallback CPU dereferenzia NULL.
+                     *
+                     * Il costo onesto e' questo: un container int8 non ha una
+                     * copia impacchettata a meta' dimensione da tenere al posto
+                     * dell'altra, quindi il risparmio di RAM che l'int4 ottiene
+                     * qui l'int8 non puo' ottenerlo. Fingere il contrario e'
+                     * esattamente cio' che ha creato il bug. */
+                    if (!keep8 && e->g && expert_is_int4) {
+                        free(e->g); e->g = e->u = e->d = NULL;
+                    }
                 }
             }
             qt_fill_wait();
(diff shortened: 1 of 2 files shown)


JustVugg:
Closing in favour of #1344, which was opened 16 hours before this and covers all three issues.

I did not check the PR list before starting — I read the issue, replied "send your fix", and then wrote my own anyway. That is duplicated effort I caused, and the apology is owed to @crichalchemist rather than to the queue.

#1344 is also better where it counts most. I wrote in #1347 that the caller-side violation could not be caught without a GPU, and #1344's `test_qwen36_tier_int8_engine.c` catches it: it includes `qwen36.c`, builds an in-memory int8 model with no container behind it, and drives `tier_warmstart()` against the shared fake backend. The thing I declared out of reach was in fact reachable, and the fake-backend header factored out of my test file is what made it reachable.

One idea from here is worth carrying over, and I have left it as a review comment on #1344 rather than a competing PR.

PR #1348, opened 2026-09-05
fix(qwen36 tier): keep the per-device block inside its buffer, and wake every waiter on shutdown

### LINKED ISSUE #1339: qwen36 CUDA tier: qt_issue overruns G.is_x with two or more GPUs (stride 8*D into a 32*D buffer)

JustVugg:
Fixes #1339 and #1340, both reported by @crichalchemist while working on #1338. Neither comes from #1334 — I checked my diff: zero lines of mine in `qt_issue`, `cv_take` or `is_x`. They are pre-existing, surfaced by the attention that PR drew.

## #1339 — the per-device block ran past its buffer

`G.is_x` was allocated `32*D` in total — one device's worth — while the addressing assumed a block each at stride `8*D`. With two or more GPUs, device 1 wrote from `8*D` to `8*D + 32*D`. Three hand-written numbers in three places (32, 8, 32), two of which disagreed.

`qt_is_offset()` and `qt_is_floats()` are now the single source: stride and capacity cannot drift apart because they are the same constant, and the buffer is sized by `ndev`.

## #1340 — shutdown woke only half the waiters

`th_stop` is watched by **two** conditions and `qt_shutdown` signalled only `cv`. Anyone parked on `cv_take` — a full queue in `enqueue_locked`, or a group in flight in the uploader — never woke, and the `pthread_join` inside `qt_shutdown` hung. Now both are broadcast: stopping must not depend on which of the two waits happened to be in progress.

## Negative controls, both bite

| restored defect | result |
|---|---|
| single `signal` on shutdown | the test **hangs** (timeout) — which is the real failure, so the hang is the diagnosis |
| `8*D` stride | two assertions fail, naming the defect |

**My first test for #1339 did not bite.** It recomputed the arithmetic from the constants instead of calling the code, so it stayed green with the wrong stride restored — it verified itself. That is why the addressing was extracted into functions: so the test calls what `qt_issue` calls. The negative control is the only reason I found out.

### DIFF
--- a/c/qwen36_tier.c
+++ b/c/qwen36_tier.c
@@ -9,6 +9,23 @@
 #include "tier.h"
 
 #define QT_MAX_DEV 8
+/* Righe di input che UN device puo' ricevere in una issue. E' anche il passo
+ * fra i blocchi per-device in G.is_x e la seconda dimensione di is_k: erano tre
+ * numeri scritti a mano (32, 8, 32) e due di loro non erano d'accordo (#1339). */
+#define QT_IS_ROWS 32
+
+/* Offset (in float) del blocco di input del device `di` dentro G.is_x, e
+ * dimensione totale del buffer. Sono una funzione e non due espressioni sparse
+ * perche' #1339 e' nato proprio da questo: il passo era scritto a mano in un
+ * punto (8*D) e la capienza in un altro (32*D), e i due numeri non erano
+ * d'accordo. Con una funzione sola il test puo' CHIAMARE cio' che chiama il
+ * codice, invece di riscrivere la formula e verificare se stesso. */
+static inline size_t qt_is_offset(int device_index, int hidden){
+    return (size_t)device_index * QT_IS_ROWS * (size_t)hidden;
+}
+static inline size_t qt_is_floats(int ndev, int hidden){
+    return (size_t)QT_IS_ROWS * (size_t)ndev * (size_t)hidden;
+}
 #define QT_QCAP 48            /* upload queue depth (staging ~1.6 MB/entry) */
 
 typedef struct {
@@ -45,7 +62,7 @@ static struct {
     uint64_t hits[QT_MAX_DEV], miss, uploads, q_full_skips;
     /* issue state of the (single) decode thread */
     int is_cnt[QT_MAX_DEV];
-    int is_k[QT_MAX_DEV][32];
+    int is_k[QT_MAX_DEV][QT_IS_ROWS];
     float *is_x;                          /* count*D input replicas per device */
     /* M3 */
     int *fill_order; int fill_cur;        /* warmstart order (heat desc) */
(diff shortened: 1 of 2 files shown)


JustVugg:
Closing in favour of #1344, which was opened 16 hours before this and covers all three issues.

I did not check the PR list before starting — I read the issue, replied "send your fix", and then wrote my own anyway. That is duplicated effort I caused, and the apology is owed to @crichalchemist rather than to the queue.

#1344 is also better where it counts most. I wrote in #1347 that the caller-side violation could not be caught without a GPU, and #1344's `test_qwen36_tier_int8_engine.c` catches it: it includes `qwen36.c`, builds an in-memory int8 model with no container behind it, and drives `tier_warmstart()` against the shared fake backend. The thing I declared out of reach was in fact reachable, and the fake-backend header factored out of my test file is what made it reachable.

One idea from here is worth carrying over, and I have left it as a review comment on #1344 rather than a competing PR.

Issue #1278, opened 2026-08-30
[Bug]: GLM-5.3-Flash chat ends on thinking.


acedogblast:
### Commit or version

dev branch (7892f58)

### Environment

- Ubuntu 24.04 LTS
- AMD EPYC 7282 16-Core Processor
- 128GB DDR4 RAM

### Reproduction steps

./coli chat --model /mnt/nvme_raid0/glm53_flash_i4 --no-think

Prompt: What are you?

### Expected behavior

GLM-5.3-Flash would respond back with what it thinks it is. Thinking would be skipped due to the "--no-think" argument.

### Actual behavior and logs
Thinks for a short while but then ends without generating final output tokens.

Screenshots of issue:
Coli chat:
<img width="1693" height="740" alt="Image" src="https://github.com/user-attachments/assets/30c6a8eb-39dc-4bb3-b54d-f28dd5d0dd5d" />
Coli web:
<img width="3359" height="1255" alt="Image" src="https://github.com/user-attachments/assets/4d2974ce-a9dc-4240-8647-fdeed6e6946a" />

Direct glm53:
```
./glm53 --model /mnt/nvme_raid0/glm53_flash_i4 --prompt "What are you?" --greedy 32
teacher_forcing 154822 279 3730 320
 (A) I am a boy. (B) I am a student. (C) I am a doctor. (D) I am fine.

What
decode 32 token in 41.4s = 0.772 tok/s (1.3 s/token)
experts hits 8378 miss 3460 bytes 48978984960
```

JustVugg:
Fixed on dev by #1282 — could you retry from there?

    git checkout dev && git pull
    cd c && make glm53
    ./coli chat --model /mnt/nvme_raid0/glm53_flash_i4 --no-think

## What was wrong

The prompt said two opposite things at once:

    <|system|>Reasoning Effort: Max ... <|assistant|><think></think>

Reason maximally, and you have finished reasoning.

The model answers that by opening a fresh `<think>`. The gateway's splitter had correctly started in **text** mode — the prompt closed the block, so what comes back should be pure answer — sees the new marker, re-enters reasoning, and files everything after it as thought. What reaches you is thinking, then no answer. Which is your report exactly.

On dev the prompt is now:

    [gMASK]<sop><|user|>What are you?<|assistant|><think></think>

No effort line. That is the same shape GLM-5.2 has produced for `--no-think` since it was written, and GLM-5.2 is the one of the two that never had this report.

## What I checked before landing on that

[...]

PR #1353, opened 2026-09-05
OpenAI API Instrumentation (e.g. seed, logprobs, echo, an array prompt)

monotophic:
Authored by Claude Opus in Claude Code, analysis in partnership with @Monotophic

The OpenAI-compatible server refuses several ordinary request fields outright
today, and the pipe it speaks to the engine over is checked in neither
direction. This changes both. It is six files — Python, one JSON test fixture,
and documentation. No engine source, no build change.

**`seed` is accepted and ignored rather than rejected.** No engine and no field
on the wire protocol reads a per-request seed, so the value genuinely has no
effect at any temperature; `docs/api.md` says exactly that instead of implying a
determinism this build cannot deliver.

**Per-token log probabilities and prompt echo are served on the glm engine.**
The engine on `dev` already implements the numeric channel — `c/decode_batch.h`
parses the `logprobs=k` and `ids=1` SUBMIT extensions and `c/colibri.c` already
emits the per-token tail on `DATA`/`ECHO` frames. Only the server side was
missing. `/v1/completions` now takes the legacy integer `logprobs` (0–32) and
boolean `echo`; `/v1/chat/completions` takes boolean `logprobs` with
`top_logprobs`. Every other engine refuses the fields with a named 400 rather
than ignoring them, and an engine build older than the SUBMIT extension is
answered with a named 503 after a bounded wait instead of leaving the request to
hang. The costs are documented rather than hidden: opting in forfeits
prefix-cache reuse for that request, the top-k table is unsorted on the wire,
and the sampled token is not guaranteed to appear in its own table.

**`prompt` accepts arrays.** A flat array of token ids is one pre-tokenized
prompt; an array of strings or an array of token-id arrays is a batch — the
shape unmodified `lm-eval` sends for tokenized loglikelihood requests, which
`dev` rejects. A batch is all-or-nothing: every member's shape is validated
before the first engine submit, members are submitted one at a time each with
its own UTF-8 decoder and its own echo reassembly, any member's failure fails
the whole request with that member named as `prompt[i]`, and nothing reaches the
client until every member has finished. A member cap and two token budgets bound
what one request can hold in memory. `group_score` — a future opt-in that would
change the response shape — is refused with a named 400 rather than silently
ignored, because ignoring it would answer a different question than the client
asked.

**The server↔engine protocol is fail-closed in both directions.** Every
`SUBMIT`/`CANCEL`/`STOP` write is checked, so a write onto a dead engine becomes
a named HTTP 500 instead of a silent connection close; upstream's frame-writer
structure and its `IMAGE`-frame ordering are untouched — the checking is
expressed inside them. In the other direction the dispatcher validates frame
grammar rather than desynchronizing, and it drains four group-scoring frame
kinds (`GRPP`/`GRPG`/`GRPS`/`GRPE`) so that a server built from this branch stays
correct if it is ever paired with an engine that emits them. **No engine in this
tree emits them** (verified by grep over the whole repository at this head: the
only hits are this server, its tests, and `docs/api.md`); the engine that would
is a group-scoring channel proposed separately, and nothing in this PR waits for
it.

**Context.** This PR is one of a set of independent contributions derived from a
single locally-verified working tree: fp8 container support (already proposed as
#1102), this server API work, four evidence-tooling and engine-instrumentation
contributions, and a group-scoring serve channel proposed later. This one
carries the OpenAI-compatible server's request surfaces and its engine-protocol
contract. It stands alone — measured, not assumed: its branch sits directly on
`dev` at `cc756a63` (its merge base with `dev` is `dev` itself), its whole suite
is green there with nothing else from the set present (`make check`, 1198 tests, skipped=102, this Mac, 2026-09-17),
and merging or declining the others does not affect it; separately, the three
programs finished so far were merged together locally and the combined suite is
green (1244 tests), so the set is also known to compose. The others are proposed
separately, each with its own evidence.

## Behavioral contract

**Everything that worked before still works.**
- A request that uses none of the new fields produces a byte-identical `SUBMIT`
  header and a byte-identical response body to `dev`'s.
- Every frame the server writes to the engine — `SUBMIT`, `SUBMIT` preceded by
  `IMAGE`, `CANCEL`, `STOP` — is byte-identical to `dev`'s, in the same order,
  under the same single lock acquisition.
- Every test method that exists on `dev` still exists here, in both server test
  modules; none was removed to make room for a new behaviour.

**`seed`**
- A request carrying `seed` on `/v1/completions` or `/v1/chat/completions` is
  processed exactly as the same request without it.
- The value reaches no wire frame and no engine, at any temperature.

**Log probabilities and prompt echo**
- `/v1/completions` accepts integer `logprobs` 0–32 and boolean `echo`; `0`,
  `false` and `null` all mean "no log probabilities" and return
  `choices[].logprobs: null` with a 200, while boolean `true` is a named 400.
- `/v1/chat/completions` accepts boolean `logprobs` plus integer `top_logprobs`
  (0–32) and returns `choices[].logprobs.content[]`; any integer `logprobs` is a
  named 400, and `echo` is refused.
- `top_logprobs` is type- and range-checked on chat even when `logprobs` is
  false or absent; a non-boolean `echo` is a named 400 on both endpoints.
- On any engine that does not implement the channel, a logprobs request is a
  named 400 — never a silent no-op.
- `logprobs` together with `stream` is a named 400.
- An engine build that never acknowledges an opted-in `SUBMIT` produces a named
  503 within `COLI_LOGPROBS_ACCEPT_TIMEOUT` (default 30 s), never a hang; the
  request is treated as cancelled rather than left pending.
- Non-finite values serialize as JSON `null`, never a clamped number. Echo
  positions are placed by the wire's own `pos` field, not arrival order.
  `text_offset` is a character offset into the returned `text`, counted from 0.
  The chosen token's log probability is read from its own record, not looked up
  by rank in the top-k table.

**Array and token-id prompts on `/v1/completions`**
- A single-member array — one string, or one token-id list, or a nested
  batch-of-one — produces exactly what the same prompt produces as a flat
  single-prompt request.
- An array must be non-empty and homogeneous (all strings, all token ids, or all
  token-id arrays), at any size; each violation is a named 400 on `prompt`.
- Token-id prompts, flat or nested, are glm-engine-only; every other engine
  returns a named 400 rather than mis-tokenizing decimal digits as text.
- A real batch returns one choice per member, `choices[i].index == i`, in order,
  with `usage` summed across members.
- Every member's shape is validated before the first engine submit, so a
  malformed member N is a clean 400 naming `prompt[N-1]` with zero submits made.
- A member the engine itself rejects after earlier members were submitted is
  still attributed to that member; a client-fault error keeps the engine's own
  status, code and `param` with only the message gaining the index; an engine
  failure the server cannot name is one un-attributed 500 for the whole batch.
- Nothing is written to the client until every member finishes; a client that
  disconnects mid-batch stops it at the member in flight.
- One member's trailing partial UTF-8 character, or its logprobs table, can
  never leak into another member's text.
- A batch is capped at 128 members; at 65,536 total prompt tokens (token-id
  members counted in tokens, string members in UTF-8 bytes as an upper bound);
  and, on the generated side, at members × the request's effective `max_tokens`
  not exceeding 65,536, checked before any engine submit. A length-1 array is
  exempt from both budgets and keeps the flat path's own oversize handling. The
  scheduler admission is released before the response is written.
- `stream: true` and `n` other than 1 are refused with a named 400 on a batch.
- `group_score` present and neither `null` nor `false` is a named 400
  (`code: "unsupported_value"`) on `/v1/completions`, flat or array, before any
  prompt intake or engine submit. `/v1/chat/completions` and `/v1/messages` do
  not read the field.

**Engine protocol**
- Every server→engine frame write is checked; a failed write is a named HTTP 500
  `engine_error` while the response is uncommitted, and ends the stream once it
  is committed — never a silent close.
- `GRPP`/`GRPG` payloads are drained under the same size bound and terminator
  check as `DATA`, and `GRPS`/`GRPE` are consumed with no effect; nothing from
  them reaches any request's event queue. A malformed group frame raises the
  dispatcher's existing named error.
- Every other frame kind the dispatcher handled on `dev` is handled identically.
- The engine child runs under the default POSIX SIGPIPE disposition, so a
  disconnected reader terminates it at its next write with no evidence records
  after the failure point.

## Capstone matrix

One decisive artifact per claim; each would fail if the claim were false. Test
citations are file:line at this branch's head.

| claim | decisive evidence |
|---|---|
| `seed` reaches no wire frame | `c/tests/test_openai_server.py:2220` `test_seed_accepted_and_absent_from_submit_frame` — the `SUBMIT` frame bytes for a seeded request are compared against the unseeded request's; any leak of the value changes them |
| requests that use none of the new fields are unchanged | two byte-level pins: `c/tests/test_openai_server.py:4382` asserts the exact `SUBMIT` header literal, and `:4495` asserts the plain completions response's field set and values |
| no frame byte the server writes changed | `c/tests/test_openai_server.py:6382` `test_wire_transcript_is_byte_identical_to_base` — the whole `SUBMIT` / `IMAGE`+`SUBMIT` / `CANCEL` / `STOP` sequence is compared to a literal captured transcript, so a reorder, a missing flush boundary or one stray byte fails it |
| logprobs are refused, never ignored, off the glm engine | `c/tests/test_openai_server.py:4328` `test_break_it_logprobs_rejected_for_non_glm_engine` |
| an old engine cannot hang an opted-in request | `c/tests/test_openai_server.py:4780` `test_old_engine_rejection_times_out_with_a_named_503` — it runs `generate()` on its own thread precisely so that a missing deadline fails this one test instead of hanging the suite |
| the documented environment variable is the one actually read | `c/tests/test_openai_server.py:4908` and `:4911` import the module in a fresh subprocess with the real variable name set and unset, so a typo in the name or a changed default cannot ship green |
| a batch of one is the flat path | `c/tests/test_openai_server.py:4137` `test_nested_batch_of_one_prompt_is_identical_to_flat` |
| a rejected batch does no engine work | `c/tests/test_openai_server.py:5301` `test_rejected_batch_makes_no_engine_submits` — the engine's call count is unchanged across three differently-malformed requests |
| a member the engine rejects mid-batch is still named | `c/tests/test_openai_server.py:5333` `test_context_exceeded_mid_batch_names_member_in_param` — `param: "prompt[1]"` with the engine's own `code: "context_length_exceeded"` preserved |
| each batch choice equals its single-prompt twin (server logic) | `c/tests/test_openai_server.py:5579` and `:5563` compare every batched choice, field by field, against the same prompt sent alone — strings and token ids |
| each batch choice equals its single-prompt twin (**real engine, one shared KV slot**) | a real glm engine and a real container on a Linux CPU host, one KV slot, `--kv-slots 1`: the same three prompts were sent as singles, as a batch, as a reversed batch, as singles again, and as a duplicate-member batch. Every `(text, finish_reason)` was identical across all five arms, and the batch's aggregate `completion_tokens` equalled the sum of the singles'. The prefix cache was live and demonstrably reached — the first request logged `prefix 0/68 token, prefill 68` and later ones logged `prefix 68/68 … prefill 0` and `prefix 40/68 … prefill 28` — so identity holds *with* another member's residue in the slot, not because the slot was never warm. All 20 requests landed on slot 0, and each batch was one engine admission. Corroborating the "no engine change" claim: the engine built from this branch's own tree is sha256-identical to the engine built from `dev`. |
| `group_score` fails closed | `c/tests/test_openai_server.py:5234` `test_group_score_opt_in_is_refused_fail_closed`, on both the flat and the array request shape |
| the group frame kinds cannot desync the pipe | `c/tests/test_openai_server.py:2018` `test_group_frames_between_data_frames_drain_leaving_second_data_intact` — each kind is interleaved between a request's own two `DATA` frames and the second must still arrive byte-for-byte |
| a token-id array prompt is budgeted by its real token count | `c/tests/test_openai_server.py:4243` `test_array_prompt_token_budget_counts_actual_tokens_for_token_id_batches` — an over-budget token-id batch gets the documented refusal and the engine sees no submit |
| a batch that names no `cache_slot` runs every member on the one slot the scheduler picked | `c/tests/test_openai_server.py:5848` `test_batch_without_cache_slot_shares_the_one_slot_the_scheduler_picked` — every recorded `generate` call carries the same slot |
| a failed `CANCEL`/`STOP` write reaching an already-committed stream ends the stream, never splices a 500 | `c/tests/test_openai_server.py:3626` `KeepAliveFramingTest.test_write_failure_reaching_the_committed_stream_ends_it_cleanly` — a real `Engine` + fake subprocess ACCEPTs, streams partial text, then fails the `STOP` write; exactly one HTTP status line reaches the client, the pre-failure text survives, no `[DONE]` and no error body are spliced in, the connection closes, and the failure is logged |
| a dead engine is a named 500, not silence | `c/tests/test_openai_server.py:6046` `test_dead_engine_submit_is_a_named_500_engine_error_not_silence` — at the pre-fix revision the `BrokenPipeError` fell into the client-hangup handler and the client saw a silent close |
| the new server tests add no cleanup-timeout waits, and no longer pay socketserver's default poll interval on every shutdown | per-test timing of `tests.test_openai_server` on this head vs `dev`: zero tests in the 2.5 s and 5 s buckets (the fixture commit's parent, before the cleanup-order and fake-pipe fixes, has 16 and 2, `a2c42bc3~1`, re-measured directly for this revision — this figure is noisy near the 2.5 s boundary on repeated runs of the identical commit, 13-14-16 across three separate measurements, while the 5 s bucket is stable at 2 every time); module wall time 9.1s vs `dev` 22.1s (both measured directly, `timetests_all.py`, this Mac). The two tests that pair an exited fake process with a real `Engine` close that fake's pipe only in `addCleanup`, after their own assertions run, not from inside the fake at exit time — an earlier version closed it eagerly and that raced the dispatcher's own EOF handling against the assertion (11/30 isolated runs failed); 30/30 and the 11 EngineWriteCheckingTest tests × 10 runs are now clean |
| the SIGPIPE policy is real, not merely asserted | `c/tests/test_openai_server.py:6433` — a real fork/pipe child under `SIG_DFL` is killed by signal 13 at its next write with no completion record after the failure point, and `:6451` pins the precondition by asserting the server passes no `restore_signals` kwarg at all |

<details>
<summary>Fuller matrix, disclosures, known gaps, origin accounting, and namespace claims</summary>

### Requirement → instrument → result

|---|---|---|
| whole suite green at this head | `make check` (clean + portable C build + C suite + Python suite) | `Ran 1198 tests … OK (skipped=102)`; the Python suite inside it took 47.5 s and the whole `make check` about 3 m 15 s — measured at `a2c42bc3` on this Mac (macOS, Apple silicon), 2026-09-17; STATUS=0 captured explicitly |
| the server module is not flaky | both server test modules run three more times | 374 tests OK each time (9.386s / 9.399s / 9.392s), zero flake |
| every commit stands on its own | per-commit test run and per-commit changed-line lint for all eleven commits | green each; 0 lint findings each |
| the new suite is not self-referential | the boundary tests read constants back from the module, but the constants themselves are pinned to literals elsewhere — `LOGPROBS_TOP_K_CAP == 32` at `:578`, `PROMPT_BATCH_CAP == 128` at `:4167`, both 65,536 budgets at `:4200` and `:5667` | a constant mutation cannot silently widen the boundary tests along with themselves |
| the engine's numeric grammar is the shipped one | the parser and the test fixtures accept the `%.6f` form the engine on `dev` prints, with non-dyadic values so a different grammar is distinguishable | accepted; `nan`/`inf`/`-inf` parse and serialize as JSON `null` |
| assembly helpers place echo positions correctly | multibyte and split-codepoint offset fixtures with literal expected values (`:817`, `:840`) | exact character offsets, monotonic, trailing incomplete sequence flushed |
| batch member isolation | a codepoint deliberately split across two members (`:5637`), and hostile echo positions mid-batch (`:5649`) | per-member decoders hold; the hostile case fails clean, bounded, never hangs |
| batch partial output is impossible | `:5310` mid-batch failure, `:5423` client disconnect mid-batch | no partial response; the disconnect stops further submits |
| the SIGPIPE claim's wording | `docs/api.md` says "signal 13 (wait status 141)"; the test asserts the child's return code is `-signal.SIGPIPE` | equivalent (a shell reports a signal-13 death as 141); the literal 141 is not itself asserted anywhere |
| suite runs on a clean upstream checkout | every fixture in the new tests is self-contained; the POSIX-only fork/pipe class is skipped by name with a reason on Windows | no test depends on anything outside the repository |
| the whole set composes | the three finished programs merged together locally and the combined suite run | `make check` exit 0, 1244 tests OK (45 skipped) — verified together locally, on top of `dev` at `cc756a63` |

### Disclosures

- **The `GRPP`/`GRPG`/`GRPS`/`GRPE` drain arms have no emitter in this tree.**
  A repository-wide grep at this head finds them only in `c/openai_server.py`,
  its test module, and `docs/api.md` — no engine source, no header. They are
  drained, never surfaced: nothing from them reaches any request. The engine
  that emits them is a group-scoring serve channel proposed separately; if you
  would rather not carry protocol tolerance ahead of its producer, the commit
  that adds them (`feat(api): drain group-scoring frames a future engine may
  emit`) is self-contained and can be dropped from this PR without touching
  anything else.
- **Environment variable and timeout documentation.** This PR introduces one
  environment variable, `COLI_LOGPROBS_ACCEPT_TIMEOUT` (default 30 seconds),
  which bounds how long an opted-in request waits for the engine's
  acknowledgement before being cancelled and answered with a named 503. It is
  documented in `docs/api.md` and pinned by its real name and its default in the
  suite rather than only through the module attribute the rest of the tests
  patch. The other public knobs this PR documents — the 32-entry top-k cap, the
  128-member batch cap, and the two 65,536 budgets — are module constants, not
  environment variables, and each is pinned to its literal value by a test.
- **Suite cost, before you measure it yourselves.** `test_openai_server.py` goes
  from 173 test methods on `dev` (`cc756a63`) to 346 here, but wall time goes
  *down*, not up: 22.1 s on `dev` to 9.1 s here (both measured directly with
  `timetests_all.py` on this Mac). The extra methods would have cost roughly 74
  additional seconds on `dev`'s own per-test overhead; almost all of that was a
  fixture defect (six cleanup helpers waiting on a thread-join before telling
  the thread to stop, a fake process not closing its pipe on exit, and every
  fixture polling `serve_forever` at socketserver's default 0.5 s instead of
  10 ms), not a cost of the new behaviour itself. The C suite and the build are
  untouched.

### Known gaps, stated plainly

None remain open on this branch. Three of the four items an earlier draft listed here —
token-id batch budgeting, `cache_slot` on a batch, and a write failure reaching an
already-committed stream — are each pinned by a test named in the capstone matrix above.
The fourth, Anthropic streaming against an engine that never accepts, is **not** pinned
here: at the maintainer's request this PR was split, and the commit that defers the
`/v1/messages` streaming 200 until the engine ACCEPTs, together with its pinning test
(`AnthropicColdPrefillTest`), live only on the parked branch
`codex/anthropic-deferred-200` for a follow-up PR. This branch's Anthropic streaming
behaviour is unchanged from `dev`: the 200 and `message_start` commit at scheduler
admission, exactly as before.

Separately, disclosed rather than fixed here: a dead engine can surface as either
"colibri engine dispatcher stopped" or "colibri engine is not running" depending on
thread timing, because `Engine.generate()` checks `dispatcher_error` before
`process.poll()`; that check order is inherited from `dev` (`62419af1`) and is not
changed by this PR.

### Origin accounting

Rebased onto current `dev` (`cc756a63`) for this revision; the rebase itself changed nothing in this branch's hunks (diff-of-diffs recorded). By commit:

| commit (chain order) | origin class |
|---|---|
| `feat(api): accept and discard per-request seed (documented no-op)` | re-expressed from our own earlier work on this surface. Adjacent consequence, disclosed in the commit: the golden-fixture capture battery's `err_seed` case is renamed `seed_accepted`, because the case now captures a 200. |
| `feat(api): serve OpenAI-style logprobs and prompt echo on the glm engine` | re-expressed, with three tightenings found by our own re-review before proposing: `top_logprobs` is validated on chat even when `logprobs` is off (it was silently ignored), `echo`'s boolean check is the same on both endpoints, and the top-k cap is pinned by its literal value rather than by the symbol it polices. Adjacent consequence: the capture battery's `err_logprobs` case becomes `logprobs_served`. |
| `feat(api): accept array and token-id prompts on /v1/completions` | re-expressed. Two documentation claims this branch had written about its own behaviour were **wrong and were corrected here before proposing**: the enumerated list of `code` values for `param: "prompt[i]"` was presented as closed when the engine's own mid-batch `context_length_exceeded` is preserved through the member rewrite (the branch's own passing test asserts it), and the multi-prompt section claimed `stream: true` was refused "the same as everywhere else on this endpoint" when the flat path refuses only the narrower `logprobs`-plus-`stream` combination and names a different `param`. Both now say what the code does, including the `param` divergence a client keying off `error.param` would hit. Adjacent consequence: the capture battery's `err_array_prompt` case becomes `array_prompt_token_ids`, because this PR moves that case from a 400 to a 200 on the glm engine — the third of the three renames. |
| `fix(api): fail closed on an unsupported group_score opt-in` | re-expressed. |
| `feat(api): drain group-scoring frames a future engine may emit` | new content, serving a channel proposed separately; disclosed above, and separable from the rest of this PR. |
| `fix(api): check every server->engine frame write` | re-expressed **inside upstream's existing frame-writer structure** rather than replacing it: the `SUBMIT` block keeps its shape and its `IMAGE`-frame ordering under one lock acquisition and gains only a not-running check, a flush, and an `OSError`→named-`RuntimeError` conversion. The five `CANCEL`/`STOP` sites, which were five identical inline blocks, call one small helper; that helper is deliberately not used by `SUBMIT`. |
| `fix(api): drop Engine.generate()'s pending-map entry on every exit` | **a pre-existing defect on `dev`, found by re-review of this branch, not one this branch introduces.** `dev`'s `generate()` has the same pending map with no `finally`: the entry is popped only on a `SUBMIT`-write failure and by the dispatcher's own `DONE`/`ERROR` handling, so a write failure on the `CANCEL`/`STOP` path leaves it behind. Bounded on `dev` — that error means the engine's stdin is broken, so the dispatcher hits EOF and clears the map — but it should not depend on that. Note for review: this commit's diff looks large and is almost entirely re-indentation; ignoring whitespace it is 11 changed lines in `c/openai_server.py` plus one test. |
| `test(api): cover the token-id prompt budget and the batch's shared slot` | carried unchanged from the previous head; unrelated to this revision's split or its fixture work. |
| `test(api): pin the committed-stream write-failure path` | **retitled with the moved test removed.** Previously titled "test(api): pin the committed-stream write-failure path and the deferred Anthropic stream" and included `AnthropicColdPrefillTest`, which pinned the deferred-200 behaviour; at the maintainer's request that test and the commit it pins (the deferred-200 change itself) were moved to the parked branch `codex/anthropic-deferred-200`, and this commit was retitled to describe only what it still contains — the committed-stream write-failure path pin. |
| `test(api): compare the lm-eval fixture digest on LF-normalised bytes` | carried unchanged from the previous head; unrelated to this revision's split or its fixture work. |
| `test(api): stop the server tests waiting on cleanup and poll timeouts` | **new in this revision**, prompted by the review's note on suite cost: six per-test server helpers registered their cleanups in the order scheduler close, shutdown, server close, thread join; `unittest` runs cleanups last-in-first-out, so the join ran first against a still-serving thread and waited its full 2 s, then shutdown paid its 0.5 s poll (11 tests). Separately, two `EngineWriteCheckingTest` cases pair a fake process that never truly exits with a real `Engine`, then simulate the engine having already died to drive the require_running refusal path; `Engine.close()` joined its reader for the full 5 s in both, because the fake's stdout pipe was never closed. Closing that pipe from inside the fake at the moment it reports as exited was tried and reverted — it let the dispatcher observe EOF and set `dispatcher_error` while the test body was still running, and `Engine.generate()` checks `dispatcher_error` before `process.poll()` (a pre-existing `dev` check order, not something this branch introduced), so the two tests intermittently saw "colibri engine dispatcher stopped" instead of the asserted "colibri engine is not running" (11/30 isolated runs failed). The pipe now stays open for the whole test body and each of the two tests closes it in `addCleanup`, registered right after `addCleanup(engine.close)` so it runs first (LIFO) and lets the dispatcher exit before the join ever has to wait — but only once the test's own assertions have already completed. A third, separate cost: every fixture in both server test modules started `serve_forever` at socketserver's default 0.5 s poll interval, so every `shutdown()` call could block for up to half a second waiting for the select loop to notice it (30 tests sat on that 0.50 s plateau (measured directly at `a2c42bc3~1` for this revision)); all 20 call sites (19 in `test_openai_server.py`, 1 in `test_anthropic_messages.py`) now pass `poll_interval=0.01`. No assertion changed; `c/openai_server.py` untouched. |

Style note: changed lines were held to the file's measured local idiom via a
diff-scoped consistency check; no surrounding code was reformatted.

### Shared-namespace claims

This PR claims the following names in namespaces other contributions could
collide with. The list is scanned against `dev` HEAD and against every open pull
request's diff immediately before push, and re-scanned if the push slips.

- **Environment variable:** `COLI_LOGPROBS_ACCEPT_TIMEOUT` (one; new).
- **API error `code` strings:** `prompt_batch_cap_exceeded`,
  `prompt_batch_token_budget_exceeded`, `batch_completion_budget_exceeded`,
  `engine_tok_ids_unsupported` (four; new). Existing codes reused unchanged:
  `invalid_value`, `unsupported_value`, `unsupported_parameter`,
  `context_length_exceeded`, `engine_error`.
- **Wire protocol: nothing new is claimed.** The `SUBMIT` extension keys this
  server now sends, `logprobs=k` and `ids=1`, are already defined and parsed by
  the engine on `dev` (`c/decode_batch.h`); this PR adds a sender, not a key.
  The four `GRP*` frame kinds are consumed only — nothing here emits them.
- **No new command-line flags, no file magic, no format ordinals.**

</details>

**Durable vs current state:** the request contracts, the error codes, the
protocol rules and the documented limits are durable. The counts (1198 tests at
this head, 346 in the server module, 1244 for the three programs merged
together) are current state, measured 2026-09-17 at `a2c42bc3` on base
`cc756a63`; the fleet run quoted above was measured 2026-09-05 on macOS (Apple silicon) and, for the KV-slot cell, on a
Linux CPU host with GCC.


monotophic:

Rebased onto current `dev` (`87c54a93`). No functional change on this branch's side — the 11-commit chain is preserved unsquashed, same messages, same content.

One file needed conflict resolution, `c/tests/test_openai_server.py`; all three hunks are the same shape (dev and this branch each added things at the same two places, no overlap):

- The stdlib import block and the `from openai_server import (...)` list: dev's `COLI_IMAGE_ROOT` work added `import os` and `_image_bytes_from_url`; this branch adds `import queue`, `LOGPROBS_TOP_K_CAP` and `logprobs_options`. Resolved as the union.
- The end of the file: dev added `ImageUrlPathGuard` and `ContextExceededMessageTest` (#1376); this branch appends its logprobs/echo test classes at the same point. Resolved as the union, dev's classes first.

`c/openai_server.py` merged with no conflict; its changed lines are identical against the new base (only hunk positions moved). Read-through of the merged file for each surface this branch touches: dev's deepseek-v41 image-expansion arm in `chat_completion` still falls through to `generation()`, so the logprobs/echo capability gate applies to it like every other family; dev's GLM-5.3 think-block change (#1278) lives in `starts_in_reasoning`, orthogonal to this branch's deferred Anthropic 200.

Re-verified on `2edcea62`: `make check` exit 0 (1155 tests); `tests.test_openai_server` and `tests.test_anthropic_messages` (378 tests) green; lint on changed lines: no findings.

@JustVugg: Checking in on the viability of this PR. Is there interest in merging this support for deeper instrumentation through the OpenAI API interface for Colibri? See my comments on PR #1102 for more context.

I've got everything I need integrated and working at my end so my container experiments are not blocked by this stuff sitting unmerged, but I would prefer not to be investing in an isolated fork and rebasing regularly to avoid drifting too far from what you are doing here. Steering input is welcome, my hope is that these contributions would be generally useful for anyone wanting to use Colibri for serious work, including model tuning/development and research.

JustVugg:
Reviewed. This is the strongest of your four and the only one that is genuinely independent: a different branch, no file overlap with the evidence stack, no engine source, and the engine you build from it is identical to the one built from dev because the diff is Python only.

It also unblocks something concrete. The engine on dev already implements the numeric channel, `logprob_tail()` in `c/colibri.c` and the `logprobs=k`/`ids=1` parsing in `c/decode_batch.h`; only the sender was missing. And the array-prompt support is what lets lm-eval send tokenised loglikelihood requests at our own models, which dev rejects outright today. The `_write_all` short-write loop on the raw stdin pipe is a real latent fix on its own: an unchecked short write desyncs engine framing.

One question before I merge, and it is the only default-path behaviour change in the diff. The `/v1/messages` streaming commit point moves from scheduler admission to the engine's ACCEPT. On an in-tree engine that is one round trip and the event order is unchanged, but against an engine old enough never to send ACCEPT, a cold multi-minute prefill now sends zero bytes where dev sent message_start plus pings, and a client with a short idle timeout would newly time out. You name this yourself and say it has no test for that pairing.

Would you split that single commit out? The rest lands immediately and the commit point can come back with its own test.

One thing to flag rather than block on: the server suite goes from 161 to 336 methods, about 96 seconds of a 140-second Python suite. That is a permanent tax on everyone's `make check`. Worth it here, because an HTTP contract with error codes and batch semantics is exactly the shape table-driven tests suit, but worth knowing.

monotophic:

@JustVugg Thank you! Revised as requested. Pushed `a2c42bc3`, rebased onto current `dev` (`cc756a63`, no conflicts).

 - The commit that moved the `/v1/messages` streaming commit point to the engine's `ACCEPT` is out, along with the one test that pinned it. The Anthropic streaming handler is byte-identical to `dev`'s again. It will come back as its own PR with the test you named: an engine that never sends `ACCEPT` behind a cold prefill, and what the client sees meanwhile. The other ten commits are unchanged (range-diff: all `=`).
 - On the suite cost — you were right to flag it, and it was worse than a tax: three fixture defects, not real work. Six of the new per-test server helpers registered their cleanups in the order scheduler close, shutdown, server close, thread join; `unittest` runs cleanups last-in-first-out, so the join ran first against a still-serving thread and waited its full 2 s, then shutdown paid the 0.5 s poll. Two engine-write tests left a fake pipe open so `Engine.close()` joined its reader for the full 5 s. And every fixture started `serve_forever` at the 0.5 s default poll interval, so each shutdown blocked half a second. One added commit fixes all three; no assertion changes, `openai_server.py` untouched. `tests.test_openai_server` on this machine: `dev` 22 s → previous head 96 s → now ≈9 s, with 346 methods.
 - PR body updated to match: the commit-point material removed, capstone anchors re-resolved at this head, and the cost paragraph corrected.

JustVugg (changes requested):
Thanks for this. I want to say up front that the shape is right: this is U7b landing against a U7a wire side that already works, the capability gate is a single flag set once from the arch, the `logprobs`/`top_logprobs` null handling is careful, and 174 tests against 1122 lines of server code is the ratio I wish every PR had. I read `c/colibri.c` and `c/decode_batch.h` alongside the diff and the server's assumptions about the SUBMIT extension arm and the ECHO field layout check out, including the 7th `gbytes` field needing to precede the first `key=value` token, which `gbytes_field` correctly emits as `" 0"`.

It is not mergeable yet. Below are the things I need fixed, worst first. I reproduced the ones marked "reproduced".

## 1. The default glm chat path returns empty logprobs, silently

`_trim_generated_records_to_text` drops every generated record when the first record's text is not a prefix of the returned text. That is the default, not an edge case: with no client `stop`, `stop_policy` returns `ignore_leading=True`, `StopFilter` swallows a leading role marker, and the prefix check fails on record 0, so the loop breaks having kept none.

Reproduced: records `[b'<|user|>', b'Hello', b' world']`, `raw_text == 'Hello world'`, result is 0 of 3 kept. The client gets a 200 with a full `message.content` and `logprobs: {"content": [], "refusal": null}`. No error, no signal.

The docstring lists the thinking-split and tool-call cases as unhandled. The ignore-leading case is a third one and it is the path most requests take.

## 2. The ACCEPT timeout violates the invariant documented twelve lines below it

The new 503 is raised from the `queue.Empty` branch of `Engine.generate()`'s event loop. That abandons an in-flight engine turn and releases the scheduler admission, and no CANCEL is ever written. The comment immediately below says exactly why not to:

> Do NOT raise here: this thread holds the scheduler admission, and releasing it before the engine confirms the cancel lets the next request SUBMIT into a pipe the busy engine is not reading, every later request then hangs silently behind the orphaned generation.

Concretely, with `kv_slots > 1`: request A is in a multi-minute cold prefill so the mux loop is not reading stdin. Request B opts into logprobs, its SUBMIT is written but unread, no ACCEPT inside 30 s, B raises 503 and frees its admission. The engine later generates B to `max_tokens` with nobody reading, and request C inherits B's `cache_slot` and gets SLOT_BUSY, then a 500.

`docs/api.md` says the request is "treated as cancelled". It is not. Either send the CANCEL and wait for confirmation before returning, or handle the missing ACCEPT somewhere that is allowed to fail.

## 3. `echo: true` deletes text the client already earned

With a stop sequence matching mid-token, `text` is rebuilt from the trimmed logprobs tokens, so the emitted prefix disappears from the response body.

Reproduced: `/v1/completions` with `echo: true, logprobs: 1, stop: ["lo"]`, one DATA frame decoding to `"Hello"`. StopFilter emits `"Hel"` and matches. The trim checks `"Hel".startswith("Hello")`, keeps 0 of 2 records, and `text = "".join(logprobs_obj["tokens"])` overwrites the body with the prompt echo alone. Without `echo`, the same trim leaves `tokens`/`token_logprobs`/`text_offset` empty while `text` is `"Hel"`, so the arrays no longer describe the returned text either.

## 4. Root cause behind 1 and 3

Realigning records to the returned text by prefix-matching decoded bytes is a symptom patch, and each further text transformation (thinking split, tool-call parse, inkling split) will need another special case on the same walk. The general fix is one level down: have `StopFilter` and the splits report how many characters of the raw stream they consumed, or have `Engine.generate()` tag each record with its raw-stream character span, so alignment becomes a lookup instead of a guess that silently yields an empty array when it misses.

I would rather see that than three more special cases, but I will take correct special cases if you prefer to keep the change contained.

## 5. `echo: null` is a new 400 for clients that never asked for echo

Verified against the real function: `logprobs_options({'echo': None}, chat=True, ...)` and the `chat=False` form both raise `APIError(400, "`echo` must be a boolean.")`. Any SDK that serializes its full request model with nulls now gets a 400 on every chat completion and every plain completion, with no logprobs involvement at all. On `/v1/chat/completions` `echo` was previously ignored entirely, so this is a new rejection of a field chat clients never meant to set.

The rule is already in your own docstring: "`top_logprobs: null` is normalized to absent, same as `logprobs: null`". Apply it to `echo`.

## 6. `LOGPROBS_ACCEPT_TIMEOUT` can kill the import

It is parsed at module level with an unguarded `float()` and no range check. `COLI_LOGPROBS_ACCEPT_TIMEOUT=30s` makes `import openai_server` die with a ValueError traceback before anything can print a diagnostic, and it takes down every importer of the module. `0` or a negative value puts `accept_deadline` at or before `time.monotonic()`, so every logprobs request 503s on the first idle poll against a healthy engine. Compare `_SALVAGE`, the only other module-level env read, which cannot raise.

## 7. One test is racy and fails deterministically off CI

`test_client_disconnect_mid_batch_stops_further_submits` closes the socket and releases the blocked engine in the same instant, so the server runs all 8 members before the kernel delivers the RST: `client_disconnected()` is False on all 8 checks, `len(engine.calls) == 8`, and `disconnect_observed.wait(5)` times out. Reproduced 4 of 4 runs; a 0.5 s sleep between `sock.close()` and `resume.set()` makes it pass with `calls == 3`.

CI is green on `a2c42bc3`, all 28 checks, so this is timing that happens to go the right way on those runners. Please make the test wait for the disconnect to be observable rather than depend on scheduling.

## 8. Smaller things, all in `c/openai_server.py`

- In the ECHO dispatcher arm, `pos = int(fields[3])` raises a bare `ValueError` instead of the named `RuntimeError` the neighbouring size/terminator/tail checks use. It lands in `_dispatch_stdout`'s blanket handler and fails every concurrent request with `invalid literal for int()`. It also makes `_order_echo_records`' own `isinstance(pos, int)` guard unreachable, so that validation is at the wrong layer.
- `batch_completion.submit_one` is a copy of `generation()`'s non-streaming single-prompt block rather than a call into a shared helper, and the copies have already diverged: `generation()` applies `split_thinking_reply` and the tool-call parse and sends inside the admission, `submit_one` does neither. Any fix to the stop/trim/echo interaction above has to be made twice or it only fixes one endpoint shape.
- One flag, `supports_logprobs_echo`, gates two unrelated capabilities: the U7a numeric channel and pre-tokenized token-id intake. They are separate keys with separate fields in `coli_submit_ext`. Two booleans set from the same `arch == "glm"` today costs one line and removes the conflation. The token-id refusal message also hardcodes "glm" instead of reading the flag it just checked.
- The GRPP/GRPG/GRPS/GRPE arms and the `group_score` 400 add wire handling for a protocol `docs/api.md` states no engine in this tree emits. The arms guess at field counts, and the `group_score` guard is scoped to `/v1/completions` only, so the same opt-in is silently accepted on chat and messages. Please drop all five until the channel exists: the dispatcher's existing `else: raise RuntimeError("invalid engine response")` is already the right fail-closed behavior.
- `_write_all` re-slices `data[written:]` every iteration, which is quadratic on exactly the large IMAGE frames it was written for, while holding `write_lock`. `view = memoryview(data)` once fixes it in the same three lines.
- `raw = [(data, record) for data, record in _order_echo_records(prompt_records)]` rebuilds every tuple for nothing. And the `if isinstance(member, str)` filter inside the sum sits in a branch that already proved every member is a string, so it is an unconditional under-count waiting to happen rather than a guard.
- For chat with thinking enabled or tools, `logprobs.content` describes tokens `message.content` does not contain, and nothing in the response marks it. It is under "Known limitations" in the docs, but a client aligning the two has no way to detect it at runtime.

## 9. Two process items

The branch needs a rebase on `dev`. The conflict is small: one region in `c/tests/test_openai_server.py`, where your `test_seed_no_longer_rejected_by_generation_options` and the `test_tool_choice_function_that_is_not_an_object_is_a_400` that landed in #1598 were added at the same insertion point. Keeping both is the whole resolution. `c/openai_server.py` merges clean.

All 11 commits carry a `Co-Authored-By` trailer. This project does not take those, so the history needs rewriting before merge.

Happy to look again as soon as the substantive items are in. The parts that matter most to me are 1, 2 and 3.

monotophic:

@JustVugg thanks for the review — all nine items are being taken, with 1, 2 and 3 first and item 4 the way you preferred (alignment by raw-stream span, not prefix match). 

My apologies for this being slow, I thought it would be quick to turn the package around with your items addressed, but I tried to keep it as a monolithic PR too far into the process AND the loss of our summer subscription subsidy meant I was working primarily with Opus 5 and we simply do not understand each other as well as needed to keep this package on track and up to standards. It was a risky tangle by Monday AM so I took a pause until fresh Fable tokens were available to sort it all out.

This branch grew well past the shape you praised, so rather than push that, we are re-cutting it into smaller pieces: `seed` accepted-and-ignored and the checked engine-write path each as their own small PR; this PR narrowed to `logprobs`/`echo`/`top_logprobs` with the span alignment; the array and token-id `prompt` intake, and a separate proposal for a full-stream logprobs field, each as their own PR after this one lands.

Nothing is pushed yet. Each piece will sit directly on current `dev` with no `Co-Authored-By` trailers. 

Thanks again, I am working on making this worth your review time.

monotophic:
@JustVugg, I closed/reopened this PR because the CI / Sanitizers (ASan + UBSan) job ran for 48 minutes and then GitHub annotated it: "The hosted runner lost communication with the server." This was at 28/29 passed checks. 

I know that this is extra CI load, if you would rather I not cycle a PR like this to re-run hung CI, I can post a comment instead when I get this kind of fail. I have gotten hung CI processes on a couple of my pushes recently and they have cleared without requiring repair of the PR itself.

monotophic:

Pushed. This PR is now the logprobs work alone, on current `dev` as three commits, and the body carries
a table mapping each of your items to what changed and to the test that fails without it. It is about
the size of the diff you reviewed; the branch that answered the review had grown to about twice that,
and it is not what is here.

Your `echo: true, logprobs: 1, stop: ["lo"]` case runs as a test now and fails on `dev`: the body is
`promptHel`, and the arrays say `["pr", "ompt", "Hel"]`.

Two things stated plainly:

- Item 1: your three records now keep two of three through the endpoint — `<|user|>` is the one
  dropped, because the client never received it, and the two that were received join back to
  `message.content` exactly. We have not reproduced that leading marker on a live glm engine, so the
  fix is covered by tests that drive the endpoint with your records, and we are not claiming a cure we
  never watched happen. Your reading of the default path is correct.
- Item 2: the ACCEPT timeout is gone rather than relocated. A logprobs request now waits exactly as
  any request waits, so nothing frees an admission before the engine's turn ends — the case you
  described cannot arise. The residue is the one you would expect: behind a cold prefill the request
  waits, quietly, for as long as the prefill takes; the body says so and offers a bounded wait as its
  own PR. The same rule covers the faults this change names — a malformed numeric tail or a bad ECHO
  position fails that one request, its admission is held to the engine's terminal frame, and the
  dispatcher keeps serving.

`/v1/brio` is untouched: no line of its path is in the diff, and its header keeps the field it has
always carried.

Happy to look at anything you want changed.

PR #1354, opened 2026-09-05
Security: confine image_url local reads and pin CI actions by SHA

dmoraesrs:
Security hardening from a review of the repository (C memory-safety, the Python server, the web/desktop front-ends, and CI). The review found nothing malicious and no RCE path via a crafted checkpoint: the C weight parsing already validates shape/offset and guards integer overflow (`st.h`, `coli_v4_layer_validate`). This PR closes the actionable, low-risk findings.

### 1. `fix(server)` — local file read via `image_url.url`
`_image_bytes_from_url` reads a local path with the server process's own permissions. On a server that binds beyond loopback (which already requires an API key), an authenticated client could send `file:///etc/passwd` and read any file the process can reach, and the `OSError` message echoed the full path back, confirming a file's existence and permissions.

- Refuse `..` outright (never needed for a real image path).
- If `COLI_IMAGE_ROOT` is set, the resolved path must stay inside it, using the same `relative_to()` check `serve_static` already uses.
- Errors are generic, so a reply no longer confirms a path or its permissions.
- The documented loopback single-user case (a plain path or `file://` URI) is unchanged.
- Covered by `ImageUrlPathGuard` (4 tests). `test_openai_server` stays green (165 tests).

### 2. `ci` — GitHub Actions supply chain
- Third-party actions were referenced by mutable tags (`msys2/setup-msys2@v2`, `hadolint/hadolint-action@v3.1.0`, `Jimver/cuda-toolkit@v0.2.19`, `ilammy/msvc-dev-cmd@v1`). A tag can be moved by whoever controls the action's account, so a compromised maintainer would run arbitrary code with the job's `GITHUB_TOKEN` — including the `contents: write` token in `release.yml`. Each is pinned to the commit SHA the tag currently points at, with the version kept in a trailing comment.
- `ci.yml` had no `permissions` block; it only reads the tree, so its token is pinned to `contents: read` as defense in depth.
- Official `actions/*` are left on tags (lower risk, avoids a massive diff).

## Validation
- `python3 -m unittest tests.test_openai_server` → 165 tests OK
- `actionlint` on the three workflows → no new errors (only pre-existing shellcheck notes)
- YAML valid across all four workflows

## Suggested follow-ups (not included — they need a decision or end-to-end testing)
- **Model-download supply chain (highest severity):** the download scripts (`download_fp8.py` and `c/tools/*`) validate only file size, not a SHA-256 / signature. `hf_hub_download` already verifies the LFS hash, but the `curl` fallback and the ModelScope path do not. `fetch_deepgemm.sh` already pins by git commit and is a good model. Left out because it rewrites download logic for hundreds of GB that cannot be validated without a full download.
- **`release.yml`:** add `actions/attest-build-provenance` (SLSA, native to GitHub) to attest the binaries' provenance; today there is only a `SHA256SUMS.txt` generated by the same pipeline (proves integrity, not provenance).
- **`--repo` without an allowlist** in the conversion scripts.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -46,7 +46,7 @@ jobs:
         shell: msys2 {0}
     steps:
       - uses: actions/checkout@v4
-      - uses: msys2/setup-msys2@v2
+      - uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2
         with:
           msystem: UCRT64
           update: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,11 @@ on:
   push:
     branches: [dev, main]
 
+# CI only reads the tree; no job here writes to the repo. Pin the token to
+# read-only so a compromised step (or action) cannot push, tag or release.
+permissions:
+  contents: read
+
 jobs:
   engine:
     name: Engine (Linux, CPU)
@@ -55,7 +60,7 @@ jobs:
       - name: Lint both Dockerfiles
         # docker/Dockerfile is not built here (see above) but it is still shipped
         # and followed by users, so it gets checked for syntax at least.
-        uses: hadolint/hadolint-action@v3.1.0
+        uses: hadolint/hadolint-action@54c9adbab1582c2ef04b2016b760714a4bfde3cf # v3.1.0
         with:
           dockerfile: docker/Dockerfile
           failure-threshold: error
@@ -170,7 +175,7 @@ jobs:
       - uses: actions/checkout@v4
       - if: matrix.os == 'windows-latest'
         id: msys2
-        uses: msys2/setup-msys2@v2
+        uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2
         with:
           msystem: UCRT64
           update: false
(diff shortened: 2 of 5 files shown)


JustVugg:
Reviewed for the 1.10.2 patch release, and it holds up:

- every pinned SHA matches the commit its tag points to (checked each of the four against the action repos' `git/ref/tags`);
- `contents: read` is enough: no job in `ci.yml` writes to the repo, and `actions/cache` does not use that scope;
- the image-path guard does what it says in all five cases I probed directly: a plain path with no root still reads, `..` is refused, a path outside `COLI_IMAGE_ROOT` is refused, a path inside it reads, and a missing file gets the generic message;
- 165 server tests green locally with your 45 on top, and the branch merges into `dev` cleanly.

The only thing missing was CI itself: as a first-time contribution it had never been allowed to run. I pushed a merge of `dev` onto your branch (no conflicts, nothing of yours changed) so the workflows exist, and approved them. If they come back green this goes in with the patch release. Thank you for a security PR that came with tests and a rationale instead of a scare.

dmoraesrs:
Ran the CONTRIBUTING local check (`make check`): portable CPU build + C unit tests pass; `python3 -m unittest tests.test_openai_server` passes 165/165 with the sanitized `_image_bytes_from_url`. `actionlint` on the three workflows reports no new errors (only pre-existing shellcheck notes). The full Python run shows one unrelated failure (`test_resource_plan`) that needs ~129 GB free disk — an environment limit, not touched by this change.

PR #1356, opened 2026-09-05
feat(engine): replace the ablation scoring mode with a checked, digest-bound one

monotophic:

**Revised 2026-09-17 in response to review: 15 files and +5,449 down to 12 files and +3,706, a 1,743-line
reduction.** The `AblateWriter` vtable is collapsed to a plain `FILE *` as asked, the cap now bounds what its name claims, and `tools/check_ablate_evidence.py` with its release-archive gate have moved here from #1355, where they had no writer.

`ABLATE_SCORE=<manifest>` replaces the decode with a teacher-forced sweep: for each item it configures the ablation cells, runs one prefill, and reads the final logits at every target position. That mode already exists on `dev`.
What it cannot do is produce evidence anyone can check. It accepts whatever a permissive whitespace scan can make of a line, skips lines it cannot parse and carries on, silently truncates an over-long cell list, and writes
records naming neither the config nor the manifest they came from. A partial run looks exactly like a complete one
— it exits 0 after writing nothing but a header.

**This PR replaces that mode's implementation, and the change is visible to anyone parsing its output.** The
artifact's schema goes from `coli-ablate/1` to `coli-ablate/2`, the mode's exit status is now its own result
instead of a constant 0, and inputs the old mode would have partly executed are now refused as a whole.
`docs/ENVIRONMENT.md` describes the mode as it now behaves.

**On replacing the mode rather than adding an alternative — the question is answered.** The review asked whether
`ABLATE_SCORE` is used by anyone outside our tree, and pre-committed: if not, replacing it is right. **We conducted
a search of the project's public GitHub and found nothing** — all upstream references, 53 issues and discussions,
and public code search. That is the strongest claim the evidence supports; it is not a claim of no use anywhere.
One fact offered without a recommendation: **the mode was written by @jeswr**, still an active contributor here.
We have not contacted them — that is the maintainer's call.

What the new mode gives you: the manifest is a defined format with an exact field count and single-space
separators, so the same content always produces the same digest; every field is bounded against the loaded config
before anything runs; anything outside that is refused as a whole, naming the record at fault, including a
duplicate item id. A CRLF terminator and a missing final newline are both accepted and normalised — a host editor
produces them without meaning to — and all three framings of the same content bind the same digest. The artifact
gains the digests of the config and the manifest, the item and target counts it expects, and a terminal record
carrying the counts actually completed, so a truncated file is detectable. Positions whose logits are not finite
are refused by name instead of being written out as unusable numbers, the top-k list holds real vocabulary entries
instead of padding a fixed width with sentinels, and the output path is created exclusively, never truncated.
Every refusal says why on stderr before the mode returns non-zero.

The first commit adds the two small foundations that make the above checkable: a classified reduction of a logit
row (`logprob_row_checked`), which reports *why* a row was unusable rather than only that it was; and a short
self-contained SHA-256 (`evidence_digest.h`) so a mode can name the exact bytes it consumed without the engine
taking on a dependency it otherwise never needs. The plain sampling path is untouched, and the new path is pinned
to agree with it on every well-behaved row.

**The reported numbers are unchanged.** The negative log-likelihood and the partition function are computed exactly
as before, in double, and are only printed at full precision now — a reader comparing old and new files sees more
digits of the same value.

## The cap: we went past the prose fix you asked for

The review asked us to make the prose match what `ABLATE_MAX_ITEM_TOKENS` actually bounds. We started there, then
measured, and found the prose was understating it: **three** unbounded allocations are reachable from a manifest,
not one, and the largest is `kv_alloc` at **177.75 GiB** at the cap — 7.4× the prefill buffer the review named.

At that point "make the comment honest" had two forms: admit our own doc is false, or make it true. The doc in this
PR says the cap exists "so that a corrupt manifest cannot ask the loader for an arbitrary allocation", and
`ABLATE_MAX_ITEM_TOKENS` is new in this PR — so the variable whose comment lied is ours.

**The cap now bounds what its name claims**, derived from a limit the engine already has rather than a new
constant: `ablate_item_token_limit() = min(CTX, ABLATE_MAX_ITEM_TOKENS)`, checked ahead of all three allocations.
A manifest over it is refused by name before anything is allocated; a manifest at the real maximum loads. No new
constant, no new environment variable.

**Disclosed, not closed:** this bounds item *length*, not item *count*. A long manifest still costs ~240 B/item,
which is ordinary loader behaviour proportional to input. The hazard was a short line demanding 178 GiB, and that
axis is shut.

The 21 GB figure in the review reproduces at `hidden` 5,120; at GLM-5.2's actual `hidden` of 6,144 the prefill
buffer is **24.00 GiB** (`2^20 * 6144 * 4 B`). Both allocations are worked out in full in the comment beside the
cap rather than left as bare constants.

(the fp8 container line in #1102, the OpenAI-compatible server hardening in #1353, the evaluation harness in #1355,
this engine block, and a transcript check tool in #1357). This one carries the engine block: the ablation scoring
mode, the row-status and digest headers, the offline checker that reads the mode's output, and their tests.
**The round-trip case now runs entirely inside this PR** — the producer and `tools/check_ablate_evidence.py` are
both here, so nothing about it waits on another PR. **It no longer carries `tools/engine_evidence.py` at all.** This revision moved the manifest helpers into
`check_ablate_evidence.py`, which left nothing here importing that module — so shipping it would have meant 524
lines of code nothing in this PR uses. It lives in #1355, which has a real consumer, and in #1357, which imports it. The others are proposed separately, each with its own
evidence. The set-level explanation and a traceability matrix covering every request across the three PRs are in the #1355 comment: https://github.com/JustVugg/colibri/pull/1355#issuecomment-5725475017.

**Behavioral contract**
- No default decode path changes what it computes: an unchanged binary and this one produce the same output, token for token, on the same container.
- The mode runs only when `ABLATE_SCORE` is set, and no other path reaches it.
- Every rejected input is rejected by name, on stderr, before a non-zero return; no input is partly executed.
- A manifest declaring more tokens for one item than `min(CTX, ABLATE_MAX_ITEM_TOKENS)` is refused **before** any of the three allocations it would otherwise size.
- The reported `nll` and partition function are bit-for-bit what `dev` computes.
- Producer output passes an offline checker written independently of the engine, for each manifest framing the
  engine accepts, and all three bind one digest.
- Everything that worked before still works: the branch builds and the existing suites pass with the rest of the
  set absent.

**Structural changes on the default load path.** The mode's digest is computed only when `ABLATE_SCORE` is set, so the load path does no new work on an ordinary run. Eight things about it are nonetheless different from `dev`, and
you should not have to find them yourself:

1. two new unconditional includes, `<stdarg.h>` and `<inttypes.h>`;
2. a new unconditional `#include "evidence_digest.h"` — a new file whose functions are plain `static`;
3. `Cfg` gained `char config_sha256[65]`, so every `Cfg` in the process grows;
4. `cfg_root` gained a third parameter;
5. the digest call inside `cfg_root` runs only when a caller passes a buffer;
6. `load_cfg` reads one environment variable on every call to decide whether to pass that buffer — one `getenv` per model load;
7. `run_score`'s prefix probe passes `NULL` for the new parameter;
8. `sample.h` gained three `static inline` functions and two types; that header is compiled into `olmoe.c` as well, where they are unused.

**Capstone matrix** — one decisive artifact per claim.

| claim | decisive evidence |
|---|---|
| the default decode path is unchanged by this PR | an engine built from `dev` and one from the integrated tree, run against the **same real GLM-5.2 container** (78 layers, 256 routed experts, hidden 6,144, int3/g64), same prompt, temperature 0, KV cache deleted before **each** arm so neither could inherit the other's prefill: **the same 24 generated tokens, id for id, on CPU and again on CUDA**. The CUDA arms are asserted to have actually used the GPU — `resident set: 625 tensors, 10.17 GB VRAM` — because an earlier pass with `CUDA_EXPERT_GB` at its default of 0 placed nothing in VRAM and silently ran on the CPU |
| the mode is unreachable without its variable | `tests/test_ablate_mode_gate.py` reads the engine source and asserts exactly one product entry into the mode plus one compile-gated adapter entry, and that `run_ablate_score` has exactly one caller inside the mode; it fails if a second call site appears |
| `nll` is still `dev`'s double-precision reduction | `tests/test_ablate_mode.c` drives a row through the real emitter whose logits are chosen so a single-precision intermediate differs, and compares the emitted text with `==` against an expectation computed in the test from `dev`'s formula; substituting the float intermediate fails it |
| the header's `config_sha256` really comes from the load path | the same suite writes a temp `config.json`, calls `load_cfg`, and compares the field against an independently computed digest **and** a pinned literal; it also asserts the field is left unset when the variable is not set |
| the manifest digest means the same thing in C and in Python | the same 64-character known answer is pinned in `tests/test_ablate_mode.c` and in this PR's Python suite, and reproduces outside both: `printf 'coli-ablate-manifest/2\n0 3 2 0 0 1 2 3\n' \| shasum -a 256` → `c63a48c375b14ca60f26c7e3c5dd36b5929ffaf669a45511c93deee6e8bbd5ed` |
| producer output survives a real round trip | an integration run drove this PR's own engine mode over a real container and fed the artifact to this PR's own checker, end to end on the tree being pushed: the checker validated the engine's own `ABLATE_OUT` against the manifest and the config, and passed |
| the refusals are real, not decorative | `tests/test_ablate_mode.c` reports **79 named checks at runtime** (`ok:` lines from a real run — a static count of `CHECK(` call sites gives 54, because several sit in loops) with no model and no weights — the loader, the writer, the three accepted framings, each named refusal, and the dispatch contract; `tests/test_logprob_status.c` reports **138** the same way, including published SHA-256 vectors at the block boundary and both padding branches. Both `PASSED (0 failures)` at the head being pushed |
| the cap refuses before it allocates, and is derived not invented | `tests/test_ablate_mode.c` section *"the per-item token limit is derived from CTX, not a bare constant"*: **cap+1 (4097 tokens) is refused, naming the CTX=4096-derived limit**, and **a manifest at the real CTX-derived maximum (4096) loads and runs**. The refusal is asserted to name the limit, not merely to occur |
| the packaged release archive imports what it ships | RAN at this head: the archive is built, and `test -f`, `ast.parse` and the new `python3 -c "import tools.check_ablate_evidence"` all pass. **Stated precisely, because the distinction matters: at this head `check_ablate_evidence.py` has no local imports at all** — the manifest helpers moved into it during this revision — so the import check currently has no dependency to catch. It is kept because it is one line and asserts importability for any dependency the tool gains later. **It is hardening a gate that could not detect importability, not a fix for a live break** |

<details>
<summary>What changed since the reviewed head, how the fleet comparison was set up, and origin accounting</summary>

**What changed since `cc80928e`, the head reviewed on 2026-09-17**

| change | origin |
|---|---|
| the `AblateWriter` vtable collapsed to a plain `FILE *`, tests driving temp files they create and remove (not literally `tmpfile(3)`). **`AblateOutputRun`, also named in the review, is NOT collapsed** — it keeps two production-path function pointers | **maintainer-requested**, partially |
| the `AblateModeRunFn` typedef and its one-entry `ablate_mode_dispatch` table removed — a function pointer with exactly one implementation, wrapping what is now a direct call | **re-review-found**, and **not named in the review**. `ablate_model_mode_run` remains as a plain static that null-checks the path and calls `run_ablate_score` |
| the cap derived and enforced ahead of all three allocations | **re-review-found** — the review asked for prose; the measurement said prose was not enough |
| `tools/check_ablate_evidence.py`, its test, and the `release.yml` gate moved here from #1355 | **maintainer-requested**, resolved differently — the review asked for them to be parked; their writer is this PR's `coli-ablate/2` mode, so they moved instead |
| a real import check added to the release-archive gate **alongside** the existing `test -f` and `ast.parse`, not replacing them | **re-review-found** — neither original executes anything, so neither can catch an import-time failure. That gate is ours, added in this PR |
| a 2.01× `logit_topk_select` regression fixed, output byte-identical | **re-review-found** |
| an `ENVIRONMENT.md`/comment contradiction corrected | **re-review-found** |

**Where the added `c/colibri.c` lines go, measured at this head (+838 / −74).** Categories are exact integer
counts and sum to 838: manifest parse / validation / digest **294 (35.1%)**, of which `ablate_manifest_load` is
**171** (`c/colibri.c:8178-8348`); writer / output path **171 gross (20.4%)**; the cap / refusal path **56 (6.7%)**,
which is new in this revision; test-only `#ifdef COLI_TEST_ABLATE_ADAPTERS` lines **35 (4.2%)**, absent from a
normal build; remainder — scoring helpers, error paths, dispatch — **282 (33.7%)**. Comments are 151 of the 838
(18%), spread across categories. The old `run_ablate_score` on `dev` is **79 lines**
(`origin/dev c/colibri.c:7929-8007`); "about 88" is the diff hunk's old-side span.

**The writer collapse, stated precisely — including the part that is not done.** `AblateWriter` and its
four-function vtable (`vprintf_fn`, `puts_fn`, `flush_fn`, `close_fn`) are **gone: 21 references at the reviewed
head, 0 here.** The writer is a plain `FILE *`, driven in the tests by temporary files they create and clean up themselves (PID-suffixed, opened with `fopen`) rather than by `tmpfile(3)` itself. **`AblateOutputRun` survives with
two function pointers** (`open_fn`, `body_fn`) which are **not** compile-gated and sit on the production path
(`c/colibri.c:8443-8466`) — they are the seam that lets the C tests run the parser with no model. The review named
both; only one is eliminated, and collapsing the second is offered as a follow-up rather than claimed as done.

**`c/Makefile` and #1102 — the precise statement.** #1356 and #1102 both add entries to `c/Makefile` and **will
conflict there — only there; the engine sources merge cleanly**, which we verified by merging them in both orders.
Whichever lands second, we re-merge `dev` and push within a day. Separately, if #1102 lands first, `quant.h` gains
`fp8_format.h` and these two rules should list it too, exactly as their sibling rules do.

**No shared file remains in this PR.** `tools/engine_evidence.py` and `tests/test_engine_evidence.py` were dropped
from it entirely once this revision left it with no importer of that module; they stay in #1355 and #1357, which do
import it. **Any merge order still works** — all six orderings of the three were built in a scratch worktree and
produce one identical tree with zero conflicts, re-verified after the drop.

Origin accounting for the original content is unchanged from the reviewed head: the mode's implementation, the two
headers and their tests are new content re-expressed onto current `dev`; one behaviour of `dev`'s is deliberately
superseded; the checker's CRLF leniency was corrected during review to match `dev`'s contract of record; two
committed pins (the `nll` value and the `load_cfg` digest) close gaps nothing guarded before. One commit lists
`evidence_digest.h` among `colibri`'s Makefile prerequisites, because upstream's Makefile-prerequisite test (#1284)
requires every header a source includes to be declared — no behaviour change. The shared logprob helper in
`c/sample.h` now promotes the logit before subtracting; no number this PR writes to disk changes, because the
evidence writer computes its own double subtraction. No surrounding code was reformatted.

</details>

**Durable vs current state:** the mode's format, its refusals, the digest binding, the derived cap and the dispatch
guard are durable. Everything measured in this body — diffstats, suite counts and timings, the cap behaviour, the
archive gates — is measured against base `cc756a63` at the head being pushed. **No evidence from an earlier
revision of this PR is carried into it** — the decode-identity and round-trip results above were regenerated
against the current integrated tree rather than quoted from the previous round.


monotophic:

Merged current `dev` (`87c54a93`) into this branch. Clean auto-merge, no conflict resolution. I read the merged `colibri.c` and `sample.h` by hand since dev landed several nearby changes (expert-FFN extraction, qwen38 CUDA tier, sparse routing, chat stats, OpenMP team sizing): this branch's added functions (`logprob_row_checked`, `logprob_from_row_checked`, the checked ablation-scoring rewrite, `load_cfg`'s config-digest binding) are intact and called the same way. Content on this branch's side is unchanged on all 15 files. `make check` exit 0 on the merged head (1143 tests); this branch's C tests and Python modules pass.

@JustVugg similar steering request to #1102, #1353, and #1355, this PR is part of the set to make the full checkpoint-quality GLM 5.2/5.3 containers work and to establish the OpenAI API instrumentation needed for container quality studies. Let me know if this PR is off target or otherwise not likely to merge. Thanks for all you do, this project is awesome :)

JustVugg:
Reviewed, and I want to answer your own question rather than pretend the review is about the code.

The code is careful and the defect it replaces is real. I read the old `run_ablate_score` on dev and it does exactly what you say: `strtol` failing means `continue`, an out-of-range cell count is silently clamped, a bad item prints to stderr and continues, and the whole thing returns 0 regardless. A manifest where every single line was malformed produced a header-only file and an exit status that said success. Anyone using that mode to make a claim about experts was building on sand.

You ask whether I want the old mode back. The answer is that this is the one place in the set that breaks the project's convention, and you say so plainly rather than hiding it: no new environment variable, but `ABLATE_SCORE` and `ABLATE_OUT` both change meaning, the schema goes from coli-ablate/1 to /2, the exit status changes, and `ABLATE_OUT` now refuses an existing path instead of truncating.

So the question back to you is narrow: is `ABLATE_SCORE` used by anyone outside your own tree? If it is, I want the new mode behind coli-ablate/2 as an opt-in and the old behaviour intact. If it is not, replacing it is right and I will take it as it stands.

Two smaller things either way.

801 lines of `colibri.c` to replace about 88, and a good part of that is the AblateWriter and AblateOutputRun indirection, which exists so the C tests can drive the parser with no model. In a project whose engines are deliberately flat single files that is a lot of abstraction bought for testability. Could the writer collapse to a plain FILE pointer with the tests driving a tmpfile?

And the new loader materialises the whole manifest in RAM where the old one streamed a line at a time. `ABLATE_MAX_ITEM_TOKENS` bounds the 4 MB token array, not the `maxT * hidden` prefill buffer, which at the cap is about 21 GB. Your comment next to it half concedes this; I would make the prose match what the cap actually bounds.

monotophic:
Set-level explanation and the full traceability matrix: https://github.com/JustVugg/colibri/pull/1355#issuecomment-5725475017

**The diff is smaller: 15 files and +5,449 → 12 files and +3,706, a 1,743-line reduction.**

### `ABLATE_SCORE` — the answer to your narrow question

You asked whether `ABLATE_SCORE` is used by anyone outside our own tree, and pre-committed: if not, replacing it is
right and you will take it as it stands.

**We conducted a search of the project's public GitHub and found nothing** — all upstream references, 53 issues and
discussions, and public code search. That is the strongest claim we can make; you can price the residual risk from there better than we can.

Note **the ablate mode was written by @jeswr**, who is still an active contributor. We have not contacted them but maybe they can jump in if they have more to say about it.

### Where the added `colibri.c` lines go

You asked why replacing about 88 lines costs 801. Two corrections, measured at the revised head: the
count is now **+838 / −74**, because this revision added the derived cap and its refusal path on top of what you
reviewed; and the old `run_ablate_score` on `dev` is **79 lines** (`origin/dev c/colibri.c:7929-8007`) — "about 88"
is the diff hunk's old-side span. Neither changes your point.

| category | lines | share of 838 |
|---|---:|---:|
| manifest parse / validation / digest (`ablate_manifest_load` is 171 of these, `c/colibri.c:8178-8348`) | 294 | 35.1% |
| writer / output path (gross added) | 171 | 20.4% |
| cap / refusal path — **new in this revision**, did not exist when you reviewed | 56 | 6.7% |
| test-only, `#ifdef COLI_TEST_ABLATE_ADAPTERS`, absent from a normal build | 35 | 4.2% |
| remainder — scoring helpers, error paths, dispatch | 282 | 33.7% |

Categories are exact integer counts and sum to 838. Comments are 151 of the 838 (18%), spread across all
categories rather than concentrated in one.

The bulk is validation the old mode did not do — which we see by reading `dev`: `strtol` failure
meant `continue`, an out-of-range cell count was silently clamped, a bad item printed to stderr and continued, and
the whole thing returned 0 regardless.

### The writer collapse — done, and one part of your ask that is NOT done

**`AblateWriter` is gone.** The four-function vtable you objected to — `vprintf_fn`, `puts_fn`, `flush_fn`,
`close_fn` — no longer exists: **21 references at the head you reviewed, 0 now.** The writer is a plain `FILE *`
and the C tests drive it through temporary files they create and clean up themselves (PID-suffixed, opened with `fopen`) — the substance of what you suggested, though not literally `tmpfile(3)`.

**But you named two things, and the second one survives.** `AblateOutputRun` still carries two function pointers,
`open_fn` and `body_fn`, and they are **not** compile-gated — they are on the production path
(`c/colibri.c:8443-8466`). They are the seam that lets the C tests run the parser with no model at all. We are
flagging it rather than letting you find it: the indirection is reduced, not eliminated. **If you want that
collapsed too, say so and it is a small follow-up.**

We also removed something you did **not** name: `AblateModeRunFn` and its one-entry `ablate_mode_dispatch` table —
a function-pointer typedef with exactly one implementation (2 references each, now 0). What remains is
`ablate_model_mode_run`, an 8-line wrapper that null-checks the path and calls `run_ablate_score` directly.

### Your asks

|---|---|---|
| "is `ABLATE_SCORE` used by anyone outside your own tree?" | **done** — searched, nothing found; replacement stands per your own condition | above |
| "Could the writer collapse to a plain FILE pointer with the tests driving a tmpfile?" | **done for `AblateWriter`, NOT for `AblateOutputRun`** — the `AblateWriter` vtable is gone (21 references → 0) and the writer is a plain `FILE *`, driven by temp files the tests create and remove (not literally `tmpfile(3)`). **`AblateOutputRun`, which you also named, still carries two function pointers on the production path**; see below. Bite re-checked: 4 distinct mutations against `test_ablate_mode.c`, each a real failure | `c/colibri.c` |
| "I would make the prose match what the cap actually bounds" | **done, and then some** — see the section below | `c/colibri.c`, `docs/ENVIRONMENT.md` |

### The cap: we went past the prose fix, and here is why

You asked for prose only. We started there, then measured, and found the prose was understating it: there are
**three** unbounded allocations reachable from a manifest, not one, and the largest is `kv_alloc` at **177.75 GiB**
at the cap — 7.4× the prefill buffer you named.

At that point "make the comment honest" had two possible forms: admit our own doc is false, or make it true. Our doc
in this PR says the cap exists "so that a corrupt manifest cannot ask the loader for an arbitrary allocation", and
`ABLATE_MAX_ITEM_TOKENS` is new in this PR — so that one is ours.

**The cap now bounds what its name claims**, derived from a limit the engine already has rather than a new constant:
`min(CTX, ABLATE_MAX_ITEM_TOKENS)`, checked ahead of all three allocations. A manifest over it is refused by name
before anything is allocated; a manifest at the real maximum loads. No new environment variable.

**Disclosed, not closed:** this bounds item *length*, not item *count*. A long manifest still costs ~240 B/item,
which is ordinary loader behaviour proportional to input. The hazard was a short line demanding 178 GiB, and that
axis is shut.

Your 21 GB figure reproduces at `hidden` 5,120; at GLM-5.2's actual `hidden` of 6,144 the prefill buffer is
**24.00 GiB** (`2^20 * 6144 * 4 B`), against **177.75 GiB** for the KV allocation. Both are worked out in full in
the comment beside the cap rather than left as bare constants, and `docs/ENVIRONMENT.md` now states the derived
bound and names all three allocations it covers.

<details>
<summary>Details — origin accounting</summary>

**maintainer-requested:** `ABLATE_SCORE` search; writer collapse; cap prose.
**re-review-found, disclosed:** the three-allocation measurement and the derived bound; removal of the
`AblateModeRunFn` typedef and its one-entry `ablate_mode_dispatch` table — a function pointer with exactly one
implementation, wrapping what is now a direct call — which **you did not name** and we removed anyway
(`ablate_model_mode_run` remains, as a plain static that null-checks the path and calls `run_ablate_score`); a 2.01× `logit_topk_select` regression fixed with
byte-identical output; and a real import check (`python3 -c "import tools.check_ablate_evidence"`) added to the
release-archive gate **alongside** the existing `test -f` and `ast.parse`, not replacing them. That gate is ours,
added in this PR, and neither of the original two executes anything, so neither catches an import-time failure. The
archive is verified correct today; the gate now asserts that rather than implying it.

**#1356 and #1102 both add entries to `c/Makefile` and will conflict there — only there; the engine sources merge
cleanly.** Whichever lands second, we re-merge `dev` and push within a day.
</details>

monotophic:
*Authored by Claude Opus 5 in Claude Code, analysis in partnership with @Monotophic.*

Merged current `dev` into this branch. The only conflict was one line in `c/Makefile`, the list of headers `colibri` is built from: `dev` had added two, this branch adds one, and the merge keeps all three. We checked the result against both sides: nothing of yours is reverted and nothing of ours is dropped.

`make check` is green on the merged tree, including this PR's own C and Python tests. The code this PR replaces is unchanged on `dev` since your review, so the review still applies as written.

One thing that postdates your review: `/v1/brio` landed, and while they are related, this feature is distinct. Brio scores a closed set of options against one state and reports a probability and an entropy. This mode ablates named experts and reports what the logits did about it at every target position — an intervention, not a scorer — so brio does not subsume it. We set this up offline-only, reachable from a manifest and nothing else. I think this is appropriate, but am open to direction if you want it wired in.

monotophic:
Rebased onto current dev to clear the conflict in `c/Makefile` and `c/colibri.c`. The same five commits, now linear; the only conflicted, hand-resolved lines are the `colibri` prerequisite line, which keeps dev's list and adds `evidence_digest.h`, and the order of two blocks at the top of `main()`, where dev's `ORACLE_STRICT` check now runs first so it still refuses to combine with `ABLATE_SCORE`. The Makefile diff also carries this branch's own `test_logprob_status` and `test_ablate_mode` rules, unrelated to the rebase. Two small fixes rode along, found by a review of the rebased tree: the branch's two test rules now list `oracle.h` the way every other rule that compiles `colibri.c` does since a recent commit added it there, and a compile-gated entry at the top of `main()` that no build could ever reach is gone, along with the gate test's assertion that it existed; that test now checks that every reference to the adapter state stays behind its compile guard. The earlier red on the CUDA syntax job was the NVIDIA apt mirror mid-sync, not the branch; this push re-runs it.

JustVugg:
Still one open question from my last review, and it decides the shape: is `ABLATE_SCORE` used by anyone outside your tree? The PR changes the meaning of two existing variables (`ABLATE_SCORE`, `ABLATE_OUT`), the output schema and the exit status. If nobody else depends on the old mode, say so and I take it as it stands; if you are not sure, the new mode goes behind coli-ablate/2 as an opt-in. Not in 1.12.1 either way: it changes existing behaviour and the answer is not in yet.

monotophic:

We found no consumer in the public repo for `ABLATE_SCORE` or `ABLATE_OUT` (and we cannot see private ones).

What we searched. Current `dev` and every other branch on this repository: nothing reads `ABLATE_OUT` or parses the `coli-ablate/1` records; the only references are the engine itself, the comment in `abl.h` and the `ENVIRONMENT.md` row. The PR and issue corpus: four hits, which are #621, which introduced the mode and framed it as "an off-by-default research instrument", this PR and its sibling #1355, and #1625, an olmoe port of the same schema opened on 2026-09-19 and closed unmerged two days later, so nothing landed. GitHub-wide code search for the two variable names and the schema string: two repositories vendor colibri's headers or docs and so carry the variable name in a comment or a table row, and neither sets `ABLATE_SCORE`, reads `ABLATE_OUT` or parses the records. The introducing author's fork: its harness branch's tools directory holds no reader for the output, its last commit is from 2026-07-31, and its author has filed nothing on this repository since #621 merged; we did not read every file on that branch. Our own lab: the only harnesses there are the ones written for the new schema.

monotophic:
@JustVugg do you have advice/direction regarding the CI situation? I'm pretty sure this isn't a legitimate failure on this PR. I have experimented with close/reopening cycles to get to clean CI for another PR, but I feel bad about the extra load on the CI services. 

I hope that all is well with you and appreciate what you do with this project.

JustVugg:
Thank you, and apologies: you had answered my question on the 18th, with the search, the writer collapse and the cap, and I asked it again on the 23rd without seeing it. Your answer meets the condition I set, so the replacement goes in as it stands; the two function pointers left in `AblateOutputRun` are fine as the test seam, no follow-up needed.

On CI: you are right, it is not this PR. The macOS job had `timeout-minutes: 15` and `make check` there now takes about 14.5 minutes, so it gets cut off inside the Python suite with nothing failed. #1737 raises it to 25. Once that is on dev I update your branch from dev (that starts a fresh run with the new limit), and merge when it is green; no need to close and reopen anything. It goes into 1.12.1. Merged onto current dev here, `colibri` builds without warnings and `test_ablate_mode`, `test_logprob_status`, `test_oracle` and `test_json` pass.

JustVugg:
A correction to my last line: this will not go into 1.12.1 after all. The release is frozen on the current `dev` head while its CI finishes, and I would rather not start another full round of runs under it. The review stands and nothing is needed from you.

Your PR is first in line after the `v1.12.1` tag. Once #1737 (the timeout fix, which also covers the Linux job) is on `dev`, I update your branch from `dev` and merge it when that run is green.

JustVugg:
v1.12.1 is tagged, so this is next. I pushed `6fe19af` to your branch: a merge of current `dev`, which now has #1261. The only conflict was the `colibri` rule's prerequisites in `c/Makefile`, where you add `evidence_digest.h` and #1261 adds `backend_xdna.h` and `$(XDNA_OBJ)`; I resolved it as the union of both. Locally the engine builds without warnings, and `test_ablate_mode`, `test_logprob_status` and the three portable XDNA tests pass. It merges when this run is green, with the 25 minute limit from #1737 in place.

PR #1357, opened 2026-09-05
test(tools): gate the logprob gap check on the engine preamble

monotophic:

**Revised 2026-09-17 in response to review. The native-MTP witness is parked and this PR is now the logprob-gap
hardening alone** — 11 files and +7,464 down to **4 files and +2,319**. Title updated to match; the previous one
still advertised the witness.

A command-line check that reads an engine transcript and says whether it is internally consistent.

`tests/check_data_logprob_gaps.py` already existed and had two weaknesses. It read a transcript without first
establishing that the transcript came from an engine it understands, so an unrecognised header was treated as no
header; and it reported problems without saying where in the byte stream they were, which makes a disagreement on a long capture tedious to localise. It now parses the engine's preamble through a shared helper and **fails loudly
when that preamble is unparsable**, the protocol grammar is exact, the numeric tail and the top-k cap are enforced,
and every reported problem cites the byte offset of the frame it concerns.

The numeric grammar accepts both the fixed six-decimal form the engine ships today and an earlier `%.17g` form
(plus `nan`/`inf`/`-inf`); a token can satisfy both spellings exactly, so form classification stays informational —
a tally in the summary line — rather than a pass/fail rule.

**What was parked, and why.** `tests/check_native_mtp_witness.py` and its test — **2,613 lines exactly** — are out
of this PR. The review was right on every number: `MTPEMIT` appears in zero files on `dev`, so the tool could only
ever exit 3 `UNSUPPORTED` on any build that exists, and carrying a reader ahead of its writer is the pattern we
identified ourselves in #1353. The work is preserved at `monotophic/colibri` `codex/parked-mtp-witness` @
`e0474748` and **is proposed again alongside the engine change that emits the record it needs** — not before.

**What invokes this tool.** A person. It is a human-invoked `argparse` tool, like `tools/check_ablate_evidence.py`,
which is why nothing in `c/Makefile` or any workflow references it and why an import walk cannot find a caller. Its
unit tests run in CI; the tool itself is run by hand against a captured transcript when diagnosing an engine. If you
would rather it were wired into something, say which and we will do that as a separate change.

(the fp8 container line in #1102, the OpenAI-compatible server hardening in #1353, the evaluation harness in #1355,
the engine's ablation scoring mode in #1356, and this transcript check). This one carries the logprob-gap check and
its suite. It imports `tools/engine_evidence.py` for the preamble grammar and **carries that file itself**, so it builds and tests standalone; once #1355 lands, that file drops out of this diff on the next `dev` merge. #1356 does not carry
it — that PR ended up importing nothing from the module, so it was dropped there. The others
are proposed separately, each with its own evidence. The set-level explanation and a traceability matrix covering every request across the three PRs are in the #1355 comment: https://github.com/JustVugg/colibri/pull/1355#issuecomment-5725475017.

**Behavioral contract**
- A transcript whose preamble cannot be parsed is a named failure quoting the offending line, never a silent pass.
- Every reported problem names the byte offset of the frame it concerns.
- Both numeric spellings the engine has used are accepted; neither is treated as evidence of anything by itself.
- Every record the check declares required is individually required — dropping any one of them fails the check.
- No model is run by anything here, and no fixture is downloaded: every case is built from literals inside the suite.

**Capstone matrix** — one decisive artifact per claim.

| claim | decisive evidence |
|---|---|
| an unparsable preamble cannot pass | `test_unparsed_banner_is_a_named_failure_quoting_the_line`, `test_unparsed_loaded_is_a_named_failure_quoting_the_line`, and `test_full_capture_with_unparsed_preamble_fails_the_whole_check` — the last drives a complete, otherwise-valid capture and requires the whole check to fail |
| the preamble gate cannot fail open | `test_preamble_returning_none_is_defensively_named_not_silently_passed` — the case where the parser returns nothing at all is asserted to be a named failure, which is the shape a "no news is good news" gate would pass |
| the frame-order gate is not decorative | the seven `test_dropping_<record>_breaks_its_own_ordinal_check` cases (`BANNER`, `LOADED`, `READY`, `STAT`, `HWINFO`, `TIERS`, `EMAP`) plus `test_required_order_matches_the_frozen_anchor`. **Before this revision every one of the seven could be dropped individually with the suite green** — the list was declared, not enforced. Each now fails on its own |
| the numeric grammar is source-true, both forms | `test_fixed6_form_is_accepted_and_reported`, `test_c17g_form_is_accepted_and_reported`, and `test_a_token_matching_both_grammars_is_ambiguous_and_still_accepted`; the fixtures use the engine's own printf formats |
| the grammar checks bite independently | `test_done_domains_and_fixed_grammar_bite_independently` — each rule fails on its own input rather than being masked by a neighbour |
| the tool and the shared grammar cannot drift apart | `test_checker_dispatch_prefixes_stay_recognized_by_the_module` — the check's own dispatch prefixes are asserted to still be recognised by `engine_evidence`, in both directions. An earlier one-directional version of this test passed while the module's dispatch was deleted outright |

<details>
<summary>Fuller matrix, what changed since the reviewed head, and origin accounting</summary>

**What changed since `e0474748`, the head reviewed on 2026-09-17**

| change | origin |
|---|---|
| `check_native_mtp_witness.py` + its test parked (2,613 lines) | **maintainer-requested** |
| five further files returned to the PR that owns each — `eval_glm.py`, `test_eval_glm.py`, `check_ablate_evidence.py`, `test_check_ablate_evidence.py`, `test_pack_python.py` | **maintainer-requested** (the de-duplication raised on #1355) |
| the seven per-ordinal frame-order cases, and the frozen order anchor | **re-review-found** — the gate declared seven required records and enforced none |
| the dispatch-prefix coupling test made bidirectional | **re-review-found** |
| the docstring now explains WHY a non-finite logprob is a real defect rather than a tolerable value | **re-review-found** — grammar-constrained decoding never zeroes or `-inf`s a vocabulary logit, so the only path to one is the total-vocabulary numerical collapse that `sample.h`'s own comments call a blow-up upstream. **Behaviour is unchanged, and this is added explanation — not the correction of a false claim**: measured across the two heads, the mentions of grammar/masking went up, not down |

|---|---|---|
| preamble gate | four named cases, including a full-capture case | loud failure in every shape |
| byte offsets | offsets asserted in the problem strings | present on every reported problem |
| exact protocol grammar | malformed and misordered global records, malformed framing | refused by name |
| frame-order lifecycle | per-record ablation, one record dropped at a time | all seven fail individually |
| numeric tail and top-k cap | literal fixtures in both spellings, non-dyadic values | accepted; classification informational |
| suite runs on `dev` plus this PR's files | `python3 -m pytest tests/test_check_data_logprob_gaps.py` | **61 passed, 110 subtests, 0.11 s** |

**Suite cost.** Parking the witness takes the two transcript-tool suites from **107 tests / 0.64 s to 61 tests /
0.11 s**, measured at `e0474748` and at this head. No test here exceeds **0.02 s** (slowest three measured at 0.02 s); there is no
fixture that waits on a timeout.

**A limitation we would rather state than have you find.** The gap this check hunts is not producible by the
engine's current multiplexing loop, so on today's builds it is regression coverage for an invariant rather than a
live detector, and the preamble check binds no binary or container digest by itself. Both facts are in the tool's
own docstring. Separately: the wire-format literals here are hand transcriptions of formats printed by
`c/colibri.c` and `c/telemetry.h`. Drift in the Python module is pinned by the tests above; **drift in the C source
is not** — no test in this branch reads a C file. That predates this revision, and fixing it would reopen the
shared file across all three PRs, so it is noted as a follow-up rather than folded in here.

Scope note: no file outside the four listed is touched, and no surrounding code was reformatted.

</details>

**Durable vs current state:** the grammar, the preamble gate and the frame-order contract are durable. The case
count (61) is current state, measured 2026-09-17 against base `cc756a63` on macOS (Apple silicon).

### DIFF
--- a/README.it.md
+++ b/README.it.md
@@ -3,7 +3,7 @@
 </p>
 
 <p align="center">
-  <a href="https://discord.gg/MAaKtQRc"><b>Discord</b></a> ·
+  <a href="https://discord.gg/FkyrEeJR"><b>Discord</b></a> ·
   <a href="README.md">English</a> · <a href="README.zh-CN.md">简体中文</a> · <a href="README.zh-TW.md">繁體中文</a> · Italiano
 </p>
 
@@ -352,7 +352,7 @@ Se ti è utile:
 - ⭐ metti una stella al repository e condividilo;
 - 🐛 apri issue con i numeri di benchmark del tuo hardware — i datapoint
   fanno avanzare questo progetto più di qualsiasi altra cosa;
-- 💬 entra nella [comunità Discord](https://discord.gg/MAaKtQRc) per discutere
+- 💬 entra nella [comunità Discord](https://discord.gg/FkyrEeJR) per discutere
   esperimenti, risultati hardware e direzioni di ricerca;
 - 💬 contattaci via GitHub issues per sponsorizzare lo sviluppo o donare hardware.
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 <p align="center">
   <a href="https://justvugg.github.io/colibri"><b>Website</b></a> ·
-  <a href="https://discord.gg/MAaKtQRc"><b>Discord</b></a> ·
+  <a href="https://discord.gg/FkyrEeJR"><b>Discord</b></a> ·
   English · <a href="README.zh-CN.md">简体中文</a> · <a href="README.zh-TW.md">繁體中文</a> · <a href="README.it.md">Italiano</a>
 </p>
 
@@ -597,7 +597,7 @@ today its numbers come from a community of real machines. If it's useful to you:
 - ⭐ star the repo and share it;
 - 🐛 open issues with benchmark numbers from your hardware — datapoints move
   this project more than anything else;
-- 💬 join the [Discord community](https://discord.gg/MAaKtQRc) to discuss
+- 💬 join the [Discord community](https://discord.gg/FkyrEeJR) to discuss
   experiments, hardware results, and research directions;
 - 💬 reach out via GitHub issues to sponsor development or donate hardware.
 
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -3,7 +3,7 @@
 </p>
 
 <p align="center">
-  <a href="https://discord.gg/MAaKtQRc"><b>Discord</b></a> ·
+  <a href="https://discord.gg/FkyrEeJR"><b>Discord</b></a> ·
   <a href="README.md">English</a> · 简体中文 · <a href="README.zh-TW.md">繁體中文</a> · <a href="README.it.md">Italiano</a>
 </p>
 
@@ -322,7 +322,7 @@ colibrì 最初由一人使用 12 核心、25 GB RAM 的笔记本开发；
 
 - ⭐ 为仓库加星并分享；
 - 🐛 以 issue 提交你的硬件 benchmark 数据——实测数据比任何其他事都更能推动项目；
-- 💬 加入 [Discord 社区](https://discord.gg/MAaKtQRc)，讨论实验、硬件数据与研究方向；
+- 💬 加入 [Discord 社区](https://discord.gg/FkyrEeJR)，讨论实验、硬件数据与研究方向；
 - 💬 若想赞助开发或捐赠硬件，请通过 GitHub issues 联系。
 
 ## 仓库结构
--- a/README.zh-TW.md
+++ b/README.zh-TW.md
@@ -3,7 +3,7 @@
 </p>
 
 <p align="center">
-  <a href="https://discord.gg/MAaKtQRc"><b>Discord</b></a> ·
+  <a href="https://discord.gg/FkyrEeJR"><b>Discord</b></a> ·
   <a href="README.md">English</a> · <a href="README.zh-CN.md">简体中文</a> · 繁體中文 · <a href="README.it.md">Italiano</a>
 </p>
 
@@ -302,7 +302,7 @@ colibrì 最初是由一人使用 12 核心、25 GB RAM 的筆電開發；
 
 - ⭐ 為儲存庫加星並分享；
 - 🐛 以 issue 提交你的硬體 benchmark 數據——實測資料比任何其他事都更能推動專案；
-- 💬 加入 [Discord 社群](https://discord.gg/MAaKtQRc)，討論實驗、硬體數據與研究方向；
+- 💬 加入 [Discord 社群](https://discord.gg/FkyrEeJR)，討論實驗、硬體數據與研究方向；
 - 💬 若想贊助開發或捐贈硬體，請透過 GitHub issues 聯絡。
 
 ## 儲存庫結構
--- a/c/tests/check_data_logprob_gaps.py
+++ b/c/tests/check_data_logprob_gaps.py
@@ -4,18 +4,24 @@
 Asserts, over a captured raw engine-stdout transcript, that EVERY generated
 DATA frame of an opted-in request carries the per-token numeric channel
 ("DATA <id> <n> <lp> <k> [tid tlp]*k") -- accepted draft tokens included.
-Speculatively-accepted draft tokens bypass the pick_tok call sites in the mux
-loop (packet Fork 5's implementation risk), so a gap would show up as a
-legacy 3-field DATA frame in the middle of an opted-in generation.
+On today's engine, an accepted speculative-draft token is emitted through
+the SAME mux_data() call as any other generated token (mux_spec_emit ->
+mux_data, passing the same logits row and requested top-k the ordinary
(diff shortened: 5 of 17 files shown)


monotophic:

Merged current `dev` (`87c54a93`) into this branch — no functional change. Dev's commits since this PR's base never touch any of the 11 files this PR adds or changes; clean auto-merge. This PR's six test modules run to completion with zero skips on the merged head (282 tests), including the release-packer test that exercises `c/tools/pack_python.py`, which dev has changed since this PR opened.

@JustVugg similar steering request to #1102, #1353, #1355, and #1356, this PR is part of the set to make the full checkpoint-quality GLM 5.2/5.3 containers work and to establish the OpenAI API instrumentation needed for container quality studies. Let me know if this PR is off target or otherwise not likely to merge. Thanks again!

JustVugg:
Reviewed, and I am asking for a split rather than a change.

The `check_data_logprob_gaps.py` hardening stands on its own. Failing loudly on an unparsable preamble instead of treating it as no preamble, and citing byte offsets, is a straightforward improvement to an existing tool.

`check_native_mtp_witness.py` is the problem, and it is not a code problem. It matches a per-emission stderr record that nothing in this repository emits. I checked the dev tree directly: `MTPEMIT` appears in zero files, against a positive control that finds `coli_omp_tune` in eleven. The engine's `MTP_DEBUG` prints something else. So those 2,613 lines, tool plus test, can only ever exit 3 UNSUPPORTED on any build that exists, and they are waiting on an engine change that is not proposed anywhere.

That is 62% of this PR. It is the same pattern you identified yourself in #1353 and offered to drop there, and I would like to be consistent about it: send the witness alongside the engine change that emits the record it needs, and the logprob-gap hardening as its own small PR now.

One question about the half that stays: neither transcript tool is referenced by `c/Makefile` or by any workflow. Only their unit tests run. What is meant to invoke them?

monotophic:
Set-level explanation and the full traceability matrix: https://github.com/JustVugg/colibri/pull/1355#issuecomment-5725475017

You were right about the redundancy, and the PR is now the small one you asked for: **11 files and +7,464 → 4 files and +2,319.**

### The witness is parked

`check_native_mtp_witness.py` and its test — 2,613 lines exactly — are out. We confirmed your numbers: `MTPEMIT`
appears in zero files on dev, so those lines could only ever exit 3 UNSUPPORTED on any build that exists, and it is
the same pattern we identified ourselves in #1353. Being consistent about it is the right call.

They are preserved at `monotophic/colibri` `codex/parked-mtp-witness` @ `e0474748`, and **they come back alongside the engine change that emits the record they need** — not before.

What is left is the `check_data_logprob_gaps.py` hardening you said stands on its own.

### "What is meant to invoke them?"

A person (or agent). These are manually-invoked argparse tools, like `check_ablate_evidence.py` — which is why nothing in `c/Makefile` or any workflow references them and why static analysis cannot see a caller. Their unit tests run in CI; the tools themselves are run by hand against a captured transcript when you are diagnosing an engine.

We think that is reasonable simplicity rather than a gap, but if you would rather they were wired into something, say
which and we will do it as a separate change.

### Your asks

|---|---|---|
| "send the witness alongside the engine change that emits the record it needs" | **parked** — un-parks when an engine emits `MTPEMIT` | `monotophic/colibri` `codex/parked-mtp-witness` @ `e0474748` |
| "the logprob-gap hardening as its own small PR now" | **done** — that is now the whole PR, 4 files | this PR |
| "neither transcript tool is referenced by `c/Makefile` or by any workflow … What is meant to invoke them?" | **answered** — human-invoked; offer above | above |

Five further files left this PR — `eval_glm.py`, `test_eval_glm.py`, `check_ablate_evidence.py`,
`test_check_ablate_evidence.py`, `test_pack_python.py` — back to the PR that owns each. That is most of the
de-duplication you asked for on #1355, landing here.

<details>
<summary>Details — origin accounting</summary>

**maintainer-requested:** the witness park; the split.
**re-review-found:** the frame-order lifecycle check in `check()` could have every one of its seven required
ordinals dropped individually with the suite green — it now fails on each. A docstring claiming non-finite logprobs
the docstring now explains WHY a non-finite logprob is a real defect rather than a tolerable value: grammar-constrained decoding never zeroes or `-inf`s a vocabulary logit, so the only path to one is a total-vocabulary numerical collapse that `sample.h`'s own comments call a blow-up upstream. **This is added explanation, not the correction of a false claim** — measured across the two heads, the mentions of grammar/masking went up, not down.
The two transcript-tool suites go **107 tests / 0.64 s → 61 tests / 0.11 s** on this branch (`test_check_data_logprob_gaps.py` plus the parked witness's tests, measured at `e0474748` and `da0c938a`).

The wire-format literals in this tool are hand transcriptions of the C source. Module drift is pinned by tests;
C-source drift is not. Noted as a limitation that predates this revision, with a follow-up in mind.
</details>

Issue #1368, opened 2026-09-06
[Bug]: 1.10.1 Windows Coli Web GLM 5.3-Flash embed_tokens tensor


claws61821:
### Commit or version

8f08dbd

### Environment

Windows 11, Ryzen 7 7800X3D, 128 GB DDR5, NVMe, 1.10.1 release binary

### Reproduction steps

1.10.1 release binary
`.\coli web --model F:\colibri\models\glm53-flash`

### Expected behavior

Colibri successfully launches chat API server and web interface and loads GLM 5.3-Flash.

### Actual behavior and logs

```shell
Colibri crashes with the report

dashboard: http://127.0.0.1:8000/  (opens automatically when the engine is ready)
model.language_model.embed_tokens.weight: tensor 'model.language_model.embed_tokens.weight' is U8/I8 ΓÇö not a float tensor; read it with st_read_raw

and provides the traceback

Traceback (most recent call last):
  File "F:\Colibri\colibri-v1.10.1-windows-x86_64\coli", line 2142, in <module>
    main()
    ~~~~^^
  File "F:\Colibri\colibri-v1.10.1-windows-x86_64\coli", line 2135, in main
    sys.exit(handler(a) or 0)
             ~~~~~~~^^^
  File "F:\Colibri\colibri-v1.10.1-windows-x86_64\coli", line 1912, in cmd_web
    cmd_serve(a)
    ~~~~~~~~~^^^
  File "F:\Colibri\colibri-v1.10.1-windows-x86_64\coli", line 1754, in cmd_serve
    openai_server.serve(a.model,a.host,a.port,model_id,a.api_key,
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          a.cap,ngen_for(a,interactive=True,family=family),engine,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          env,a.cors_origin,
          ^^^^^^^^^^^^^^^^^^
          a.max_queue,a.queue_timeout,a.kv_slots,allowed_hosts=a.allowed_host,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          family=family)
          ^^^^^^^^^^^^^^
[...]

Issue #1365, opened 2026-09-06
[Bug]: coli doctor reports 2 missing core tensors on a healthy GLM-5.3-Flash: the required list is the GLM-5.2 layout


JustVugg:
Reported on Discord, reproduced here.

`coli doctor` fails `model.required` with **"2 required core tensor(s) are missing"** on GLM-5.3-Flash. The reporter hit it from both sources before asking:

- the pre-converted `Justvugg/GLM-5.3-Flash-colibri-int4-g64`
- their own `coli convert` against `zai-org/GLM-5.3-Flash`

Same result, which is the tell: the check never looks at the weights, only at names.

## Cause

`doctor.py` held three literal tensor names:

```python
REQUIRED_CORE_TENSORS = (
    "model.embed_tokens.weight",
    "model.norm.weight",
    "lm_head.weight",
)
```

Those are the GLM-5.2 layout. GLM-5.3-Flash's root is the vision wrapper, so its language model is nested:

```
model.embed_tokens.weight            not present
model.norm.weight                    not present
lm_head.weight                       present

actually present:
  model.language_model.embed_tokens.weight
  model.language_model.norm.weight
```

Two of three names miss, hence exactly "2". Verified against three real checkpoints locally.

**Nothing is missing from either download.** The model runs; the tiny-oracle gate in CI passes on it.

## Scope

`doctor` reports rather than gates, so nobody is blocked: `coli chat` works and every other check is green. The cost is a healthy model being called broken, and in this case someone re-downloading 195 GB to test the theory.

It is not specific to Flash. Every engine discovers its tensor prefix at load time, and `qwen38.c:2392` probes the nested name before falling back to the flat one. `doctor` was the only place that treated the prefix as a constant, so any future family that nests will hit this the same way.

## Fix

Match on ROLES rather than literal names, prefix-agnostically: a token embedding, a final norm outside the layer stack, an output head (or `tie_word_embeddings`). That covers families that do not exist yet without another hand-edit, which is the actual defect here.

PR follows.

JustVugg:
Fixed on `dev` at `4a48f12` (#1366).

Verified against the merged code, not the branch:

```
GLM-5.3-Flash (the reported case)   PASS, no missing roles
GLM-5.3-Flash streaming             PASS, no missing roles
```

Both were reporting 2 missing before.

[...]

Issue #1359, opened 2026-09-05
[Bug]: release packer does not follow the imports of scripts it packages for subprocess launch


monotophic:
### Commit or version

dev @e1efc68

### Environment

Observed/verified on MacBook Pro M5 Max macOS 26.6 with Python 3.12.2
Also observed in the published Linux release archive and CI on ubuntu-latest.

### Reproduction steps

On a clean checkout of dev at e1efc68, from the `c` directory:

    # 1. what the archive would contain
    python3 tools/pack_python.py . /tmp/dist
    # -> 14 Python files reachable from coli (copied)

    ls /tmp/dist/tools/convert_fp8_to_int4.py   # present
    ls /tmp/dist/tools/iq3_pack.py              # No such file or directory

    # 2. the gate does not see it
    python3 tools/pack_python.py . /tmp/dist --check
    # -> 14 Python files reachable from coli (all present); exit 0

    # 3. the consequence, from inside the archive.
    #    This is exactly the pair of lines at tools/convert_fp8_to_int4.py:216-217.
    cd /tmp/dist && python3 -c \
      "import sys,os; sys.path.insert(0, os.path.abspath('tools')); import iq3_pack"
    # -> ModuleNotFoundError: No module named 'iq3_pack'

### Expected behavior

The fix is in #1355 (https://github.com/JustVugg/colibri/pull/1355, "feat(tools): offline ablation-evidence checker and an evidence-bound eval harness"), in the
commit "build(tools): follow the imports of subprocess-launched scripts": it
pushes subprocess-reached scripts onto the same import-following queue as every
other reached file. That PR's test suite fails if the recursion is removed, and
it asserts against the real tree rather than a fixture, because every fixture
case in that module would have passed on dev today.

### Actual behavior and logs

```shell
Authored by Claude Opus in Claude Code, analysis in partnership with @Monotophic

`c/tools/pack_python.py` reaches a file two ways: by following imports out from
`coli`, and by spotting scripts launched as subprocesses, written as
`os.path.join(TOOLS, "x.py")`. A subprocess-reached script is copied into the
archive, but its own imports are not followed, so a module that only such a
script imports is left out.

On dev at e1efc68 this happens once. `tools/convert_fp8_to_int4.py` is packaged
because `coli` launches it. It imports `tools/iq3_pack.py` at line 217, inside
`quant_e8()`, after a `sys.path.insert`. Nothing else in the reachable set names
[...]

PR #1372, opened 2026-09-06
release: 1.10.2

JustVugg:
Version bump to 1.10.2 and the changelog for it. Everything else is already on `dev`.

- `1.10.1` → `1.10.2` in the six places the version lives: `c/version.py`, the four README badges, the site's "Currently shipping" line. The `1.10.1` in `desktop/src-tauri/Cargo.lock` is `httparse`/`hyper`, not ours.
- `CHANGELOG.md` had stopped at 1.7.0. The `[Unreleased]` block was Qwen3.8, which shipped as 1.10.0 on 2026-08-31 per that release's notes, so it is labelled as such; 1.10.1 is recorded as the packaging repair it was; 1.10.2 lists what `dev` accumulated since, grouped by what a user would look for.

The GitHub release text is in the PR's first comment, ready to paste when tagging.

Contract tests (version ↔ site ↔ READMEs) green. The dev→main merge and the tag stay with the maintainer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,59 @@
 All notable changes to colibrì are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/).
 
-## [Unreleased]
+## [1.10.2] — 2026-09-06
+
+Patch release. Three of these fixes answer reports made against 1.10.1 in the
+days after it shipped.
+
+### Fixed
+
+- `coli convert` picks the converter from the checkpoint's `config.json`. It ran
+  GLM-5.2's converter on everything; on GLM-5.3-Flash that quantized the nested
+  embedding and the engine refused the result hours later inside `coli web`.
+  An option the target converter does not take is refused, not dropped (#1368,
+  #1369). The converter also refuses a checkpoint it cannot serve, with a
+  per-family pointer (#1305).
+- `coli doctor` no longer reports two missing core tensors on a healthy
+  GLM-5.3-Flash: it matches roles, prefix-agnostically, instead of GLM-5.2's
+  literal names (#1365, #1366).
+- The release archive ships every file `coli` reaches: `iq3_pack.py`, its grid
+  data file, and `tools/convert_glm53.py`, which had never been packaged
+  (#1359, #1364).
+- The RSS guard counts anonymous memory, not reclaimable page cache, so a mapped
+  container no longer evicts experts to free memory it was not using (#1350).
+- `serve` honors CANCEL while a turn is still running (#1336), and the
+  disconnect scenario is built rather than hoped for (#1329).
+- Metal: bit-exact fp8-e4m3 decode (#1346). Qwen3.6: tokenizer merges in both
+  spellings (#1319). macOS: Homebrew prefixes found when `brew` is off the PATH
+  (#1320). DeepSeek V4 on macOS: real CPU and memory in HWINFO (#1308).
+- `coli doctor` omits the GPU plan for a CPU-only engine (#1322); a missing core
+  tensor explains itself on every fatal path (#1318); GLM-5.3-Flash `--no-think`
+  is the template's lowest effort level, not a shape of ours (#1327).
+
+### Changed
+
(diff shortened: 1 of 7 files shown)


JustVugg:
Patch release. Three of these fixes answer reports made against 1.10.1 in the days after it shipped. No new engine; one behaviour change, opt-in.

## If you hit one of these on 1.10.1, this is the fix

**`coli convert` produced a GLM-5.3-Flash container the engine refused** (#1368). It ran GLM-5.2's converter on everything; Flash nests its language model under the vision wrapper, so the embedding matched no rule, fell through to the generic fallback, and was quantized. The engine was right to refuse it, hours later, inside `coli web`. `coli convert` now reads the checkpoint's `config.json` and picks the converter from it. GLM-5.2's path is unchanged. An option the target converter does not take is refused, not silently dropped (#1369). And `tools/convert_glm53.py`, the right tool for Flash, is in the archive for the first time (#1364).

**`coli doctor` said two core tensors were missing from a healthy GLM-5.3-Flash** (#1365). The check held three literal tensor names taken from GLM-5.2. It now matches roles, prefix-agnostically, so it holds for any family that nests, including ones not written yet (#1366). Nothing was ever missing from those downloads.

**The archive was missing files `coli` reaches** (#1359): `iq3_pack.py` and its grid, reachable only through a subprocess-launched script. The packer follows that edge now, and data files next to a module (#1364).

## Also fixed

- RSS guard counts the memory the process owns, not page cache the kernel can reclaim, so a mapped container no longer evicts experts to free memory it was not using (#1350)
- `serve` honors CANCEL mid-turn (#1336); the disconnect scenario is built, not hoped for (#1329)
- Metal: bit-exact fp8-e4m3 decode (#1346)
- Qwen3.6: tokenizer merges in both spellings (#1319)
- macOS: Homebrew prefixes found when `brew` is off the PATH (#1320); DeepSeek V4 reports real CPU and memory on macOS (#1308)
- `doctor` omits the GPU plan for a CPU-only engine (#1322); a missing core tensor explains itself on every fatal path (#1318)
- GLM-5.3-Flash `--no-think` is the template's lowest effort level, not a prompt shape of ours (#1327)
- The converter refuses a checkpoint it cannot serve, with a per-family pointer (#1305)

- GLM-5.3-Flash `serve` can use the 16 KV slots the engine has; the registry declared 1 (#1283)
- **Security:** `image_url` local reads refuse `..`, and with `COLI_IMAGE_ROOT` set the path must resolve inside it, so an authenticated client of a non-loopback server cannot read arbitrary files through the image API; errors no longer confirm a path exists (#1354, dmoraesrs). CI runs with `contents: read` and third-party actions are pinned by SHA

## Supported checkpoints, stated plainly

- **GLM-5.3 (744B-class)** runs on the `colibri` engine. It is the same base model as GLM-5.2 (Z.ai say so; the configs differ only by the transformers version that wrote them), so the family is now named `GLM-5.2/5.3` rather than announcing one as the other (#1367).
- **DeepSeek V4 Flash REAP-150B** (`puwaer/DeepSeek-V4-Flash-0731-reap-150b`, 85 GB, 132 of 256 experts) loads with the same engine and no conversion (#1310), confirmed by a second user. It is documented, and the banner reports it by its measured geometry, `43L x 132E`, not as 284B (#1371).

## Changed

- qwen36 VRAM tier promotes int8 experts instead of reserving for nothing (#1334)
- ...and the three tier bugs that surfaced with it are fixed: an `is_x` overrun with two or more GPUs, a shutdown that could hang, and a use-after-free on int8 containers (#1339, #1340, #1341; fix #1344 by crichalchemist)
- Kimi K3 stops loading a DSA indexer nothing reads (#1335)
- Opt-in `COLI_MAP_EXPERTS=1`: experts served through a per-shard file mapping (#1325). Off by default; output byte-identical either way

## Docs

- Windows DeepSeek V4 users are led to the release launcher, `coli.cmd`, instead of a source build (#1291)

## Build

- Makefile lists the headers each engine includes as prerequisites (#1284, #1349)
- Site and READMEs carry the eight families with real RAM figures under a contract test (#1302)

PR #1382, opened 2026-09-07
feat(web): reasoning depth selector for GLM (off/low/medium/high/max)

### LINKED ISSUE #1311: [Feature]: Add web toggle for GLM5.x reasoning levels
### Problem

Currently coli web only has a simple on/off switch for reasoning. The new GLM5.3 (flash and regular) no longer supports disabling reasoning and is in part what is causing the bug in #1278.

GLM 5.2 has 3 reasoning modes: "Non-thinking", "High", and "Max".
GLM 5.3 has 3 reasoning modes: "Low", "High", "Max"
Source: https://z-cdn-media.chatglm.cn/prompts-rich-media-resources/5.3-flash-blog/H1hAKXnDMx.png

### Proposed solution

This feature requests asks for the following:

- Add the 3 options for reasoning adjustment on the web interface
- Correctly handles the differences between 5.2 and 5.3 so that when running a 5.3 model there is no option to disable reasoning.

### Alternatives considered

Temperately remove the reasoning toggle when running GLM 5.3 so that reasoning is always enabled to a default value of either "low" or "high". 

### Scope and compatibility

Web interface.

dmoraesrs:
Closes #1311.

The web UI only had a boolean reasoning toggle, but the GLM chat template already accepts a level: `render_chat` maps `reasoning_effort` (low/medium/high/xhigh) onto the words the model understands (Low/Medium/High/Max) when `enable_thinking` is on. So the web client could ask for "some reasoning", never how much.

## What changed

- **Level selector instead of on/off.** The sidebar toggle becomes a selector with `Off / Low / Medium / High / Max`. Each maps onto what the engine renders (`enable_thinking` + `reasoning_effort`).
- **Backward compatible.** `reasoning_effort` is sent only when reasoning is on, so a plain on/off client (and any non-GLM backend) behaves exactly as before.
- **GLM 5.3 constraint honored.** GLM 5.3 cannot disable reasoning, so `Off` is dropped from the choices for those models, and switching to one while `Off` is selected lifts the level to `High` — mirroring the model's own rule instead of sending a level it will ignore. Detection is by model id (`/5\.3/`).
- **i18n.** Level strings added for all five locales (en, de, it, zh-CN, zh-TW).

## Mapping (UI level -> request)

| UI | enable_thinking | reasoning_effort | GLM renders |
|----|-----------------|------------------|-------------|
| Off | false | (omitted) | Non-thinking |
| Low | true | low | Low |
| Medium | true | medium | Medium |
| High | true | high | High |
| Max | true | xhigh | Max |

## Validation
- `npm run build` (tsc -b + vite build) — clean, 0 type errors
- `npm test` (vitest) — 33 passed, including two new cases in `api.test.ts` (reasoning_effort sent when on / omitted when off)

## Notes / open questions for the maintainer
- 5.2-vs-5.3 detection is by model-id substring, the only version signal the web client has today. If the server exposed the family on `/v1/models`, the check could key off that instead — happy to follow that up if preferred.
- The alternative in the issue (drop the toggle entirely on 5.3) is a subset of this: here 5.3 keeps the Low/High/Max choices and only loses `Off`.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -46,6 +46,24 @@ const message = (role: ChatMessage["role"], content: string): ChatMessage => {
   return { id, role, content }
 }
 
+/* Reasoning depth for GLM. The engine renders enable_thinking + reasoning_effort;
+   these levels map onto the words GLM understands (Low/Medium/High/Max), and "off"
+   turns thinking off entirely. GLM 5.3 cannot disable reasoning, so "off" is dropped
+   for it -- the model's own constraint, mirrored in the control rather than sent and
+   silently ignored. */
+type ReasoningLevel = "off" | "low" | "medium" | "high" | "max"
+
+const REASONING_EFFORT: Record<Exclude<ReasoningLevel, "off">, string> = {
+  low: "low", medium: "medium", high: "high", max: "xhigh",
+}
+
+const modelForcesReasoning = (model: string) => /5\.3/.test(model)
+
+const reasoningLevelsFor = (model: string): ReasoningLevel[] =>
+  modelForcesReasoning(model)
+    ? ["low", "medium", "high", "max"]
+    : ["off", "low", "medium", "high", "max"]
+
 export default function App() {
   const { t, locale, setLocale, locales } = useLocale()
 
@@ -61,7 +79,7 @@ export default function App() {
   const [model, setModel] = useState(() => stored(localStorage, "colibri.model", "glm-5.2-colibri"))
   const [temperature, setTemperature] = useState(0.7)
   const [maxTokens, setMaxTokens] = useState(4096)
-  const [thinking, setThinking] = useState(false)
+  const [reasoning, setReasoning] = useState<ReasoningLevel>("off")
   const [cacheSlot, setCacheSlot] = useState(0)
   const [conversations, setConversations] = useState<Record<number, ChatMessage[]>>({ 0: [] })
   const [health, setHealth] = useState<HealthResponse | null>(null)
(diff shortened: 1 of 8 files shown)


dmoraesrs:
Ran the CONTRIBUTING local check (`make check`): the portable CPU build completes and the C unit tests pass; Python stdlib tests pass 669/670. The one failure is `test_resource_plan.test_auto_tune_mtp_off_when_disk_low_hit`, which fabricates multi-GB shards and hit `OSError: [Errno 28] No space left on device` on my box (disk at 99%) — an environment limit, and this change is web-only (it does not touch C or Python). The web gate is green: `npm run build` (0 type errors) and `npm test` (33 passed, incl. two new api.test.ts cases).

dmoraesrs:
Update on validation: re-ran `make check` on Ubuntu 24.04 / gcc 13.3 (matching the Linux CI), on a box with disk to spare. Result: **681 tests, `OK`, 0 failures** — the earlier `test_resource_plan` failure was purely my Mac being at 99% disk, not the change. Build compiles; the only compiler warnings are pre-existing baseline ones in `deepseek_v4.c`/`qwen38.c` (this PR is web-only and touches no C). Web gate remains green (`npm run build` 0 type errors, `npm test` 33 passed).

JustVugg:
Reviewed against the gateway on dev: the model ids are glm-5.3-flash-colibri and glm-5.2-colibri, so the /5\.3/ check hits Flash and leaves the large GLM-5.3 on the 5.2 template, where Off is a real state; reasoning_effort only travels with thinking on; the mapping table matches render_chat. One nit for a follow-up, not a blocker: on 5.3 the server renders medium as High (its template has Low/High/Max), so the Medium entry is a duplicate of High there; reasoningLevelsFor could drop it for models that force reasoning. Merging. Thank you, this is what #1311 asked for.

Issue #1331, opened 2026-09-04
[Bug]: CUDA VRAM expert tier silently no-ops on int8 checkpoints -- reserves budget, promotes zero experts, no error


jmpmachado:
### Commit or version

8f08dbd22a4b80d02402221f1663cc5d4e7aaf89 (current `main` at filing time). First noticed during a downstream fork's full-file audit of `c/qwen36.c`/`c/qwen36_tier.c` at v1.7.0, and independently re-confirmed against this commit before filing.

### Environment

Code-reading finding (static analysis of the control flow below), not a single-machine repro — the defect is in the interaction between two functions and does not depend on host/GPU specifics beyond "any CUDA device + any int8 checkpoint".

### Problem description & steps to reproduce

`COLI_CUDA=1` against an **int8** checkpoint initializes the VRAM tier, reserves per-device VRAM budget, and marks experts as "planned" for promotion — but promotes **zero** experts for the process's entire lifetime, with no error, warning, or log line distinguishing this from a correctly-functioning (but cold) tier.

Root cause is a mismatch between two functions that disagree about when an expert is eligible for VRAM promotion:

- `qt_plan_fill()` (`c/qwen36_tier.c:281-306`) reserves VRAM budget (`G.used[hd]+=G.exp_bytes`) and sets `QSlot.planned=1` **unconditionally** for every candidate expert, based purely on VRAM budget and heat order — it never inspects whether that expert actually has an int4 (`g4`) weight buffer.
- `qt_note_planned()` (`c/qwen36_tier.c:310-315`), the function that actually stages and uploads a planned expert, gates on `if(!G.on || !g4) return;` — so for an int8 checkpoint, where `g4` is always `NULL` (the int4 load path that populates it is only taken for int4 containers — see `c/qwen36.c` around the two load paths), every single "planned" expert silently no-ops here and is never staged, never uploaded, never counted as a failure.

Net effect: the tier reports itself as active (`[qtier] CUDA VRAM expert tier active: ...`), reserves and holds VRAM budget for the whole run, and produces a hit-rate/eviction summary at the end that will simply show 0 promotions — a resource path that reports success while doing nothing, on exactly the checkpoint format (int8) many users run day to day.

[...]

PR #1388, opened 2026-09-07
qwen36 tier: QT_MAX_ROWS, placement pin, ownership-based int8 free (follow-up to #1344)

crichalchemist:
Follow-up to #1344, addressing the review there from @kreuzzelg and @JustVugg. Three commits on current `dev`.

## 1. Review response (original implementation by @mfethe1, crichalchemist/colibri#1)

- **`QT_MAX_ROWS`** replaces the seven literal `32`s in `qwen36_tier.c` that had to agree (`is_k`, the `K>` clamp, `tg/tu/td`, `rows[]`, the topk clamp message, `is_x_floats`, the `qt_issue` stride). #1339 was three of them disagreeing. Since #1344 the `is_x` allocation moved below the auto-placer; the constant follows it there.
- **Placement pin** in `test_qwen36_tier_multidev`: asserts device 0's block starts at `G.is_x` and device 1's at `G.is_x + QT_MAX_ROWS*D`, plus the `sizeof`-based stride check. They detect different drifts: a stride reverted to `8*D` trips the pin (in-bounds and disjointness both still pass on a single-row issue), while widening `is_k` without touching the stride trips only the `sizeof` check. Both kept.
- **Warmstart log line and `docs/qwen36-cuda-tier.md`** say which container they describe: on an int8 container nothing is freed since #1341, and the 40 → 29 GB RSS row is an int4 measurement with no int8 analogue.

The shutdown-test watchdog from the original follow-up is already on `dev` via fa8142d, so this PR does not touch `test_qwen36_tier_shutdown.c`.

## 2. The #1341 free keys on ownership, not on format

@JustVugg's suggestion from the review. `if (!keep8 && expert_is_int4 && e->g)` encoded today's correspondence between weight format and memory ownership; a third format that also aliases `e->g` would reintroduce the use-after-free without anyone reading a format flag as an ownership statement. Now:

```c
if (!keep8 && e->g && wg != (const uint8_t *)e->g) { free(e->g); ... }
```

Do not free what was just handed over. Same behaviour on both current containers (`test_qwen36_tier_int8_engine` covers both), correct by construction for the next one.

## 3. One warning

`cnt` in the placement pricing loop (ff13134) was set and never read; `-Wunused-but-set-variable` flagged it in every test build that includes `qwen36_tier.c`. Removed.

## Verified

macOS 13 x86_64 (2017 iMac, i7-7700K), Apple clang, fake CUDA backend: `make qwen36` and all seven `tests/test_qwen36_tier_*` binaries with 0 warnings, all passing (autoplace, fp8, int8, int8_engine, invariants, multidev, shutdown). Not built on MinGW/UCRT64 locally; the Windows job here is the proof for that.

### DIFF
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -2787,8 +2787,13 @@ static void tier_warmstart(Model *m, int expert_is_int4) {
              * slot_ensure_int8() unable to bring the expert back: it returns
              * early without g4, and the CPU fallback in the decode loop then
              * dereferences NULL. Keep it (#1341). COLI_KEEP_INT8 keeps its
-             * meaning for int4. */
-            if (!keep8 && expert_is_int4 && e->g) { free(e->g); e->g = e->u = e->d = NULL; }
+             * meaning for int4.
+             * The condition is ownership, not format: do not free what was
+             * just handed over. int4 handed g4 (wg != e->g), so the int8
+             * copy is spare; int8 handed e->g itself, so it stays. A future
+             * format that also aliases e->g is then correct without anyone
+             * remembering to extend a format check here. */
+            if (!keep8 && e->g && wg != (const uint8_t *)e->g) { free(e->g); e->g = e->u = e->d = NULL; }
         }
     }
     qt_fill_wait();
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -2793,8 +2793,17 @@ static void tier_warmstart(Model *m, int expert_is_int4) {
     }
     qt_fill_wait();
     free(wpl); free(wpe); free(planned);
-    fprintf(stderr, "[qtier] warmstart (parallel): all %d experts in RAM (int8 only for non-residents), %d in VRAM -- %.1f s\n",
-            cap_total, wn, now_s()-t0);
+    /* The parenthesis used to read "int8 only for non-residents", which was
+     * true only while the int8 copy of every resident was freed. Since #1341
+     * that free is int4-only: on an int8 container every resident keeps its
+     * weights, so the RSS saving the old line implied does not exist there.
+     * Say which container this is instead of promising a saving the reader
+     * will not see. */
+    fprintf(stderr, "[qtier] warmstart (parallel): all %d experts in RAM (%s), %d in VRAM -- %.1f s\n",
+            cap_total,
+            expert_is_int4 ? "int8 copy dropped for residents, kept for non-residents"
+                           : "int8 container: all experts keep their weights in RAM",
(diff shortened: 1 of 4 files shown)


kreuzzelg (approved):
Thanks for taking all three points from the #1344 review; this closes them.

**`QT_MAX_ROWS`.** All seven sites now read the one constant, and the `sizeof`-based check in the multidev test ties `is_k`'s row capacity to the replica stride, which is exactly the pair #1339 let drift. The two placement pins (`records[dev0].x == G.is_x`, `records[dev1].x == G.is_x + QT_MAX_ROWS*D`) are the assertion the original test lacked: in-bounds and disjoint both pass on a single-row issue under the old `8*D` stride, these do not. One literal is left in the test itself, `float val[32]` and its loop just below the new pins; harmless, but since the point of the PR is that the number lives in one place, `QT_MAX_ROWS` there too.

**Ownership free.** `!keep8 && e->g && wg != (const uint8_t *)e->g` is the right predicate: int4 hands `g4` (spare int8 copy, free it), int8 hands `e->g` itself (keep it), and a third format that aliases `e->g` is safe without a format check. The `wg == NULL` case still leaves via the `continue` above it, so `qt_note_planned` has always staged before anything is freed. Correct by construction, as the description says.

**Log line and docs.** Both now say which container they describe; the 40 → 29 GB row is labelled as int4-only. Good.

**On real hardware, two cards** (RTX 3070 + Quadro RTX 4000, driver 580.173, CUDA 12.0), this PR plus #1390 applied on `dev` `6f9117f`: the CI `qwen36_tiny` fixture as an int8 container (`--ebits 8`, `./qwen36 8 8`), `COLI_CUDA=1`, default warmstart: 40/40 runs, 16/16 tokens, `uploads 64 / miss(CPU) 0 / VRAM hit rate 100 %`, banner `int8 container: all experts keep their weights in RAM`. Same fixture as int4 (`--ebits 4`, `./qwen36 8 4`): 20/20, 16/16, banner `int8 copy dropped for residents, kept for non-residents`, `resident 64/64`. All seven `tests/test_qwen36_tier_*` pass. (Caveat as in #1390: `dev`'s warmstart has an unrelated one-in-twelve segfault in `ehit_mark`'s lazy allocation, see #1331; the runs above carry a local lock for it.)

Approve. My open qwen38 tier branch adds to `qwen36_tier.c` too (fp8 mode, dense residents) but not to `qt_issue` or the `is_x` sizing, so rebasing onto this is mechanical; I will do it once it lands.

PR #1395, opened 2026-09-07
Opt-in exact verify mode for speculative batches (COLI_EXACT_VERIFY=1)

rybruscoe:
Follow-up to #689. Opt-in, env-gated, verify batches only, default paths untouched.

**What it does.** With `COLI_EXACT_VERIFY=1`, during draft+verify forwards (`g_spec_live`) the CPU MLA-absorb attention core accumulates its two dots (q.k score and p.v context) exactly: every product is formed from the integer mantissas, binned by exponent, folded into one wide integer and rounded once. Summation order, SIMD width and contraction flags can't change the bits, so a verify row decides a near-tie the same way on every host. Off by default, and the exact branch is a separate `if`; the `else` is the existing code, so the default build path doesn't move. The tq1 / TQ / kv8 context paths are left alone.

**Files.** `c/exact_dot.h` (the accumulator, header-only, needs `__int128`), `c/tests/test_exact_dot.c` (registered in `TEST_BINS`), the two gated branches in `c/colibri.c`, four lines in the Makefile.

**Token-exact check.** glm_tiny oracle, decode mode, `DRAFT=4`, 5 runs each:

| | matching tokens | tok/s (median of 5) |
|---|---|---|
| `DRAFT=4` | 20/20 | 9090 |
| `DRAFT=4 COLI_EXACT_VERIFY=1` | 20/20 | 5657 |

Measured on a box that was also running a training job, so the absolute tok/s wobble run to run; the ratio (about 0.6x) is the number to read. Same tokens either way on this oracle, which is what you'd expect: the exact path only changes bits on near-ties, and the tiny model doesn't produce one in 20 tokens. The number I don't have is a flip caught on the big model at n=64; I don't have the hardware to run GLM-5.2 here, so that check is yours if you want it.

**Cost.** The tiny model is the worst case for the tok/s hit, since attention is most of its forward; on a real MoE the expert matvecs dominate and the gated dots are a small slice. On the dot itself it's about 7x the float loop at -O3 (2.5 ns/elem vs 0.38, from the unit test). `make test-c` passes; no new warnings.

**Unit test covers:** exact vs long-double reference, near-tie rounding (nearest-even), subnormals, cancellation, NaN/inf propagation matching the float classification.

Happy to move the gate name or narrow the scope further if you'd rather.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -478,6 +478,7 @@ TEST_EXCLUDE = test_uring test_deepseek_v4 test_v4_ownership test_v4_serve_frami
 TEST_BINS    = $(addprefix tests/,$(addsuffix $(EXE),$(filter-out $(TEST_EXCLUDE),$(TEST_RULES))))
 ifneq (,$(LINUX))
 TEST_BINS += tests/test_uring$(EXE)
+TEST_BINS += tests/test_exact_dot$(EXE)
 endif
 ifeq ($(COLI_V4_SUPPORTED),1)
 TEST_BINS += tests/test_v4_hybrid_policy$(EXE) tests/test_k3_fill_budget$(EXE) tests/test_v4_bank_pair$(EXE)
@@ -1723,6 +1724,9 @@ tests/test_rss_anon$(EXE): tests/test_rss_anon.c
 # Windows; questo test gira nei tre job e fallisce dove la misura vale 0.
 tests/test_mem_available$(EXE): tests/test_mem_available.c compat.h
 	$(CC) $(CFLAGS) tests/test_mem_available.c -o $@ $(LDFLAGS)
+
+tests/test_exact_dot$(EXE): tests/test_exact_dot.c exact_dot.h
+	$(CC) $(CFLAGS) tests/test_exact_dot.c -o tests/test_exact_dot$(EXE) $(LDFLAGS)
 
 tests/test_798_guards$(EXE): tests/test_798_guards.c st.h json.h compat.h route_trace.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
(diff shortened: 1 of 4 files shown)


JustVugg:
One line away from green, and the change itself is good.

The failure is `test_every_engine_rule_lists_the_headers_its_source_includes` in `c/tests/test_makefile_deps.py`: you added `#include "exact_dot.h"` to `c/colibri.c` but the `colibri$(EXE):` rule does not list it, so a change to your header would not rebuild the engine. Verified on dev: `exact_dot.h` is absent from that prerequisite list.

Append it and this should go green everywhere.

The design is right, and the gating in particular: both branches are `if(exact_verify_on() && g_spec_live)` with the existing code as the else, so the default path is untouched. The cost you report, about 7x on the dot itself, is honest and the sort of number people usually leave out.

Two things worth a line in the docs rather than a change. The `clat` exact branch is additionally guarded against the quantised-KV paths, so it silently falls back there and a user on int8 KV would not get the exactness they asked for. And you say yourself you could not exercise a real near-tie flip on GLM-5.2, which is the honest thing to say and worth keeping in the body.

rybruscoe:
Thanks — that was it. Pushed:

- `exact_dot.h` is now in the `colibri$(EXE):` prerequisite list (`test_makefile_deps` passes locally).
- Rebased onto current `dev` so the branch is just the three commits; engine builds clean, `test_exact_dot` still ALL PASS.
- The two limits you flagged are now in the docs rather than only in this thread: a paragraph under "Speculative decoding, honestly" in the README says the context dot stays on the float path with a quantised KV cache (tq1 / TQ / int8), so `COLI_EXACT_VERIFY` does not provide exactness there, and that a real near-tie flip on GLM-5.2 has been argued but not yet caught. The `clat` branch comment says the same in the code.

CI on the new head is sitting at "action required" (fork run needs your approve-and-run); locally `make check` and the Python suite (1300 tests) are green on the rebased tree.

If you'd rather the quantised-KV case print a one-line notice at startup instead of falling back silently, that's a three-line change — say the word.

PR #1399, opened 2026-09-08
feat(trunk): add TRUNK_RESIDENT_LAYERS=N mmap streaming for the dense trunk

SebaWag:
## Summary

Implements the mmap-view design from #826: **`TRUNK_RESIDENT_LAYERS=N`** — an opt-in knob that keeps the **TOP N dense/attention layers resident** (qalloc) and loads the remaining bottom layers' dense tensors as **read-only mmap views** of the safetensors, reusing the `map_of_fd()` registry that `COLI_MMAP` already uses for experts.

**Unset = byte-identical current behaviour.** This is the "run at all vs run fast" lever the issue asks for: on a host where the dense trunk does not fit in RAM, the mmap'd layers page from disk instead of OOM-killing the process.

## Why only 4 sites (not the ~99 the issue estimated)

The matmul sites never observe *how* a `QT`'s `q8/q4/s` pointers were obtained — the loader hands them a file-backed mmap view and they keep working unchanged:

1. **Loader** — new `qt_load_mmap()` + per-tensor `qt_load_ex()` in the `model_init_range` layer loop (top-N resident, bottom streamed)
2. **Planarize** — skipped on mmap views (`qt_planarize` reorders int4 in place; a read-only mapping cannot be rewritten). Same rule the expert `COLI_MMAP` path already applies
3. **Accounting** — `resident_bytes` uses `qt_rb()` (0 for mmap views), so `cap_for_ram()` / `expert_avail()` hand the freed RAM back to the expert cache budget automatically
4. **Env/guard** — `TRUNK_RESIDENT_LAYERS` parsing (validated ≥ 0) + GPU backends refuse (CPU-only phase 1)

New flags: `QT.mmap_view` (free paths must never free interior mmap pointers) and `Layer.trunk_mmap`.

## Measurement (condition #2 from #826)

Host where the trunk does not fit, reproduced: ARM Neoverse-N1 (Oracle A1), deterministic GLM-MoE fixture from the repo's `make_glm_bench_model.py` tooling (2.25B params, 14 layers, 12 dense `first_k_dense_replace`, `glm_moe_dsa`), real int4 gs64 container. Memory capped via cgroup v2 at 900 MB — below the 1128 MB resident trunk.

| Config | memory cap | resident dense | Result |
|---|---|---|---|
| Baseline (knob unset) | — | 1128.45 MB | loads, decode 0.7 tok/s, cache hit 78% |
| Baseline (knob unset) | 900 MB | *(1128 MB does not fit)* | **OOM-killed** (cgroup `memory.max` events: 350) |
| `TRUNK_RESIDENT_LAYERS=0` | 900 MB | **64.06 MB** | **runs under the cap** |

Load time under the knob: 0.03s vs 0.49s resident (mmap path does not copy weights). Output is **byte-identical** between knob unset and `TRUNK_RESIDENT_LAYERS=0`:

```
knob unset              : 1783 1729 6792 609 1800 4507 2266 209
TRUNK_RESIDENT_LAYERS=0 : 1783 1729 6792 609 1800 4507 2266 209
```

*Full N sweep (n_layers → n_layers/2 → 0) RSS/tok-s curve to follow in a follow-up comment.*

## NUMA note (as requested in #826)

The mmap path loses the `>1MB` NUMA interleave that `qalloc()` applies under `COLI_NUMA`. The resident path keeps it — the top-N resident layers still go through `qalloc()` unchanged.

## Notes

- **CPU-only phase 1.** Metal/CUDA/Vulkan refuse when the knob is set (a backend quietly reading a pointer to a non-resident layer is the #813 class of bug — refuse, never degrade). Widening to GPU tiers can be its own issue.
- MTP row / DSA indexer / embed / lm_head / norms stay resident (phase-1 scope).
- Segment-engine layer ranges: the knob applies to full-model loads; segment loads keep current behaviour.

Closes the design discussion in #826 (implementation now ready for review).


JustVugg:
Reviewed. This is the design from #826 done the way it was asked: opt-in, byte-identical when unset, the matmul sites untouched, the freed trunk handed back to the expert budget through `qt_rb()`, and the cgroup measurement is the right shape (1128 MB resident does not fit in 900 MB and gets killed; 64 MB under the knob runs, same tokens). Two things I checked on my side so they are not open questions: the expert reader's `fadvise DONTNEED` calls are range-scoped, so they do not evict the trunk pages even though dense tensors and experts share shards; and none of the kernels use aligned SIMD loads, so the 4-byte alignment check in `qt_load_mmap` is sufficient.

Three changes before merging, the first is a real bug:

1. **Segment and edge adapters free the views.** `glm_segment_qt_destroy` and `glm_edge_qt_destroy` do `free(tensor->q8); free(tensor->q4); free(tensor->s)`, and `mmap_ok` in `model_init_range` depends only on the layer index and the global, not on whether this is a full-model load. So with the knob set, a segment load gets mmap views and later frees interior pointers of a shard mapping. The description says segment loads keep the current behaviour; the code does not enforce it. Please guard both destroy paths with `if(tensor->mmap_view) return;` (cheap, and it makes the `mmap_view` comment true) and gate the knob on the full-model path so a range load never produces a view in the first place.
2. **The measurement is half delivered.** The table has tok/s for the baseline only. The cost of paging the trunk every token is the number that decides whether anyone turns this on; the N sweep (n_layers, n_layers/2, 0) with tok/s and RSS belongs in this PR, not in a follow-up.

And a test, since the fixture you used (`make_glm_bench_model.py`) already runs in CI: `TRUNK_RESIDENT_LAYERS=0` against unset, identical tokens and a lower `resident_bytes`, gated on the fixture like the other tests. A few seconds, and it is what keeps the next refactor of the loader from silently breaking the view path.

With those in, and the tok/s numbers, I merge it.

SebaWag:
**Measurement — N sweep on a host where the resident trunk does not fit.**

ARM Neoverse-N1, 23 GB RAM. Fixture: GLM-MoE-DSA (14 layers, 12 dense, hidden 4096, int4 gs64, 1.3 GB container). Command `./colibri 4 4 4` (expert cap auto-raised 4->16 since RAM allows), 16 greedy tokens.

**RAM cap (cgroup v2 `memory.max` = 900 MB) — the #826 scenario:**

| config | result | resident dense | cgroup peak | tok/s |
|---|---|---|---|---|
| baseline (knob unset) | **OOM-killed** (SIGKILL, peak hits the cap) | — | 900 MB | — |
| `TRUNK_RESIDENT_LAYERS=0` | **runs** | 64.06 MB | 177 MB | 1.3 |

**N sweep — warm cache (pages stay cached):**

| N | resident dense | tok/s | major faults |
|---|---|---|---|
| 14 (all resident) | 1128.45 MB | 1.3 | 0 |
| 7 | 518.63 MB | 1.4 | 0 |
| 3 | 170.16 MB | 1.5 | 0 |
| 0 | 64.06 MB | 1.4 | 0 |

**N sweep — cold cache (`echo 3 > drop_caches` before each run), i.e. the paging cost made visible:**

| N | resident dense | tok/s | major faults |
|---|---|---|---|
| 14 | 1128.45 MB | 1.4 | 10 |
| 7 | 518.63 MB | 0.7 | 210 |
| 3 | 170.16 MB | 0.5 | 284 |
| 0 | 64.06 MB | 0.5 | 355 |

Reading: with the trunk hot in page cache the knob is essentially free (0 major faults, flat tok/s). Once the pages are evicted the cost is real — 10 -> 355 major faults and decode 1.4 -> 0.5 tok/s. Under a RAM cap there is no tradeoff left to make: the resident path is OOM-killed and the mmap path runs.

Also in this push: the two destroy guards (`glm_segment_qt_destroy`, `glm_edge_qt_destroy`), the knob gated to the full-model path (`allow_trunk_mmap`), the Spanish->Italian comment translation, and the regression test `test_trunk_resident_layers.py` (`TRUNK_RESIDENT_LAYERS=0` vs unset: identical tokens, lower `resident dense`), gated on a quantized fixture in the CI `efficiency` job.

JustVugg:
All four items are in: the destroy guards and the full-model gate close the segment/edge hole, the comments are translated, the test runs in the efficiency job on a real int4 fixture, and the sweep says exactly what the knob costs (free with the trunk in page cache, 1.4 to 0.5 tok/s with the cache dropped, and under a RAM cap the alternative is being killed). That is the honest shape of a run-at-all lever. Merging; #826 closes with it.

PR #1402, opened 2026-09-08
serve: continue a trailing `assistant` turn instead of answering in a new one

### LINKED ISSUE #1401: [Feature]: continue a trailing `assistant` turn instead of answering in a new one
### Problem

I'd like to use a trailing assistant message to supply the beginning of a response and have Colibrì continue it, as I can with llama.cpp. For Colibrì, I think this would be particularly useful to allow resuming assistant answers that were stopped mid-generation.

Currently, the gateway treats that message as a completed turn and appends a new assistant generation cue. The model then starts a new answer, often repeating the opening. For example, on GLM-5.3-Flash int4:

```
messages: [user "What is the capital of France?",
           assistant "The capital of France is"]

today  →  "The capital of France is **Paris**. It's also the country's largest city…"
```

The opening is restated rather than continued. Nothing errors, but I'd expect the returned text to continue the supplied opening, for example ` **Paris**.`.

llama.cpp calls this assistant prefill (`--prefill-assistant`). I'm using *continuation* here to distinguish it from Colibrì's prompt-processing phase, also called prefill.

### Proposed solution

Render the trailing assistant turn as a **past** turn and stop — no generation cue. This is not a new prompt shape, but instead `add_generation_prompt=False`, the other branch of the template's own conditional:

```jinja
{%- if add_generation_prompt -%}
    <|assistant|>{{- '<think>' -}}
{%- endif -%}
```

Every renderer in `openai_server.py` hard-codes that flag to `True`. The proposal is to let one of them reach the `False` branch.

The proposed opt-in is a server setting, `COLI_CONTINUE_ASSISTANT=1`, in the shape of `COLI_THINK` and `COLI_TOOL_SALVAGE`: env-only, default off, one row in `docs/ENVIRONMENT.md`. Clients would only need to send a message list ending in an assistant message; no custom request field is required.

I'm opening the implementation alongside this issue so there's a concrete patch to discuss. I'd be happy to adjust the trigger if you'd prefer continuation to be automatic rather than behind a server setting; the implementation starts with an opt-in for GLM-5.3-Flash because it's the only family I use and can test.

I'd like to see continuation supported across the other families too, but this seemed like a useful first patch. Each handwritten renderer needs its own template comparison and validation; the `glm53` renderer matches the reference template in the added continuation case, and 40 sampled requests on GLM-5.3-Flash produced the expected continuations or clean endings.

### Alternatives considered

- **Continue automatically when the last message is from the assistant.** This is the interface I'd prefer longer term. For this first patch, it would also mean deciding how unsupported families should behave, so I've kept it opt-in.
- **Add a request field, e.g. `x_colibri_prefill`.** This would allow per-request control, but would require handwritten requests or clients with custom support.
- **Use `/v1/completions` with a manually rendered prompt.** This provides a workaround, but requires the client to handle the template and loses the chat gateway's stop policy, tool parsing, and conversation cache-slot routing.

### Scope and compatibility

- Python gateway and documentation changes only; no engine, wire protocol, model format, or platform changes. The default CPU path remains dependency-free, and CUDA code is untouched.
- Initially supports the `glm53` family. With continuation enabled, requests to continue a turn on another family return 400.
- Tool use is excluded from continuation in this patch. Empty assistant content, trailing whitespace, and a missing preceding turn are also rejected.

enitimeago:
## Summary

Fixes #1401.

A request ending in an assistant message now continues that message by default, instead of treating it as a completed turn and appending a new assistant generation cue. This supports resuming interrupted answers or supplying the beginning of an answer through an existing chat client, without a custom request field.

For example, a request ending in:

```text
user:      What is the capital of France?
assistant: The capital of France is
```

can return ` **Paris**.` to append to the supplied opening.

Continuation covers all nine families: GLM-5.3, the base GLM renderer, Qwen3.8, Qwen3.6, OLMoE, DeepSeek V4, DeepSeek V4.1, Inkling, and Kimi K3. It applies to streaming and non-streaming requests on `/v1/chat/completions` and `/v1/messages`. Set `COLI_CONTINUE_ASSISTANT=0` to restore the previous behavior on both endpoints.

## Implementation

The gateway detects a trailing assistant message and renders its final turn open, without a fresh generation cue. Renderers with explicit turn terminators also omit the final terminator; earlier turns retain their existing rendering. GLM continuation only needs the generation cue suppressed.

Continuation requires nonempty assistant text and a preceding turn. It rejects trailing whitespace, tool definitions, and `tool_calls` on the final assistant message with a descriptive 400 response. The whitespace check prevents templates from silently changing the supplied ending, and the tool restriction preserves the parsers' assumption that they receive an assistant turn from its beginning.

The reasoning splitter starts in content mode for a continued answer, so generated answer text is not incorrectly returned as `reasoning_content`. This preserves #1278's GLM-5.3 starting-state rule for new turns.

Kimi constructs its prompt inside the engine. The gateway therefore sends the final assistant turn as a new `C` record in its `K3CHAT1` payload, and `kimi_k3.c` leaves the response open without closing the turn or adding a new cue. The engine rejects any subsequent message record. **Kimi continuation requires the updated gateway and engine together:** an older engine rejects the new record. Ordinary requests retain the existing record format.

## Validation

- [x] `make -C c check`
- [x] CUDA validation: N/A — no CUDA changes.
- [x] Performance measurements: N/A — no performance claims.

The added coverage includes:

| Coverage | Scope |
|---|---|
| Reference-template comparisons | GLM-5.3, Qwen3.8, Qwen3.6, GLM-5.2, and OLMoE; the continuation reference removes the final terminator where necessary. GLM-5.2's comparison covers thinking-off mode only. |
| Literal expected-prompt tests | DeepSeek V4 and Inkling, following their existing renderer conventions in both thinking modes. |
| Earlier-turn regression | All eight string renderers, in both thinking modes; only the final assistant turn is opened. |
| HTTP integration test | A non-streaming OpenAI chat request on `glm53`, using a fake engine; checks the submitted continuation prompt and the returned content/reasoning fields. |
| Kimi C tests | Build and decode the prompt with the tiny tokenizer, compare the resulting XTML with expected strings, and reject message records following the open turn. |

The standalone template scripts require Jinja2 and downloaded reference files. Each records its reference provenance and returns exit code `2` when skipped. The Kimi checks exercise engine-side framing; they are not an independent token-ID comparison against the checkpoint encoder.

The latest targeted check passed all 14 continuation Python tests and the Kimi C test on b7593ee. The subsequent amendment to 3e07979 changes documentation and a docstring only. The five template scripts skipped in that check because Jinja2 was unavailable. These targeted results do not replace a full build and test run for the expanded stack.

### Earlier GLM-5.3 validation

The original GLM-5.3 patch passed `make check` on macOS/arm64 and Debian 13/x86_64, the 26-case template comparison, and the token-exact oracle (32/32 teacher-forcing positions and 20/20 greedy tokens). Those results predate the family extension and Kimi C changes. The Debian baseline also emitted compiler warnings, so those results did not establish a zero-warning build.

Previously reported model runs used GLM-5.3-Flash int4 on CPU with `--gpu none --ram 20 --ctx 32768`, temperature 0.7, and top_p 0.9. Across both API endpoints, 36 sampled sentence/list/code openings continued successfully and four completed openings ended cleanly; all 40 returned no reasoning content.

Those runs used v1.10.2 plus the earlier patch, before the rebase onto #1278, and all measured requests had thinking enabled. They are historical functional evidence, not measurements of the expanded stack. The old disabled Anthropic arm returned a 400 that this PR no longer introduces, so it is not a baseline comparison for the current behavior. The harness and recorded reports are available to share.

I can only run GLM-5.3-Flash with real weights. The other eight families have offline coverage, without model-generation validation; review of their prompt conventions remains especially useful.

## Compatibility

### Additional notes

- Default-on continuation changes how trailing assistant messages are handled on both chat endpoints. `COLI_CONTINUE_ASSISTANT=0` restores their previous behavior, including accepting those messages without the continuation-specific checks.
- Kimi adds an engine-side record type; the other families use their existing prompt transport.
- The default CPU runtime remains dependency-free. No CUDA or model-format changes are included.


enitimeago:
Updated this PR to preserve existing behavior on both `openai` and `anthropic` endpoints when `COLI_CONTINUE_ASSISTANT` is off. The earlier revision also rejected trailing assistant messages on Anthropic `/v1/messages` with the switch disabled; upon further inspection, that behavior change should be considered separately from adding opt-in continuation. I’ve updated the description and rerun the checks.

JustVugg:
Direction settled in #1401: all families (one open-turn branch per renderer, string tests against the vendored templates where we have them) and automatic trigger (trailing non-empty assistant turn = continue, COLI_CONTINUE_ASSISTANT=0 to disable). Tell me there whether you extend this PR or I add the other renderers on top of it; review follows that.

enitimeago:
Updated the branch to cover all eight families and enable continuation by default, as agreed in #1401.

A trailing assistant message with nonempty text continues the supplied turn; `COLI_CONTINUE_ASSISTANT=0` restores the previous behavior on both endpoints. The existing validation checks remain.

The original GLM-5.3 commit changed from e9789ff to da69755 only to document the template reference and return exit code `2` when the template test skips. Its implementation is unchanged. The remaining commits add the automatic trigger, one commit per additional family, and regression coverage.

Kimi is included. Its gateway renderer sends a new `C` record, and `kimi_k3.c` leaves the final assistant response open without appending the closing markers or a fresh generation cue. Kimi continuation therefore requires the updated engine as well as the gateway; an older engine rejects the new record. Ordinary requests keep the existing record format.

The test coverage is:

- Template comparisons for GLM-5.3, Qwen3.8, Qwen3.6, GLM-5.2, and OLMoE. Where the reference closes the final turn, the continuation comparison removes that terminator. GLM-5.2's reference comparison covers thinking-off mode only. These are standalone scripts requiring Jinja2 and downloaded reference files, whose provenance is recorded in each script.
- Literal expected-prompt tests for DeepSeek V4 and Inkling, in both thinking modes, following their existing renderer conventions.
- A cross-family regression checking that earlier assistant turns remain unchanged when the final turn is opened, covering the seven string renderers in both thinking modes.
- A non-streaming `/v1/chat/completions` test on `glm53` using a fake engine, checking that the handler sends the continuation prompt and returns generated text as `content`, without `reasoning_content`.
- Kimi C tests that build and decode the prompt using the tiny tokenizer, compare it with expected XTML, and reject messages following an open turn. These exercise engine-side framing; they are not an independent token-ID oracle against the checkpoint encoder.

Real-model validation remains limited to the previously reported GLM-5.3-Flash runs. I haven't run the other seven checkpoints, and the offline tests don't establish how those models generate from the supplied opening. Review of those family-specific prompt conventions would be especially useful.

JustVugg:
The design is settled and you shipped all eight families the same day, so this is just drift. `c/openai_server.py` has moved 3 commits since (COLI_IMAGE_ROOT, DeepSeek V4.1 tool calling, family registration), `c/kimi_k3.c` 2, `c/tests/test_openai_server.py` 2.

One substantive addition comes with the rebase rather than after it: dev has since registered DeepSeek V4.1 as a ninth family, so it needs its own open-turn branch alongside the eight you already did.

Issue #1401 is still open and nothing on dev continues a trailing assistant turn, so this is wanted as it stands.

enitimeago:
Rebased onto `dev` at `dd9cad8`, preserving the existing eight-family continuation behavior and automatic trigger alongside the upstream dispatch and tool-fallback changes.

Added DeepSeek V4.1 as the ninth family. Its continuation renderer omits the final assistant EOS and fresh generation cue while preserving V4.1's thinking-mode framing. Coverage includes literal expected prompts in both thinking modes and the shared regression checking that only the final assistant turn is opened.

Validation: the C suite — including the Kimi `test_k3_chat_tools` open-turn framing — and the Python continuation suites (`TrailingAssistant*`, the V4.1 expected-string pins, and the cross-family "only the final turn is opened" regression) all pass. GLM-5.3 continuation was separately verified byte-for-byte at explicit `low`, `high`, and `xhigh` effort levels; a new tip commit adds those comparisons.

Real-model validation remains limited to the earlier GLM-5.3-Flash runs. V4.1 has offline prompt coverage only, not how the model resumes from it — a look at the V4.1 convention would be welcome.

enitimeago:
Aside, in case it's of interest:

I prototyped exposing this in the web chat UI. A Continue button on the last assistant message resends the transcript ending on that turn, so the model resumes it instead of answering fresh — the client-side surface for `COLI_CONTINUE_ASSISTANT`. It reuses the existing streaming path (`api.ts` unchanged); the whole thing is ~55 lines across `App.tsx` and the locale files.

Tested end-to-end against a local GLM-5.3-Flash build of this branch, and continuing an already-completed turn correctly does nothing, so the button only shows when the last turn is actually open (hit `max_tokens`, or stopped by hand):

Screenshot 2026-09-21 at 19-17-06 colibrì

Completing the turn then hides the button:

Screenshot 2026-09-21 at 19-26-19 colibrì

It depends on this PR's server side, so I've kept it separate — happy to open a follow-up web PR (or file an issue first, your call) once this lands.

(I also considered whether it might better to replace the disabled send button with the continuation button, rather than exposing the continuation button in the transcript — let me know if you have a preference.)

JustVugg:
Merged current dev into your branch: the one conflict was docs/api.md, where the max_tokens ceiling paragraph (#1641) landed in the same place as your continuation section; both are kept, ceiling first. On the merged tree the 243 gateway and Anthropic tests and test_k3_chat_tools pass here. Merging once CI is green; it goes into 1.12.1 with a CHANGELOG entry, and #1401 closes with it.

PR #1404, opened 2026-09-09
fix(qwen36 tier): offer int8 experts on the decode path (#1391)

### LINKED ISSUE #1391: [Bug]: qwen36 tier: an int8 container never promotes on the decode path (QT_NO_WARMSTART=1 stays at 0 uploads)
### Commit or version

`dev` at f18a54a (also reproduced on the #1338 branch, which is `dev` plus the Vulkan tier).

### Environment

Observed through the Vulkan tier on a Radeon Pro 580 under MoltenVK (macOS 13 x86_64), but the gate is in the engine before any backend is involved, so the CUDA tier is affected identically.

### Reproduction steps

`ci.yml`'s Qwen3.6 tiny fixture, converted at `--ebits 8` (int8 experts), engine built with a tier backend, and a 400-token reference so the run outlives the uploader's first copy (a 16-token run ends before anything could upload in either mode, which hides the difference):

```
python3 tools/make_qwen36_tiny.py --out qwen36_tiny --ref-mode full --emit-ref qwen36_tiny/ref_full.json
python3 tools/convert_qwen36.py --model qwen36_tiny --out qwen36_tiny_c --ebits 8
COLI_DENSE_I8=0 SNAP=qwen36_tiny_c COLI_VULKAN=1 VK_EXPERT_GB=auto COLI_TIMERS=1 QT_NO_WARMSTART=1 ./qwen36 8 8 ref_long.json
```

Then the same fixture converted at `--ebits 4`, same command.

### Expected behavior

With `QT_NO_WARMSTART=1` the tier fills on first use for either container format, as it does for int4, and as the warmstart already does for int8 since #1331/#1334.

### Actual behavior and logs

int8 container, 400 tokens, no warmstart: nothing is ever enqueued, the tier reports itself active and sits at 0 % for the life of the process.

```
[qtier] resident 0/64 experts | uploads 0 | miss(CPU) 6464 | q_skips 0
[qtier] VRAM hit rate: 0.0 % | LFRU swaps 0
```

int4 container, same command:

```
[qtier] resident 64/64 experts | uploads 64 | miss(CPU) 901 | q_skips 126
[qtier] VRAM hit rate: 86.1 % | LFRU swaps 0
```

Cause: both decode-path call sites in `c/qwen36.c` only offer an expert to the tier when it has a packed int4 copy:

```c
if (e->g4) qt_note(layer, idx[kk], e->g4, e->u4, e->d4, e->gs, e->us, e->ds);      // ~line 1969
if (ps->g4) qt_note(lnext, eid, ps->g4, ps->u4, ps->d4, ps->gs, ps->us, ps->ds);   // ~line 2374, prefetch
```

On an int8 container `e->g4` is NULL (there is nothing to pack), so `qt_note` is never called. The warmstart was taught to hand the tier `e->g` in that case (`wg = expert_is_int4 ? e->g4 : e->g`); the decode path was not. Same shape as #1331, one call site over.

Fix shape: mirror the warmstart at both call sites, `expert_is_int4 ? e->g4 : e->g` (the tier accepts int8 pointers since #1334 and refuses grouped-scale int8 at init). A test in the shape of `test_qwen36_tier_int8_engine.c`, driving the decode path on an in-memory int8 model against the fake backend and asserting `fake_uploads > 0`, would fail today and pass with the change. Happy to send that if the reading is agreed.

Petsku01:
Fixes #1391.

## Root cause (verified on dev @ 1ccee43)

The two decode-path offer sites gate on `g4` only:

- `moe()` resident offer (c/qwen36.c:1969): `if (e->g4) qt_note(...)`
- pilot-prefetch lookahead (c/qwen36.c:2374): `if (ps->g4) qt_note(...)`

On an int8 container `e->g4` is NULL (nothing to pack), so `qt_note` is never called. With `QT_NO_WARMSTART=1` the tier reported itself active and sat at 0 uploads / 0 % hit rate for the life of the process. The warmstart was taught the right pointer choice in #1331 (`wg = expert_is_int4 ? e->g4 : (const uint8_t *)e->g`); the decode path was not.

## Fix

One shared decision, three callers:

- `tier_offer_slot()` in c/qwen36.c: offer the packed int4 pointers when present, else — when the container is int8 (`g_expert_is_int4`, set by main's existing nbytes probe, same signal tier_warmstart receives) — the live int8 RAM weights. The tier accepts int8 pointers since #1334 (fmt=1; grouped-scale int8 is refused at init, so int8 containers are always per-row here).
- `moe()` resident offer and the prefetch lookahead both call it; the test calls it too, so the sites cannot drift apart.
- Read-only offers: never frees, never rewrites. The #1341 ownership rules in tier_warmstart are untouched.

## Test: test_qwen36_tier_int8_decode (new, Makefile rule auto-picked by TEST_RULES)

Drives the production `tier_offer_slot` against the fake CUDA backend on an in-memory model, both container formats, `QT_NO_WARMSTART=1`:

| | old gate (`if (g4)`) | with this fix |
|---|---|---|
| test_qwen36_tier_int8_decode, int8 case | **fails 4/5** (no offers, 0 uploads, 0 % resident) | passes 8/8 |
| same test, int4 case | passes | passes (packed g4 pointer still parked) |

The int8 failure on the old gate is exactly the #1391 report, reproduced deterministically. All seven existing tier tests pass, `make qwen36` builds with 0 warnings.

## Review

Two-model blind cross-review (GPT-5.6-Sol + Grok-4.5, different families). Both flagged the same core issue — the test originally mirrored the offer expression in a local copy — fixed by moving the decision into qwen36.c's `tier_offer_slot()` and calling it from both production sites and the test. Grok's doc-drift catch applied: the `qt_note` header comment now names the int8 offer.

## Verification limits

No CUDA hardware here (WSL): verified with the fake-CUDA-backend tier tests (the same pattern the tier test suite already uses; `test_qwen36_tier_int8_engine` proves the engine-level fake path) plus CI's GPU jobs as the hardware gate. The change is offer-policy only: pointer choice before `qt_note`, no GPU-API, thread, or free/ownership changes.

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1711,6 +1711,10 @@ tests/test_qwen36_tier_fp8$(EXE): tests/test_qwen36_tier_fp8.c tests/qwen36_fake
 tests/test_qwen36_tier_int8_engine$(EXE): tests/test_qwen36_tier_int8_engine.c tests/qwen36_fake_cuda.h qwen36.c qwen36_tier.c qwen36_tier.h backend_cuda.h cli_args.h st.h json.h compat.h
 	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
 
+# #1391: il decode path deve offrire gli esperti int8 al tier, non solo il warmstart
+tests/test_qwen36_tier_int8_decode$(EXE): tests/test_qwen36_tier_int8_decode.c tests/qwen36_fake_cuda.h qwen36.c qwen36_tier.c qwen36_tier.h backend_cuda.h cli_args.h st.h json.h compat.h
+	$(CC) $(CFLAGS) -DCOLI_CUDA $< -o $@ $(LDFLAGS)
+
 tests/test_serve_poll$(EXE): tests/test_serve_poll.c serve_poll.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
--- a/c/qwen36.c
+++ b/c/qwen36.c
@@ -955,6 +955,10 @@ static void matmul_q_batch(float *y, const float *x, const int8_t *q,
 /* Group-scaled int8 GEMV: one f32 scale per `gs` input elements per row
  * (gs64 expert containers). Row layout of `scale`: [O][I/gs] row-major. */
 static int g_expert_gs = 0;   /* set from qwen36_meta.json (expert_gs) at load */
+/* 1 = expert container packs int4 (tier fmt=4); 0 = int8 per-row (tier fmt=1).
+ * Same signal main's nbytes probe and tier_warmstart receive; the decode path
+ * needs it to offer int8 experts (#1391): on an int8 container e->g4 is NULL. */
+static int g_expert_is_int4 = 1;
 static void matmul_q_gs(float *y, const float *x, const int8_t *q, const float *scale,
                         int I, int O, int gs) {
     int ng = (I + gs - 1) / gs;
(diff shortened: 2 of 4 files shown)


JustVugg:
Reviewed: the decision lives in one place and both decode sites plus the test call it; the prefetch offer only fires for slots slot_indexed() already returns, i.e. loaded, and the tier's cap == n_experts precondition (#1341 notes) means those slots are never recycled, so the live int8 pointer is as safe here as in the warmstart. The test fails 4/5 on the old gate and passes with the fix, which is the reproduction #1391 asked for. CI green. Merging; thank you for keeping it read-only and for not touching the ownership rules.

Issue #1413, opened 2026-09-10
[Bug]: test_fp8_e2e_repack_load cannot build its harness on Windows (_cc_flags omits -D_FILE_OFFSET_BITS=64 and -fopenmp)


dajiaohuang:
### Commit or version

`173c75758a075bcd350648e7292938fd53f610cb` (`dev`, v1.10.2)

### Environment

Windows 10 Pro N 19045, MinGW-w64 UCRT (WinLibs GCC 16.2.0, posix-seh), Python 3.13.13 with `torch 2.12.0.dev20260408+cu128` and `numpy`. CI's Windows job is MSYS2/UCRT64; the same `_WIN32` path is taken on both.

### Reproduction steps

`torch` must be importable, otherwise the module skips at import and the problem hides:

```sh
python -c "import torch"                                  # must succeed
cd c && python -m unittest tests.test_fp8_e2e_repack_load
```

### Expected behavior

The module's four tests run, or skip with a stated reason. `docs/FORMATS.md:380` and the module docstring both describe this as the pin that the fp8 container declaration stays inert, so on a machine that has `torch` it should be doing that work.

### Actual behavior and logs

```
ERROR: setUpClass (test_fp8_e2e_repack_load.Fp8RepackLoadE2ETest)
AssertionError: e2e harness build failed:
c/compat.h:89:2: error: #error "_FILE_OFFSET_BITS=64 required on Windows (add -D_FILE_OFFSET_BITS=64 to CFLAGS)"
   89 | #error "_FILE_OFFSET_BITS=64 required on Windows (add -D_FILE_OFFSET_BITS=64 to CFLAGS)"
Ran 4 tests in 14.6s
FAILED (errors=1)
```

`setUpClass` compiles `tests/test_fp8_e2e_loader.c` through `_cc_flags()`, which builds its own flag list "closely enough to the Makefile" — but on Windows it passes neither of the two flags `c/Makefile:106` passes there:

```
CFLAGS = -D_FILE_OFFSET_BITS=64 -O3 -march=$(ARCH) -fopenmp -Wall -Wextra ...
```

The `#error` in `compat.h` is deliberate and correct — a 32-bit `off_t` would silently wrap offsets past 4 GB — so the harness cannot be built on Windows without `-D_FILE_OFFSET_BITS=64`.

[...]

PR #1418, opened 2026-09-10
test(v4): make the ExpertStore knobs reach getenv() on Windows

### LINKED ISSUE #1416: [Bug]: test_deepseek_v4 leaves three scratch directories in c/ on every Windows make check (setenv() never reaches getenv())
### Commit or version

`173c75758a075bcd350648e7292938fd53f610cb` (`dev`, v1.10.2)

### Environment

Windows 10 Pro N 19045, MinGW-w64 UCRT (WinLibs GCC 16.2.0, posix-seh). Native
MinGW binary — no MSYS shell involved. CI's Windows job is MSYS2/UCRT64 and takes
the same `_WIN32` path.

### Reproduction steps

```sh
cd c && make check                              # or: make test-c
git status --short
```

### Expected behavior

`make check` leaves the working tree as it found it. `c/tests/test_deepseek_v4.c`
creates its scratch directories with `mkdtemp()` and removes them again at the end
of each of the three ExpertStore tests.

### Actual behavior and logs

Three directories per run, one per test, each holding a `.coli_usage` the test
never asked for — and the suite reports `ok` throughout:

```
$ git status --short
?? c/colibri-v4-pool-a46820/
?? c/colibri-v4-pool-a65640/
?? c/colibri-v4-scaling-a46820/
?? c/colibri-v4-scaling-a65640/
?? c/colibri-v4-store-a46820/
?? c/colibri-v4-store-a65640/

$ ls -la c/colibri-v4-store-a46820/
-rw-r--r-- 1 ... 74  .coli_usage
$ head -3 c/colibri-v4-store-a46820/.coli_usage
-1 1 7
-2 1 1277378958
0 0 3
```

That is a real `route_trace.h` history written by the engine while the test ran.
They accumulate: twelve directories after four runs of the suite here.

### Why

The three ExpertStore tests set the knobs they need with `setenv()`:

```c
setenv("COLI_V4_AUTOPIN", "0", 1);
setenv("COLI_V4_SAVE_USAGE", "0", 1);
setenv("COLI_V4_ROWS16", "0", 1);
```

On Windows `c/compat.h:330` maps that onto `SetEnvironmentVariableA`:

```c
static inline int compat_setenv(const char *name, const char *value, int overwrite){
    if(!overwrite && getenv(name)) return 0;
    return SetEnvironmentVariableA(name, value) ? 0 : -1;
}
#define setenv(name,value,overwrite) compat_setenv(name,value,overwrite)
```

`SetEnvironmentVariableA` updates the Win32 process environment block. `getenv()`
reads the CRT's own copy of it, populated once at startup, and never sees the
change. So all three calls are **no-ops as far as the engine is concerned**:

| knob | read by | what the no-op does |
|---|---|---|
| `COLI_V4_SAVE_USAGE` | `c/deepseek_v4.c:8096` (`hot_usage_save`) | the engine writes `<scratch>/.coli_usage`, the file the switch exists to suppress |
| `COLI_V4_AUTOPIN` | `c/deepseek_v4.c:8843` | the usage history is loaded and pinned at store open |
| `COLI_V4_ROWS16` | `c/deepseek_v4.c:8305` (`hot_rows16_wanted`) | packing uses the rows16 kernel instead of the reference matvec |

The cleanup then loses the evidence:

```c
store->ops->destroy(store);
unlink(path);          /* removes model.safetensors */
rmdir(directory);      /* fails ENOTEMPTY -- .coli_usage is still inside -- and nobody looks */
```

`rmdir` returns `ENOTEMPTY` (41 on UCRT) and the return value is discarded, so the
directory survives and the test reports success.

This is not a new trap in this repo — it is written down twice already, with the
fix:

- `c/tests/test_qwen36_ctx.c:27-33` — *"compat.h maps setenv() to SetEnvironmentVariableA, which updates the Win32 environment block -- but getenv() reads the CRT's own copy and never sees it, so the value under test would silently not arrive."*
- `c/tests/test_inkling_shared_batch.c:16-19` — same, using `_putenv_s`.

Six test files already carry that helper. `test_deepseek_v4.c` was left on the
bare `setenv()` spelling.

The cause is isolated by setting the variable in the **environment** instead
(before the process starts, so the CRT copy has it) — that suppresses the file on
the unpatched binary too:

| on pristine `173c757` | leftovers |
|---|---|
| nothing set | 3 directories |
| `COLI_V4_SAVE_USAGE=0` | none |
| `USAGE_SAVE=0` | none (the second gate, `route_trace.h:340`) |

So this is the test's plumbing, not the engine saving history it should not save.

### Why it is worth fixing

1. Every Windows `make check` adds three untracked directories to `c/`, one per
   pid, and they accumulate — twelve after four runs here. They show up in
   `git status`, and `c/` is the directory the C tests run from.
2. The three tests do not run the configuration they document on Windows. On
   Linux CI (`ci.yml` → `make test-c`, which includes `tests/test_deepseek_v4`)
   the same file runs with `COLI_V4_AUTOPIN=0`, `COLI_V4_SAVE_USAGE=0` and
   `COLI_V4_ROWS16=0` actually in effect, so one commit exercises two different
   configurations depending on the platform — in the tests whose whole point is
   deterministic expert-store accounting.

### Fix

Use the pattern the sibling tests already established — `_putenv_s` on Windows,
`setenv` elsewhere — for the nine call sites, and make the cleanup check its own
`rmdir` so a knob that does not arrive fails the test instead of leaking. PR
follows.

### Scope

`test_deepseek_v4.c` is the file where this has an observable cost, so that is
what the PR touches. These call `setenv()`/`unsetenv()` without the helper and
are presumably the same trap:

```
c/tests/test_deepseek_v4.c          <- fixed in the PR
c/tests/test_olmoe_matmul_q.c       unsetenv("IDOT")
c/tests/test_qwen36_tier_autoplace.c
c/tests/test_qwen36_tier_fp8.c
c/tests/test_qwen36_tier_int8.c
c/tests/test_qwen36_tier_invariants.c
c/tests/test_qwen36_tier_multidev.c
c/tests/test_qwen36_tier_shutdown.c
c/tests/test_v4_ownership.c
c/test_expert_store_registry.c      <- filed separately, with its own evidence
```

I have not measured a consequence for the rest, so I am not guessing at them
here — worth a look separately.

`c/test_expert_store_registry.c` is the same trap but a different failure (it
does not build on Windows at all, and the obvious repair would leave a test that
links and asserts nothing), so it has its own issue and PR rather than being
smuggled in here.

(The two production hits, `c/colibri.c:10794-10810` and `c/inkling.c:2426-2430`,
are not this bug: those values are for the re-exec'd child on Linux and for the
libgomp constructor of a *new* process, and the Win32 block is what a child
inherits. The re-exec is `#ifdef __linux__`, so the Windows path only sets them
for nothing.)

dajiaohuang:
## Summary

Fixes #1416.

On Windows, `make check` leaves three scratch directories behind in `c/`, every run — `colibri-v4-store-<pid>/`, `colibri-v4-pool-<pid>/`, `colibri-v4-scaling-<pid>/`, each holding a `.coli_usage` the test never asked for. Four runs on this machine left twelve of them. The suite reports `ok` throughout.

The three ExpertStore tests set the knobs they need like this:

```c
setenv("COLI_V4_AUTOPIN", "0", 1);
setenv("COLI_V4_SAVE_USAGE", "0", 1);
setenv("COLI_V4_ROWS16", "0", 1);
```

`compat.h:330` maps `setenv()` onto `SetEnvironmentVariableA`, which updates the **Win32** environment block. `getenv()` reads the **CRT's** copy of it, which was populated at startup and is never updated by that call. So on Windows all three knobs are no-ops for the code under test: `hot_usage_save()` (`c/deepseek_v4.c:8096`) does `getenv("COLI_V4_SAVE_USAGE")`, gets `NULL`, and writes `<scratch>/.coli_usage`. The cleanup then loses the evidence, because `rmdir()` on a directory that is no longer empty returns `ENOTEMPTY` and nothing was reading the result.

This trap is already written down twice in this repo, with the fix:

- `c/tests/test_qwen36_ctx.c:27-33`
- `c/tests/test_inkling_shared_batch.c:16-19`

Both define an `env_set()` helper that uses `_putenv_s` on Windows. `test_deepseek_v4.c` was left on the bare `setenv()` spelling.

### The change

Two helpers at the top of `c/tests/test_deepseek_v4.c`, then nine call sites and three cleanups:

1. **`env_set()`** — `_putenv_s` on Windows, `setenv` elsewhere, the same shape as the two files above. All nine `setenv("COLI_V4_*")` calls now go through it, so the three tests run, on Windows, the configuration they already run on Linux CI.
2. **`scratch_remove()`** — unlinks the fixture and **checks the `rmdir`**, failing the test with the errno if the directory survived. This is what makes the next regression of this kind loud instead of silent: a directory that outlives its own test now means a knob did not arrive.

Both are worth having together. `env_set()` alone fixes today's symptom; without the checked `rmdir`, the symptom can only ever be noticed by someone reading `git status`.

The four **error-path** `rmdir(directory)` calls in the same file (lines 818, 833, 862, 1040 on this base) have the same shape but run only after the test has already failed; they are left alone to keep this change to the path that leaks on every run.

## Validation

- [ ] `make -C c check`

Not a CUDA change; no performance claim. The first box is left unticked, for the same reason as my other two PRs on this repository: this machine has no binary named `make` (GNU make ships as `mingw32-make.exe`) and does have `torch` installed, so the suite's number here is not the number CI would print. With a GNU `make` on `PATH`:

| | result |
|---|---|
| pristine `dev` @ `173c757` | `Ran 928 tests … FAILED (errors=1, skipped=68)` |
| this branch | `Ran 928 tests in 203.312s … FAILED (errors=1, skipped=68)` |

The single error is identical on both, and is not this PR's: `ERROR: setUpClass (test_fp8_e2e_repack_load.Fp8RepackLoadE2ETest)` — the Windows CFLAGS bug in that harness, filed as #1413 with its own PR. On CI it does not fire, because the Windows job installs no `torch` and the module skips at import. So the gate differs by zero tests, and I would rather show the number than tick a box it does not support.

### The test itself

Native MinGW build (WinLibs GCC 16.2.0, UCRT64), run in an empty directory with no env vars preset:

**pristine `dev` @ `173c757`** — exit 0, three directories left behind:

```
$ ls
colibri-v4-pool-<pid>  colibri-v4-scaling-<pid>  colibri-v4-store-<pid>
$ ls -la colibri-v4-store-<pid>
-rw-r--r-- ... 74  .coli_usage
```

**this branch** — exit 0, nothing left behind:

```
$ ls
out.txt
```

And end to end, after a full `make check` in each of the four worktrees that were built here:

| worktree | `c/colibri-v4-*` directories left behind |
|---|---|
| pristine `173c757`, two gate runs | 6 |
| `fix/fp8-e2e-windows-cflags`, four gate runs | 12 |
| `fix/glm53-serve-cancel`, one gate run | 3 |
| **this branch, one gate run** | **0** |

Twenty-one directories across the three unpatched trees, none in the patched one — and all three ExpertStore tests still report `ok` inside the gate.

**Negative control**, this branch with `env_set()`'s Windows branch put back to the pre-fix spelling — exit 1:

```
scratch directory colibri-v4-store-<pid> survived its test (errno=41): the engine
wrote into it, so the knobs this test sets did not reach getenv()

FAIL: test_expert_store
```

`errno=41` is `ENOTEMPTY`. The control fails for the right reason and names it.

### Why the two settings are the same test

Setting the variable in the **environment** (before the process starts, so the CRT copy has it) prevents the leftovers on the unpatched binary too:

| on pristine `173c757` | leftovers |
|---|---|
| nothing set | 3 directories |
| `COLI_V4_SAVE_USAGE=0` | none |
| `USAGE_SAVE=0` | none (`route_trace.h:340`, the second gate) |

That is what isolates the cause to the `setenv` spelling rather than to the engine saving history it should not save — and it is why the fix is in the test and not in `hot_usage_save`.

## Compatibility

Test-only change; no production code is touched. `_putenv_s` is the CRT function the two sibling tests already use on Windows, and the non-Windows branch is the same `setenv()` the file used before.

---

### DIFF
--- a/c/tests/test_deepseek_v4.c
+++ b/c/tests/test_deepseek_v4.c
@@ -5,6 +5,7 @@
 #include "../native_quant_fp4_rows16.h"
 
 #include <assert.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
 #include <math.h>
@@ -14,6 +15,36 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+
+/* compat.h maps setenv() onto SetEnvironmentVariableA, which updates the Win32
+ * environment block -- but getenv() reads the CRT's own copy of it and never
+ * sees the value.  The three ExpertStore tests below set knobs that the engine
+ * reads with getenv(), so on Windows the setenv() spelling was a no-op and they
+ * ran with the very defaults they meant to override.  The same trap is spelled
+ * out in test_qwen36_ctx.c and test_inkling_shared_batch.c; _putenv_s updates
+ * the copy getenv() reads. */
+static void env_set(const char *name, const char *value) {
+#ifdef _WIN32
+    _putenv_s(name, value);
+#else
+    setenv(name, value, 1);
+#endif
+}
+
+/* mkdtemp() hands out a scratch directory, the engine appends <snap>/.coli_usage
+ * to it whenever it saves its expert history, and rmdir() then fails on a
+ * directory that is no longer empty.  Nothing was listening to that failure, so
+ * a knob that did not arrive left the directory behind and the test still said
+ * ok.  Removing the fixture and checking the rmdir turns that back into a
+ * failure. */
+static int scratch_remove(const char *directory, const char *fixture) {
+    unlink(fixture);
+    if (rmdir(directory) == 0) return 0;
+    fprintf(stderr, "scratch directory %s survived its test (errno=%d): the "
+                    "engine wrote into it, so the knobs this test sets did not "
+                    "reach getenv()\n", directory, errno);
+    return 1;
+}
 
 /* ==== begin test_deepseek_v4_attention_cache.c ==== */
 /* umbrella headers */
@@ -611,9 +642,9 @@ static int test_expert_store(void) {
     /* Native MinGW binaries do not resolve the MSYS /tmp mount. */
     char directory[] = "colibri-v4-store-XXXXXX";
     char path[256], error[256];
-    setenv("COLI_V4_AUTOPIN", "0", 1);
-    setenv("COLI_V4_SAVE_USAGE", "0", 1);
-    setenv("COLI_V4_ROWS16", "0", 1);
+    env_set("COLI_V4_AUTOPIN", "0");
+    env_set("COLI_V4_SAVE_USAGE", "0");
+    env_set("COLI_V4_ROWS16", "0");
     if (!mkdtemp(directory)) { perror("mkdtemp"); return 1; }
     snprintf(path, sizeof(path), "%s/model.safetensors", directory);
     if (write_fixture(path) != 0) { perror("write_fixture"); return 1; }
@@ -780,8 +811,7 @@ static int test_expert_store(void) {
         return 1;
     }
     store->ops->destroy(store);
-    unlink(path);
-    rmdir(directory);
+    if (scratch_remove(directory, path)) return 1;
     puts("DeepSeek-V4 ExpertStore tests: ok");
     return 0;
 }
@@ -808,9 +838,9 @@ static int test_expert_store_prefill_pool(void) {
     enum { LAYERS = 2, EXPERTS = 8, SLOTS_PER_LAYER = 6 };
     char directory[] = "colibri-v4-pool-XXXXXX";
     char path[256], error[256];
-    setenv("COLI_V4_AUTOPIN", "0", 1);
-    setenv("COLI_V4_SAVE_USAGE", "0", 1);
-    setenv("COLI_V4_ROWS16", "0", 1);
+    env_set("COLI_V4_AUTOPIN", "0");
+    env_set("COLI_V4_SAVE_USAGE", "0");
+    env_set("COLI_V4_ROWS16", "0");
     if (!mkdtemp(directory)) { perror("mkdtemp pool"); return 1; }
     snprintf(path, sizeof(path), "%s/model.safetensors", directory);
     if (write_fixture_layers(path, LAYERS, EXPERTS)) {
@@ -956,8 +986,7 @@ static int test_expert_store_prefill_pool(void) {
         failed = 1;
     }
     store->ops->destroy(store);
-    unlink(path);
-    rmdir(directory);
+    failed |= scratch_remove(directory, path);
     if (failed) return 1;
     puts("DeepSeek-V4 ExpertStore prefill pool: ok "
          "(A/B bytes 816->408, second sweep=0 reads, warm entries + decode reserve retained)");
@@ -1030,9 +1059,9 @@ static int run_expert_miss_scaling_case(const char *directory, int experts,
 static int test_expert_store_miss_scaling(void) {
     char directory[] = "colibri-v4-scaling-XXXXXX";
     char path[256];
-    setenv("COLI_V4_AUTOPIN", "0", 1);
-    setenv("COLI_V4_SAVE_USAGE", "0", 1);
-    setenv("COLI_V4_ROWS16", "0", 1);
+    env_set("COLI_V4_AUTOPIN", "0");
+    env_set("COLI_V4_SAVE_USAGE", "0");
+    env_set("COLI_V4_ROWS16", "0");
     if (!mkdtemp(directory)) { perror("mkdtemp scaling"); return 1; }
     snprintf(path, sizeof(path), "%s/model.safetensors", directory);
     if (write_fixture_experts(path, 256)) {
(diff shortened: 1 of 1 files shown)


JustVugg:
All four are merged: #1415, #1418, #1419 and #1414. Thank you for the quality of this batch. Each report named the exact line, the mechanism (the CRT copy of the environment that SetEnvironmentVariableA never touches, the rmdir that failed with nobody listening, the harness flags that only diverged from the Makefile on Windows), why CI could not see it, and a fix that matched the tree's own conventions; and the glm53 CANCEL work reads the serve protocol as carefully as anyone has, CANCEL and STOP kept apart with a test for each. This is the kind of contribution that makes the Windows build a first-class platform rather than a port.

One follow-up your three reports point at, which I will take unless you want it: compat.h's setenv shim should use _putenv_s (it updates both the CRT copy and the Win32 block), so the six per-test helpers become unnecessary and the next test cannot fall into the same trap.

PR #1425, opened 2026-09-10
Add GGUF general reader and converter into OlMoE safetensors files

zameo94:
## Summary
This feature add a general GGUF `reader` and a `converter` that allow gguf's owner to run gguf file into Colibrì, after the conversion into safetensors files. The feature has been builded uses this [OlMoE gguf](https://huggingface.co/allenai/OLMoE-1B-7B-0125-Instruct-GGUF/blob/main/OLMoE-1B-7B-0125-Instruct-Q2_K.gguf) as example.

I know that Colibrì already has the possibility to use this LLM with safetensors. The hope is that this feature can be a pioneer integration, opening the door of GGUF into Colibrì. For that reason, I didn't implement a pure C GGUF reader, but a compromise, since the GGUF is still a standard in local LLMs' world.

The reader and all the converter files are wrapped into a CLI written in Python.

## Motivation

OLMoE and most other open weights are widely distributed as GGUF files (llama.cpp ecosystem), while Colibri consumes safetensors containers. Today the only way in is to start from the official HF safetensors and run a converter; a user who already has a GGUF has no path.
This PR adds an offline GGUF importer that turns a GGUF OLMoE checkpoint into a normal Colibri container, so c/olmoe.c reads it unchanged through st.h. No changes to the C engine or the runtime: the output is a standard container (model-*.safetensors + config.json + tokenizer.json).
The change is layered so the generic parts are reusable for other families and only the model-specific mapping lives per architecture

## Design
The conversion is a pipeline of four stages; only the first two are GGUF-specific, which is what makes the reader reusable for other families.

1. `Read the container (structure).` The GGUF header, metadata key/values, and the tensor index (name, dimensions, type, byte offset) are parsed and validated. Nothing is decoded yet: this stage only locates each tensor and confirms the file is well-formed, so a malformed or hostile file is refused before any data is touched.
2. `Decode the numbers.` Each tensor's raw bytes are turned back into float32. Plain storage types (F32/F16/BF16) map directly; block-quantized types (Q2_K/Q3_K/Q4_K/Q5_K/Q6_K/Q8_0) are de-quantized block by block using the scales stored in each block. An unsupported type is refused loudly rather than mis-decoded.
3. `Map and reshape to the engine's layout.` GGUF stores dimensions reversed and uses llama.cpp naming; this stage renames tensors to what the engine expects and reshapes the flat data back into the original **[out, in]** orientation. Routed experts are sliced out of the fused per-layer tensors, re-quantized row-wise to int8 with the exact math the engine's kernel uses, and merged into one packed weights blob plus a per-row scale array.
4. `Emit the Colibri container.` The resulting tensors are written as **safetensors shards**, alongside a **config.json** reconstructed from the GGUF metadata and a **tokenizer.json** rebuilt from the embedded tokenizer data. Colibri then reads that directory exactly as it reads any other container: **st.h** indexes the shards, the engine streams the experts from disk, and no C code changes.

Everything happens offline, once, before Colibri ever runs. The GGUF file itself is left untouched unless the user explicitly asks to delete it after a successful conversion.

## Architecture
`The code has been written entirely by LLMs (Deepseek V4 Flash and Kimi K3). However, the idea, technical realization, the system design, design implementation, tests by code and manual, and general direction were under human (me :)) control and review `

GGUF file
   │
   ├── `gguf_reader.py`        generic: header, metadata, tensor index, raw bytes,
   │                         bounds/format validation  (new in this PR: all 374 lines)
   ├── `gguf_dequant.py`       generic: decode ggml block types -> float32
   └── `gguf_olmoe_profile.py` OLMoE-only: name mapping, layout, config,
                             int8 expert merge, tokenizer reconstruction
   │
`convert_gguf_to_olmoe.py`     single user-facing CLI orchestrating the above
   │
container for c/olmoe.c (via st.h, untouched)

**Input formats (GGUF) we decode to float32**

Raw types (1 element = N bytes, direct decode):

| type | meaning | bytes/element |
|---|---|---|
| `F32` | 32-bit float | 4 |
| `F16` | 16-bit float | 2 |
| `BF16` | bfloat16 | 2 |

Block-quantized types (dequantized to f32):

| type | nominal bits/value | block | bytes/block |
|---|---|---:|---:|
| `Q2_K` | ~2.5 | 256 values | 84 |
| `Q3_K` | ~3.4 | 256 | 110 |
| `Q4_K` | ~4.5 | 256 | 144 |
| `Q5_K` | ~5.5 | 256 | 176 |
| `Q6_K` | ~6.6 | 256 | 210 |
| `Q8_0` | ~8.5 | 32 | 34 |

**Output formats (Colibri container)**

| location | stored type | notes |
|---|---|---|
| 1-D dense (norms, router…) | `F32` | 4 bytes/elem |
| 2-D dense (attention, embed, head) | `F16` | 2 bytes/elem |
| expert `merged_weight` | `I8` | row-wise int8, `gate|up|down` concatenated |
| expert scales `.qs` | `F32` | one scale per row (`inter+inter+hidden`) |

**Recognized but not yet decodable**

`Q4_0`, `Q4_1`, `Q5_0`, `Q5_1`, `Q8_1`, `Q8_K`, the `IQ`/`TQ` families and
`MXFP4`: the reader knows their geometry in its type table but `gguf_dequant`
refuses them loudly instead of mis-decoding. (`Q8_0`, `Q4_K` and `Q5_K` are
decoded, and the live cross-check against llama.cpp's `gguf-py` passes at zero
tolerance.)

## Converter CLI
`python3 c/tools/convert_gguf_to_olmoe.py --input model.gguf --output /path/to/container`
- --input: a .gguf file or a directory holding exactly one .gguf (single-file; multi-shard is a follow-up).
- --output: created with missing parents.
- --dry-run: prints the plan, writes nothing, creates no directory.
- --remove-source-file: deletes the source .gguf only after a fully successful conversion.
- --overwrite: rebuild an existing container from scratch. Without it, the run refuses when --output already holds model-*.safetensors — it never silently skips or overwrites work.
- An incomplete expert trio (a layer missing gate/up/down) is refused in the pre-flight, before `config.json` and `tokenizer.json` are written.
- --norm-topk-prob, --qk-permuted: the two GGUF-absent switches described below (defaults target the official OLMoE checkpoints; --qk-permuted is off because the official OLMoE GGUF is already in the HuggingFace layout).
- --flush-every N: output shard flush size (default 512), same knob as convert_olmoe_merged.py.

## Validation on the real checkpoint
allenai/OLMoE-1B-7B-0125-Instruct Q2_K (~2.5 GB): every checked tensor matches the official HF safetensors within quantization error, and the engine produces correct answers.

The q/k RoPE layout was verified directly against the official safetensors. Running the converter's dense path twice — `--qk-permuted` off and on — against `allenai/OLMoE-1B-7B-0125-Instruct` (BF16, layer 0), from `OLMoE-1B-7B-0125-Instruct-Q2_K.gguf`:

| tensor | `--qk-permuted` off (default) | `--qk-permuted` on |
|---|---|---|
| `model.layers.0.self_attn.q_proj.weight` | 4.70e-02 | 2.50e-01 |
| `model.layers.0.self_attn.k_proj.weight` | 3.49e-02 | 3.00e-01 |

With the flag off the residual is just the source Q2_K quantization error (baseline on tensors that are never permuted: `v_proj` 1.78e-2, `o_proj` 4.09e-2); with the flag on it grows 5-9x. The official OLMoE GGUF is therefore already in the HuggingFace layout, because llama.cpp's `OlmoeModel` converter does not permute `q_proj`/`k_proj` (only the dense `OlmoModel` does) and `LLM_ARCH_OLMOE` runs NEOX RoPE, the same half-offset pairing as HF `rotate_half` and `c/olmoe.c`'s `rope_head`.

## Known limitations
- Input is a single .gguf; sharded GGUF (multi-file) is not handled yet.
- Dequantization covers Q2_K/Q3_K/Q4_K/Q5_K/Q6_K/Q8_0 + the raw types (F32/F16/BF16); Q4_0, Q4_1, Q5_0, Q5_1, Q8_1, Q8_K, the IQ/TQ families and MXFP4 are recognised by the reader's type table but not yet decodable — the converter refuses rather than guess.
- norm_topk_prob is not encoded in GGUF; the default (false) targets the official OLMoE checkpoints, with --norm-topk-prob for the alternative. q/k are taken as stored: the official OLMoE GGUF is already in the HuggingFace layout, because llama.cpp's OlmoeModel converter does not permute q_proj/k_proj (only the dense OlmoModel does) and LLM_ARCH_OLMOE runs NEOX RoPE, the same half-offset pairing as HF rotate_half and c/olmoe.c's rope_head. --qk-permuted exists only for a GGUF produced by a permuting converter.
- Quality is bounded by the source quantization (a Q2_K import is a Q2_K model).
- Expert re-quantization is second-generation (GGUF → float → int8); measured extra error ≤ 0.39% of row-max, dominated by the source GGUF.

## Screenshoot
CLI example
cli_example

Colibrì running the converted GGUF file
colibri_running_con_file

## About .npz binary file
The dequantization and quantization tests are pinned by committed NumPy golden fixtures: `c/tests/fixtures/gguf_dequant_golden.npz` (deterministic random raw blocks for `Q2_K`/`Q3_K`/`Q4_K`/`Q5_K`/`Q6_K`/`Q8_0` plus the float32 output produced by llama.cpp's `gguf-py` reference) and `c/tests/fixtures/olmoe_quantize_row_golden.npz` (deterministic inputs plus the int8 weights/scales from the torch reference `convert_olmoe_merged.py`). Both run with only numpy. A dedicated CI job (`gguf`) installs numpy, safetensors and `gguf` and runs all four GGUF test modules, so the numerical evidence actually executes instead of skipping; the live cross-checks against `gguf-py` and the torch reference additionally run where those packages are present. torch is deliberately not installed in CI, so the job stays lightweight.

## Validation

- [x] `make -C c check`

## Compatibility


JustVugg:
Thank you for this, and for the write-up: the design section made the review a
lot faster than the diff would have on its own. I read the whole thing and ran
it locally before answering. The short version is that the decomposition is the
right one and I want this in the project. Below is what I verified, what I would
like changed first, and where I think it goes next.

## What I checked myself

- The four test modules pass here, 67 tests, and CI on this branch is green
  across the matrix.
- The regression surface is zero: ten files, all new, not one line of existing
  code touched, no C.
- `gguf_reader.py` reads the way a parser of untrusted input has to. Every read
  is bounds-checked against the real file size, counts and string lengths are
  capped, strings are UTF-8 validated, every tensor range is verified to lie
  inside the file, and an unknown ggml type is refused rather than mis-sized.
  This was the part I was most prepared to be unhappy with, and I am not.
- Pinning the dequantizers to a golden produced by the reference implementation,
  and re-comparing live when `gguf-py` happens to be installed, is exactly the
  standard we hold our own oracles to. A 12 KB npz of plain arrays is fine.
- `restore_rope_layout` is right, including the part that is easy to get wrong:
  the permutation is not an involution, and you say so in the docstring instead
  of leaving it to be discovered.

On the authorship note: what matters here is that the result is reviewable, and
it is. The comments explain why rather than what, which is more than I can say
for a lot of hand-written conversion code.

## What I would like before merging

**1. The quantizations people actually have.** The set today is Q2_K, Q3_K,
Q6_K. Missing are Q4_K and Q5_K, which is what most published GGUFs are, and
Q8_0, which is nearly free to add. The good news is that the K-superblock
machinery you already wrote is most of the work: the 6-bit scale unpacking is
shared, so each new type is a small decoder plus a golden. I would rather ship
the importer once it covers what a user is likely to be holding.

**2. The price on disk, written down.** The output stores experts as row-wise
int8, which is 8 bits per weight against roughly 2.9 for a Q2_K source. On this
machine the existing OLMoE int8 container is 7.0 GB and the Q2_K GGUF is 2.5 GB,
so an import is about three times the file it came from, at the quality of the
file it came from. That is a fair trade for what colibri gives back, but it is
the kind of thing a user should read in the docs rather than discover from
`df`.

**3. One front door.** Conversion here is driven by `coli convert` and the
family registry, and a standalone script is a second entry point that will drift
from the first. I would like this to become something like `coli import`, with
the profile chosen from `general.architecture` rather than from the script name.
While you are there: have the profile validate its own output through
`resolve_model()` before reporting success, so a container that the planner
cannot resolve fails at conversion time instead of hours later.

**4. The test that is missing.** There is already an OLMoE tiny oracle in CI.
If the test generated a small synthetic GGUF, converted it, and ran that same
oracle on the resulting container, the claim "the engine reads this" would be
checked by CI rather than by a screenshot. That would make this the best-tested
converter in the tree.

**5. Two small ones.** A metadata array nested inside itself recurses without a
depth cap: a 60 KB file with 5000 levels raises `RecursionError`, which escapes
the `GGUFError` contract the rest of the reader keeps so carefully. A depth
argument on `_read_metadata_value` fixes it. And the importer needs a page in
`docs/`, since it is a new user-facing path.

## Where this goes

You called it a pioneer integration and I agree with the framing, so here is
where I think the road leads, in case you want to keep walking it.

**Formats.** Mechanical, and your structure already absorbs it: one isolated
decoder and one golden per type. The IQ family is the tedious one because of the
codebook tables.

**Profiles.** This is the real work, but less than it looks: llama.cpp
normalizes names into `blk.N.*`, so attention, FFN and norms can share a base
map and a family only declares its own oddities. The fragile part is not the
weights, it is the tokenizer. GGUF does not carry the pre-tokenizer
configuration, only a name like `llama-bpe`, and the profile currently writes a
fixed ByteLevel, which is correct for OLMoE and will not be in general. The rule
that scales is: use the original `tokenizer.json` when the user has it, and
treat reconstruction as the fallback with its limits stated.

**Reading GGUF in place.** Converting cuts against what this engine is for. We
stream experts off the disk without a conversion step, and an import costs the
user a second copy and a second generation of quantization error. Two things
make the direct path realistic:

- MXFP4 is already ours. A GGUF MXFP4 block is 32 values in 17 bytes, one e8m0
  scale byte plus sixteen nibble bytes on the same e2m1 grid that `quant.h`'s
  `matmul_mxfp4` already consumes for Kimi K3 and for DeepSeek V4.1's experts.
  Between their layout and ours there is no arithmetic at all, only a byte
  shuffle into two planes. An MXFP4 GGUF can be imported with zero loss today
  and read in place tomorrow.
- A `gguf.h` next to `st.h`. If the engine can index a GGUF the way it indexes
  safetensors, the user points colibri at the file they already have. No copy,
  no requantization. It needs a k-quant kernel to be general, but for MXFP4 the
  kernel exists right now.

Your reader and your dequantizer are needed on both roads, which is the main
reason I want this merged rather than redirected.

None of the five points above is a rewrite, and I am happy to take 5 myself if
you would rather not. Tell me which ones you want to pick up and I will stay out
of your way on those.

zameo94:
Thanks for your comment.
It's cool you did talk about real reader with `gguf.h`, that was my first idea and approach. I didn't continue in that way because I wanted to be sure that GGUF implementation was in the future of Colibrì project, so I didn't want to create new C engines e push a lot of files without knowing that, I came with this converter compromise for that reason.

Of course, that implementation will be the natural evolution of the GGUF implementation in this project, making the things more serious, easy to use and hopefully appreciated from the community.

About the five points you mentioned:

I've picked up two of the five points; here's where things stand and what I'm passing back to you.

**Done:**

Point 1 — Q4_K, Q5_K, and Q8_0 are now in gguf_dequant.py (each a small decoder over the existing K-superblock scale machinery), covered by the live golden reference in CI. The importer now handles the quantizations people actually hold.

Point 5a — The metadata reader now caps array nesting (MAX_METADATA_DEPTH=64) instead of letting a deep array drive a RecursionError, so it stays inside the GGUFError contract. Added a refusal test.

**Passing these back to you:**

Point 3 (the coli import front door, profile chosen from general.architecture, and output validated through resolve_model()) — This depends most on the registry's conventions and on how you want the front door shaped. Rather than guess at the API and produce code that drifts from your intent, I'd rather you take it.

Point 4 (the missing test) — Same as point 3: I couldn't locate a dedicated "OLMoE tiny oracle" job in CI, so I'm handing this back to you to implement or wire up to the right workflow, instead of guessing, with the risk of adding more caos.

Documentation:

Point 5b (docs page for the importer) and point 2 (it's a doc, so I would include this in the same point) — This should describe the coli import front door as it will actually work and the disk cost on paper. Once the rest of the implementation is completed and the importer's shape is settled, I'd be happy to write and implement the documentation myself, if you don't want to take it.

Let me know if there's anything else you need on the two completed points!

JustVugg:
Reviewed. The parser is genuinely careful: every read is bounded against the file size, every tensor range and offset alignment is validated before any payload is touched, and the caps on counts, rank and string length are all present. The dequantisers were checked against the reference implementation by hand, including the Q3_K details that are easy to get wrong, and on a machine with gguf-py installed the live cross-check passes at zero tolerance for all six types. `gguf_fixture.py` being an independent writer makes the reader tests a two-implementation cross-check rather than a tautology.

Three things in the body do not match the code, and one of them is output-critical.

The body says Q8_0, Q4_K and Q5_K are refused loudly. They are not: all three are registered and correct. The body understates the code, which is the good direction, but it should be fixed.

The architecture section annotates `gguf_reader.py` as already in dev. It is not; this PR adds all 374 lines.

And the code contradicts itself on RoPE. `gguf_olmoe_profile.py` says llama.cpp's OLMoE converter permutes q_proj and k_proj like Llama so the permuted tensors must be restored, while `convert_gguf_to_olmoe.py` leaves `--qk-permuted` off by default with help text saying the official GGUF is already in the HuggingFace layout. Both cannot be true, and if the docstring is right the default conversion emits wrong q and k and the model is quietly degraded with no diagnostic.

So the question before anything else: which is correct for the official OLMoE GGUF? Please show the check, for example the maximum absolute difference between a converted `model.layers.0.self_attn.q_proj.weight` and the official safetensors tensor, with and without the flag.

One small change regardless: move the expert-trio completeness check out of `convert_experts` and into the pre-flight beside the existing missing-name check, so an incomplete GGUF is refused before config.json and tokenizer.json are written rather than after.

And a note on what actually gates: CI installs no numpy, so upstream only `test_gguf_reader.py` runs. The dequant pin, the profile tests and the converter end-to-end all skip. A job that installs numpy, safetensors and gguf would make the numerical evidence real.

zameo94:
Here is my latest work.

About the check you asked me to share. I ran the converter's actual dense path twice, with --qk-permuted off and on, against the official safetensors (allenai/OLMoE-1B-7B-0125-Instruct, BF16), using OLMoE-1B-7B-0125-Instruct-Q2_K.gguf:

| tensor | --qk-permuted off | --qk-permuted on |
|---|---|---|
| layers.0.self_attn.q_proj.weight | 4.696655e-02 | 2.503662e-01 |
| layers.0.self_attn.k_proj.weight | 3.491211e-02 | 3.004150e-01 |

With the flag off the residual is just Q2_K quantization error (baseline on tensors that are never permuted: v_proj Q3_K = 1.78e-2, o_proj Q3_K = 4.09e-2). With the flag on it grows ~5-9x, which is the permutation. Moreover stored vs permute(official) equals restored vs official exactly (0.2503 / 0.3004), proving the stored tensor is unpermuted. So the official GGUF is already in the HuggingFace layout, the default (flag off) is correct, and the gguf_olmoe_profile.py docstring was the wrong one — fixed in this PR.

- I updated the body of this PR.
- I updated the code docs inside my files.
- I moved the expert-trio completeness check out of convert_experts and into the pre-flight, so an incomplete GGUF is refused before config.json/tokenizer.json are written instead of after.
- I added a new CI job to run my tests and install the required libraries only for this flow. I also added a new `.npz` fixture so the quantizer check runs without installing `pytorch`, saving resources during tests.

PR #1431, opened 2026-09-11
fix(build): guard malloc_trim to glibc for musl libc compatibility

### LINKED ISSUE #1430: [Bug]: colibri does not build agains musl libc
### Commit or version

fd93c41aa6ae2c7d1cc1a1e2d6b79dbe6d341708

### Environment

tested on Alpine Linux 3.23.2 in a container with 30 threads on a dual Xeon E5-2680 v2 server. 200 GB RAM

### Reproduction steps

```bash
cd  colibri/c
./setup.sh
```

### Expected behavior

The program compiles successfully

### Actual behavior and logs

```shell
🐦 colibrì — setup
  gcc: 15.2.0 · 30 core
  OpenMP: ok
  building (ARCH=native)…
colibri.c: In function 'rss_guard':
colibri.c:8374:5: error: implicit declaration of function 'malloc_trim' [-Wimplicit-function-declaration]
 8374 |     malloc_trim(1024);
      |     ^~~~~~~~~~~
make: *** [Makefile:727: colibri] Error 1
```

iiEliJas:
Fixes #1430.

## Problem

`colibri.c:8374` calls `malloc_trim(1024)` inside `#ifdef __linux__`, but musl
libc defines `__linux__` without declaring `malloc_trim`, which is a GNU
extension, not POSIX. Builds on Alpine Linux (musl) fail with:

    colibri.c: In function 'rss_guard':
    colibri.c:8374:5: error: implicit declaration of function 'malloc_trim'

## The fix

Guard `malloc_trim` with `#ifdef __GLIBC__` instead of `#ifdef __linux__`.
`__GLIBC__` is defined by glibc but not musl, so the call is compiled only
where it exists. `#include <malloc.h>` moves to the same guard since it is
only needed for `malloc_trim`. `<sys/syscall.h>` stays under `__linux__`
(musl supports it for NUMA `mbind`).

## Verified

- `make colibri` on glibc: builds clean, no warnings
- musl: `__GLIBC__` is not defined, so `malloc_trim` is skipped, no build error


iiEliJas:
Closing in favor of a new PR with correct branch targeting dev (was forked from main, causing extra commits to appear)

JustVugg:
Thank you for the report and the fix in the same day. Reviewed by building it,
not by reading it, and the change is right. What follows is what I verified,
what I could not, and the one thing I would like added.

**Verified on glibc.** Builds clean, and the call survives where it should:

```
$ nm -C colibri | grep malloc_trim
                 U malloc_trim@GLIBC_2.2.5
```

So the guard does not quietly disable the RAM guard's trim on the platform that
has it, which was the thing worth checking.

**Verified that the fix is complete, not partial.** I grepped the engines for
the usual glibc-only suspects: `mallinfo`, `malloc_usable_size`, `backtrace` and
`execinfo.h`, `qsort_r` (different signature on musl), `strerror_r` (GNU vs POSIX
return type), `fopencookie`, `gnu_get_libc_version`, `error_at_line`.
`malloc_trim` is the only hit in the tree, one call site. So this is the whole
of the musl problem, not the first of ten.

**Verified the nesting.** The new `#ifdef __GLIBC__` stays inside the
`__APPLE__ || __linux__ || __FreeBSD__` block, so macOS, FreeBSD and MinGW
behave exactly as before, where `__linux__` already excluded `malloc.h`. And
`__GLIBC__` is defined by the time the preprocessor reaches line 42, since
`time.h`, `limits.h`, `pthread.h` and `unistd.h` all precede it. Both details
are easy to get wrong and neither is.

**What I could not verify, and it matters.** There is no musl toolchain on my
machine and no Alpine leg in CI, so nothing on our side can prove this builds
against musl. Two asks:

1. @chouettevan, if you can rebuild this branch on Alpine 3.23 and paste the
   result, that closes #1430 on your machine rather than on my reasoning.
2. Separately from this PR: the container job builds on debian-slim, so CI
   compiles glibc only. A small Alpine build leg would stop this class of
   regression from coming back, and it is cheap. I will open that myself unless
   you want it.

**One thing to add to the patch, one line of comment.** On musl the RSS guard
now drops its cached experts and never asks the allocator to return the freed
arenas to the OS. That is fine here and it is worth saying why: an expert slab
is tens of megabytes, well above any allocator's mmap threshold, so free()
already munmaps it. Without that sentence the next reader sees a guard that does
less on one libc and cannot tell whether it was reasoned or forgotten.

Approving the workflow runs, since you are a first-time contributor and they
were parked. Merging once they are green.

PR #1432, opened 2026-09-11
tests: drop the per-test _putenv_s helpers now that compat.h's setenv works

### LINKED ISSUE #1429: tests: drop the per-test _putenv_s helpers now that compat.h's setenv works
Follow-up to #1420, deliberately kept out of #1428.

Fourteen test files carry a private `env_set`/`env_unset` pair that calls `_putenv_s` on Windows and `setenv` elsewhere, written because `compat.h`'s `setenv` wrote only the Win32 block and `getenv` never saw it. That is fixed on dev, so each of them can call `setenv`/`unsetenv` directly once the file includes `compat.h`.

```
test_expert_store_registry.c        tests/test_qwen36_ctx.c
tests/qwen36_fake_cuda.h            tests/test_qwen36_dense_batch.c
tests/test_qwen36_tier_int8_decode.c tests/test_qwen36_tier_int8_engine.c
tests/test_inkling_shared_batch.c   tests/bench_inkling_shared_batch.c
tests/test_omp_tune.c               tests/bench_qwen36_dense_batch.c
tests/test_deepseek_v4.c            tests/test_rans.c
tests/test_route_trace.c            tests/test_stops.c
```

Not urgent and not a behaviour change: the helpers are correct as they stand. The reason to do it at all is that fourteen spellings of one platform difference is how the difference survives the next time someone reads the code. The reason not to do it today is that several of these files are being edited by open pull requests, so it is a pass to make when the queue is quiet.

`tests/test_compat_env.c` is the gate that keeps the shim honest once they are gone.

texasich:
## Summary

`compat.h`'s `setenv` now writes through `_putenv_s` (and keeps the Win32 block in sync), so `setenv()` followed by `getenv()` in the same process works on Windows. The duplicated environment helpers and Windows overrides the tests carried because of the old shim are redundant: replace them with direct `setenv()`/`unsetenv()` calls and `compat.h` includes.

The fourteen files listed in #1429, plus the six qwen36 tier tests: they set their environment through `setenv()` and inherited the Windows half from `qwen36_fake_cuda.h`, so with the header's shim gone they include `compat.h` directly. Rules whose sources newly include `compat.h` list it as a prerequisite, and the registry test builds with `-D_FILE_OFFSET_BITS=64` (`compat.h` requires it on Windows).

No intended change to these tests' environment-setting behavior: every converted call passes `overwrite=1` and discards the return value.

Fixes #1429.

## Validation

- [x] `make -C c check` (see notes)
- [ ] CUDA changes were tested with `make -C c cuda-test` (not applicable, no CUDA changes)
- [ ] Performance claims include hardware, commands, and repeatable measurements (not applicable, no performance claims)

Notes (isolated Linux sandbox, GCC 15.3.0, 2 vCPU):

- All C tests pass except sandbox-environment failures that also reproduce on pristine `dev` with the same command: `test_inkling_shared_batch` (bf16 batch-vs-scalar bit-exactness) and `test_rss_anon` (/proc RSS accounting) fail consistently, and `test_uring` / `test_v4_serve_framing` fail intermittently (io_uring and socket fixtures under this sandbox). No failure is unique to this branch.
- The Python suite additionally hits environment failures here (a cluster of errors plus the runner being resource-killed mid-run). The Makefile- and compat-sensitive Python tests (`test_makefile_deps`, `test_makefile_platform`, `test_cuda_test_makefile`, `test_env_defaults`) all pass standalone.
- Aside from the environment-class failures above, every converted test passes in the sandbox (`test_rans`, `test_route_trace`, `test_stops`, `test_omp_tune`, the qwen36 tier family, `test_compat_env`). Windows behaviour is exercised by the UCRT64 CI job, the authoritative run for the affected paths.

## Compatibility

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1350,7 +1350,7 @@ tests/test_kv_disk$(EXE): tests/test_kv_disk.c colibri.c st.h json.h tok.h tok_u
 tests/test_e8_kernel$(EXE): tests/test_e8_kernel.c quant.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
-tests/test_rans$(EXE): tests/test_rans.c rans.h
+tests/test_rans$(EXE): tests/test_rans.c rans.h compat.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 # Structured-mutation fuzz for the rans record parser/decoder under
@@ -1604,7 +1604,7 @@ V4_TEST_LINK_OBJS += COLI_V4_UNIT_GPU.o backend_loader_dsv4.o
 V4_TEST_EXTRA_TARGETS = COLI_V4_UNIT_GPU.o backend_loader_dsv4.o
 endif
 
-tests/test_deepseek_v4$(EXE): tests/test_deepseek_v4.c deepseek_v4.c deepseek_v4.h \
+tests/test_deepseek_v4$(EXE): tests/test_deepseek_v4.c deepseek_v4.c deepseek_v4.h compat.h \
 		Makefile.deepseek-v4.units Makefile.deepseek-v4
 	$(MAKE) -f Makefile.deepseek-v4 ARCH=$(ARCH) deepseek-v4-test-objs \
 		COLI_V4_UNIT_ST.o COLI_V4_UNIT_CONFIG.o COLI_V4_UNIT_MATH.o COLI_V4_UNIT_SPARSE_ATTENTION.o \
@@ -1656,7 +1656,7 @@ tests/test_v4_serve_framing$(EXE): tests/test_v4_serve_framing.c deepseek_v4.c \
 
 endif
 
-tests/test_route_trace$(EXE): tests/test_route_trace.c route_trace.h
+tests/test_route_trace$(EXE): tests/test_route_trace.c route_trace.h compat.h
 	$(CC) $(CFLAGS) $< -o $@ $(LDFLAGS)
 
 tests/test_cli_args$(EXE): tests/test_cli_args.c cli_args.h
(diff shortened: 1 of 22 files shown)


JustVugg:
This is the cleanup from #1429 and it is done the right way round, which was the
part I was most worried about. Reviewed by building and running it, plus a
mechanical check on the thing that cannot be observed from Linux.

**The ordering is right.** The branch point is 899e7c2, the merge of #1428, so
the shim these helpers were working around actually works underneath this. Had
it gone in first, every one of these tests would have kept compiling on Windows
and silently stopped setting anything, which is the exact failure the helpers
existed to avoid.

**What I ran.**

```
make test-c                                   ALL PASS  (0 failures)
make -f Makefile.deepseek-v4 deepseek-v4-test-registry    ALL OK
make tests/bench_qwen36_dense_batch           builds
```

**What I checked mechanically rather than by eye.**

- No call changed its overwrite semantics. The old helpers always overwrote; the
  diff introduces no `setenv(..., 0)` anywhere.
- Every removed line is either a helper definition or a helper call site. Nothing
  lost an assertion or a branch in the process.
- The `-D_FILE_OFFSET_BITS=64` in Makefile.deepseek-v4 is correct and I checked
  why it is needed rather than assuming: compat.h carries an explicit `#error`
  for it on Windows, because a 32-bit off_t wraps offsets past 4 GB and reads the
  wrong weight bytes. The main CFLAGS already defines it, so only that
  hand-rolled rule was missing it.

**One thing to consider changing, one thing that is not your problem.**

1. Nine of the files you edited call `setenv` without including compat.h. I
   confirmed with `gcc -MM` that all nine reach it transitively through the
   engine source they include, and that in every one the first include sits far
   above the first call, so the macro is defined in time on Windows. It works.
   It is also invisible: someone reordering includes in `inkling.c` a year from
   now breaks a test on a platform their CI does not compile. Since the whole
   point of the change is one spelling in one place, I would rather see
   `#include "../compat.h"` in those nine too. Your call; it is not a blocker.

2. Not your problem but it belongs in this thread: **no CI job compiles the C
   tests on Windows.** `make test-c` runs on the Linux leg only, and the only
   test built under msys2 is `tests/test_compat_env`, which I added yesterday.
   So the entire class of change this PR makes is unobservable in CI on the
   platform it is about. I will look at a Windows leg that builds at least the
   env-touching tests.

**Merge order.** `tests/bench_inkling_shared_batch.c` is one of the files you
edited, and on this branch it cannot be built at all: dev's Makefile has that
target with no recipe, so make falls through to its builtin rule and fails. That
is #1433, which I am merging first. I verified that with its one line applied on
top of your branch the bench builds and runs, so nothing else is needed from you.

Approving the parked workflow runs. Merging after #1433 and once they are green.
Thank you for picking this up the day it was filed, and for reading the
dependency lines in the Makefile instead of only the includes.

PR #1433, opened 2026-09-11
build: restore the missing recipe for bench_inkling_shared_batch

texasich:
## Summary

`tests/bench_inkling_shared_batch` has had its prerequisites line in `c/Makefile` but no recipe since a774efd, so GNU make fell back to its implicit rule with none of the repository flags (no `-fopenmp`, no include paths), and the target could not build the way its siblings do. It is only reachable explicitly (`make tests/bench_inkling_shared_batch`), which is why the break went unnoticed: nothing in CI builds it.

Restore the recipe the sibling `tests/test_inkling_shared_batch` target already carries; prerequisites unchanged. The added line is the same single compile command as its sibling, so nothing else is touched.

## Validation

- [x] `make -C c check` (see notes)
- [ ] CUDA changes were tested with `make -C c cuda-test` (not applicable, no CUDA changes)
- [ ] Performance claims include hardware, commands, and repeatable measurements (not applicable; this only fixes the bench's build, it reports no measurement)

Notes (isolated Linux sandbox, GCC 15.3.0):

- `make tests/bench_inkling_shared_batch` builds clean (0 warnings) with the recipe restored, matching the flags every sibling test uses; without it, the same command never builds (the builtin rule emits implicit-declaration errors). The bench output itself is unchanged; this restores the build only.
- `make -C c check` exhibits the same sandbox-environment failures with or without this change (the bench is not part of the check target).

## Compatibility

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1279,6 +1279,7 @@ tests/test_inkling_shared_batch$(EXE): tests/test_inkling_shared_batch.c inkling
 # Reproducible local A/B for the shared-expert prefill path.  This is timing
 # evidence, not a CI gate: build and run it explicitly on the target CPU.
 tests/bench_inkling_shared_batch$(EXE): tests/bench_inkling_shared_batch.c inkling.c st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h omp_tune.h route_trace.h kv_prefix.h serve_codec.h $(INK_CUDA_OBJ) $(METAL_OBJ)
+	$(CC) $(CFLAGS) $< $(INK_CUDA_OBJ) $(METAL_OBJ) -o $@ $(LDFLAGS)
 
 tests/test_inkling_cache_index$(EXE): tests/test_inkling_cache_index.c inkling.c st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h omp_tune.h route_trace.h kv_prefix.h serve_codec.h $(INK_CUDA_OBJ) $(METAL_OBJ)
 	$(CC) $(CFLAGS) $< $(INK_CUDA_OBJ) $(METAL_OBJ) -o $@ $(LDFLAGS)


JustVugg:
Reviewed by reproducing it rather than by reading, and it is a real defect with
a one-line fix. Merging this first of the three open build changes.

**Reproduced on dev.** The target exists with a full prerequisite list and no
recipe, so make falls through to its own builtin rule:

```
make: *** [<builtin>: tests/bench_inkling_shared_batch] Error 1
```

It compiles the .c alone, without inkling.c or the CUDA/Metal objects, fails at
the first engine symbol, and leaves a 6 MB half-built artifact behind. Same trap
as a bare `make qwen36`.

**Origin.** `git log -S` puts it in a774efd, the shared-expert prefill batching
change: the dependency line was added, the command line under it was not.

**Verified with the patch.** Builds to a 200 KB binary and runs:

```
inkling shared bf16: S=16 D=2048 I=1024 shared=2 bytes=24.0 MiB
scalar 0.040678 s  batch 0.010421 s  speedup 3.90x  calls 96 -> 6
inkling shared int4-g64: S=16 D=2048 I=1024 shared=2 bytes=6.8 MiB
scalar 0.032825 s  batch 0.010070 s  speedup 3.26x  calls 96 -> 6
```

The recipe matches its siblings exactly, `$(INK_CUDA_OBJ)` and `$(METAL_OBJ)`
included.

**Why nobody noticed.** `TEST_RULES` derives the gates from rules named
`test_*`, so a `bench_*` target is never built by `make test-c`. That is the
right design, and the cost of it is exactly this: a bench can rot silently. Not
something to change for one case, but worth knowing.

One consequence that decides the merge order: #1432 edits
`tests/bench_inkling_shared_batch.c`, and on that branch the file still cannot
be compiled. This goes in first, then that one can actually build what it
touched. I checked by applying your line on top of #1432: the bench builds and
runs there too.

Approving the workflow runs (first-time contributor), merging once they are
green. Thank you, this is the kind of report that is worth more than its size.

PR #1437, opened 2026-09-11
ci: add Alpine/musl build leg + comment on malloc_trim guard

iiEliJas:
Adds an Alpine (musl libc) CI job to catch build regressions like #1430 early.

Also adds the one-line comment requested in #1435 review explaining why the `malloc_trim` guard is safe on musl.


iiEliJas:
Superseded by maintainer's Alpine CI PR. Will open a separate PR for the comment only.

PR #1438, opened 2026-09-11
docs: add malloc_trim comment for musl

iiEliJas:
Carries over the comment requested in #1435 review.


JustVugg:
Exactly the line I asked for, and it says the right thing: the guard does less on musl because free() already hands a tens-of-megabytes slab back to the OS, not because anyone forgot. Merging.

Worth knowing what landed alongside it: #1436 added an Alpine leg to CI, and it is green on this branch. colibri and olmoe now build against musl on every pull request, and the job asserts the binary carries no reference to malloc_trim. Your report is a gate now, not a fixed bug.

PR #1447, opened 2026-09-12
fix(mirror verify): support complete mirrors without a receipt

trigger2k20:
## Summary

`coli mirror verify` currently fails with `receipt_missing` whenever `.colibri-mirror.json` is absent, even when the mirror already contains a complete copy of all model shards.

This is especially relevant for models or engines that do not generate `.coli_usage`. In that case, `coli mirror stage` cannot create a usage-ranked partial-mirror receipt, which means that an otherwise complete mirror cannot be verified.

There is also an inconsistency in the current CLI implementation: `mirror verify` already requires `--model`, but the model path is not used when the receipt is missing.

## Current behavior

For a complete manually created mirror:

```bash
coli mirror verify \
  --model /path/to/model \
  --mirror /path/to/full-mirror
```

Colibri currently returns:

```json
{
  "schema": "colibri.partial-mirror.v1",
  "ready": false,
  "reason": "receipt_missing",
  "failures": []
}
```

This happens even when every model shard is already present in the mirror.

The current verification path effectively stops here when the receipt does not exist:

```python
if not receipt_path.is_file():
    return {
        "schema": SCHEMA,
        "ready": False,
        "mirror_root": str(mirror),
        "reason": "receipt_missing",
        "failures": [],
    }
```

## Proposed behavior

This PR adds a fallback verification path for complete mirrors.

When `.colibri-mirror.json` exists, nothing changes: Colibri continues to use the existing receipt-based SHA-256 verification.

When the receipt does not exist, `mirror verify` now uses the already-required `--model` argument, together with optional `--source-dir` arguments, to determine which model shards should exist in the mirror.

The fallback verifies that:

- every expected model shard exists in the mirror;
- each mirrored shard has the expected file size;
- each mirrored shard has a valid Safetensors header;
- the mirror is reported as ready only if all expected shards pass verification.

A successful receipt-less verification reports:

```json
{
  "ready": true,
  "verification_mode": "full_mirror"
}
```

If even one required shard is missing or invalid, verification still fails.

## Why this is useful

A complete mirror should not depend on routing history or on the partial-mirror staging workflow.

`.coli_usage` is useful for deciding which hot shards should be copied when creating a **partial mirror**. It should not be required to verify a **complete mirror** that already contains all model shards.

This also makes use of information that the CLI already requires:

```bash
coli mirror verify \
  --model /path/to/model \
  --mirror /path/to/mirror
```

Previously, `--model` was required for the command but was not used by the receipt-less verification path.

## Compatibility

The existing partial-mirror workflow is unchanged.

If `.colibri-mirror.json` exists, Colibri continues to perform the existing receipt-based SHA-256 verification. The new full-mirror fallback is only used when no receipt exists.

This means that the stricter SHA-256 verification used for staged partial mirrors is preserved.

## Tests

The full test suite was run locally after applying the patch:

```text
Ran 724 tests in 60.234s

OK (skipped=39)
```

The patched source was then rebuilt and installed locally as Colibri `1.10.2`.

Regression tests are included for both relevant cases:

1. A complete mirror without `.colibri-mirror.json` or `.coli_usage` is successfully verified.
2. A receipt-less mirror with a required shard missing is rejected.

## Real-world validation

The change was also tested against a real full mirror of:

```text
```

The model consists of 62 Safetensors shards with approximately 194.7 GB of mirrored model data.

Verification was run with:

```bash
coli mirror verify \
  --model "/Users/sascha/lllm-colibri/Justvugg/GLM-5.3-Flash-colibri-int4-g64" \
  --mirror "/Volumes/big2/modelstorage/lllm-colibri-mirror/Justvugg/GLM-5.3-Flash-colibri-int4-g64"
```

The patched implementation successfully verified the complete receipt-less mirror:

```json
{
  "schema": "colibri.partial-mirror.v1",
  "ready": true,
  "mirror_root": "/Volumes/big2/modelstorage/lllm-colibri-mirror/Justvugg/GLM-5.3-Flash-colibri-int4-g64",
  "verification_mode": "full_mirror",
  "file_count": 62,
  "mirrored_bytes": 194727335256,
  "failures": []
}
```

This confirms that the fallback works with an actual large multi-shard Colibri model and not only with the regression fixtures.

No model files, generated binaries, or benchmark artifacts are included in this change.


JustVugg:

PR #1451, opened 2026-09-12
v41: register deepseek_v41 family (engine stub fails closed)

fegato:
## Summary

## Validation

- [ ] `make -C c check`

## Compatibility


fegato:
Closing this: #1453 landed the family the right way -- the engine, its own
tiny-fixture oracle and the conformance manifest entry -- so the stub-registration
premise here is obsolete, and I would rather not add two files the tree does not
reference. Thanks for taking V4.1 on.

PR #1453, opened 2026-09-12
DeepSeek V4.1 Flash: engine, reference, and the family wired through

JustVugg:
DeepSeek V4.1 Flash runs on colibri: 552B, 510 GB on disk, streamed from an SSD on
a CPU box. No conversion — the released checkpoint is already fp8 (dense) and fp4
(experts), and both formats are read natively; the fp4 expert layout is byte-identical
to the mxfp4 the Kimi K3 engine already reads.

## What is in it

A new engine, `c/deepseek_v41.c`, and the family wired through everything that makes a
family real in this repo: the registry and the planner, the gateway, the Makefile, CI,
the release workflow, the docs and the site.

The architecture is not V4 with a new number. Every one of these is new here:

- **Engram**, two n-gram memories of 384M rows (203 GB) that never enter RAM — rows are
  read one at a time behind a small LRU, because n-gram traffic is Zipfian.
- **DSA indexer** with a two-level candidate source, **hyper-connections** (mHC),
  **MQA sliding-window plus compressed** attention, a **compressor** with group slots.
- **fp4 experts** with per-32 group scales, **fp8 dense** with 32x32 ue8m0 tiles.
- A **32-layer vision tower** with a 3x3 aligner.
- **DSpark**, the checkpoint's own speculative head: 3 stages, 128 experts, blocks of 5,
  verified in one batched forward with a rollback for the rejected rows.
- **Tool calling** in the checkpoint's DSML format.

## How it is checked

`c/tools/dsv41_ref.py` is a torch-only CPU reference, written because the vendor's own
reference needs tilelang kernels that do not run on CPU. `tools/make_dsv41_tiny.py`
builds a container with the same tensor names, dtypes and block-scale conventions as the
released checkpoint at toy dimensions, so the engine reads the real format.

CI holds the engine to that reference **token for token** at three cache capacities, on a
forty-token prompt as well as the default eight (the block boundaries are at 32), under
all three `V41_SPEC_FORCE` modes including a partly-verified block, and asserts the
vision tower matches to 5e-06. It also asserts that the two index-key policies *disagree*,
so the vendor's odd-but-shipped default cannot be "tidied" by accident.

## Measured on the real checkpoint

Verified on a 552B checkpoint on real hardware, not only on the fixture. Every
measurement cold, page cache dropped before each run, same prompt and seed:

| | wall | disk | expert matmul | attention | tok/s |
|---|---|---|---|---|---|
| first working version | 78.7 s | 36.9 s | 15.0 s | 19.1 s | 0.305 |
| batched expert reads | 42.4 s | 13.0 s | 8.4 s | 16.1 s | 0.566 |
| vector attention kernel | 41.1 s | 12.9 s | 7.8 s | 15.9 s | 0.584 |
| blocked attention matrices | 29.5 s | 12.7 s | 9.1 s | 3.4 s | 0.812 |
| expert-major MoE | **25.1 s** | 10.4 s | 7.6 s | 3.4 s | **0.957** |

Three changes did that, and each is bit-exact against what it replaced:

1. A step's expert misses go to the device **together** instead of one behind another —
   88.5 GB per turn was arriving at 2.47 GB/s where the same blocks go at 6.8 GB/s with
   eight in flight. `V41_READ_DEPTH` sets the depth; 8 is the measured knee.
2. Attention reads its matrices **once per block of positions**, not once per position.
   wq_b is 42 MB and so is wo_b; prefill of a 26-token prompt was reading each of them
   twenty-six times per layer.
3. The MoE is **expert-major**: route the whole block, reduce the six-per-position draws
   to distinct experts, apply each to every position that asked. Contributions are kept
   apart and summed in rank order, so the arithmetic is the same arithmetic.

A five-turn chat session on the released checkpoint runs at 1.14 to 1.58 tok/s and
answers correctly — a factual question, a reasoning trap, C code, Italian, and a domain
question — with an image going through the gateway into the vision tower on the same
engine.

Two ways of hiding the reads behind the compute were built, measured and **removed**;
both are written up in `docs/deepseek-v41.md` so the third attempt starts from the
numbers. A `posix_fadvise(WILLNEED)` hint blocks against a device the demand reads have
already saturated. A pool of reader threads does overlap — the wait on disk falls from
10.4 s to 1.0 s — but the expert matmul rises from 7.6 s to 18.1 s and the turn is 7%
slower, because a 5.9 MB pread is a kernel-side copy competing for the memory bandwidth
the matmuls are already bound by.

## Two traps worth knowing

The Engram hash multipliers exceed a double's mantissa, so `prepare_dsv41.py` stores them
as decimal **strings** and the engine refuses a numeric sidecar: a rounded multiplier
hashes every n-gram into a different, perfectly valid-looking row.

`shared_attn.index_k` is a module global in `model.py`, republished only when a
compression group completes, so a decode step with a half-filled group scores against
another layer's cache. That is what DeepSeek ships and what every published number for
this model was measured with, so it is the default here; `V41_INDEX_OWNER=1` selects the
tidy reading, which is a different model.

## Not claimed

`coli run` is deliberately unwired for this family, as it is for qwen36 and qwen38: the
engine speaks the SERVE protocol and nothing else, so a one-shot has nowhere to go but
the gateway. `coli chat`, `coli serve` and `coli web` all work. No GPU path.


jarnnont1-art (approved):

jarnnont1-art:
> Measured

PR #1454, opened 2026-09-12
fix: enable qwen36 GPU tier when building with HIP=1

yuripourre:
## Summary

- Fix `make -C c qwen36 HIP=1` on Linux ROCm: the target linked `backend_cuda.o` but still used `NOCUDA_LDFLAGS`, which dropped `-lstdc++` and failed at link time with `DSO missing from command line`.
- Mirror the existing `CUDA=1` branch so HIP builds also compile `qwen36_tier.c` and link with the full `CFLAGS`/`LDFLAGS` set (including `-DCOLI_CUDA`, `-lamdhip64`, and `-lstdc++`).

## Problem

The qwen36 Makefile switch only treated `CUDA=1` as a GPU build. With `HIP=1`, `$(CUDA_OBJ)` (including `backend_cuda.o`) is still linked into `qwen36`, but the build fell through to the CPU-only branch:

- `QWEN36_LDFLAGS = $(NOCUDA_LDFLAGS)` — no `-lstdc++` / HIP runtime flags
- `QWEN36_TIER_SRC` empty — GPU expert tier not compiled

Repro (Linux + ROCm):

```sh
make -C c qwen36 HIP=1 HIP_ARCH=gfx1201 HIPCC=/usr/bin/hipcc
```

Code changes assisted by Cursor.


JustVugg:

JustVugg:
The `Linux` failure here (`expert did not become resident during warmstart`, in test_qwen36_tier_multidev) is not yours. Your run was built against a `dev` snapshot taken in the middle of a merge sequence that landed #1390 and #1388, and #1388 changes assertions in that exact test. Both tests pass on settled `dev` (six consecutive runs of test_qwen36_tier_multidev, five of test_qwen36_tier_fill_wait). I have re-run the failed jobs against current `dev`. Nothing to change on your side; the four Makefile lines are the right shape.

yuripourre:
@JustVugg thanks for clarifying, I will be more mindful in the next PRs.

Thank you so much for merging as well.

PR #1455, opened 2026-09-12
feat(glm53): support COLI_MODEL_MIRROR and disk-weighted expert reads

trigger2k20:
# feat(glm53): support COLI_MODEL_MIRROR and disk-weighted expert reads

## Summary

This PR adds `COLI_MODEL_MIRROR` support to the GLM-5.3 (`glm53`) engine and enables deterministic, disk-weighted expert reads across the primary model directory and one or more mirror directories.

Before this change, the generic Colibri engine supported mirrored model storage, but the dedicated `glm53` loader bypassed that infrastructure. Setting `COLI_MODEL_MIRROR` therefore had no effect on actual GLM-5.3 expert/shard reads.

The implementation integrates the existing `st.h` replica infrastructure into `glm53` while preserving primary-only operation when no mirror is configured.

## Problem

GLM-5.3 uses its own model loader and expert streaming/cache implementation.

Although Colibri already exposes:

- `COLI_MODEL_MIRROR`
- `COLI_DISK_WEIGHTS`
- `SNAP_MIRROR` as a legacy alias
- multi-replica support in `st.h`

the `glm53` path initialized its safetensors index using:

```text
load_cfg(&m->c, dir);
st_init(&m->S, dir);
```

but did not register model mirrors.

The GLM-5.3 expert table also stored and subsequently used only the primary shard file descriptors.

As a result, configuring a mirror did not cause GLM-5.3 inference to read expert data from that mirror.

## Reproduction before this patch

Test model:

```text
```

Model characteristics:

```text
GLM-5.3-Flash
321B parameters
~195 GB on disk
45 layers
288 experts/layer
top-8 routing
62 safetensors shards
```

Primary:

```text
/Users/sascha/lllm-colibri/Justvugg/GLM-5.3-Flash-colibri-int4-g64
```

Mirror:

```text
/Volumes/big2/modelstorage/lllm-colibri-mirror/Justvugg/GLM-5.3-Flash-colibri-int4-g64
```

The mirror contains all 62 shards and was independently verified.

`fs_usage` was used while running GLM-5.3 with `COLI_MODEL_MIRROR` configured.

Before this patch, approximately 22,397 `RdData` events were observed from the primary model path.

No corresponding GLM-5.3 `RdData` activity occurred against the mirror path.

This confirmed that `COLI_MODEL_MIRROR` was not being consumed by the `glm53` expert I/O path.

## Root cause

The dedicated `glm53` implementation bypassed the generic mirror setup and expert routing used by the standard Colibri engine.

`st.h` already provides the required underlying replica primitives:

```text
st_mirror_reset()
st_mirror_add()
st_fd_rep()
st_direct_fd_rep()
st_mirror_count()
st_path_for_fd()
```

`st_mirror_add()` also already validates replica files by size and safetensors header and permits partial mirrors.

The missing part was integration of those primitives into `glm53`.

### Mirror discovery

`glm53` now reads:

```text
COLI_MODEL_MIRROR
```

with:

```text
SNAP_MIRROR
```

as the legacy fallback.

Multiple mirror directories may be supplied using `;` or `,` separators.

Mirror directories identical to the primary model directory are ignored.

Each configured mirror is registered through `st_mirror_add()`.

This retains the existing `ST_MAX_MIR` replica limit from `st.h`.

### Deterministic expert routing

Each expert is assigned to a replica using a deterministic hash of:

```text
(layer, expert_id)
```

The routing hash follows the same scheme used by the generic Colibri mirror implementation.

Repeated accesses to the same expert therefore resolve to the same storage device rather than randomly moving between drives.

### Disk weights

Explicit drive weights can be configured using:

```text
COLI_DISK_WEIGHTS
```

Weights correspond to:

```text
primary,mirror1,mirror2,...
```

For example:

```text
COLI_DISK_WEIGHTS=3,2
```

produces approximately:

```text
primary: 60%
mirror:  40%
```

of the deterministic routing space.

### Automatic bandwidth probing

When valid explicit disk weights are not supplied, `glm53` benchmarks the available replicas at startup.

The probe:

- selects a sufficiently large shard
- uses 19 MiB blocks
- performs 8 reads
- uses aligned offsets
- uses direct file descriptors where available
- falls back to buffered reads when required
- measures elapsed time with `CLOCK_MONOTONIC`

Measured throughput is converted into routing weights.

Example from the test system:

```text
[GLM53 MIRROR] probe drive 0: 6533 MiB/s
[GLM53 MIRROR] probe drive 1: 4541 MiB/s
[GLM53 MIRROR] 2 drives | routing 59% / 41% (startup bandwidth probe)
```

The automatically selected 59/41 distribution is consistent with independent storage benchmarking of the two devices.

### Whole-expert replica preflight and fallback

Before an expert is read from a selected replica, all six pieces required by that expert are preflighted.

If any required piece is unavailable on the selected replica, the **entire expert** falls back to primary storage.

This guarantees that one expert is never assembled from pieces located on different replicas.

The same selected replica is subsequently used by both the mmap and pread paths.

This behavior is explicitly covered by the GLM-5.3 mirror regression tests.

### Replica-aware mmap

The mmap path resolves the selected replica file descriptor before mapping expert data.

Previously, mmap operations were tied to the primary shard descriptor.

The existing CPU mmap fast path remains preserved.

### Replica-aware pread

The non-mmap expert read path uses the same selected replica selected by the whole-expert preflight.

This keeps mmap and pread routing consistent.

## Independent storage benchmark

Using Colibri's `iobench` against the same model shard:

Primary storage:

```text
~7.27 GB/s
~2.7 ms/block
```

Mirror storage:

```text
~4.77 GB/s
~4.2 ms/block
```

This corresponds to a primary-to-mirror bandwidth ratio of approximately:

```text
1.52 : 1
```

An explicit weight configuration of:

```text
COLI_DISK_WEIGHTS=3,2
```

is therefore a reasonable approximation for this system.

The sum of the two independent measurements is approximately 12.0 GB/s. This is not presented as expected inference bandwidth; it only demonstrates the available independent storage bandwidth.

### Explicit weights

Configuration:

```text
COLI_DISK_WEIGHTS=3,2
```

Startup reported:

```text
[GLM53 MIRROR] replica 1: /Volumes/big2/modelstorage/lllm-colibri-mirror/Justvugg/GLM-5.3-Flash-colibri-int4-g64 (62/62 shards)
[GLM53 MIRROR] 2 drives | routing 60% / 40% (COLI_DISK_WEIGHTS)
```

No startup bandwidth probe was performed when explicit valid weights were supplied.

This verifies that explicit `COLI_DISK_WEIGHTS` takes precedence over automatic probing.

### Automatic weights

With `COLI_DISK_WEIGHTS` unset:

```text
[GLM53 MIRROR] replica 1: /Volumes/big2/modelstorage/lllm-colibri-mirror/Justvugg/GLM-5.3-Flash-colibri-int4-g64 (62/62 shards)
[GLM53 MIRROR] probe drive 0: 6533 MiB/s
[GLM53 MIRROR] probe drive 1: 4541 MiB/s
[GLM53 MIRROR] 2 drives | routing 59% / 41% (startup bandwidth probe)
```

Both drives were benchmarked and the measured bandwidth was used to derive the routing distribution.

## Controlled cold A/B validation

Following maintainer feedback, an additional controlled cold-cache A/B test was performed on real Apple Silicon hardware.

Both runs used:

- the same GLM-5.3-Flash model
- the same prompt
- the same generation length
- the same expert-history state
- the same Metal configuration
- `--ram 96`
- `--cap 8`
- `--ctx 8192`
- `--temp 0`
- `--ngen 256`
- `--no-tune-profile`
- `sudo purge` followed by a 5-second delay before each run
- usage-history saving disabled so the benchmark did not modify routing history

Prompt:

```text
Explain how virtual memory works in a modern operating system, including page tables, TLBs, page faults, and memory mapping.
```

### Mirror disabled

```text
256 tokens / 138.2 s
1.853 tok/s

expert cache hits:   20,845
expert cache misses: 68,556

logical expert bytes:
970,463,379,456
```

### Mirror enabled, 60/40 routing

```text
256 tokens / 141.1 s
1.815 tok/s

expert cache hits:   20,845
expert cache misses: 68,556

logical expert bytes:
970,463,379,456
```

A repeated mirror-enabled run instrumented with `fs_usage` produced:

```text
256 tokens / 138.8 s
1.844 tok/s
```

The controlled runs therefore do **not** establish a defensible inference-throughput improvement from mirroring.

The measured decode differences are small enough to be treated as normal run-to-run variation.

The purpose of this PR is consequently not to claim higher inference throughput. It is to make the existing Colibri mirror semantics functional in the dedicated GLM-5.3 engine and allow expert I/O to be distributed across configured replicas.

## Filesystem read distribution

The instrumented mirror-enabled cold run was captured with macOS `fs_usage`.

Observed filesystem read data:

```text
PRIMARY
events: 259,538
bytes:  109,736,882,176
GiB:    102.20
share:  63.76%

MIRROR
events: 152,496
bytes:  62,367,682,560
GiB:    58.08
share:  36.24%

TOTAL
bytes:  172,104,564,736
GiB:    160.28
```

The observed filesystem read-byte distribution was therefore:

```text
primary: 63.76%
mirror:  36.24%
```

This is reasonably consistent with the configured deterministic 60/40 routing policy.

These values are **filesystem read bytes observed by macOS `fs_usage`**. They are not presented as measurements of physical SSD bus traffic.

Most importantly, GLM-5.3 generated substantial expert read activity from both configured model locations.

Before this patch, the dedicated GLM-5.3 engine read expert data only from the primary model storage.

## Fallback validation

Fallback behavior was validated at both the safetensors replica layer and the GLM-5.3 expert-routing layer.

The safetensors mirror runtime tests cover replicas with:

- divergent safetensors headers
- different/short file sizes
- empty or missing mirror files
- rejected replica file descriptors

The GLM-5.3 mirror tests additionally verify that the complete six-piece expert is preflighted before mmap/pread access.

If any required piece is unavailable on the selected replica, the whole expert resolves to the primary replica.

This prevents mixed-replica expert assembly.

Runtime storage validation:

```text
[MIRROR] tmp_mirror_c/model.safetensors: header differs from the primary copy — file skipped
[MIRROR] tmp_mirror_d/model.safetensors: size differs from the primary copy — file skipped

safetensors mirror tests: ok
```

## Automated tests

The dedicated regression test:

```text
c/tests/test_glm53_mirror.py
```

contains 10 tests covering:

- deterministic expert routing
- expected 3:2 weighted distribution
- routing hash constants
- explicit `COLI_DISK_WEIGHTS`
- startup bandwidth probing
- whole-expert replica preflight
- partial-replica fallback logic
- replica-aware mmap
- ensuring mirror support does not disable mmap
- replica-aware pread

Latest targeted validation after rebasing onto current `dev` and adding the whole-expert preflight:

```text
GLM-5.3 mirror tests:
10/10 passed

GLM-5.3 Metal routed source tests:
5/5 passed

safetensors mirror runtime tests:
OK

git diff --check:
clean

GLM-5.3 METAL=1 build:
successful
```

The Metal source-level test was also made whitespace-tolerant so that it validates the preserved CPU mmap fast path without depending on a specific one-line source formatting.

That test-only adjustment does not change inference or mirror-I/O behavior.

## Compatibility

When no mirror environment variable is configured, GLM-5.3 continues using the primary model storage only.

Existing single-disk operation therefore remains unchanged.

The implementation reuses the existing `st.h` mirror infrastructure rather than introducing a second storage-replica abstraction.

## Known limitation

The patch handles missing or invalid expert pieces on a replica before starting the expert read and falls the whole expert back to primary storage.

It does **not** currently recover from a physical I/O error that occurs after a valid replica read has already started.

`st_pread_full()` currently terminates on read error or unexpected EOF, so implementing runtime retry/failover for an I/O failure occurring during an active read would require changes below the GLM-5.3 routing layer.

This PR intentionally does not change those generic `st.h` error semantics.

## Test platform

```text
Apple Mac Studio
Apple M4 Max
128 GB unified memory
macOS
GLM-5.3-Flash 321B
62 safetensors shards
~195 GB model size
```

## Current PR state

Current head:

```text
0a3fa16b987736411b385e9f62d05ff7d3481b8a
```

The PR currently consists of three commits.

The latest commit is a test-only adjustment that makes the existing CPU mmap source-level assertion tolerant of equivalent multiline formatting.

No additional inference or mirror-I/O behavior was introduced by that final commit.

## Scope

This PR adds GLM-5.3 integration for the existing Colibri model-mirror infrastructure.

It does not claim that mirrored storage universally improves GLM-5.3 inference throughput.

The controlled cold A/B measurements show approximately equivalent decode performance on the tested M4 Max system, while the filesystem measurements confirm that expert reads are successfully distributed across both configured model locations.


JustVugg:

JustVugg:
Taking your other two first: #1456 (the `/proc/meminfo`-on-macOS collapse) and #1457 (routing telemetry) are both going in, they are self-contained and their claims check out against the rest of the tree.

This one I want to review on its own rather than in a batch, because unlike those two it changes the expert read path, and 494 lines of it. Parity with `colibri.c`'s `COLI_MODEL_MIRROR` and `COLI_DISK_WEIGHTS` is clearly worth having for GLM-5.3, so this is a scheduling note, not a doubt about the goal.

Two things that would speed that review up:

- **A measured datapoint.** The mirror's whole premise is that splitting expert reads across two devices beats one, in a ratio worth computing. A cold A/B on real hardware, caches dropped, same prompt and seed, with and without the mirror, reporting tok/s and the bytes read per device, turns the weighting from a plausible heuristic into a measured one. The README is deliberately cautious about the dual-SSD path for exactly this reason.
- **What happens when the mirror is wrong.** Short shard, stale copy, disappearing mount. I would like to see the failure be a clean fall back to the primary drive rather than a wrong read, and a test that pins it.

Please rebase once #1456 and #1457 land, since all three touch `glm53.c`.

trigger2k20:
Thanks — agreed on both review points.

For the mirror-failure case I want to keep two failure classes separate.

The glm53 routing layer already preflights all six pieces of a routed expert on the selected replica. If any piece is unavailable, the whole expert is routed back to the primary; pieces from different replicas are never mixed. I’ll add a regression test that explicitly pins that behavior with an incomplete mirror.

A physical read failure after st_pread_full() has already started is different: today st_pread_full() treats an I/O error or unexpected EOF as fatal for every engine. Adding transparent retry/failover there would change the generic storage semantics rather than just glm53 mirror routing, so I would prefer not to broaden this PR into that unless you want the generic st.h behavior changed as well.

I’ll also add the requested controlled hardware datapoint: same model, prompt, generation length and cap, single-primary versus 3:2 primary/mirror routing, with the filesystem-cache state controlled as closely as macOS permits. I’ll report decode tok/s plus physical read bytes/events for each device.

I’ll rebase this after #1456 and #1457 have landed, as requested.

trigger2k20:
Rebased #1455 onto current dev and added/pinned the whole-expert fallback behavior.

I also ran the requested cold A/B on an Apple M4 Max with 128 GB unified memory using GLM-5.3-Flash int4-g64. Both runs used the same prompt and settings (--ram 96 --cap 8 --ctx 8192 --temp 0 --ngen 256), with sudo purge before each run.

Cold A/B results

* Mirror OFF: 1.853 tok/s
* Mirror ON (60/40): 1.815 tok/s
* Mirror ON repeat while tracing I/O: 1.844 tok/s

I therefore don’t see a defensible throughput improvement from the mirror on this particular workload. The observed difference is within roughly 0–2% run-to-run variation, so I would not claim a speedup from this measurement.

The traced mirror-on run confirms that reads are actually distributed across both devices. These are filesystem read bytes observed by macOS fs_usage:

* Primary /dev/disk3s5: 109,736,882,176 bytes (102.20 GiB, 63.76%)
* Mirror /dev/disk11s1: 62,367,682,560 bytes (58.08 GiB, 36.24%)
* Total: 172,104,564,736 bytes (160.28 GiB)

This is reasonably close to the configured COLI_DISK_WEIGHTS=3,2 routing ratio.

Failure/fallback behavior

The existing storage-level mirror tests exercise divergent headers, short copies and missing mirror files. The GLM53 tests now also pin the whole-expert preflight/fallback semantics. If any shard required for the selected replica is unavailable or rejected, that expert read falls back to the primary copy rather than mixing pieces from different replicas.

Validation after the rebase/fallback change:

* test_st_mirror: OK
* test_glm53_mirror.py: 10/10 OK
* make -C c glm53 METAL=1: OK

Current PR head: 89eeb698e1dddb85bebb0d3d991ed6ebf43da7a4.

PR #1458, opened 2026-09-12
fix(build): clean up Clang compiler warnings

trigger2k20:
Clean up Clang compiler warnings

## Branch and commit

- Branch: `fix/clang-compiler-warnings`
- Commit: `e79aee8`
- Commit message: `fix(build): clean up Clang compiler warnings`
- Base used for development: `fd93c41aa6ae2c7d1cc1a1e2d6b79dbe6d341708` (`v1.10.2`)

## Summary

This change removes Apple Clang compiler warnings from the portable CPU/test build without globally suppressing warning classes.

The cleanup covers GLM-5.3, Qwen3.8 Edge adapter initialization, the expert-store registry/test initialization, and DeepSeek V4 CPU/GPU conditional compilation. The changes are intended to be warning cleanup only; no performance claim is made.

### GLM-5.3

- Explicitly initializes the `Mat.vk` field to `NULL` in the three affected matrix initializers.
- Keeps `hc_mult` declarations only in code paths where the value is actually used, avoiding unused-variable diagnostics in alternate/Segment builds.

### Qwen3.8

- Converts the `ColiEdgeAdapter` initializer to designated initialization.
- Initializes the `reserved_fn` array explicitly.
- Removes the scalar-brace/missing-field diagnostics and makes the adapter initializer less dependent on struct field order.

### Expert store

- Explicitly initializes the additional `ColiExpertStore` field in `test_expert_store_ops.c`.
- Makes `COLI_ESR_EXPORT` feature-detect `externally_visible` instead of unconditionally applying an attribute unsupported by Apple Clang.
- Retains the `used` attribute on GCC/Clang builds.

### DeepSeek V4

- Restricts `gpu_batch`, `cols`, and `gpu_compute` state to GPU-tier compilation where those values are consumed.
- Removes unused `scales`/`qdq` scratch allocations and their size accounting.
- Keeps the CPU-only path free of GPU-only unused-variable diagnostics.

## Compiler diagnostics addressed

The original Apple Clang build/test output contained warnings in these groups:

1. Three `glm53.c` `missing field 'vk' initializer` diagnostics.
2. One GLM-5.3 Segment-build `unused variable 'H'` diagnostic.
3. Qwen3.8 Edge adapter `braces around scalar initializer`.
4. Qwen3.8 Edge adapter `missing field 'reserved_fn' initializer`.
5. Expert-store test `missing field 'gpu' initializer`.
6. Four `expert_store_registry.c` `unknown attribute 'externally_visible' ignored` diagnostics.
[...]

PR #1462, opened 2026-09-13
feat(glm53): add Metal routed MoE acceleration

trigger2k20:
## Summary

This PR adds an opt-in Metal-accelerated routed-MoE path for GLM-5.3 Flash on Apple Silicon.

The goal is to accelerate the actual GLM-5.3 routed expert path — Gate → Up → clamped SwiGLU → Down → route-weighted scatter — while preserving the existing CPU implementation as the fallback path.

The implementation operates directly on the resident expert slots in Apple unified memory and keeps the existing CPU path unchanged when Metal is disabled or unavailable.

## What changed

- Added GLM-5.3 routed-MoE support to the existing Metal backend.
- Added a GLM-5.3-specific clamped SwiGLU Metal kernel.
- Added support for the grouped-int4-gs64 expert representation used by GLM-5.3 Flash.
- Added 16 KiB-aligned resident expert slots suitable for Metal registration.
- Added Metal registration of resident GLM-5.3 expert slots.
- Added runtime accounting for:
  - Metal MoE attempts
  - successful Metal MoE executions
  - CPU fallbacks
  - processed rows
- Preserved the existing CPU implementation as fallback.
- Added source-level regression tests and numerical CPU↔Metal correctness tests.

## Numerical validation

The clamped SwiGLU kernel was first validated independently against the CPU reference:

```text
n=16
limit=10

max_abs = 1.90734863e-06
max_rel = 6.09632238e-08

OK
```

A second oracle validates the complete routed-MoE path:

```text
Gate
  → Up
  → clamped SwiGLU
  → Down
  → route-weighted scatter
```

using the actual grouped-int4-gs64 expert representation.

Result:

```text
max_abs = 2.79396772e-09
max_rel = 2.77803934e-09
worst   = 138

OK
```

This provides an end-to-end numerical comparison of the Metal expert path against the existing CPU implementation.

## Runtime validation

Real GLM-5.3 Flash runs were also instrumented to verify that the Metal path was actually executed rather than silently falling back to CPU.

The expected invariant is:

```text
metal_moe_attempts == metal_moe_ok
metal_moe_fallback == 0
```

This was satisfied during the benchmark runs.

Across three 256-token benchmark runs, 44,702 routed-MoE Metal executions completed with zero CPU fallbacks.

## Performance

Test system:

```text
Apple M4 Max
128 GB unified memory
macOS
GLM-5.3-Flash
--ram 96
```

An initial controlled 32-token CPU/Metal comparison showed:

```text
CPU:    0.469 tok/s
Metal:  0.621 tok/s
```

approximately +32% decode throughput for that configuration.

After moving the complete routed-MoE path to Metal, expert residency (`--cap`) became an important tuning parameter.

32-token sweep:

```text
cap    Metal expert cache    decode
1       0.6 GB               1.215 tok/s
2       1.2 GB               1.941 tok/s
4       2.4 GB               2.045 tok/s
8       4.8 GB               2.121 tok/s
16      9.5 GB               2.146 tok/s
32     19.0 GB               1.796 tok/s
64     38.1 GB               1.679 tok/s
148    88.0 GB               1.304 tok/s
```

Longer runs show that the optimum depends on workload length.

128-token run:

```text
cap 8     1.666 tok/s
cap 12    1.631 tok/s
cap 16    1.555 tok/s
cap 20    1.584 tok/s
cap 24    1.599 tok/s
```

256-token run:

```text
cap 4     1.691 tok/s
cap 8     1.718 tok/s
cap 12    1.776 tok/s
```

The best longer run measured so far on this M4 Max configuration is therefore:

```text
--cap 12 → 1.776 tok/s
```

These measurements are intentionally not used to establish a universal default `--cap`. The optimum depends on the model, Apple Silicon generation, available unified memory, and workload.

## Validation against current dev

The branch was rebased onto current upstream `dev` before final validation.

Full test suite:

```text
Ran 855 tests in 57.355s

OK (skipped=72)
```

Additional validation included:

- GLM-5.3 Metal build
- source-level Metal tests
- clamped-SwiGLU CPU↔Metal oracle
- complete routed-MoE CPU↔Metal oracle
- real GLM-5.3 Flash inference
- runtime Metal execution counters
- zero observed CPU fallbacks
- `git diff --check`

## Scope

This PR is deliberately limited to GLM-5.3 Metal routed-MoE acceleration.

It does not change the default CPU execution path, model precision, or router semantics. Metal remains opt-in and the CPU implementation remains available as fallback.

Tested commit:

```text
a214824
feat(glm53): add Metal routed MoE acceleration
```

### DIFF
--- a/c/Makefile
+++ b/c/Makefile
@@ -1122,11 +1122,10 @@ NOCUDA_CFLAGS  = $(filter-out -DCOLI_CUDA -DCOLI_ANS,$(CFLAGS))
 NOCUDA_LDFLAGS = $(filter-out -lcudart -lstdc++ -lcuda -L$(CUDA_HOME)/lib64 \
                    -Wl$(comma)-rpath$(comma)$(CUDA_HOME)/lib64,$(LDFLAGS))
 
-# GLM-5.3-Flash: CPU puro, esperti in streaming dal contenitore int4 gs64.
-# Nessun oggetto CUDA/Vulkan/Metal fra le dipendenze perche' il motore non ne
-# ha ancora: quando li avra', questa riga somigliera' a quella di kimi_k3.
-glm53$(EXE): glm53.c cli_args.h st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h serve_poll.h quant.h hyper_connections.h delta_attention.h sparse_index.h vision_tower.h backend_vulkan.h edge_adapter_internal.h edge_adapters.h edge_runtime.h segment_adapter_internal.h segment_adapters.h segment_runtime.h $(VK_OBJ) $(VK_SPV)
-	$(CC) $(CFLAGS) glm53.c $(VK_OBJ) -o glm53$(EXE) $(LDFLAGS)
+# GLM-5.3-Flash: routed experts stream from the int4-gs64 container.
+# METAL=1 accelerates resident matrices and routed MoE; CPU remains fallback.
+glm53$(EXE): glm53.c cli_args.h st.h json.h tok.h tok_unicode.h tok_unicode_o200k.h compat.h serve_poll.h quant.h hyper_connections.h delta_attention.h sparse_index.h vision_tower.h backend_metal.h backend_vulkan.h edge_adapter_internal.h edge_adapters.h edge_runtime.h segment_adapter_internal.h segment_adapters.h segment_runtime.h $(METAL_OBJ) $(VK_OBJ) $(VK_SPV)
+	$(CC) $(CFLAGS) glm53.c $(METAL_OBJ) $(VK_OBJ) -o glm53$(EXE) $(LDFLAGS)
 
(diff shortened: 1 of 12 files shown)


JustVugg:
Reviewed and merging. What I checked, and what I could not:

**Checked.** The non-Metal path is untouched by construction: `g_metal_ready` is set only when `COLI_METAL=1` is in the environment on a Metal build, the slot allocator keeps the zero-copy mmap fast path whenever it is not, and on Linux/Windows the branch does not even compile in. Built glm53 here without Metal: clean apart from the `missing initializer for field 'vk'` warnings that `dev` already has (#1458 fixes those). The 20 source-level tests pass here. The `coli run` change is correct against the engine's own argv contract: a bare numeric argument in any position is the cap and zero means "decide it yourself", so `cap_for_launch(a.cap, e, 0)` reaches the engine as `0` by default and as `N` when the user asked, which also fixes an explicit `coli run --cap N` being silently ignored on the one-shot path. CI 28/28, including Windows and macOS.

**Not checked, because I have no Apple hardware.** The Metal kernels themselves. What stands in for that is your two oracles (2.8e-09 max abs on the full gate/up/SwiGLU/down/scatter path against the CPU implementation) and the runtime invariant `attempts == ok, fallback == 0` across 44,702 executions. The two C tests (`test_glm53_clamped_metal.c`, `test_glm53_moe_block_metal.c`) have no Makefile rule, so CI does not build them anywhere; that is fine for a Linux `make check` but it means they only run when someone with a Mac runs them by hand. Worth a follow-up that wires them as a `metal-check` target the way kimi_k3 does.

Since this is opt-in and the blast radius is exactly the people who set `COLI_METAL=1` on glm53, which today is nobody, it goes into the release. The cap sweep in the description (the optimum moving with workload length) is the kind of datapoint `docs/benchmarks.md` wants; a PR adding the table there would be welcome.

Issue #1464, opened 2026-09-13
[Bug]: RTX 3080 sm_86, WSL2 — "hybrid batched block failed in MoE" on every prompt


RichBigDreamer:
### Commit or version

fd93c41

### Environment

GPU: NVIDIA RTX 3080 10GB, sm_86 (Ampere)
OS: Windows 11, WSL2 Ubuntu
Colibri: v1.10.2 (fd93c41)
Model: DeepSeek-V4-Flash (160GB official checkpoint)
CUDA: 13.3 (Windows) / 12.8 (WSL2)

### Reproduction steps

1. Build deepseek-v4 engine with CUDA=1 via WSL2
2. Run: CUDA_DENSE=1 python3 coli chat --model /path/to/deepseek-v4-flash --ram 32
3. Type any prompt

### Expected behavior

Model generates a response

### Actual behavior and logs

```shell
request failed: hybrid batched block failed in MoE

Startup log shows GPU loading correctly:
v4_gpu tier=dense-matvec device=0
NVIDIA GeForce RTX 3080 10.7 GB sm_86

But every inference attempt fails immediately with hybrid batched block error regardless of flags tried: CUDA_DENSE=1, COLI_CUDA_MOE_BATCH=0, COLI_CUDA_ATTN_BATCH=0, DSV4_HYBRID=0

RTX 3080 sm_86, WSL2 — "hybrid batched block failed in MoE" on every prompt

GPU: NVIDIA RTX 3080 10GB, sm_86 (Ampere)
OS: Windows 11, running via WSL2 Ubuntu
Colibri: v1.10.2 (fd93c41)
Model: DeepSeek-V4-Flash (160GB official checkpoint)
CUDA: 13.3 (Windows) / 12.8 (WSL2)

GPU loads successfully — startup log shows v4_gpu tier=dense-matvec device=0. But every inference attempt fails with:

request failed: hybrid batched block failed in MoE

Tried the following — all fail with same error:

CUDA_DENSE=1
COLI_CUDA_MOE_BATCH=0
COLI_CUDA_ATTN_BATCH=0
DSV4_HYBRID=0
All combinations of the above

Any workaround for Ampere 10GB?
```

JustVugg:
Thanks for the precise report. I can tell you what that message is, and it is not what it looks like.

[...]

PR #1468, opened 2026-09-13
fix(deepseek-v4): the MoE step says why it failed, instead of "failed in MoE"

JustVugg:
For #1464.

`hybrid batched block failed in MoE` is the generic tail of the layer block. `moe_token_pipeline` and `v4_moe_batch_union` can fail in some thirty places (an expert that would not read, an upload that would not land, the GPU expert group refusing, a routing table missing, a plain `malloc`) and every one returned a bare `-1`, so the block could not tell an out-of-VRAM card from a bad read over `/mnt/c`, and neither could the reporter after a day of flags.

Every failure site now records a reason (layer, expert where there is one, the step) through `moe_fail()`, thread-local and cleared on entry, and both block tails append it:

```
hybrid batched block failed in MoE: layer 12: the GPU expert group of 6 experts refused (a CUDA allocation or launch failed; check nvidia-smi for free VRAM)
block computation failed in MoE: layer 3: reading expert 137 from the expert store failed
```

No behaviour change on any success path: only the text in the error buffer differs. All of it lives in the `COLI_V4_UNIT_BLOCK_HYBRID` unit, which needed its own `<stdarg.h>`.

**Gate:** `tests/test_v4_moe_reason_source.py` pins the shape (no silent `-1` left in the two functions except the job markers, reason cleared on entry, both tails read it); it fails on the tree before this commit. The tiny V4 oracle stays token-exact: target long, CLI, serve, prefix reuse.

Not part of 1.11.0 (#1467 is already open); this ships in the next release.


skhari1989-cmyk (commented):
Thank you for the support and the resolution.

PR #1485, opened 2026-09-14
fix(coli): coli run on an OLMoE model never told the engine which model

kevin9327:
## Summary

`coli run --model <OLMoE directory> "prompt"` does not run: the engine answers that it has no model and the launcher exits 1.

```
$ coli run --model olmoe_tiny_c "abc"
colibri: this is the OLMoE engine, and it was started without a model.
The engine is not the program you run directly -- the launcher is:
...
(exit 1)
```

`olmoe.c` takes its model directory from `SNAP` and nothing else (`if (!snap) { coli_print_launcher_help("OLMoE"); return 1; }`). The one-shot branch in `cmd_run` builds the child's environment with `env_for_engine()`, which only sets `SNAP` for GLM (through `env_for()`), and it passes no model on argv either. So the engine never learns the directory. `chat`, `serve` and `web` go through the gateway, which sets `SNAP` itself, which is why only `run` is affected. It works by accident only when `SNAP` is already exported in the shell.

The change: set `SNAP` to the model directory in that branch, as an absolute path, the way the `deepseek_v4` and `glm53` one-shot branches pass theirs.

## Reproduced

Tiny OLMoE container built as `olmoe-tiny-check` builds it, `SNAP` not set in the shell, engine built from this tree:

`dev` @ `e92c139`:

```
colibri: this is the OLMoE engine, and it was started without a model.
exit=1
```

this branch:

```
olmoe chat — SNAP=...\olmoe_tiny_c, ctx=4096, TEMP=0.70, NUCLEUS=0.95
resident weights loaded in 0.0s | RSS after load: 0.00 GB
> <generated text>
exit=0
```

## Tests

`OlmoeRunTest` (new, in `tests/test_launcher_dispatch.py`): calls `main()` with `coli run --model <dir> hello`, so the real argument parser and `cmd_run` run, with `resolve_model` returning the registry's OLMoE family and `subprocess.run` mocked. It removes any exported `SNAP` first, since the child would inherit it and hide the bug, and asserts the engine is launched with `SNAP` equal to the model directory, `CHAT=1`, and the prompt on stdin. On `dev`:

```
FAIL: test_the_engine_is_given_the_model_directory (tests.test_launcher_dispatch.OlmoeRunTest.test_the_engine_is_given_the_model_directory)
AssertionError: None != 'C:\\Users\\...\\olmoe' : the engine was launched without its model directory
```

Passes on this branch. The module runs in `make check` and in the Windows-focused Python job.

## Validation

- [x] `make -C c check`

Not a CUDA change and no performance claim. Windows 11 x86-64, WinLibs GCC 16.2.0, Python 3.11:

- `python -m unittest tests.test_launcher_dispatch`: 8/8, both with the parent in UTF-8 mode and outside it (cp949).
- `make -C c check`: C suite passes; Python `Ran 986 tests`, `errors=6, skipped=90`: the same six `test_cuda_test_makefile` errors untouched `dev` shows on this box (they call a `make` binary it does not have, and pass with one on `PATH`). The new test is the one extra.

## Compatibility

---


JustVugg:
Thank you: your PR named the bug first and its diagnosis is the one in the fix. On today's dev the test here fails for a reason unrelated to the fix (the mocked subprocess.run returns an object without .stdout, which the current one-shot path reads), and #1503, which sets SNAP for every engine from the same place and is green on the current dev, carries the same change. Closing this one in favour of #1503, with your report credited in #1501; the three other PRs of yours from this batch are merged (#1483, #1484, #1489).

Issue #898, opened 2026-08-08
[Bug]: `coli tune` banners DeepSeek V4 but calibrates with the GLM engine and fails


Blakeolson21:
## Commit or version

v1.5.0 (`8f512fc8c2f48ffa18cd624cd4a5bcaae4a4abfc`). The code path is unchanged on current `dev` (`98cab3b` at the time of writing): reproduced there too.

## Environment

Ubuntu 26.04 LTS (kernel 7.0.0-28-generic), AMD Ryzen 7 6800H (8C/16T), 32 GB RAM, no GPU in use, gcc 15.2.0, weights on NVMe (WD Green SN350 1 TB, ext4).

## Reproduction steps

```sh
# build the engines
make -C c
make -C c deepseek-v4

# point tune at a DeepSeek-V4-Flash weights directory
cd c
./coli tune --model /path/to/DeepSeek-V4-Flash-0731
```

## Expected behavior

Either the tuner calibrates with the DeepSeek V4 engine, or `coli tune` refuses the model up front with a clear message that tuning is not wired for this engine yet — the same way `coli run` already dispatches per architecture.

## Actual behavior and logs

The banner correctly identifies the model ("DeepSeek V4 Flash · 284B MoE"), but calibration launches the GLM C engine, which rejects the V4 config and the command exits 1:

```
$ ./coli tune --model /srv/models/DeepSeek-V4-Flash-0731
  [PLAN] RAM 26.6 GB · cap 0/layer · CPU

    (\      colibri v1.5.0
     )·>     tiny engine, immense model
    / \      DeepSeek V4 Flash · 284B MoE · 167 GB on disk
             tune · measured execution profile

  ──────────────────────────────────────────────────────────
  fixed replay: 16 tokens · 2 run(s) per candidate
  only quality-preserving scheduling knobs are eligible

tuning failed: calibration failed (1)
== GLM C engine (glm_moe_dsa), cache=8 experts/layer | experts@8-bit dense@8-bit | idot: avx2 ==

[OMP] hot-thread tuning: re-exec once (COLI_NO_OMP_TUNE=1 to skip)
this engine requires n_group=1 (GLM-5.2)

$ echo $?
1
```

(stdout and stderr captured into one file, so the child engine's banner and the `n_group=1` line appear after the parent's failure message; the GLM engine lines come from the calibration subprocess.)

[...]

PR #1489, opened 2026-09-14
fix(coli): coli bench ran the GLM engine on any model

kevin9327:
## Summary

`coli bench` on any model that is not GLM-5.2/5.3 starts the GLM engine on it and fails with an engine error that does not say what went wrong:

```
$ coli bench --model olmoe_tiny_c hellaswag --limit 1
[hellaswag] 1 questions
total requests: 2 (answer options)
running: ...\c\colibri.exe 64
  [engine] this engine requires n_group=1 (GLM-5.2)
ENGINE ERROR (exit 1)
```

The launcher's banner for the same run reads `OLMoE · 7B · 1 MB on disk`.

`tools/eval_glm.py` scores through the log-likelihood protocol of the `colibri` binary (`--glm GLM`), which no sibling engine speaks. `cmd_bench` passes it every model regardless of family, so the launcher names OLMoE in its banner and then hands the container to the GLM engine. It is the shape #898 fixed for `coli tune` ("calibration launched the GLM engine, which died on 'this engine requires n_group=1'"), one command over.

The change: `cmd_bench` resolves the family and refuses anything other than `glm` before a dataset is fetched or a process is started, with the wording `coli run` already uses for a family it cannot drive:

```
coli bench is not wired for OLMoE; its harness scores the GLM-5.2/5.3 engine only
```

`glm` (GLM-5.2/5.3, the `colibri` binary) is the family `eval_glm.py` drives and is unaffected. Every other family is refused, including GLM-5.3-Flash, which has its own `glm53` engine.

## Reproduced

Tiny OLMoE container built as `olmoe-tiny-check` builds it, one local `hellaswag.jsonl` line so nothing is downloaded, engines built from this tree. The output above is `dev` @ `e92c139`. This branch:

```
coli bench is not wired for OLMoE; its harness scores the GLM-5.2/5.3 engine only
exit=1
```

and no engine process is started.

## Tests

`BenchDispatchTest` (new, in `tests/test_launcher_dispatch.py`): runs `cmd_bench` on a directory whose `config.json` says `olmoe`, with the model check, banner and `subprocess.call` mocked, and asserts the refusal and that nothing is launched. On `dev`:

```
FAIL: test_a_non_glm_model_is_refused_before_the_glm_engine_runs (tests.test_launcher_dispatch.BenchDispatchTest.test_a_non_glm_model_is_refused_before_the_glm_engine_runs)
AssertionError: 'coli bench is not wired for OLMoE' not found in '0'
```

Passes on this branch. `test_bench_and_convert_use_windows_project_python` used a model path with no `config.json`; it now resolves to the GLM family explicitly and still checks the same thing.

## Validation

- [x] `make -C c check`

Not a CUDA change and no performance claim. Windows 11 x86-64, WinLibs GCC 16.2.0, Python 3.11:

- `python -m unittest tests.test_launcher_dispatch`: 8/8, both with the parent in UTF-8 mode and outside it (cp949).
- `make -C c check`: C suite passes; Python `Ran 986 tests`, `errors=6, skipped=90`: the same six `test_cuda_test_makefile` errors untouched `dev` shows on this box (they call a `make` binary it does not have, and pass with one on `PATH`). The new test is the one extra.

## Compatibility

---


codeCraft-Ritik (commented):
Refusing unsupported families upfront before fetching datasets or spawning binaries mirrors the pattern from #898 well.

How does cmd_bench handle unrecognized model families or malformed/missing config.json files?

If a custom model directory lacks a model_type / family field in its configuration, does it cleanly fail with the refusal message or will it raise an uncaught KeyError / FileNotFoundError?

Ensuring that any unresolved family falls through to this refusal path (or a dedicated 'unknown model family' error) rather than crashing unhandled will keep CLI output clean

PR #1492, opened 2026-09-14
docs: add Japanese REAMDE

eltociear:
## Summary

I created Japanese translated README.

## Validation

- [ ] `make -C c check`

## Compatibility


JustVugg:
Thank you for the translation. Retargeted to dev (every PR lands there; main only receives releases). Two things before it can go in: the file mixes versions (it names v1.11.0 in the banner but the install line still says colibri-v1.8.0-linux-x86_64.tar.gz), so please translate from the current README on dev so the version, the Discord invite and the nine-family list match; and once that is done the release flow needs the version line in this file too (we bump it in every README at release time), so please keep the banner line in the same shape as README.md. CI will run on the update.

eltociear:
Thanks for the review! I rebuilt the branch on top of `dev` as a single commit (the conflicts are gone) and re-synced `README.ja.md` with the current `README.md` on dev:

- the Discord invite and the nine-family list match dev
- the banner line has the same shape as README.md (`🐦 colibri v1.11.0 — GLM-5.2 · 744B MoE · int4 · streaming CPU`), so the release bump will catch it
- brought over the recent dev changes: the `docs/benchmarking.md` link and table row, the single-file `out-mtp-00000.safetensors` note in the MTP head check, and the `expert_ffn.h` entry in the tree
- added the 日本語 link to the other four READMEs

One note on the versions: the install line in `README.md` on dev also still says `colibri-v1.8.0-linux-x86_64.tar.gz` while its banner says v1.11.0. I kept the Japanese file identical to the English one, so if that line should change, it probably needs updating in every README.

JustVugg:
Merged, thank you. One follow-up, since `dev` moved while this was open: the English README now has the 1.12.0 changes (the "Brio mode: ask a closed question" subsection after the first `coli web` block, the new dashboard screenshot and the brio screenshot, the docs table row for `docs/brio.md`). `README.ja.md` predates them. If you can bring it in line with the current `README.md` on `dev` in a small follow-up, the Japanese page will match the release.

PR #1496, opened 2026-09-14
coli run: set SNAP for the OLMoE one-shot path

tuhinbatra1990:
## Problem

`coli run --model <olmoe container> "prompt"` always fails:

```
colibri: this is the OLMoE engine, and it was started without a model.
The engine is not the program you run directly -- the launcher is:
...
```

while `coli doctor`, `coli chat` and `coli serve` on the same directory all work.

`olmoe.c` reads its model directory from the `SNAP` environment variable and exits 1 without it (`main()` in `c/olmoe.c`). `env_for()` sets `SNAP` for glm, and the gateway (`openai_server.py`) sets it for chat/serve/web, but the olmoe branch of `cmd_run` builds its environment from `env_for_engine()`, which never sets it. Same defect shape as #805: a mechanism that reaches one path and not its sibling.

## Fix

Set `SNAP=a.model` on the olmoe branch of `cmd_run`, matching `env_for()`. One line plus a comment.

## Test

Extended the existing `test_olmoe_run_uses_its_engine_and_writes_one_prompt_line` in `c/tests/test_v4_cli.py` to assert `SNAP` is forwarded to the engine.

- unpatched tree: the new assertion fails with `KeyError: 'SNAP'`
- patched: `python -m unittest tests.test_v4_cli` → 18 tests OK

End-to-end on Windows 11 (i7-1255U, 32 GB, CPU only) with the v1.11.0 prebuilt `olmoe.exe` and a container converted with `tools/convert_olmoe_merged.py --repo allenai/OLMoE-1B-7B-0125-Instruct`:

```
$ python coli run --model D:\models\olmoe --ngen 30 --temp 0 "Name three primary colors."
> The three primary colors are red, blue, and yellow. ...
```

No manual `SNAP=` export needed any more. The rest of the Python suite is unchanged by this PR (the modules that fail on my box, `test_cuda_test_makefile`, `test_glm53_*`, `test_kimi_usage_cli`, `test_openai_tools_v41_e2e`, fail identically on the unpatched tree and need `make`/a compiler or a UTF-8 default locale).


JustVugg:
Thanks, this is the right fix, and #1485 arrived with the same one-line change and a launcher test a few hours earlier; #1503 then moves the SNAP assignment into env_for_engine for every engine. Closing this one as a duplicate so the credit goes to the first PR; nothing wrong with yours.

Issue #1502, opened 2026-09-14
Experiment report: GLM-5.2 on AMD Strix Halo (Ryzen AI MAX+ 395, 128 GB unified) — 0.26-0.38 tok/s, CPU-bound not disk-bound, and the Vulkan expert tier double-counts unified memory


huppiflupp:
End-to-end datapoint for **GLM-5.2 on an AMD Strix Halo APU** (Ryzen AI MAX+ 395, 128 GB
unified memory, Radeon 8060S / gfx1151), with the Vulkan backend. Posting the negative
results too, as CONTRIBUTING asks — the headline is that **this machine class is
CPU-compute-bound, not disk-bound**, which is the opposite of what `coli doctor` reports,
and the Vulkan expert tier cannot fix it because on unified memory it *duplicates* weights
instead of sharing them.

**TL;DR — 0.26–0.38 tok/s decode.** It runs, it answers correctly, it terminates cleanly.
It is roughly 5x slower than the project's own ~1.8 tok/s reference for a 128 GB CPU-only
desktop, despite this box having a GPU and comparable RAM.

## 1. Environment

| | |
|---|---|
| commit | `f028d26b422144ed4a69ad9aeaee2553ce0f9572`, `coli info` → v1.11.0 |
| CPU | AMD Ryzen AI MAX+ 395, 16 c / 32 t (Zen 5), 124 GB usable unified RAM |
| GPU | Radeon 8060S (gfx1151, RDNA 3.5), RADV, Vulkan 1.4.354 — **unified memory, no dedicated VRAM** |
| storage | Lexar SSD ARES 2 TB NVMe, btrfs (`compress=zstd:1`), model dir set `chattr +C` before download |
| OS | Nobara Linux 44, kernel 7.2.0-202.fc44 |
| model | `mastouri/GLM-5.2-colibri-int4-g64-with-int8-mtp`, 406 GiB, MTP head verified int8 |
| build | `make -C c glm VK=1 ARCH=native` (gcc 16.2.1) — built clean, no patches |

MTP-head check: this container ships the head as a **single** `out-mtp-00000.safetensors`
(9,959,321,520 B), not the three files the README's `ls -l <model>/out-mtp-*` check expects.
The sizes sum to within 768 B of the documented int8 triple, and the safetensors header shows
777 U8 tensors at exactly 1 byte/element — so it is int8. The README check may want updating.

## 2. What it does

```bash
[...]

PR #1505, opened 2026-09-14
gpu: gate grouped_s4_wmma on the s4 fragment type and sm_75, not on WMMA (#1499)

### LINKED ISSUE #1499: HIP/ROCm: COLI_CUDA_TC_INT4 dispatches an empty grouped_s4_wmma kernel, silently returning stale expert output (gfx1151)
## Summary

On a Linux HIP/ROCm build with the rocWMMA headers installed, `COLI_CUDA_TC_INT4=1`
opens the `grouped_s4_wmma` branch while the kernel body is compiled out. The three
launches succeed, write nothing, and `ctx->y` silently keeps stale data from the
previous call. No error is raised.

This is exactly the failure mode the comment at `c/backend_cuda.cu:1906` warns about,
but the guard it describes does not cover this case: it tests rocWMMA *availability*,
not the `__CUDA_ARCH__ >= 750` threshold the kernel body actually requires.

The in-tree test that covers this path (`tests/test_backend_cuda.cu:399-402`) **passes**
on the affected configuration — for the wrong reason. Details in "Why CI does not catch
this" below.

## Environment

| | |
|---|---|
| commit | `f028d26b422144ed4a69ad9aeaee2553ce0f9572` (2026-09-13), `coli info` reports v1.11.0 |
| GPU | AMD Radeon 8060S (Ryzen AI MAX+ 395, Strix Halo), **gfx1151** |
| HIP | 7.1.52802-9999, clang 20.0.0.rocm |
| rocWMMA | rocwmma-devel 7.1.0-5.fc44 |
| OS | Nobara Linux 44, kernel 7.2.0-202.nobara.fc44.x86_64 |

## Mechanism

1. `c/backend_gpu_compat.h:38-39` — rocWMMA header found, so:
   ```c
   #define COLI_GPU_HAS_WMMA        1
   #define __CUDA_ARCH__            700
   ```
2. `c/backend_cuda.cu:707` — the kernel body is guarded by `#if __CUDA_ARCH__ >= 750`.
   With 700, **the body is compiled out**. The header comment at lines 13-15 already
   notes this ("the grouped_s4_wmma kernel has a stricter guard of `__CUDA_ARCH__ >= 750`,
   so it stays a no-op even with `__CUDA_ARCH__=700`").
3. `c/backend_cuda.cu:1910` — the host-side gate:
   ```c
   tc=tc&&!pin_small_batch&&COLI_GPU_HAS_WMMA&&all_s4&&D%32==0&&I%32==0&&D%8==0&&I%8==0;
   ```
   `COLI_GPU_HAS_WMMA` is 1, so with `all_s4` the branch is taken.
4. `c/backend_cuda.cu:1927`, `:1928`, `:1931` — three empty launches. `ctx->y` is never
   written; the caller receives whatever the buffer held before.

Step 2 and step 3 disagree: the body is keyed on `>= 750`, the dispatch on rocWMMA presence.

## Evidence

**(a) `__CUDA_ARCH__` really is 700 device-side, so the body really is absent.**
A probe including `backend_gpu_compat.h`, built with the project's own HIP flags:

```
HOST:   COLI_GPU_HAS_WMMA=1   __CUDA_ARCH__=700
DEVICE: __CUDA_ARCH__=700     body-compiled-in(>=750)=0
```

**(b) The gate opens anyway.** One `fprintf` after `c/backend_cuda.cu:1910`, running the
stock `tests/test_backend_cuda.cu`:

```
[dbg] env_tc=0 pin_small=0 has_wmma=1 all_s4=0 D=4   I=2   -> tc=0
[dbg] env_tc=0 pin_small=0 has_wmma=1 all_s4=1 D=32  I=32  -> tc=0
[dbg] env_tc=1 pin_small=0 has_wmma=1 all_s4=1 D=32  I=32  -> tc=1   <-- empty kernel dispatched
[dbg] env_tc=1 pin_small=1 has_wmma=1 all_s4=1 D=32  I=32  -> tc=0
```

**(c) The output is stale, not computed.** Poisoning the destination with `-12345.0`
immediately before the call at `tests/test_backend_cuda.cu:401`, then comparing:

```
[dbg] bit-identical to scalar4: YES
tensor4[0..3] = -0.118602 -0.0417721 -0.0459986 0.0613381
scalar4[0..3] = -0.118602 -0.0417721 -0.0459986 0.0613381
```

A real W4A4 int4xint4 tensor-core GEMM cannot reproduce the fp32 scalar path bit-for-bit —
that is precisely why the assertion uses a 0.30 relative-RMS tolerance. Bit-equality proves
the values were never recomputed: they are the previous scalar call's result, still resident
in `ctx->y`.

## Reproduction

```bash
git clone https://github.com/JustVugg/colibri && cd colibri/c
# note: -lamdhip64 is required manually, see the second finding below
/usr/bin/hipcc -O3 -std=c++17 -x hip --offload-arch=gfx1151 -w -fPIE \
  backend_cuda.cu tests/test_backend_cuda.cu -o backend_cuda_test \
  -L/usr/lib64 -lamdhip64
./backend_cuda_test; echo "exit=$?"     # exits 0 — but see below
```

To observe the defect, apply either instrumentation:

```c
/* after c/backend_cuda.cu:1910 */
if(getenv("COLI_DEBUG_TC")) fprintf(stderr,"[dbg] env_tc=%d pin_small=%d has_wmma=%d all_s4=%d D=%d I=%d -> tc=%d\n",
    getenv("COLI_CUDA_TC_INT4")?atoi(getenv("COLI_CUDA_TC_INT4")):0,pin_small_batch,(int)COLI_GPU_HAS_WMMA,all_s4,D,I,tc);

/* before the call at c/tests/test_backend_cuda.cu:401 */
for(int i=0;i<64;i++)tensor4[i]=-12345.f;
/* ...and after it: */
std::fprintf(stderr,"[dbg] bit-identical: %s\n", std::memcmp(tensor4,scalar4,sizeof(tensor4))?"NO":"YES");
```

## Why CI does not catch this

Two independent reasons, and the first hides the second:

1. **`make hip-test` does not link.** The `cuda-test` target (`c/Makefile:898-901`) does not
   add `-lamdhip64` under `HIP=1`:
   ```
   ld.lld: error: undefined symbol: hipStreamSynchronize
   ld.lld: error: undefined symbol: hipHostMalloc
   ld.lld: error: undefined symbol: hipEventCreate
   ```
   So the kernel test suite cannot be executed on Linux ROCm at all. (`gpu-compile` only
   compiles, it never runs the binary.)

2. **Even when linked manually, the assertion is self-satisfying.** The check at
   `tests/test_backend_cuda.cu:402` compares `tensor4` against `scalar4`, and `scalar4`
   was produced by the immediately preceding `coli_cuda_expert_group` call at line 390 —
   which used the same device buffer. The stale data the empty kernel leaves behind *is*
   the reference value. The test compares the buffer against itself.

Poisoning the destination buffer before the call (as above) makes the test meaningful; a
regression test for this should also assert that the tensor-core result is **not** bit-identical
to the scalar path, since a lossy W4A4 path must differ.

## Suggested fix

Gate the dispatch on the same threshold the kernel body uses, rather than on rocWMMA
availability. rocWMMA exposes no sub-byte fragment types at all (there is no `s4` / `int4`
equivalent to `nvcuda::wmma::experimental::precision::s4` anywhere in its headers), so under
HIP this is a hard zero:

```c
/* c/backend_gpu_compat.h, in the rocWMMA branch next to the __CUDA_ARCH__ 700 define */
#define COLI_GPU_HAS_S4_WMMA 0   /* rocWMMA has no sub-byte fragments; 700 < 750 */
/* ...and 1 (or a compute-capability test) on the CUDA side */

/* c/backend_cuda.cu:1910 */
tc = tc && !pin_small_batch && COLI_GPU_HAS_WMMA && COLI_GPU_HAS_S4_WMMA && all_s4 && ...
```

## Scope

- `COLI_CUDA_TC_INT4` is opt-in and off by default, so this is not a default-path hazard.
- Requires `fmt=2` (per-row int4) experts. `fmt=4` (grouped/gs64) keeps `all_s4=0` and is unaffected.
  - The **GLM engine is reachable**: `c/colibri.c:1902` promotes `fmt=2` to `fmt=4` only when group
    scales are detected, so a per-row int4 container stays at `fmt=2`.
  - **Qwen3.6 is immune**: `c/qwen36_tier.c:551` uploads any int4 container as `wfmt=4` regardless
    of container, so `all_s4` is never true there.
- The consequence is silently wrong expert output rather than a crash, which is why it seems
  worth reporting despite the narrow trigger — the README states placement must change speed,
  never semantics.

## Second, related finding

`make hip-test` / `make cuda-test HIP=1` is broken on Linux ROCm (missing `-lamdhip64`, quoted
above). Adding the HIP runtime to the test link line makes the suite build and run on gfx1151;
with that and the poisoning fix, the suite would have caught the issue above. Happy to split
this into its own issue if preferred.

JustVugg:
Closes #1499.

Under HIP with rocWMMA installed, `COLI_GPU_HAS_WMMA` is 1 and `__CUDA_ARCH__` is defined 700, so the `COLI_CUDA_TC_INT4` dispatch took the `grouped_s4_wmma` branch while the kernel body (guarded by `__CUDA_ARCH__ >= 750`, and using a sub-byte fragment type rocWMMA does not have) was compiled out: three empty launches, `ctx->y` kept the previous call's output, no error. The in-tree test passed against its own stale buffer.

- `backend_gpu_compat.h`: `COLI_GPU_HAS_S4_WMMA` (0 in both HIP branches, 1 under CUDA).
- `backend_cuda.cu`: the gate requires that flag and `compute_major*10+compute_minor >= 75`, mirroring the body's own guard, instead of WMMA availability alone. `COLI_CUDA_TC_INT4` stays opt-in and off by default.
- `tests/test_backend_cuda.cu`: the destination is poisoned before the tensor-core call, so an empty kernel now fails the 0.30 RMS check instead of matching itself.
- `Makefile`: the GPU test binaries link `-lamdhip64` under `HIP=1` (`GPU_TEST_LIBS`, empty for CUDA), so `make hip-test` builds on ROCm; the reporter's second finding.

No AMD hardware here: the CUDA side is unchanged in behaviour on sm_75+ (the flag is 1 and the capability check matches the body's guard), the CI `gpu-compile` leg compiles it, and @huppiflupp offered to verify on gfx1151.


huppiflupp:
Ran this on real gfx1151. **The s4 gate fix verifies: `backend_cuda_test` passes.**

## Hardware / toolchain

Ryzen AI MAX+ 395 / Radeon 8060S (gfx1151, RDNA 3.5), ROCm 7.1 / HIP 7.1.52802,
`rocwmma-devel` 7.1.0 **installed** (so `COLI_GPU_HAS_WMMA` is 1 — the condition that
triggered the original issue), Nobara 44, kernel 7.2.0-202.fc44.

```
make hip-test ROCM_HOME=/usr HIPCC=/usr/bin/hipcc HIP_ARCH=gfx1151
```

## Result

| test | result |
|---|---|
| `backend_cuda_test` (carries the s4 gate + poisoned-destination check) | **pass** |
| `ragged_attention_test` | pass |
| `cuda_fmt_trap_test` | pass |
| `fp8_warp_test` | pass — 0 mismatches, mutated LUT entry correctly bites (1), S-invariance 0, NaN injection 0 |
| `absorb_determinism_test` | pass — `batch_differing=0/145`, `ragged_differing=0/29` |
| `fp8_cuda_test` | pass **after a one-line include guard**, see below |
| `weights_owned_test` | pass — 8 injected H2D failures, free-mem delta 0 bytes |
| `mxfp4_cuda_test` | **does not build under HIP**, see below |

The link fix works — `-L/usr/lib -lamdhip64` is on every test link line now, and the suite gets
much further than before (previously it died immediately on `undefined symbol:
hipStreamSynchronize`).

Incidental confirmation that the gate is doing the right thing rather than the kernel having
become live: the HIP build still emits `unused parameter 'y' / 'x' / 'xscale' / 'desc' / 'K' /
'O' / 'which'` for `grouped_s4_wmma` at `backend_cuda.cu:705-706`, i.e. the body is still
compiled out — and the test now passes anyway, which is exactly the intent.

## Two remaining HIP build gaps in the test files (not in your fix)

**1. `#include <cuda_runtime.h>` before the compat header.** Six test/bench files do this and so
never see the HIP mapping:

```
tests/test_fp8_cuda.cu      tests/test_router_cuda.cu    tests/test_grouped_g4_cuda.cu
tests/test_mxfp4_cuda.cu    tests/bench_dsv4_deepgemm.cu tests/bench_dsv4_mxfp8.cu
```

`__HIP_PLATFORM_AMD__` is not yet defined at that point; `__HIPCC__` is. One-line guard, verified
to build and pass on this box:

```c
#if !defined(__HIPCC__)
#include <cuda_runtime.h>   /* under HIP the compat header in backend_cuda.cu provides these */
#endif
```

With that, `fp8_cuda_test` builds and reports `fp8 oracle: ... 0 mismatches / fp8 API: ...
0 mismatches / OK`.

**2. `test_mxfp4_cuda.cu` needs more than the include guard.** After it, two further errors:

```
tests/test_mxfp4_cuda.cu:73,74:  error: no matching function for call to 'isinf'
                                 (ambiguous against HIP's __DEVICE__ isinf(float)/isinf(double))
tests/test_mxfp4_cuda.cu:206:    error: use of undeclared identifier 'cudaGetDeviceCount'
tests/test_mxfp4_cuda.cu:206:    error: use of undeclared identifier 'cudaSuccess'
```

`cudaGetDeviceCount` and `cudaSuccess` have no mapping in `backend_gpu_compat.h`. I stopped here
rather than guess at the intended fix — happy to test a patch. Note the Makefile comment already
calls MXFP4 "expected to fail" on some hosts, so this may be known and acceptable.

## Offer

The box stays available. If you want `make hip-test` green end to end on gfx1151, send a patch
for the two items above and I will run it. I can also re-run anything else that needs AMD
silicon — this is the same machine as JustVugg/colibri#1502.

PR #1513, opened 2026-09-14
vulkan: reserve the expert tier's RAM on integrated GPUs (#653 for the VK path)

huppiflupp:
Fixes the unified-memory double-accounting from #1502 §4. Measured on the same box: **2.8x faster on GLM-5.2**, on a configuration that previously died during init.

## The defect

On an integrated GPU the Vulkan expert tier's `HOST_VISIBLE|DEVICE_LOCAL` allocation is the same physical memory `expert_avail()` and `cap_for_ram()` hand to the pin set and the LRU. Neither knows about the other.

#653 already fixed this for the CUDA tier, by subtracting the placed bytes from `g_mem_avail_boot` afterwards. **That approach cannot work for the Vulkan tier**, and this is the part I had wrong in #1502 before reading the init order: `vk_registry_fill()` runs at the *end* of init, long after both budget decisions. The old log shows it plainly:

```
[VK] expert tier active: ... (budget 3000)     <- count known
[PIN] auto: budget 94.7 GB -> pinning 47.3 GB  <- decided
[RAM_GB=113.4 auto] projected peak 113.3 GB    <- decided
[VK] expert tier: 3000 hot experts (56.62 GB)  <- only now placed
<run dies, no error>
```

So this reserves the planned size **before** the decisions instead of subtracting the placed size after.

## The change

`backend_vulkan.c` already computes `deviceType` at device selection (`:304`) but uses it only to rank candidates. `coli_vk_device_integrated()` exposes it the way `coli_cuda_device_integrated()` does — no new detection, it just stops discarding what is already known. Verified on this box: RADV reports `VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU` for the 8060S.

Then in `colibri.c`, before `usage_load`/`expert_avail`, the planned tier size (`g_vk_budget` x a routed layer's row width) comes off the snapshot. Discrete GPUs are not `INTEGRATED` so it is a no-op there, as with #653.

## Measured, GLM-5.2 int4-gs64, Ryzen AI MAX+ 395 / Radeon 8060S, 124 GB unified

| `COLI_VK_EXPERTS` | decode | hit rate | from VK tier | RSS |
|---|---|---|---|---|
| 320 (default) | 0.34 tok/s | 69.6 % | 17.0 % | 82.1 GB |
| **3000** | **0.94 tok/s** | **76.1 %** | **66.0 %** | 40.1 GB |

```
[VK] integrated/unified memory: expert tier will share physical RAM;
     RAM budget snapshot reduced by 63.70 GB (3000 experts planned) -> MemAvailable=65.1 GB
[PIN] auto: budget 38.6 GB -> pinning 7.8 GB
[RAM_GB=57.3 auto] projected peak 55.8 GB
[VK] expert tier: 3000 hot experts resident (56.62 GB VRAM, 24.3s)
prefill 14 tokens in 30.19s | decode 24 tokens in 25.55s (0.94 tok/s)
```

The 3000 case is not just survivable now, it is **2.8x the default** — two thirds of expert reads come from the tier instead of disk. My advice in #1502 ("do not raise `COLI_VK_EXPERTS` on a UMA APU") was only correct while the planner was blind to it; I will correct that comment.

## Known imprecision

The reservation is `g_vk_budget` x `expert_bytes_row()` at a mid layer. Measured: **63.70 GB reserved against 56.62 GB placed, +12.5 %**, erring high. Over-reserving costs some LRU cache; under-reserving would resurrect the OOM, so high is the safe side — but if you would rather price it per candidate row the way `pin_count_for_budget()` does, the ranked list is not built until `vk_registry_fill()` and would need hoisting. Happy to do that instead if you prefer it exact.

Stacks cleanly on current `dev` (`6cea190`). Same machine as #1502 and #1511, available for anything you want re-run.


huppiflupp:
## Updated — and the 2.8x in the original description was wrong. Corrected number: +29 %

Two changes since I opened this: a clamp that fixes a regression I introduced, and a properly
controlled measurement that replaces a confounded one. **The fix itself is unchanged in intent;
only my claims about it needed fixing.**

### 1. I introduced a new failure mode, and my own sweep found it

Pricing the reservation off `COLI_VK_EXPERTS` was wrong because that value is a **request**, not a
placement — `vk_registry_fill()` stops early when the device-local budget runs out:

```
4500 requested -> 95.55 GB reserved, 66.0 GB actually placed   (+45 % over)
6000 requested -> 127.40 GB reserved -> MemAvailable floored at 1.5 GB -> run died in init
```

So the first version of this patch traded one silent init death for another. Now clamped by
`coli_vk_mem_budget()` minus `COLI_VK_RESERVE_GB`, plus a 35 % floor so the host planner is never
starved regardless of the request. The 1 GB emergency floor is removed — it hid the starvation
instead of preventing it.

```
320  -> 6.79 GB reserved,  320 placed
3000 -> 63.70 GB reserved, 3000 placed
4500 -> 83.14 GB reserved, 3496 placed, 1.17 tok/s
6000 -> 83.14 GB reserved, 3496 placed, 1.18 tok/s   <- saturates instead of growing, survives
```

### 2. The 2.8x was a measurement artifact

The original description compared 320 against 3000 from two runs in sequence. The second run
inherited the first one's warm page cache **and** its updated `.coli_usage`. Re-running 320 later
gave 0.83, then 0.91, then 0.92 tok/s as the history matured — the "2.8x" was mostly that drift.

Alternating A/B, three rounds, on a machine whose history had stopped moving:

| round | 320 | 4500 (3496 placed) |
|---|---|---|
| 1 | 0.92 | 1.19 |
| 2 | 0.93 | 1.19 |
| 3 | 0.92 | 1.20 |
| **mean** | **0.923 tok/s** (hit 76.0 %) | **1.193 tok/s** (hit 82.7 %) |

**+29 %**, both arms reproducing to ±0.01.

### 3. There is a valley, which is worth knowing

| requested | placed | tok/s | pin | vk |
|---|---|---|---|---|
| 320 | 320 | 0.91 | 38.4 % | 26.8 % |
| 3000 | 3000 | **0.88** | 0.0 % | 76.7 % |
| 4500 | 3496 | 1.17 | 0.0 % | 80.5 % |

At 3000 the reservation has taken the pin set's budget (38.4 % -> 0) but the tier is not yet large
enough to make up for it — worse than the default. The win only appears once the tier carries
~80 % of reads. Anyone tuning this on an APU should go high or stay at the default, not stop in
between. That also explains why 320 -> 1500 measured flat.

### What I claim and what I do not

**Structural, independent of cache or history:** without this patch the planner projects a peak
that ignores the tier (113.3 GB projected while 56.6 GB was already held on a 124 GB machine) and
high `COLI_VK_EXPERTS` values die during init with no error. With it they run.

**Measured on one box, one model, one workload:** +29 % at the high end. I would not generalise it
— the valley above shows the effect depends on how mature the pin history is, and on this hardware
the pin set at the default setting had become strong enough that 3000 was a regression.

Sorry for the churn on the numbers. The patch has been rebuilt and re-measured after each change;
the current state is what the tables above show.

c/backend_vulkan.c
```diff
@@ -1533,6 +1533,19 @@ void coli_vk_tensor_free(ColiVkTensor *t) {
...
+ * expert cache draws from, so whoever sizes that cache has to know. Mirrors
+ * coli_cuda_device_integrated() (#653) for the Vulkan path. */
+int coli_vk_device_integrated(void) {
+    if (!G.phys) return 0;
+    VkPhysicalDeviceProperties p;
+    vkGetPhysicalDeviceProperties(G.phys, &p);
+    return p.deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU ? 1 : 0;
+}
+
 void coli_vk_shutdown(void) {
```
skhari1989-cmyk: The class file of the mode will be entitled for the  classified common file.

huppiflupp:
## Discrete-GPU check for the no-op claim

The description says the reservation is a no-op on discrete GPUs because `deviceType` is not
`INTEGRATED`. That was a code reading; here it is measured, on a second machine.

Intel Core Ultra 245K box with an **RTX 5080 (sm_120, 16 GB)**, enumerated through the same
`vkGetPhysicalDeviceProperties` call the patch relies on:

```
NVIDIA GeForce RTX 5080
  deviceType = DISCRETE
  Heap 0: 17.1 GB DEVICE_LOCAL
  Heap 1: 24.7 GB

Intel(R) Graphics (ARL)
  deviceType = INTEGRATED
  Heap 0: 24.7 GB DEVICE_LOCAL      <- system RAM, presented as device-local
```

So `coli_vk_device_integrated()` returns 0 for the 5080 and the reservation never fires — the
discrete path is untouched, as intended.

### One case I have *not* measured, and it is the interesting one

That machine has **both**: a discrete GPU and an integrated one, and the integrated one advertises
**24.7 GB of DEVICE_LOCAL that is actually system RAM**. Exactly the trap this patch exists for.

`vk_pick_device()` ranks `DISCRETE (4) > INTEGRATED (3)`, so it should select the 5080 and the
patch should stay quiet. I believe that is correct but I could not run it end to end — that box
has neither `glslc` nor a CUDA toolkit installed, so neither `VK=1` nor `CUDA=1` builds there
without adding packages, and it is not my machine to install on.

Worth flagging because the failure mode would be quiet: if anything ever causes the integrated
device to be selected on a hybrid machine — `COLI_VK_DEV2`, a driver that hides the discrete card,
a future ranking change — then the reservation *should* fire, and the current code would do the
right thing only because `G.phys` happens to be the selected device rather than device 0. That is
the correct dependency, but it is worth being deliberate about: the patch reads the device the
engine actually uses, not the first one enumerated.

If you want that path exercised I can ask the machine's owner about installing shaderc, or test it
here by forcing `COLI_VK_DEV2` on the APU — say which is more useful.

JustVugg:
Merged into dev. The reservation logic reads correctly against #653's CUDA counterpart and the clamp to what the device can actually place is the right side to err on; the Lavapipe job in CI covers the compile. Thanks for the measured record on the 8060S.

monotophic:

Datapoint from a second Strix Halo (Radeon 8060S, RADV, 32 cores, 123 GB), on the configuration that died for us
on 2026-09-06 with the same symptom you describe (kernel OOM at first prefill, `anon-rss` ≈92 GiB):

| arm | binary | env | result |
|---|---|---|---|
| A | `dev` @ edd4f48d (before #1513), inside a cgroup capped at 88 GB so only the engine could be killed | `OMP_NUM_THREADS=16 PIN=<container>/.coli_usage PIN_GB=0 COLI_VULKAN=1 COLI_VK_EXPERTS=2000`, int3-g64 container | tier placed (2000 experts, 28.31 GB), `[RAM_GB=111.9 auto]`; OOM-killed at the first prefill, anon-rss 87.6 GiB (cgroup peak 88.0 GiB), 0/5 requests answered |
| B | `dev` @ 87c54a93 (with #1513), same env, uncontained | `[VK] integrated/unified memory: … reduced by 33.03 GB (2000 experts requested) -> MemAvailable=94.2 GB`; `[RAM_GB=82.9 auto]` (projected peak 81.9 GB); tier placed 28.31 GB; 5/5 requests answered (147 tokens), MemAvailable never below ~20 GB |
| C | same as B with `COLI_VULKAN` unset | no `[VK]` lines; `[RAM_GB=111.8 auto]` unchanged from before — the no-op holds |

Reserve vs placed: 33.03 GB reserved for 28.31 GB placed (ratio 1.17, +4.72 GB). The device-local clamp did not bind
here (RADV exposes a single 42.5 GiB DEVICE_LOCAL heap that is GTT-backed; placeable ≈42.5 GB, so a ~33 GB request is under it). The over-reservation costs LRU cache, not correctness, and is in the same direction as the +12.5% you report.

Second box (an identical Strix Halo SKU, its own usage history), same three arms: pre-fix binary killed at first prefill (anon-rss 87.7 GiB in its 88 GiB scope); with #1513 `reduced by 33.03 GB (2000 experts requested) -> MemAvailable=93.9 GB`, `[RAM_GB=82.6 auto]`, tier placed 28.31 GB, 5/5 answered with MemAvailable never below ~19 GB; no-Vulkan control `[RAM_GB=111.5 auto]` unchanged. Same 4.72 GB over-reservation, same non-binding clamp.

One caveat on arm A: the cgroup contained the kill (only the engine died; the box recovered within seconds) but not the whole host-visible footprint — the box's MemAvailable dipped to 7.6 GB for one sample while the scope sat at its 88 GiB cap, consistent with GTT pages being charged outside the scope.

Decode rate on the fixed configuration was 0.19–0.48 tok/s over five 32-token prompts (CPU+Vulkan tier on this box,
not a throughput setup); we have no pre-fix rate to compare because the pre-fix run never answers.

Captures (request/response bodies, both stderr streams, a 5-second MemAvailable sampler, dmesg, binary digests) are
retained and available on request.

PR #1521, opened 2026-09-14
hyper_connections: parallelise the mHC mix and expand, bit-identical (+3.9%)

rcklomp:
`coli_hc_pre`'s mix is `mix_count` independent dot products over the flattened input (24 rows at `hc=4`, ~400k MACs), and `coli_hc_post`'s expand writes disjoint outputs. Both were serial, and both run per token on every layer.

## Why this is bit-identical rather than merely close

Each row keeps its own scalar reduction exactly as written; the `parallel for` only distributes whole rows across threads. No summation order changes anywhere, so this is a work-distribution change, not a SIMD reduction reorder.

Verified on this branch against `dev` — GLM-5.3-Flash int4-g64, same prompt, `--greedy 40 --logits`:

| | |
|---|---|
| `teacher_forcing` | identical |
| `last_logits` | **byte-identical** |
| generated text | identical |

## Speed

EPYC 7F32 (8c/16t, `OMP_NUM_THREADS=8 OMP_PLACES=cores`), CPU path, model warm in page cache, arms interleaved **P,Q,Q,P** so cache warmth cannot favour either side:

| | tok/s |
|---|---|
| pristine | 1.315, 1.308 |
| patched | 1.362, 1.364 |

**+3.9%**, spread within each arm 0.5%, and reversing the order reproduces the gap.

Worth stating because it nearly went in the PR: a first, uncontrolled pair read **+20%**. That was the pristine arm running first on a colder page cache, not the patch. The alternating order is there to stop exactly that.

## A negative result kept in the comment

Replacing the two per-call `malloc`s with fixed-size stack arrays (`hc_mult` is bounded to 8 by every loader) looks like free money and is **not numerics-neutral** on this box. Once `hc`'s range is knowable at compile time — whether from the array size or from a defensive `hc <= 8` check — GCC re-rounds these reductions. Bisected across `-fno-tree-vectorize`, `-ffp-contract=off`, `#pragma GCC unroll 1`, an aliasing barrier and `__attribute__((noipa))`; none restored bit-identity, while keeping the mallocs and adding only the parallelism is bit-identical. So the mallocs stay and only the parallelism lands. The comment records this so the next person doesn't spend the afternoon I did.

## Scope

`coli_hc_pre` / `coli_hc_post` are used by `glm53` and the DeepSeek-V4 engines. Measured on `glm53`; the change is in the shared header and is numerics-neutral for every caller, but I have only run the oracle on `glm53`, so treat V4 as "unchanged by construction, not measured".

The `#pragma`s are `#ifdef _OPENMP`-guarded, so a non-OpenMP build is untouched.


JustVugg:
Verified on the code: each row keeps its own scalar reduction and both call sites (glm53 run_layers, the V4/V4.1 wrappers) are serial, so there is no nesting and the bit-identity claim holds by construction. The oracle in CI will confirm it on the tiny model.

One request before merging: the two new comments refer to "G10", "G7's router fix" and "G8's compiled max_active_levels probe", which are labels from your own notes and mean nothing to a reader of this tree. Please reword them to what they describe (the stack-array attempt and why it was not numerics-neutral; the row-level parallelism; that nested parallelism was checked) without the labels. The negative result itself is worth keeping.

rcklomp:
Done in `4cdc736` — comment-only, no code change.

Both comments now say what they mean instead of naming a downstream fork's rows: that replacing the two mallocs with fixed-size stack arrays was tried and is not numerics-neutral (with the bisection that established it — `-fno-tree-vectorize`, `-ffp-contract=off`, `#pragma GCC unroll 1`, an aliasing barrier, `__attribute__((noipa))`), that the parallel-for distributes whole rows rather than reordering any reduction, and that nested parallelism was checked separately. The negative result is kept, as you asked.

You are right that the labels should not have been there. They are meaningful in the fork this came from and opaque here, and I did not re-read the comments with a reader of *this* tree in mind before opening the PR. #1524 carries the same risk — I have checked it and it refers only to `coli_kda_scratch_floats`, `omp_get_thread_num` and the call sites by name, with no fork labels, so nothing to change there.

Rebuilt after the edit; `glm53` compiles clean.

rcklomp:
Correction to my previous comment: I said #1524 was free of the same problem. That was wrong, and I checked it properly only after saying so. It carried two more `G4` labels, a link to `tools/hot-expert/README.md` (a path that does not exist in this tree), the rig's name and "see the record". Fixed in `fdedf17` on that branch — comment-only, rebuilt clean — and reworded to the substance: why parallelising only the head loop leaves the function flat, why sharing the accumulator needs an oracle rather than eyeballing the output, the 2.25 ms/call measurement, and why there is no `num_threads()` cap.

To stop this recurring rather than just apologising for it, I have added a lint to the fork that scans the added lines of a PR diff for references that only mean something downstream — row labels, fork-only paths, the rig's name, "see the record" — and refuses the branch if it finds any. Validated by running it against this PR's original commit: it reproduces exactly the four lines you flagged. All three of my open branches here lint clean now.

Thanks for catching it, and for the verification on the bit-identity claim.

PR #1522, opened 2026-09-14
feat(deepseek-v41): read the expert stream from more than one drive

LiRunGuo:
## Summary

`deepseek_v4.c` has had `COLI_MODEL_MIRROR` and `O_DIRECT` expert reads for many
releases. `deepseek_v41.c` -- its direct successor, and the largest container the
project ships at 510 GB -- never got either. That is the "a mechanism reached one
engine and never its siblings" shape, twice, in the engine whose docs already say the
expert stream "is most of the engine".

Three changes, all about the same fact -- an expert is read once and never wanted
again:

**`COLI_MODEL_MIRROR` / `COLI_DISK_WEIGHTS` / `SNAP_MIRROR`** register byte-identical
read-only copies on other drives and split the experts across them with a flat-index
hash (the clustering fix `deepseek_v4.c` documents, including why an XOR of two small
ids clustered the hot subset onto one replica). `st_mirror_add` already refuses a copy
whose size or safetensors header differs, so every data offset stays valid on any
replica, and a partial mirror keeps the shards it does not carry on the primary.

**`COLI_MODEL_DIRS`** is the other half of the same idea and the cheaper one: the
container split across drives as *distinct* shards, **no second copy**. At 510 GB that
is the only multi-drive layout most people can build -- two ~256 GB drives instead of
one 512 GB -- and GLM, V4 and Kimi K3 all accept it while V4.1 did not. One line
(`st_init` -> `st_init_multi`), plus the docs and a gate.

**`V41_DIRECT`** (default on, as `COLI_V4_DIRECT` is in V4) reads the expert tensors
with `O_DIRECT`. The engine already calls `posix_fadvise(DONTNEED)` on every expert
after reading it, which is the engine's own admission that a buffered read pays for
the bytes twice and then discards the cache it paid for.

### The shared primitive, and why it is not a bounce buffer

`st_read_range_rep()` in `st.h` is the replica-aware, O_DIRECT-capable range read.
O_DIRECT constrains address, offset and length to the block size, and a safetensors
range begins wherever the writer put it. `deepseek_v4.c` and `colibri.c` solve that by
reading into an aligned scratch buffer and `memmove`-ing the payload out; on a turn
that moves 88.5 GB of experts, that copy is most of what the change was for. This does
the opposite: the caller reserves `ST_DIRECT_ALIGN` bytes of scratch **behind** the
destination, the read starts at the enclosing block boundary, one aligned pread
carries the block-aligned bulk and one short buffered pread carries the sub-block
tail. The payload lands exactly where the slot wants it and nothing is copied after.

The alignment is verified rather than assumed: a caller that did not reserve the
prefix, a platform without `O_DIRECT`, a replica whose twin failed to open, and a
device that refuses the call all fall back to the buffered path, which reads the same
bytes.

`tests/test_st_range_rep.c` drives that contract directly -- multi-block ranges,
sub-block tails, the EOF clamp where the aligned window would run past the end, a
replica read, a divergent replica (refused), a replica with no copy of the shard
(stays on the primary), and a caller that reserved nothing. Guard bytes past the
payload prove the window arithmetic never writes outside it.

## Measurements, and the mistake I made measuring them

**Read the correction first, because it is bigger than the result.** I first measured
this A/B with the engine's default OpenMP team, which on a 208-logical-CPU host is
pathological -- 93% of cycles inside libgomp, a turn that takes 409.9 s instead of
37.2 s (that is a separate defect, PR #1517). It said the disk phase is 5% of a turn
and this work is worth ~2.5%. Re-run with the team sized:

| arm | expert disk | rate | whole turn | decode |
|---|---:|---:|---:|---:|
| `P` one NVMe | 16.07 s | 6.17 GB/s | 37.2 s | 12.86 s |
| `M3` mirror on a second controller | **12.66 s** | **7.83 GB/s** | **34.0 s** | **11.00 s** |
| `MP` mirror, startup probe picks the split | 13.44 s | 7.38 GB/s | 36.3 s | 12.19 s |
| `M2` mirror, both copies behind one RAID0 | 16.36 s | 6.06 GB/s | 37.9 s | 12.96 s |
| `D0` buffered reads | 31.14 s | 3.18 GB/s | 52.7 s | 21.01 s |

208-core Xeon 8555C, released 510.3 GB checkpoint, cold cache before every run, arms
interleaved, 3 runs per arm, medians, **every run token-exact 16/16 in both regimes**.

So `O_DIRECT` is **-48% of the disk phase and +63% of the decode rate**, and the split
is **-21% and +17%**. Validated manifests: `dsv41-direct.json` (+41% whole-turn),
`dsv41-mirror.json` (+12%).

Three things in that table are the point, not the percentages:

- **The engine now reaches the device's own ceiling.** 6.17 GB/s against the 6.34 GB/s
  `c/iobench` measures one `nvme0n1` at depth 4. `M3` clears it at 7.83, which is only
  possible if the split is genuinely using two devices.
- **`M2` is the control.** Both copies behind the `md0` RAID0 pair is indistinguishable
  from one NVMe, so the variable is the number of independent *filesystems*, not the
  number of drives. Two NVMe behind one device add nothing here; a second filesystem
  adds 28% of read rate.
- **The probe is not better than a pinned split.** `MP` lands between the two with more
  variance (52/52/51% of bytes to the mirror, disk 12.81 / 13.44 / 13.72 s). On a pair
  of equal drives `COLI_DISK_WEIGHTS=1,1` is the better choice; the docs already
  recommend that without saying why.

Full record with both regimes: `docs/experiments/dsv41-expert-io-2026-09-14.md`. Raw
per-run engine stderr: `dsv41-expert-io-2026-09-14-raw.txt`.

I kept the 208-thread numbers in the record rather than deleting them. They are real
measurements of a real configuration, the phase breakdown looks entirely reasonable in
both, and that is exactly why the wrong ratio was easy to publish. The record now leads
with a correction block.

## Validation

- [x] `make -C c check` -- full suite
- [x] the tiny oracle token-exact at caps 1/2/8/49, with a mirror, with
      `COLI_MODEL_DIRS`, and for both `V41_DIRECT` arms
- [x] a CI step holding the mirror, the split and the buffered path to the same token
      stream as the primary, and asserting a *divergent* mirror copy is refused rather
      than trusted
- [x] the released 510.3 GB checkpoint end to end; all 88 files verified against the
      upstream manifest; 36 runs across both regimes, every one token-exact
- [x] not applicable -- no CUDA changes

## Compatibility

- [x] default CPU build remains dependency-free -- `open` flags, one `posix_memalign`
      per cache slot, and a header-local helper
- [x] no model files, generated binaries or benchmark artifacts are included
- [x] defaults unchanged in observable behaviour: `V41_DIRECT=1` reads the same bytes
      by a different path, `COLI_MODEL_MIRROR`/`COLI_MODEL_DIRS` are opt-in, and every
      arm of the A/B is token-exact against the same stream
- [x] Windows/macOS: `st_direct_fd_rep` returns the `compat_open_direct` twin, the
      alignment check falls back to buffered where `O_DIRECT` does not exist, and
      `posix_memalign` is `compat.h`'s `_aligned_malloc` shim

## Reviewer notes

- `st.h` gains the primitive rather than `deepseek_v41.c` owning it, because V4 and GLM
  need exactly it and both currently carry their own copy-and-memmove version.
  Migrating them is a separate, riskier change in much larger files.
- `COLI_MODEL_DIRS` came along because it is the same user problem with a cheaper
  answer; say the word if you would rather split it out.
- This PR and PR #1517 are independent. #1517 is the CPU-side fix that makes this one
  measurable; neither needs the other to be correct, though the numbers in this record
  were taken with `OMP_NUM_THREADS=32` on the command line.


JustVugg:
Reproduced the O_DIRECT half here before merging, since it changes the default. Released V4.1 Flash checkpoint, Ryzen 7 PRO 8700GE (8 cores), two NVMe in one md RAID0, cold page cache before every run, 26 prompt + 24 generated tokens, cap 8, OMP_NUM_THREADS=8, two runs per arm interleaved:

| arm | expert disk | matmul | attention | turn | tok/s |
|---|---|---|---|---|---|
| dev | 21.17 / 21.11 s | 13.2 / 13.4 s | 8.4 / 8.4 s | 49.2 / 49.3 s | 0.488 / 0.487 |
| this PR, V41_DIRECT=0 | 21.12 / 21.12 s | 13.6 / 13.5 s | 8.2 / 8.2 s | 49.3 / 49.2 s | 0.487 / 0.488 |
| this PR, V41_DIRECT=1 (default) | 12.75 / 12.77 s | 15.4 / 15.2 s | 8.6 / 8.3 s | 43.4 / 42.8 s | 0.554 / 0.561 |

Disk phase -40%, turn -12%, tokens identical across all six runs. The buffered arm of the PR is indistinguishable from dev, so the gain is the read path and not the slot refactor. The mirror split is not measurable on this box (both drives sit behind one RAID0, your M2 control), so I am taking your two-filesystem numbers as reported. Will merge once CI is green.

PR #1524, opened 2026-09-15
delta_attention: parallelise the KDA conv and head loops, bit-identical (+3.0%)

rcklomp:
`coli_kda_step` ran entirely serially. Two loops dominate it and both are embarrassingly parallel:

* the **causal convolution** — each channel owns its own slice of `window` and one element of `mixed`, and the loop carries `3*width` SiLU calls (24,576 `expf` per call at GLM-5.3's shape);
* the **head loop** — each head owns its slice of `state` and of `out`.

Parallelising only the head loop leaves the function flat, because the conv loop is then the serial half. Both need a team.

## The one shared thing

Heads share exactly one object, the `memory` accumulator. The scratch tail therefore becomes one `v_dim` slice **per thread** and each head indexes its own. Sharing it corrupts every head's read-back *while still emitting plausible tokens*, which is why this is verified against an oracle rather than by reading the output and nodding.

## API note — why this is not a pure drop-in

`coli_kda_scratch_floats()` now returns a **thread-count-dependent** size. Two conditions follow, and both are documented on the function:

1. the buffer must be sized under the same OpenMP thread count the call will see;
2. `coli_kda_step` must not be called from **inside** a parallel region — the head loop indexes the scratch tail by `omp_get_thread_num()`, so an outer team's larger ids would run off the end.

Both hold for every caller in the tree today. `glm53.c` sizes it at session open (`glm53.c:2264`) and calls it from `kda_layer`, whose only call path (`glm53.c:2347`) has no enclosing parallel region; `tests/test_delta_attention.c` sizes it the same way. If you'd rather not take that contract, the alternative is a fixed `omp_get_max_threads()` allocation independent of the caller — say the word and I'll rework it.

## Verification

Against `dev`, GLM-5.3-Flash int4-g64, same prompt, `--greedy 40 --logits`:

| | |
|---|---|
| `teacher_forcing` | identical |
| `last_logits` | identical |
| generated text | identical |

Speed — EPYC 7F32 (8c/16t, `OMP_NUM_THREADS=8 OMP_PLACES=cores`), CPU path, model warm, arms interleaved **P,Q,Q,P**:

| | tok/s |
|---|---|
| pristine | 1.311, 1.308 |
| patched | 1.350, 1.348 |

**+3.0%** end to end. The isolated kernel gain is much larger than that — this is one step inside a token that also does the projections, MoE and attention — but the end-to-end number is the one worth quoting.

## Not run

`tests/test_delta_attention` includes a generated fixture (`glm53_kda_case.h`, produced by `c/tools/make_glm53_tiny.py` against HuggingFace's `Glm5NextTextModel`) which I did not stand up here, so that test was not executed. Flagging it rather than staying quiet — the end-to-end oracle above exercises the same code at real shapes, but it is not the same check.


JustVugg:
Merged into dev, together with #1521. Both reword commits verified comment-only before merging; the scratch contract (sized under the same thread count, never called from inside a parallel region) holds for the one caller in the tree, and with #1517 the team is fixed as the first statement of main, so the two agree. Thanks for the lint on your side.

PR #1526, opened 2026-09-15
fix(registry): make the plan and the engine agree about the engine

LiRunGuo:
Two places where the plan and the engine disagreed about the engine, and nothing
failed at the point of divergence.

Both are the same shape: a mechanism every other engine adopted, that one engine
did not, with the planner trusting the registry and the engine not.

### 1. `deepseek_v41` never read `CTX`

Every family declares its context variable, `coli --ctx` sets it, and
`resource_plan.py` sizes the buffers it implies from that number:

| family | declared | engine reads it |
|---|---|---|
| glm53 | `GLM53_MAXT` | yes |
| glm | `CTX` | yes |
| inkling | `CTX_MAX` | yes |
| kimi | `K3_MAXT` | yes |
| olmoe | `CTX` | yes |
| qwen36 | `Q36_MAXT` | yes |
| qwen38 | `Q38_MAXT` | yes |
| deepseek_v4 | `CTX` | yes |
| **deepseek_v41** | **`CTX`** | **no** |

The engine always used the checkpoint's own ceiling (1,048,576 positions), which
sizes the compressed KV and index keys of the four source layers and both rope
tables:

| context | compressed KV + index keys |
|---:|---:|
| 4,096 (the registry's default) | 0.02 GiB |
| 32,768 | 0.20 GiB |
| 1,048,576 (what the engine allocated) | **6.25 GiB** |

**6.23 GiB of unbudgeted allocation at the default context**, plus ~0.5 GiB of rope
tables the planner does not model at all. On a machine sized to the plan, that is
the difference between fitting and not. And `--ctx` was silently inert: a user could
ask for 32k and get buffers for 1M with no message either way.

Measured, with the fix:

| | load | peak RSS |
|---|---:|---:|
| no `CTX` (unchanged, ceiling kept) | 10.13 s | 24.69 GB |
| `CTX=32768` | **6.23 s** | **17.83 GB** |

-6.86 GiB and -3.9 s of load, which is the 6.23 GiB predicted above plus the rope
tables. **The generated tokens are byte-identical across no-CTX, `CTX=32768` and
`CTX=262144`** -- this is a capacity knob, not a numerics knob.

The clamp only ever asks for less: `CTX` above the checkpoint's ceiling warns and
keeps the ceiling, below 2 warns and is ignored.

### 2. Two CPU-only engines claimed an accelerator

`resource_plan.py` refuses a VRAM budget and drops the GPUs for an engine that
declares `supports_accelerator=False`, and its comment says why: *"a CPU-only engine
has no VRAM tier: announcing one anyway writes a line in the plan that nobody can
execute."* The flag defaults to `True`, so an engine has to opt out -- and the two
CPU-only engines never did:

| family | GPU refs in source | build rule | declared |
|---|---:|---|---|
| olmoe | 0 (its own comment: "CPU-only (no CUDA/Metal backend)") | `NOCUDA_LDFLAGS` | `True` |
| **deepseek_v41** | **0** | no backend object | `True` |

`coli plan` therefore offered DeepSeek V4.1 Flash:

```
VRAM   296.0 GB hot tier · ~15744 experts · 0:NVIDIA H200, 1:NVIDIA H200, ...
hit    100% projected expert residency
```

on eight H200s the engine cannot address. After the fix:

```
RAM    905.5 GB budget · 11.5 GB dense · 11.8 GB runtime · 296.0 GB warm experts · cap 384/layer
VRAM   no supported GPU detected · CPU path
```

which is both true and better advice: the RAM tier absorbs the budget instead of a
phantom VRAM tier holding it, and the auto-tune hint becomes `COLI_NUMA=1
multi-socket CPU-only` rather than a GPU placement suggestion.

## Validation

- [x] `make -C c check` -- full suite
- [x] new gate `tests/test_registry_engine_agreement.py`, checked in **both
      directions** for both checks (each fix reverted, each caught)
- [x] `deepseek_v41` token-exact 16/16 and byte-identical with no `CTX`,
      `CTX=32768` and `CTX=262144`
- [x] `CTX` above the ceiling and below the minimum both warn and behave
- [x] zero warnings on the rebuilt engine

## The gate, and why it is derived rather than listed

`tests/test_registry_engine_agreement.py` has two assertions:

1. **every engine reads its declared `context_env`** -- this one is a direct list
   walk over the registry, so a new family that forgets is caught on arrival;
2. **an engine whose build rule links no backend object must not claim
   `supports_accelerator`** -- derived from the Makefile rather than a list of
   known-CPU engines, so it cannot rot when a backend is added or removed.

I checked that the second one fails when `olmoe`'s flag is put back, and the first
when `deepseek_v41` stops reading `CTX`. A gate that cannot fail is not a gate, and
this is the second one in this series I have had to repair for that reason.

## Scope, honestly

- **Two engines, not a class.** The other seven agree with their registry entries
  today; the gate is what keeps that true rather than a claim that it always was.
- **The 6.23 GiB is this checkpoint's number.** It scales with the model's own
  ceiling and source-layer count, which is why the fix is "honor the variable"
  rather than "allocate less by default".
- **I did not change the default.** With no `CTX`, the engine still uses the
  checkpoint's ceiling and behaves exactly as before. Making the default the
  registry's 4,096 would be a behaviour change for existing users and is a
  separate conversation -- though it is where another 6 GiB and ~4 s of load
  would come from.
- **The VRAM flag is a plan-level fix only.** It does not add a backend; it stops
  the plan from promising one.


JustVugg:
Merged into dev. Reproduced here before merging, on the released V4.1 Flash checkpoint (Ryzen 7 PRO 8700GE, 8 cores, 61 GB, cold page cache, 26 prompt + 24 generated tokens, cap 8, OMP_NUM_THREADS=8):

| | peak RSS (engine STAT) | tok/s |
|---|---|---|
| no CTX | 24.77 GB | 0.490 |
| CTX=32768 | 17.91 GB | 0.491 |

Generated tokens identical in both runs, so the clamp is a capacity knob only, as you said. The registry gate reads correctly on this tree too: both CPU-only engines now agree with their build rules.

Issue #1501, opened 2026-09-14
[Bug]: coli run never sets SNAP for non-GLM engines, OLMoE exits 1 with 'started without a model'


alobo2:
**Version:** v1.11.0 (Windows release zip; same code path on macOS).
**Model:** OLMoE-1B-7B int8 container.

`coli run --model <olmoe dir>` prints `started without a model` and exits 1. `coli serve --model <same dir>` and `coli chat` work.

Cause: the launcher's `env_for_engine()` only sets `SNAP` for the GLM engines; for OLMoE (and, reading the code, the other non-GLM engines) the child process starts with no `SNAP` in its environment. `openai_server.py` sets it itself around line 2938, which is why `serve` and `chat` are fine.

**Workaround:** export `SNAP=<model dir>` before `coli run`.

Two smaller things noticed on the same path, in case they belong in the same fix: the `chat` path prints no timing line for OLMoE, so tok/s has to be read through the gateway's `/profile`; and the model-less failure exits 1 without naming the missing variable, which cost an hour of guessing.

JustVugg:
Fix in #1503: env_for_engine sets SNAP from --model for every engine (an explicit SNAP still wins), and the engine's no-model message now names the variable. The chat timing line for OLMoE is a separate item; the private-engine chat prints the engine's own counters when the DONE frame carries them, which olmoe now does, so I will check that path after this lands.

JustVugg:
Merged into dev: #1503 (333e760). Ships in the next release; closing here, reopen if a dev build still shows the symptom.

PR #1529, opened 2026-09-15
fix(coli): forward --model as SNAP on non-GLM direct launches

yablokolabs:
env_for_engine() builds the environment for every direct engine launch (run, bench, chat-spawn, tune) but never set SNAP, which every non-GLM engine getenv()s for its weights (olmoe.c, kimi_k3.c, inkling.c, qwen36.c, glm53.c, deepseek_v4.c). Only env_for() (GLM one-shots) and the gateway spawn in openai_server.py set it — which is why serve/chat always worked.

Repro, any machine: coli run --model <olmoe-int8 dir> "hello" dies in the engine's "started without a model" exit. deepseek_v4 is unaffected (takes the model via argv); glm53 goes through the same broken function, so this repairs it too (only olmoe live-verified here — no GLM-5.2 weights on this box).

The fix mirrors env_for(): the explicit --model always wins over a stale exported SNAP, so an old SNAP export in the parent shell can never hijack a run aimed elsewhere. getattr-guarded for the partial test Namespaces.

Tests: extended test_olmoe_run_uses_its_engine… to assert SNAP reaches the spawned env, plus test_env_for_engine_forwards_model_as_snap and test_explicit_model_wins_over_exported_snap. All three fail on unfixed code; test_v4_cli + test_cli_output (51 tests) and test_stop_scope + test_datapoint (28 tests) green with the fix. Also live-verified: olmoe-int8 one-shot generation succeeds with SNAP unset.


JustVugg:
Thank you, the diagnosis is exactly right, and the same fix is already in dev: #1503 (merged 2026-09-14) sets SNAP from --model inside env_for_engine for every engine, with an explicit SNAP still winning, and #1485 by another contributor reported it first (#1501). Closing this one as already fixed; nothing wrong with the patch.

PR #1530, opened 2026-09-15
harden: add integer overflow check in kimi_k3.c

anupamme:
## Summary
Harden input handling in `c/kimi_k3.c` (flagged by multi_agent_ai).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `c/kimi_k3.c:287` |
| **Assessment** | Defensive hardening |
| **CWE** | CWE-190 |

**Description**: The falloc() and fcalloc() functions perform multiplication n*sizeof(float) without checking for integer overflow. When n is large enough (e.g., n > SIZE_MAX/sizeof(float)), the multiplication wraps around, resulting in a small allocation. Subsequent writes to the undersized buffer cause heap corruption.

## Threat Model Context

This is a Python library - vulnerabilities affect applications that import this code.

## Changes
- `c/kimi_k3.c`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


c/kimi_k3.c
```diff
@@ -284,8 +284,12 @@ static double rss_gb(void){ struct rusage r; getrusage(RUSAGE_SELF,&r);
     return r.ru_maxrss/(1024.0*1024.0);
 #endif
 }
-static float *falloc(int64_t n){ float *p=malloc((size_t)n*sizeof(float)); if(!p){fprintf(stderr,"OOM %lld floats\n",(long long)n);exit(1);} return p; }
-static float *fcalloc(int64_t n){ float *p=calloc((size_t)n,sizeof(float)); if(!p){fprintf(stderr,"OOM %lld floats\n",(long long)n);exit(1);} return p; }
+static float *falloc(int64_t n){
+    if(n<0 || (uint64_t)n > SIZE_MAX/sizeof(float)){ fprintf(stderr,"alloc size overflow: %lld floats\n",(long long)n); exit(1); }
+    float *p=malloc((size_t)n*sizeof(float)); if(!p){fprintf(stderr,"OOM %lld floats\n",(long long)n);exit(1);} return p; }
+static float *fcalloc(int64_t n){
+    if(n<0 || (uint64_t)n > SIZE_MAX/sizeof(float)){ fprintf(stderr,"alloc size overflow: %lld floats\n",(long long)n); exit(1); }
```
sheraz4196, line 291: `calloc` already guards against `nmemb * size` overflow per C11 and returns NULL, so this check is redundant here (unlike in `falloc` where it's useful). Also, if `n == 0` is a valid caller input, you may want to allow it since `calloc(0, ...)` is implementation-defined. Non-blocking; LGTM.

Issue #1518, opened 2026-09-14
[EXPERIMENT] The OpenMP team is the whole ballgame on a many-core host: 14x on V4.1, from a call four engines already make


LiRunGuo:
`omp_tune.h` has carried the physical-core team sizing since #718 (+2.3x on a 16C/32T
part, from the thread count alone). Four engines call it: `colibri.c`, `inkling.c`,
`kimi_k3.c`, `olmoe.c`. Four were written afterwards and never did: **`glm53.c`,
`qwen36.c`, `qwen38.c`, `deepseek_v41.c`**.

There is no error when that happens. This issue is the measurement of what it costs.

## Machine and method

2x Intel Xeon Platinum 8555C, 52 cores/socket, 2 threads/core = **104 physical cores,
208 logical CPUs**, 4 NUMA nodes, 1007 GiB RAM. Ubuntu 24.04, gcc 13.3.0, libgomp,
one Samsung MZQL2960HCJR NVMe.

Released `deepseek-ai/DeepSeek-V4.1-Flash`, no conversion, 48 shards, 510.3 GB, all 88
files verified against the upstream manifest. Engine's one-shot oracle mode, 25 prompt
tokens, 16 generated, `cap=8`, page cache evicted per run, `ref.json` holding the 16
reference ids so **every run below is token-exact 16/16**.

PR: https://github.com/JustVugg/colibri/pull/1517. Record: `docs/experiments/dsv41-omp-team-2026-09-15.md`.

## 1. The profile

`perf record` on the shipped default, before changing anything:

| overhead | where |
|---:|---|
| 38.65% | `libgomp.so.1.0.0` +0x256c0 |
| 35.85% | `libgomp.so.1.0.0` +0x258a0 |
| 4.91% + 4.63% + 4.35% + 4.05% + ... | the same two instruction streams, sampled at successive IPs (skid) |
| 1.24% | `[k] native_queued_spin_lock_slowpath` |

**93% of cycles in the OpenMP runtime.** ~720 parallel regions per decoded token -- three
per expert application, 240 applications a token -- each only a few thousand rows wide,
so the barrier is a first-order cost rather than an overhead term.

## 2. The sweep

One variable, cold cache, all runs token-exact:

| `OMP_NUM_THREADS` | expert disk | expert matmul | attention | prefill | decode | tok/s |
|---:|---:|---:|---:|---:|---:|---:|
[...]

PR #1534, opened 2026-09-15
qwen36: physical-core OpenMP team sizing on the direct-launch path

huppiflupp:
Adds the one `coli_omp_tune_threads()` call that `qwen36.c` is missing, which
#1518 identified. The justification is narrower than that issue's, though, and
I want to be upfront about it.

### Why this is *not* the +25% it looks like

`c/coli:458` already sets `OMP_NUM_THREADS` from `physical_cpu_count()` for
exactly this set of engines, under the comment *"OMP_NUM_THREADS for the sister
engines that do not size their own team"*. So **through the launcher, this PR
changes nothing** — the helper returns at its own `getenv("OMP_NUM_THREADS")`
guard and doesn't even print its `[OMP]` line.

I found this the embarrassing way: my first A/B showed +23.7% and I was about to
report it as this patch's effect. The tell was the missing `[OMP]` line, not the
number, which looked entirely plausible.

**What this PR actually fixes is the direct-launch path** — running `./qwen36`
instead of going through `coli`. There, nothing sizes the team and libgomp falls
back to `nproc`, i.e. logical cores.

### Measurement

To measure the C call rather than the launcher, I temporarily gated the launcher
line behind an env var so it would stay silent, then A/B'd the same patched
binary with and without `COLI_NO_OMP_TUNE=1`. The helper demonstrably fired only
in the treated arm:

```
[OMP] qwen36: 16 physical-core threads instead of 32 logical CPUs
```

Host: Ryzen AI MAX+ 395 (Strix Halo, Zen5), 16C/32T, 128 GB unified, colibri
f028d26, Qwen3.6 int4-gs64, `--cap 256 --ram 40`. 3 alternating rounds, 2 timed
requests each, first discarded as warm-up:

| | median | round medians |
|---|---|---|
| without (32 logical) | 18.94 tok/s | 18.62, 18.94, 18.95 |
| with (16 physical) | **23.72 tok/s** | 23.80, 23.72, 23.57 |

**+25.3%**, 3/3 rounds positive, no overlap.

### On the statistics

The two requests inside a round share a process and a cache state, so they are
not independent — the unit is the round, which leaves **n = 3**. An exact paired
sign/permutation test bottoms out at **p = 0.125** there, so this experiment
alone is suggestive, not significant.

Pooling it with the launcher-arm A/B — same treatment (32 logical vs 16
physical), reached by the other route — gives **n = 7, 7/7 positive, +24.4%,
exact permutation p = 0.0078**, round-median ranges 18.62–19.16 vs 23.30–23.80,
no overlap.

The treated arm lands at 23.72 against the launcher's 23.42: the same operating
point. That is the expected result for a call that *replaces* the launcher's
work rather than adding to it, and it is the main reason I believe the n = 3
result despite the p-value.

### No-op guarantees, verified against the built binary

| condition | `[OMP]` line |
|---|---|
| direct launch, nothing set | 1 |
| `OMP_NUM_THREADS=8` | 0 |
| `COLI_NO_OMP_TUNE=1` | 0 |

Also a no-op without SMT (`phys >= logical` returns silently) and when the
physical-core count is not determinable, per `omp_tune.h`'s own rule against
guessing.

### Scope

Only `qwen36.c`. #1518 also lists `glm53.c`, `qwen38.c` and `deepseek_v41.c`; I
have not measured those and would rather not patch what I cannot test. `qwen36`
is also the engine whose `attn_sc_thr = omp_get_max_threads()` sites (`ensure_kv`
and the segment path) run after `main()` starts, so they inherit the sized team
rather than keeping a second, contradictory count.


huppiflupp:
Closing this — #1517 landed the same `qwen36.c` call and is already in `dev`,
so there is nothing left here to merge.

For anyone arriving from #1518: @LiRunGuo's PR covers all four engines and was
verified by the maintainer on an 8C/16T Ryzen. This one was `qwen36` only.

Two things from here that might still be worth having, neither blocking:

1. The comment block in this PR says explicitly that the call is a **deliberate
   no-op under `coli`**, because the launcher already sets `OMP_NUM_THREADS`
   from `physical_cpu_count()` for exactly these engines. #1517's version does
   not state that, and it is the part that cost me an hour — my first A/B showed
   +23.7% and I nearly reported it as the patch's effect. The tell was the
   missing `[OMP]` line, not the number, which looked entirely plausible. Happy
   to send that as a comment-only PR if you want it in the tree.

2. The measurement in this PR isolates the C call from the launcher (launcher
   line temporarily gated so it could not mask it): **18.94 -> 23.72 tok/s,
   +25.3%** on 16C/32T Zen5, 3 paired rounds. n=3 bottoms out at p=0.125, so on
   its own it is suggestive rather than significant; pooled with the
   launcher-arm A/B it is n=7, 7/7 positive, exact permutation p=0.0078. Left
   here for the record rather than repeated elsewhere.

Thanks @LiRunGuo for the credit in #1517 — and for taking the framing seriously
rather than the headline number.

Issue #1550, opened 2026-09-15
[Feature]:Add Dark mode switch button


sparshsingh12312:
## Problem Statement

There is no option to switch to dark mode in the Appearance section

## Proposed Solution

  I want to add an option to switch to dark mode on the landing page itself.

## Additional Context

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/c2a9f9e4-f07b-4b34-8020-b4285e8e75b0" />

### Proposed solution

  I want to add an option to switch to dark mode on the landing page itself.

sparshsingh12312:
I would like to address on this feature? @JustVugg

JustVugg:
If you want to completely revamp it to make it better, I’m open to suggestions. I think a different site design is possible.

sparshsingh12312:
I just submitted my PR regarding the new landing pagr with light to dark switch mode button. @JustVugg

Nagulanvelu:
I’d like to work on improving and revamping the dark mode feature rather than duplicating the existing implementation. I can explore a cleaner UI, better theme consistency, and improved user experience. Would it be okay if I take this approach?

JustVugg:
Yes 👍🏻

sparshsingh12312:
Look at my PR #1551 @JustVugg

JustVugg:
Done on `dev`, in both places.

The landing page got a light and dark theme in #1635, built on the mechanism @sparshsingh12312 proposed in #1551: the `colibri_theme` key, the `data-theme` attribute set before first paint so it never flashes, the `D` shortcut, the sun and moon crossfade, and the operating system preference followed only until the visitor chooses. The dashboard under `web/` got its own switch as part of the redesign in #1633.

Both #1551 and #1574 are closed, not because the work was wrong but because the two files they touch were rewritten by those two pull requests; the comments there say what was kept and what is still open to pick up.

Thanks to both of you for pushing this.

PR #1551, opened 2026-09-15
Added dark_mode to the website.

sparshsingh12312:
## Summary

This PR addresses the lack of dark mode configuration options on the landing page and refactors several core application layouts (Dashboard, Kanban, Timeline, UI Buttons) that used hardcoded backgrounds and colors, leading to visibility issues in dark mode.

## Related Issue
#1550 

## Validation

-`make -C c check`
-CUDA changes were tested with `make -C c cuda-test` (if applicable)
- Performance claims include hardware, commands, and repeatable measurements
- Performance claims include a validated experiment manifest with raw evidence

## Compatibility

-The default CPU build remains dependency-free
- No model files, generated binaries, or benchmark artifacts are included

## Proof
Screenshot 2026-09-15 at 10 25 52 PM


sparshsingh12312:
Could you please merge my PR as it is required for the betterment of your websites UI.

sparshsingh12312:
Is my PR merged or not because I still cannot see. the changes on the website. @JustVugg

JustVugg:
Thank you for this, and sorry it is being closed rather than merged: the reason is the file, not the work.

The landing page was rewritten in #1635 (merged): the models became a searchable list instead of a card grid, a brio mode section was added, and the version and family names are now generated from `c/version.py` and `c/family_registry.py` so they cannot go stale again. That rewrite touches almost every line of `site/index.html`, so there was no way to land your patch on top of it.

**Your dark mode is the one that shipped.** #1635 uses your mechanism, not a different one:

- the same `colibri_theme` key in `localStorage`
- the same `data-theme` attribute, set by an inline script before first paint so the theme never flashes
- the same `D` keyboard shortcut, with the same guard that skips it while an input is focused
- the same sun and moon icons crossfading with a rotate and scale, and the same 300 ms `theme-transition` class
- the same rule that the operating system preference is followed only until the visitor chooses for themselves

The commit and the pull request body both say where it came from. If you want to keep going on the site, the theme is done but two things next to it are open: the page has no mobile layout yet, and the model list has no per-model page behind each row.

One separate note: the LICENSE, NOTICE and README part of this pull request is already on `dev`, so that half had nothing left to apply.

Issue #1370, opened 2026-09-06
Perplexity vs Ollama's Q4_K_M on the same tokens: our int4 experts lose 2.5–3.4 %, the int8 trunk buys nothing — two questions about the format


kreuzzelg:
I measured output quality against Ollama for the first time, on Qwen3.6-35B-A3B, and the result raises two questions about the container format that only you can answer. Data first, then the questions. Nothing here is a request to change course; it is a request for your read.

## The measurement

wikitext-2-raw test split, **identical tokens on both sides** (297,054 with the container's `tokenizer.json`, 297,054 with `llama-tokenize` on Ollama's GGUF, same ids), scored the way `llama-perplexity` scores: 16 chunks of 512, first half context, second half scored, natural log. Colibrì side through `qwen36`'s `PPL=1` teacher-forcing mode, one chunk per process; Ollama's own Q4_K_M blob through a llama.cpp CPU build (which needed four local loader patches to read Ollama's GGUF — Ollama's converter targets Ollama's engine; patches listed at the end, none of them touch arithmetic).

| | perplexity | scored tokens |
|---|---|---|
| **Ollama Q4_K_M** (`qwen3.6:35b`, 23.9 GB) | **7.147** | 4080 |
| Colibrì gs64 int4 experts, int8 trunk (22 GB) | 7.325 | 4096 |
| Colibrì per-row int4 experts, int8 trunk (20 GB) | 7.388 | 4096 |
| Colibrì per-row int4 experts, **f32** trunk (`COLI_DENSE_I8=0`) | 7.389 | 4096 |

Paired over the same 16 chunks: Q4_K_M is better in 15 of 16 against gs64 and 14 of 16 against per-row; mean NLL difference +0.025 nats (standard error 0.006) and +0.033 (0.010). The int8 trunk against the f32 trunk: −0.0001 ± 0.0025, i.e. nothing.

Sanity checks before trusting it: the patched llama.cpp reports the right hyperparameters (40 layers, 16/2 heads, 256 experts top-8, DeltaNet 128/32/16, 35.95 B params, Q4_K – Medium), and its greedy continuation of my standard prompt begins word for word like Colibrì's and diverges only after ~25 tokens, as two quantizations should. A misloaded model would not score 7.1.

## What I read from it, and where I was wrong

[...]

PR #1552, opened 2026-09-15
qwen38: the int8 trunk carries one scale per 64 weights (grouped fmt 1 on the GPU), perplexity-neutral

kreuzzelg:
Follow-up to #1424 (merged): one commit on `dev`. Also takes the doc note from the #1424 review (greedy-identical is a measurement, not a guarantee; the two reference knobs).

## Why

#1424 quantizes the dense trunk to int8 per row. On wikitext-2 (8 x 512, 2048 scored tokens) that costs +1.9 %
perplexity (1.880 against 1.845 BF16). Probing the rows showed the error is spread over the trunk, not one
matrix class: the expert tier is neutral, lm_head contributes nothing, keeping the routers in BF16 recovers 0.3
points, and the GPU's summation order is 0.2 of the 1.9 (1.877 on the CPU reference path). One scale per 64
weights along the input brings it to **1.852 (+0.4 %)**, the same eight chunk NLLs on the GPU and on the CPU
reference path, greedy text identical to BF16 over 100 tokens.

## What changes

- **Backend:** fmt 1 uploaded with a group size takes the grouped int4 branch of `quant_matmul` (scales
  `[O][ceil(I/gs)]`, group subtotal times scale); the per-element `i / gs` becomes a shift when gs is a power of
  two (75 -> 69 ms resident-mm per token on the 553 trunk matrices; per-row was 63). The upload records gs for
  fmt 1 as it does for fmt 4. No new symbols, the DLL loader table is unchanged.
- **Tier:** `qt_dense_init(q, sc, I, O, device, gs)`; gs > 0 uploads through `upload_g` and the budget counts the
  scale table; `qt_dense_matmul` passes the handle's gs. The backend's cached-tensor check refuses a grouped
  tensor asked with gs 0 -- the first hardware run of this branch failed all 553 matrices exactly that way, so
  the fake backend now mirrors the check and the dense test has the case.
- **Engine:** `Q38_TRUNK_GS` (default 64, 0 = per row) drives the quantizer, the CPU reference path
  (`Q38_TRUNK_CPU_INT8`), the selftest, the offered bytes and the placement sum.
- **Planner:** the trunk estimate carries the scale table (+1/16).
- **Docs:** knob, trunk paragraph, the perplexity table.

## Measured (RTX 3070, cap 224, the 315+100-token prompt of #1424)

| | per row (#1424) | gs 64 |
|---|---:|---:|
| trunk in VRAM | 3.96 GiB | 4.21 GiB |
| resident experts / VRAM share of routes | 352 / 22.3 % | 312 / 20.6 % |
| resident-mm / lm-head, ms per token | 63 / 5 | 69 / 7 |
| decode tok/s | 1.91 | 1.88 |
| greedy text, 100 tokens, against BF16 | identical | identical |
| selftest, 553 matrices, GPU against CPU int8 of the same rows and scales | 1e-7 | 1.5e-7 |
| perplexity wikitext-2 | 1.880 (+1.9 %) | **1.852 (+0.4 %)** |

Tests: the ten `tests/test_qwen36_tier_*` (dense with the grouped case), `test_qwen38_tier_engine` (trunk grouped
on the fake tier, cosine 0.99999), Python suites; CUDA builds of qwen36 and qwen38.

Thanks to [Claude Code](https://claude.com/claude-code) for the help along the way.

https://claude.ai/code/session_01X2HPETYXNwT4RmQkdMMyPS


kreuzzelg:
@JustVugg after #1668 this PR no longer fits as it stands, and I would rather ask before reshaping it.

#1668 quantizes the trunk once on the CPU, one scale per row, and the GPU placement uploads those same rows -- one copy, `idot.h` on the CPU, `qt_dense_init` on the card. This PR puts one scale per 64 weights on the GPU only, i.e. a second quantization of the trunk, which is exactly what #1668 removed. So it conflicts on `qwen38_core.h`, `qwen36_tier.[ch]` and the fake, and a textual rebase would bring the second copy back.

What the group scales buy, on the released checkpoint, wikitext-2, 2048 tokens (#1552 text, GPU trunk; the CPU reference path measured the same within 0.2 points):

| trunk | ppl |
|---|---:|
| BF16 | 1.845 |
| int8, one scale per row | 1.880 (+1.9 %) |
| int8, one scale per 64 | 1.852 (+0.4 %) |

Your four 512-token doc chunks in #1668 put the per-row cost at +0.5 % perplexity; on wikitext it is four times that, and the group scales recover most of it. Decode cost on the card was 1.91 -> 1.88 tok/s.

The shape that would keep one copy: `q38_trunk_quantize` writes `[O][I/64]` scales instead of `[O]`, `matmul_q_idot` takes a `gs` and accumulates as your K1b kernel already does (exact int32 per group, `fmaf((float)group_int, scale[g], acc)` in ascending g -- the maddubs/vpdpbusd inner loop is untouched, the group boundary is every 64 bytes), and the GPU takes the same rows with the fmt 1 grouped path this PR already has (`gs` in the backend's cached-tensor check). `Q38_TRUNK_GS=0` would keep today's per-row rows and kernel as the A/B. Roughly: `idot.h` + the two quantizers + the test on both sides, two to three days, one PR replacing this one.

Two questions before I do that:

1. Do you want group scales in `idot.h` at all, or is the per-row trunk the format you intend to keep for both engines (qwen36's dense trunk from #1664 would take the same `gs` for free, though I have not measured its perplexity there)?
2. If yes, on the released checkpoint or the tiny fixture -- the fixture's matrices are all below `Q38_TRUNK_MIN_KB`, so the oracle would need the `Q38_TRUNK_MIN_KB=0` pass you added in #1668 to exercise the grouped path.

If the answer to 1 is no, I will close this PR; the perplexity table stays in #1370 as the record.

PR #1556, opened 2026-09-15
Detect Metal GPUs in macOS resource plans

karlem:
## Problem

While diagnosing slow `coli chat` inference on an M1 Pro, we ran:

```sh
COLI_MODEL=/path_to/glm52_i4 ./coli plan --ram 22 --ctx 512
```

The output incorrectly said:

```text
VRAM   no supported GPU detected · CPU path
```

This was unexpected because `make metal-test` succeeded on the same machine, confirming that the Metal backend and Apple GPU were available.

## Cause

`build_plan()` calls `discover_gpus()`. Before this change, discovery only tried:

- `nvidia-smi`
- `rocm-smi`

Neither command can detect Apple Metal GPUs, so `coli plan` silently reported no supported GPU and showed the CPU path.

## Fix

Add a macOS-specific discovery path using:

```sh
system_profiler SPDisplaysDataType -json
```

`coli plan` now detects and displays the Apple Metal GPU. It is reported as unified memory, without inventing a separate VRAM budget or applying CUDA-style placement settings.

## Summary
- discover Apple Metal devices through system_profiler on macOS
- report Metal unified-memory hardware without inventing a separate VRAM budget
- cover the macOS discovery path and planner rendering

## Validation
- python -m unittest tests.test_resource_plan.ResourcePlanTest


JustVugg:
Merged current dev into your branch: format_plan's VRAM line now also prints the int8 trunk bytes on dev, so the VRAM branch of your Metal/VRAM split keeps that; the Metal branch is yours as written. The 72 planner tests pass on the merge. Merging once CI is green.

PR #1560, opened 2026-09-15
fix(coli): coli run on OLMoE sent the prompt in the locale's code page

kevin9327:
## Symptom

`coli run` on an OLMoE model dies before the engine starts when the prompt holds a character the console's code page cannot encode, e.g. on Korean Windows (cp949):

```
$ coli run --model olmoe_tiny_c "café"
...
  File "c/coli", line 1335, in cmd_run
    result=subprocess.run(
...
  File ".../subprocess.py", line 1143, in _stdin_write
    self.stdin.write(input)
UnicodeEncodeError: 'cp949' codec can't encode character '\xe9' in position 3: illegal multibyte sequence
```

An emoji does the same on cp949/cp932, and any non-ASCII character does it in the C locale with UTF-8 mode off. Where the code page *can* encode the text (é on cp1252, Hangul on cp949), there is no error, but the engine receives the code page's bytes instead of UTF-8 and answers a garbled prompt.

## Root cause

`cmd_run`'s OLMoE branch feeds the prompt to the engine's stdin chat loop with `input=prompt+"\n", text=True`. Text mode encodes with the locale's preferred encoding, but `olmoe.c` reads raw bytes from stdin and tokenizes them as UTF-8. The DeepSeek V4 branch of the same function already writes the prompt as UTF-8 (to a prompt file on Windows, covered by `test_windows_v4_run_passes_chinese_prompt_as_utf8_file`). The OLMoE branch was the one left in text mode.

## Fix

Pass the prompt to the engine as UTF-8 bytes: `input=(prompt+"\n").encode("utf-8")`, with no `text=True`. stdout and stderr are not captured on this path, so nothing else changes.

## Tests

- New `c/tests/test_coli_run_olmoe_prompt.py` runs the real `cmd_run` in a child Python whose locale is not UTF-8 (`PYTHONUTF8=0`, `PYTHONCOERCECLOCALE=0`, `LC_ALL=C`, or the Windows ANSI code page). It points `cmd_run` at a fake engine that records the bytes it reads on stdin, then asserts the engine got `"café — 世界 😀\n"` as UTF-8. The prompt reaches the child as ASCII-escaped JSON, so the child's own argv decoding is not what is under test.
- `test_v4_cli.test_olmoe_run_uses_its_engine_and_writes_one_prompt_line` now expects the bytes `b"hello world\n"` and no `text` argument.

```
FAIL: test_the_engine_reads_the_prompt_as_utf8 (tests.test_coli_run_olmoe_prompt.ColiRunOlmoePromptTest.test_the_engine_reads_the_prompt_as_utf8)
...
  File "C:\...\colibri\c\coli", line 1335, in cmd_run
    result=subprocess.run(
...
  File "...\subprocess.py", line 1143, in _stdin_write
    self.stdin.write(input)
UnicodeEncodeError: 'cp949' codec can't encode character '\xe9' in position 3: illegal multibyte sequence

Ran 1 test in 7.737s
FAILED (failures=1)
```

## Verification

- `tests.test_coli_run_olmoe_prompt`, `tests.test_v4_cli`, `tests.test_family_registry`: 68 tests OK, both in the cp949 locale and with `PYTHONUTF8=1`.
- End to end with the real engine and a tiny OLMoE checkpoint (`make_olmoe_tiny.py` + `convert_olmoe_merged.py` + `make_edge_tiny_tokenizer.py`, as in CI), in cp949: `coli run --model olmoe_tiny_c --ngen 4 "café 😀 한국어"` now exits 0 and the engine answers. Before the fix, the same command raised the traceback above.
- `make -C c check` (Windows, gcc 16.2, `PYTHONUTF8=1`): 1062 tests, 10 errors. All 10 are environmental on this machine, and all 65 tests in those three modules pass when rerun alone:
  - 6 in `test_cuda_test_makefile`: `make` was not on PATH; they pass with it;
  - 3 in `test_autotune` and 1 in `test_cli_output`: 5 s and 10 s subprocess timeouts while the whole suite loads the box.


JustVugg:

PR #1561, opened 2026-09-15
fix(autotune): coli tune on GLM died decoding the engine's UTF-8 output

kevin9327:
## Symptom

`coli tune` on a GLM model dies in calibration, before measuring anything, on any console whose code page cannot decode the engine's UTF-8 output (Korean or Japanese Windows: cp949/cp932; the C locale with UTF-8 mode off). Reproduced on Windows (cp949) with the `glm_tiny` fixture from `tools/make_glm_oracle.py` plus a tiny tokenizer:

```
$ coli tune --model glm_tune_model --tokens 4 --repeats 1 --timeout 120 --cap 8
  File "...\subprocess.py", line 1599, in _readerthread
    buffer.append(fh.read())
UnicodeDecodeError: 'cp949' codec can't decode byte 0x98 in position 781: illegal multibyte sequence
...
  File "c/coli", line 1255, in cmd_tune
    profile,path=run_tune(
  File "c/autotune.py", line 437, in run_tune
    replay, _ = create_replay(engine, cap, base_env, prompt, tokens, timeout)
  File "c/autotune.py", line 392, in create_replay
    output = proc.stdout + "\n" + proc.stderr
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

## Root cause

`autotune._run` captures the engine's stdout and stderr with `text=True`, so the pipes are decoded in the locale's preferred encoding. The GLM engine writes UTF-8. Every calibration and replay run sets `PROF=1`, and on a prompt of 8 or more tokens it prints `[prefill] layer 1/78 · 12 token · +0.10s` (U+00B7). The profile verdict it prints carries an em dash (U+2014). In a code page that cannot decode those bytes, the reader thread raises. `communicate()` then returns `None` for that stream, and `create_replay` fails on `None + str`. Where the code page happens to decode every byte (cp1252), there is no error, but the output is mojibake.

The non-GLM path is unaffected: it measures through `openai_server.Engine`, which already reads bytes and decodes them as UTF-8 with `replace`.

## Fix

In `_run`, decode the engine's output as UTF-8 with `errors="replace"` instead of the locale's encoding. `_run` is the only place that path captures engine output (calibration and every replay run).

## Tests

New `c/tests/test_autotune_engine_output_locale.py` drives `autotune` in a child Python whose locale is not UTF-8 (`PYTHONUTF8=0`, `PYTHONCOERCECLOCALE=0`, `LC_ALL=C`, or the Windows ANSI code page). The fake engine speaks the GLM calibration and replay lines and writes the `·` progress line and the em-dash verdict as UTF-8. The test asserts two things:
- `run_tune` on the GLM path completes;
- `_run` returns the engine's lines decoded correctly, which also fails on cp1252, where there was no crash, only mojibake.

```
FAIL: test_glm_tune_reads_utf8_engine_output_in_a_non_utf8_locale (...AutotuneEngineOutputLocaleTest...)
    self.assertIsNone(report["error"], where)
AssertionError: "TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'" is not None : child locale cp949

Ran 1 test in 22.015s
FAILED (failures=1)
```

## Verification

- `tests.test_autotune_engine_output_locale` + `tests.test_autotune`: 26 tests OK in the cp949 locale; with `tests.test_cli_output` added, 59 OK with `PYTHONUTF8=1`.
- End to end, Windows cp949, real `colibri` engine on the `glm_tiny` fixture (`tools/make_glm_oracle.py`, plus `tools/make_edge_tiny_tokenizer.py --vocab-size 256`): `coli tune --model glm_tune_model --tokens 4 --repeats 1 --timeout 120 --cap 8` now runs calibration, every candidate and both confirmation runs, and exits 0 (`✓ accepted omp-8 ...`, profile written). Before the fix, the same command raised the traceback above.
- `make -C c check` (Windows, gcc 16.2, `PYTHONUTF8=1`): 1066 tests, 1 error, in `test_makefile_platform.test_windows_nt_without_uname_selects_mingw_build`. That error is a local harness artifact, not this change. The test runs `make` with a stripped PATH, and the `make.exe` copy I put on PATH for this run could not load its DLLs (0xC0000135). Without that copy the test skips, as it does on this machine normally.


JustVugg:

PR #1562, opened 2026-09-15
fix(datapoint): fresh-process mode talked to legacy engines in the locale's code page

kevin9327:
## Symptom

`tools/datapoint.py --mode fresh-process` with a legacy engine (olmoe, inkling, colibri) dies on a console whose code page is not UTF-8 (Korean or Japanese Windows: cp949/cp932; the C locale with UTF-8 mode off). It fails in either of two ways:

- **A non-ASCII `--prompt`** raises `UnicodeEncodeError` before the engine starts. A prompt the code page *can* encode (é on cp1252) reaches the engine as non-UTF-8 bytes instead.
- **An engine line the code page cannot decode** raises in the reader thread, even with an ASCII prompt. An example is colibri's `[prefill] layer 1/78 · 12 token` progress line. `proc.stdout` comes back `None`, and the datapoint dies with `TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'`.

## Root cause

`run_fresh_engine` runs the engine with `subprocess.run(cmd, input=stdin_input, capture_output=True, text=True, env=env)`. Text mode encodes stdin and decodes stdout and stderr in the locale's preferred encoding. The engines read and write UTF-8. The persistent mode is unaffected, because it goes through `openai_server.Engine`, which uses bytes and decodes them as UTF-8.

## Fix

Use `encoding="utf-8", errors="replace"` on that call. The prompt is written as UTF-8, and the engine's output is read as UTF-8. The qwen38 branch already writes its prompt file as UTF-8. Nothing else changes.

## Tests

New `c/tests/test_datapoint_fresh_locale.py` runs `run_fresh_engine` in a child Python whose locale is not UTF-8 (`PYTHONUTF8=0`, `PYTHONCOERCECLOCALE=0`, `LC_ALL=C`, or the Windows ANSI code page). The fake legacy engine records the bytes it reads on stdin and writes the `·` progress line, the load/RSS line and an em-dash answer as UTF-8. The test asserts three things:
- the engine got `"café — 世界 😀\n"` as UTF-8;
- the load and RSS values are parsed;
- no exception is raised.

```
AssertionError: "UnicodeEncodeError: 'cp949' codec can't encode character '\\xe9' in position 3: illegal multibyte sequence" is not None : child locale cp949
FAILED (failures=1)
```

and, with the same test and an ASCII prompt (the output-decoding half alone):

```
AssertionError: "TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'" is not None : child locale cp949
FAILED (failures=1)
```

## Verification

- `tests.test_datapoint_fresh_locale` + `tests.test_datapoint`: 21 tests OK, both in the cp949 locale and with `PYTHONUTF8=1`. The stdin line ending is still the platform's (`\r\n` on Windows, as before), and the engine strips it; the test compares the prompt bytes without it.
- End to end, Windows cp949, real `olmoe` engine on the tiny OLMoE checkpoint CI builds: `tools/datapoint.py --mode fresh-process --engine ./olmoe.exe --snap olmoe_tiny_c --max-new 4 --no-evict --prompt "café 😀"` now exits 0 and prints the decode and workload tables. Before the fix it died with the `UnicodeEncodeError` above.
- `make -C c check` (Windows, gcc 16.2, `PYTHONUTF8=1`): 1066 tests, 1 error, in `test_makefile_platform.test_windows_nt_without_uname_selects_mingw_build`. That error is a local harness artifact, not this change. The test runs `make` with a stripped PATH, and the `make.exe` copy I put on PATH for this run could not load its DLLs (0xC0000135). Without that copy the test skips.


JustVugg:

PR #1563, opened 2026-09-15
fix(inkling): the dense int4 converter could not finish on Windows

kevin9327:
## Symptom

`docs/inkling.md` points a RAM-constrained Inkling user to `tools/convert_inkling_dense_int4.py`. The converter cannot finish in two situations:

**On Windows, every run** dies right after printing the plan, before writing anything:

```
      densa da convertire: 2 tensori  {'int8': 1, 'raw': 1}
      input  bf16/f32 :    0.00 GB
      output stimato   :    0.00 GB   (35%)
Traceback (most recent call last):
  File "c/tools/convert_inkling_dense_int4.py", line 206, in main
    free = os.statvfs(DIR).f_bavail * os.statvfs(DIR).f_frsize
AttributeError: module 'os' has no attribute 'statvfs'
```

**On a stdout that cannot encode "✅"** (Korean or Japanese Windows: cp949/cp932), the run gets past free space, writes the container, and then dies on the success line. It exits 1 and never prints the quantization-error report:

```
  File "c/tools/convert_inkling_dense_int4.py", line 284, in main
    print(f"\n\u2705 scritto {a.out}  {sz/1e9:.2f} GB  in {time.time()-t0:.0f}s")
UnicodeEncodeError: 'cp949' codec can't encode character '\u2705' in position 2: illegal multibyte sequence
```

In the C locale with UTF-8 mode off, the first progress line already raises (`…`).

## Root cause

- `os.statvfs` is POSIX-only.
- The progress lines print `…` and `✅` through a stdout encoded in the console's code page.

## Fix

- Measure free space with `shutil.disk_usage(DIR).free`, which is portable and returns the same figure as `f_bavail * f_frsize` on POSIX.
- Escape characters stdout cannot encode with `sys.stdout.reconfigure(errors="backslashreplace")`, the same treatment `tools/rans_verify.py` and `tools/k3_repack.py` already have. The encoding itself is left alone; a UTF-8 stdout is unaffected.

## Tests

New `c/tests/test_convert_inkling_dense_portable.py` writes a tiny shard (an embedding classified int8 and a norm passthrough) and runs the real converter in a child whose stdout is not UTF-8 (`PYTHONUTF8=0`, `PYTHONCOERCECLOCALE=0`, `LC_ALL=C`, or the Windows ANSI code page). It asserts four things:
- exit status 0;
- `dense-int4g64.safetensors` is written;
- its `__metadata__` carries `fmt=1;orig=8x64` and `fmt=raw`;
- the quantization-error report is printed.

The test is skipped without numpy, like the other converter tests.

```
AssertionError: 1 != 0 : [1/3] indicizzo 1 shard (solo header)\ufffd\ufffd
      densa da convertire: 2 tensori  {'int8': 1, 'raw': 1}
...
  File "C:\...\c\tools\convert_inkling_dense_int4.py", line 206, in main
    free = os.statvfs(DIR).f_bavail * os.statvfs(DIR).f_frsize
AttributeError: module 'os' has no attribute 'statvfs'
FAILED (failures=1)
```

With only the free-space line fixed, the same test fails on the second cause:

```
  File "...convert_inkling_dense_int4.py", line 284, in main
    print(f"\n\u2705 scritto {a.out}  {sz/1e9:.2f} GB  in {time.time()-t0:.0f}s")
UnicodeEncodeError: 'cp949' codec can't encode character '\u2705' in position 2: illegal multibyte sequence
FAILED (failures=1)
```

## Verification

- `tests.test_convert_inkling_dense_portable`: OK, both in the cp949 locale and with `PYTHONUTF8=1`.
- End to end, Windows cp949, on the tiny Inkling checkpoint from `tools/make_tiny_inkling.py` (the same generator the Inkling oracle job uses): `convert_inkling_dense_int4.py --dir tiny_inkling` now exits 0. It prints free space (`spazio libero: 123.12 GB`), writes `dense-int4g64.safetensors` (160 tensors, 48 attention int8 and 2 embed/head int8), prints the success line as `✅ scritto ...`, and ends with the quantization-error report. Before the fix, the same command died on `os.statvfs`.
- `make -C c check` (Windows, gcc 16.2, `PYTHONUTF8=1`): 1066 tests, 1 error, in `test_makefile_platform.test_windows_nt_without_uname_selects_mingw_build`. That error is a local harness artifact, not this change. The test runs `make` with a stripped PATH, and the `make.exe` copy I put on PATH for this run could not load its DLLs (0xC0000135). Without that copy the test skips.


JustVugg:

PR #1569, opened 2026-09-16
tests: skip the cuda-test recipe checks when make is absent

SurefireStudios:
## Summary

`CudaTestMakefileTest` shells out to `make` to dry-run the `cuda-test` recipe but never checks that `make` is there, so on a host without it the six recipe tests raise `FileNotFoundError` — an absent prerequisite dressed up as a regression.

Five test files reach for `make`; four guard it and this is the one that does not, so this applies `test_makefile_platform.py`'s existing idiom: `MAKE = shutil.which("make")`, `@unittest.skipUnless(MAKE, ...)`, and invoke by the resolved path.

`test_setup_openmp_probe_does_not_require_tmp` only reads `setup.sh`, so it moves to its own class rather than being skipped along with the rest.

## Validation

- [x] `make -C c check` — via this repo's own `check` and `CI` workflows run unmodified on a fork: **28 jobs green**, Linux/Windows UCRT64/macOS, sanitizers, musl, ARM and every tiny-oracle gate.
- [ ] CUDA changes were tested with `make -C c cuda-test` — n/a, no CUDA code changed.
- [x] Performance claims include hardware, commands, and repeatable measurements — n/a, no performance claim.
- [x] Performance claims include a validated experiment manifest with raw evidence — n/a.

`python -m unittest discover -s tests -p 'test_*.py'` on Windows 11 / Python 3.12 with no `make` on PATH: **before** `FAILED (errors=6)`, 819 ran / 132 skipped; **after** `OK`, 813 ran / 138 skipped.

The check that matters is the opposite one — a guard that silently disabled six build-contract assertions on CI would be worse than the errors it replaces — so I confirmed unmodified `dev` (`b5968eb`) and this branch both report `934 ran, 119 skipped` in the Linux `Python tests` job.

## Compatibility

- [x] The default CPU build remains dependency-free — test-only change; no engine or build code touched.
- [x] No model files, generated binaries, or benchmark artifacts are included.

A scan turned up two other bare-name invocations and I left both alone: `test_mw_ur3_raw_adapter.py`'s `make` call is inside an already-`@unittest.skip`ped test, and `test_supervisor_scope.py`'s `bash` sits behind `sys.platform.startswith("linux") and shutil.which("flock")`.


SurefireStudios:
Fork CI results, as promised in the description. Both of this repo's workflows, unmodified, on `SurefireStudios/colibri`.

**`check` — all green**

```
success  Linux                                     success  Windows UCRT64
success  macOS (colibri + V4 platform gate)        success  Linux (generated GLM-5.3 oracle contract)
success  Linux (generated tiny K3 vendor oracle)   success  Linux (generated tiny V4 target oracle)
```

**`CI` — all 22 jobs green**, including `Engine (Linux, CPU)`, `All engines (linux/macos/windows)`, `Sanitizers (ASan + UBSan)`, `musl libc (Alpine)`, `ARM (engines + NEON kernel exactness + tiny oracle)`, every tiny-oracle gate, `Python tests` and `Python tests (Windows focused)`.

**The check that actually matters.** A guard that quietly disabled six build-contract assertions on CI would be worse than the errors it replaces, so I ran the baseline too — the same `CI` workflow on unmodified `dev` (`b5968eb`), same runner image, and compared the Linux `Python tests` job:

| | Ran | skipped |
| --- | --- | --- |
| baseline, unmodified `dev` | 934 | 119 |
| this branch | 934 | 119 |

Identical. Where `make` exists, all six tests still execute — the skip only engages on a host that cannot run them at all.

PR #1573, opened 2026-09-16
lfs: configure .gitattributes and contributor guidance

### LINKED ISSUE #1572: [Feature]: Adopt Git LFS for test fixtures, large dataset artifacts, and media assets
### Problem

As Colibri expands support for multi-family MoE architectures, multimodal weights (e.g., GLM-5.3-Flash and DeepSeek V4.1 vision), web data files (such as `experts.json` at ~3.5MB), and test assets, managing binary artifacts directly in standard git commits introduces several challenges:

1. **Repository Bloat**: Committing large binary files or growing test datasets directly to the git tree permanently inflates the `.git` packfile history, slowing down clones and fetches for all contributors.
2. **Oracle Fixture Dependency Friction**: Oracle fixtures (like `glm_tiny`) currently require generation via `python3 c/tools/make_glm_oracle.py`, which depends on a full PyTorch environment. Distributing small pre-built golden fixtures or test shards directly in git is currently avoided due to repo size concerns, forcing contributors to install heavy external tooling just to run oracle tests.

---

### Proposed solution

Configure **Git LFS (Large File Storage)** with a `.gitattributes` specification to track current and future large binary artifacts:

1. **`.gitattributes` Configuration**: Track specific patterns for binary fixtures, sample model shards, and heavy static assets (e.g., `c/fixtures/**`, test checkpoints, and large data files).
2. **Contributor Documentation**: Update `CONTRIBUTING.md` with guidelines on:
   - Installing `git-lfs` (`git lfs install`).
   - Using `GIT_LFS_SKIP_SMUDGE=1` for contributors who only need the pure C core and want ultra-fast, lightweight clones without pulling test weights.
3. **CI Integration**: Ensure GitHub Actions workflows (`.github/workflows/`) configure `actions/checkout` with `lfs: true` (or selective `git lfs pull`) only when running oracle/test suites that require the fixtures.

---

### Alternatives considered

- **Dynamic Download Scripts**: Fetching fixtures/weights at runtime via `curl` / `wget` from Hugging Face or GitHub Releases. *Drawback:* Requires network access during test runs, depends on external URL availability, and complicates offline development.
- **On-the-fly Generation via PyTorch**: Current approach for `glm_tiny`. *Drawback:* Requires contributors to set up a heavy Python + PyTorch environment even when working exclusively on C/CUDA engine changes.
- **Direct Git Tracking**: Committing binary fixtures directly. *Drawback:* Causes permanent bloat in git history.

---

### Scope and compatibility

- **Zero-dependency CPU Path**: Completely preserved. The engine compilation and runtime in `c/` remain 100% dependency-free pure C.
- **Developer Workflow**: Users without `git-lfs` installed can still clone, build, and run the engine; they will only see text pointer files unless they opt into pulling the LFS assets.
- **Platforms**: Git LFS is standard across Linux, macOS, and Windows.

Krishna-sm:
## Summary

Configures Git LFS tracking for binary artifacts, model weights (`*.bin`, `*.safetensors`, `*.gguf`, etc.), and test fixtures via `.gitattributes`, and adds contributor guidance in `CONTRIBUTING.md`.

This resolves #1572 by:
- Adding `.gitattributes` with LFS filter definitions for large binary files and fixtures while preserving LF line ending normalization.
- Documenting how contributors working strictly on the pure C CPU engine can use `GIT_LFS_SKIP_SMUDGE=1` for lightweight, instant clones without pulling test weights.

## Validation

- [x] `make -C c check` (all C engine tests pass)
- [ ] CUDA changes were tested with `make -C c cuda-test` (N/A)
- [ ] Performance claims include hardware, commands, and repeatable measurements (N/A)
- [ ] Performance claims include a validated experiment manifest with raw evidence (N/A)

## Compatibility


JustVugg:
Thank you for turning the proposal into a patch so quickly. I am not going to merge this one, and I want to give the reasons properly because two of them are specific to what is already in the tree and one is about who pays.

**`*.svg` is text.** LFS-tracking it with `-text` turns every SVG into an opaque blob: no diffs, no review of what changed inside an icon, no line-ending normalisation. That line should come out regardless of what happens to the rest.

**The media patterns would break a clone that has no git-lfs.** The repository already carries twenty PNG, SVG and ICO files, the largest being `docs/media/colibri-atlas.png` at 896 KB, then 602 KB and 437 KB. `.gitattributes` does not migrate anything, so those stay ordinary git objects while any future commit of them becomes a pointer. That leaves a tree where some images are LFS and some are not, and a contributor who clones without `git lfs install` gets text files where the README and the site expect images. The `GIT_LFS_SKIP_SMUDGE=1` advice in CONTRIBUTING makes that outcome the documented default for anyone who wants a fast clone.

**GitHub bills LFS bandwidth to the repository owner.** On a public repository every clone that pulls an LFS object draws on that quota, and the free allowance is small. Turning it on is a standing cost that scales with how popular the project gets, which is the wrong way round.

And the premise is worth checking against the numbers: the repository is about 15 MB. There is no packfile bloat to solve today. The oracle fixtures the parent issue mentions are generated rather than committed, which is a deliberate choice, and the friction it causes is a PyTorch dependency, not a git one. If we want to remove that friction the direct fix is to ship the fixtures as a release asset or let `make` fetch them, which costs no LFS quota and no `.gitattributes` at all.

What I would take from this branch today is the line-ending half:

```
* text=auto eol=lf
*.bat text eol=crlf
*.cmd text eol=crlf
```

That is useful on its own, it has none of the above problems, and after the Windows code-page work in #1560 to #1563 it is overdue. If you want to send that as its own small PR I will merge it.

Closing this one and #1572 as not planned, with the reasoning here rather than in a label.

PR #1574, opened 2026-09-16
Add dark mode switch button

### LINKED ISSUE #1550: [Feature]:Add Dark mode switch button

Nagulanvelu:
## Summary

- Added a dark/light mode switch button to the web interface
- Added light theme styling using CSS variables
- Added theme persistence using localStorage
- Kept dark mode as the default theme
- Added Sun/Moon icons for the theme switch

## Testing

- `npm run build` passes successfully
- `git diff --check` passes
- Manually tested dark/light mode switching
- Verified theme preference persists after page refresh

Fixes #1550


JustVugg:
Thanks for this. The toggle itself is clean: no new dependencies, no external requests, and the state handling is straightforward. Two things to fix before it can go in.

**1. Dark-mode users get a white flash on every page load.**

Removing `class="dark"` from `web/index.html` means the document paints with the light theme first, and the class is only applied later, once React has mounted and the `useEffect` has run. The theme is read from `localStorage` at that point, so anyone who chose dark sees a full-page white flash on every navigation.

The fix is to decide the theme before the first paint, in a small inline script in `web/index.html`, and let React take over afterwards:

```html
<script>
  try {
    var t = localStorage.getItem('colibri-theme') === 'light' ? 'light' : 'dark';
    document.documentElement.classList.add(t);
  } catch (e) { document.documentElement.classList.add('dark'); }
</script>
```

The `useEffect` can then stay exactly as it is.

**2. The `localStorage` read in the `useState` initializer is unguarded.**

```js
const [theme, setTheme] = useState<"dark" | "light">(() => {
  return localStorage.getItem("colibri-theme") === "light" ? "light" : "dark"
})
```

Accessing `localStorage` throws, rather than returning null, when site data is blocked: a private window with third-party storage disabled, or an embedded webview. That exception happens during render, so the whole dashboard fails to mount instead of falling back to a default. Wrapping the read in try/catch, as the inline script above does, covers both.

Minor: the indentation of the added `useState` and `useEffect` blocks does not match the surrounding code.

Once the flash is gone this is good to merge.

JustVugg:
Thank you, and sorry to close this one rather than merge it.

This pull request themes the dashboard under `web/`, and that whole surface was redesigned in #1633 (merged): a new workspace layout, a navigation dock, and a brio page on top of it. The redesign brings its own light and dark theme with a switch in the sidebar, so the two changes collide in `App.tsx` and `index.css` with no sensible way to keep both.

Two things from your patch are already in, by other routes:

- the LICENSE, NOTICE and README copyright lines are on `dev` already, so that half had nothing left to apply
- the theme itself is there, and the landing page got one too in #1635, built on the mechanism @sparshsingh12312 proposed in #1551

You said on #1550 that you wanted to revamp the theme rather than duplicate it, and that is still worth doing on the new surface: the redesign ships a working switch, not a finished theme. The dashboard has no high-contrast variant, the charts in the Brain and Profiling pages still pick their colours per component instead of from tokens, and nothing remembers the choice across the desktop build. Any of those on top of the current `dev` would land cleanly.

PR #1580, opened 2026-09-17
Fix Windows Qwen36 CUDA DLL tier build

LTCjRet:
## Summary

This fixes two Windows-specific gaps that prevented qwen36.exe from using the CUDA DLL tier.

Changes

Add a Windows qwen36 alias so make qwen36 CUDA_DLL=1 reaches the intended qwen36.exe rule instead of GNU Make’s implicit C-file rule.

Compile qwen36_tier.c under CUDA_DLL=1, linking its CUDA ABI calls through backend_loader.o.

Add the missing coli_cuda_available_device_count loader typedef, optional DLL resolution, and forwarding wrapper. The wrapper loads the DLL before querying because Qwen tier device discovery occurs before coli_cuda_init().

Add a .build-config prerequisite so a prior CPU-only qwen36.exe rebuilds when CUDA-DLL configuration changes.

Add loader/header parity coverage and Windows Qwen CUDA-tier documentation.

## Validation

On Windows 11 with an RTX 4070 Ti (sm_89), CUDA 13.2, MSVC Build Tools, and MSYS2 MinGW GCC:

make qwen36 CUDA_DLL=1 ARCH=native

successfully compiled qwen36.c, qwen36_tier.c, and backend_loader.o.

The new loader-parity test passed. A real Qwen3.6-35B-A3B int4-gs64 smoke test confirmed:

[gpu] MoE experts -> CUDA VRAM tier
[qtier] resident 4889/10240 experts

The existing coli_cuda.dll exported coli_cuda_available_device_count and was reused unchanged.

## Compatibility

- Windows x86-64 with the `CUDA_DLL=1` backend: fixes the Qwen3.6 CUDA-tier build and runtime path.
- Linux `CUDA=1` and `HIP=1` build recipes are unchanged.
- No model-container or CUDA-kernel changes are included.
- The new DLL symbol is resolved optionally, so an older `coli_cuda.dll` still loads; automatic device discovery returns zero in that case, while an explicit `COLI_GPUS=0` selection remains available.


LTCjRet:
Real-hardware validation on Windows 11 Pro / Ryzen 7 7800X3D / 64 GB RAM / RTX 4070 Ti 12 GB (`sm_89`) / CUDA 13.2 / MSYS2 GCC 16.1.

Model: `Kreuzzelg/qwen36-35b-a3b-colibri-i4-gs64`  
Commit: `b690cfa`  
Prompt: identical 17-token MoE explanation prompt  
Generation limit: 64 tokens  
Cache cap: 256 per layer  
Threads: 8

| Configuration | TTFT | Decode | Peak RSS |
|---|---:|---:|---:|
| CPU-only | 4.56 s | 4.00 tok/s | 18.76 GB |
| Patched Windows CUDA DLL tier | 1.81 s | 5.92 tok/s | 37.00 GB |

CUDA startup confirmed:

```text
[gpu] MoE experts -> CUDA VRAM tier
[qtier] resident 4889/10240 experts
[qtier] VRAM hit rate: 47.7%
The CUDA run used 9.24 GB VRAM. This is a single-run validation datapoint, not a median benchmark claim; I can provide repeated cold/warm runs if useful.

**Comment by JustVugg** [comment]
Thank you for this, and the two Windows gaps you describe are real. The problem is the base, not the patch: this targets `main`, and on `dev` both of them are already closed by #1533.

Concretely, `dev` already has the `fn_available_device_count` typedef, the struct field and the `RESOLVE_OPT` line your diff adds, and the Makefile already reaches the tier under CUDA_DLL:

```make
ifneq (,$(filter 1,$(CUDA) $(CUDA_DLL) $(HIP) $(HIP_DLL)))
QWEN36_TIER_SRC = qwen36_tier.c
```

with a comment citing #1533 for the same symptom you hit. Merged as is against `dev` it would duplicate the typedef and the struct member.

Could you retarget onto `dev` and rebase? Every PR here targets `dev`; `main` only moves on release. After the rebase I expect most of this to vanish, and I would like to see what is left, because the `.build-config` prerequisite and `docs/qwen36-cuda-tier.md` look worth keeping on their own merits. Full comparison of the three PRs on this issue: https://github.com/JustVugg/colibri/issues/1577#issuecomment-5710394786

LTCjRet:
Thank you for the clarification. I’ll retarget the PR to `dev` and rebase the branch onto the current `dev` tip.

I’ll remove the portions already covered by #1533, then preserve and revalidate any remaining `.build-config`, documentation, or test improvements against `dev`. I’ll update this PR once the cleaned diff is pushed.

LTCjRet:
Rebased onto the current `dev` tip and force-pushed the cleaned branch.

The PR now retains only:
- the Windows bare `make qwen36` alias;
- the `.build-config` prerequisite so CUDA-DLL configuration changes rebuild `qwen36.exe`;
- the loader/header parity regression test;
- the Windows Qwen CUDA-tier documentation.

The parity test passes on the rebased tree. Thank you for the guidance.

JustVugg:
Merged current dev into your branch: the qwen36 rule's prerequisite list had grown on dev (decode_batch.h, idot.h), so I kept dev's line and added your alias block and the .build-config prerequisite on top. The loader part of the change was already on dev via #1533, which is why the net diff is the Makefile, the parity test and the docs. Merging once CI is green.

PR #1591, opened 2026-09-17
perf(olmoe): O(1) LRU victim selection under g_pilot_mx (#1050)

### LINKED ISSUE #1050: [perf] olmoe: O(cap) eviction scan under g_pilot_mx makes tok/s non-monotonic in cache size

Frank-zhu0404:
## Summary

Fixes the O(cap) expert-cache eviction scan under `g_pilot_mx` that made OLMoE tok/s **non-monotonic** in cache size (#1050).

Victim selection now uses an intrusive least-recent-first list of unpinned resident slots:
- **Hit path:** O(1) re-link to the MRU end
- **Miss / eviction:** O(1) read of `ev_head` (the common case that dominated mid-cap bookkeeping)
- **All-pinned fallback:** keeps the legacy O(cap) scan (rare under HOT budgets; not the measured pathology)

Pilot speculation still refuses pinned victims (`allow_pinned=0`), matching the previous pilot scan.

## Test plan

- [x] `make tests/test_olmoe_victim_index && ./tests/test_olmoe_victim_index` — basic order, oracle parity over mixed ops, hot-path fallback_scans=0
- [x] `make tests/test_olmoe_cache_index && ./tests/test_olmoe_cache_index` — existing index regression still green

Closes #1050


JustVugg:
Retargeted to `dev`: every pull request lands there, `main` only receives releases. The branch now conflicts with `dev` (the olmoe cache code moved after this was opened), so a rebase onto current `dev` is needed before it can be reviewed for merge. Once it applies cleanly I will take a look at the O(1) victim selection against #1050.

PR #1597, opened 2026-09-17
fix(serve): a tool call whose arguments are not an object answered 500 "engine failed"

kevin9327:
## What happens

A `/v1/chat/completions` request that replays an assistant turn whose tool call
carries `"arguments": "[1, 2]"` comes back as:

```
HTTP/1.1 500 Internal Server Error
{"error":{"message":"The colibri engine failed to process the request.",
          "type":"server_error","param":null,"code":"engine_error"}}
```

The engine was never asked anything. `render_chat` raised before the prompt was
built:

```
  File "c/openai_server.py", line 1612, in render_chat
    for key, value in (args or {}).items():
AttributeError: 'list' object has no attribute 'items'
```

`do_POST`'s catch-all turns that into the 500 above. OpenAI SDKs retry a 5xx,
so a malformed request is replayed as a server fault.

## Why it is a gap and not a policy

Both GLM renderers already decided what a bad `arguments` means: an `arguments`
string that does not parse becomes `{}` (the `except` branch), and the falsy
cases are covered by `(args or {})`. The only shape left uncovered is
`arguments` that parses into something truthy that is not an object --
`"[1, 2]"`, `"5"`, `'"text"'`, or an already-decoded list -- and that is the one
that crashes.

Every sibling renderer in the same file survives the identical message:

| renderer | `arguments: "[1, 2]"` |
| --- | --- |
| `render_chat` (GLM-5.2) | **500** |
| `_glm53_tool_calls` (GLM-5.3) | **500** |
| `_qwen38_tool_calls` | 200 (guards with `isinstance(args, dict)`) |
| `_k3_call_records` (Kimi) | 200 (J record fallback) |
| `render_chat_v4` / `_dsv41_tool_calls` | 200 |
| `render_chat_olmoe` / `_inkling` / `qwen36` | 200 |

`anthropic_to_openai` answers 400 for the equivalent `tool_use.input`, so
`/v1/messages` never produces this shape itself.

## The change

Complete the guard in the two renderers that were missing it, so a non-object
`arguments` takes the same path the unparseable one already takes. A tool call
with a real object argument renders byte-for-byte as before.

## Verification

```
$ python -m unittest discover -s tests -p "test_openai_server.py" -t tests -k not_an_object
# before: FAILED (errors=6) -- 5 subTests AttributeError, HTTPTest "HTTP Error 500"
# after:  Ran 2 tests ... OK
```

Full module and neighbours, on Windows (Python 3.11):

```
tests/test_openai_server.py        173 OK (baseline) -> 175 OK
tests/test_anthropic_messages.py    28 OK
tests/test_cli_output.py            34 OK
tests/test_family_registry.py       58 OK
tests/test_chat_images.py            3 OK
```

The new `HTTPTest` case drives a real `APIServer` over a socket, so the
assertion is the HTTP status the client sees, not just the renderer's return
value.


JustVugg:
Merged current dev into your branch to clear the conflict in tests/test_openai_server.py: a sibling fix for a non-object tool_choice landed next to your test, both regressions are kept. Merging once CI is green; it goes into 1.12.1.

PR #1602, opened 2026-09-18
metal: fall back to MTLCopyAllDevices() when the default device is nil

emma-assistant-1990:
## The problem

`backend_metal.mm` acquires its device with `MTLCreateSystemDefaultDevice()` and returns 0 if that is nil. That call resolves the **system default** device, which is not guaranteed to be available to a process even when usable Metal hardware is present.

When it returns nil the backend reports Metal as unavailable and the engine silently falls back to the CPU path — on exactly the hardware the Metal backend was written for. The symptoms are indirect and easy to misread as "this Mac has no supported GPU":

- `coli doctor` → `[skip] accelerator.gpu  no supported GPU detected; CPU path is available`
- `make metal-test` → `Metal unavailable (skipping)`

The second one is the more costly symptom: the Metal suite *skips* rather than fails, so CI and local runs look green while the backend is untested and unused.

## Reproduction

Apple M2 Ultra, 192 GB, macOS 14.6, `launchctl managername` = `Aqua`, WindowServer running, user on the console session:

```objc
MTLCreateSystemDefaultDevice()  // nil
MTLCopyAllDevices()             // 1 device: "Apple M2 Ultra"
```

Not a sandbox artifact — I ran the probe both inside and outside a sandbox and got nil either way, so this is the API's own behaviour rather than an entitlement issue.

## The fix

Fall back to the first enumerated device when the default is nil. Five lines, no behaviour change on machines where the default device already resolves.

## Verification

Before and after, in the same tree, same command:

```
unpatched: metal backend tests -> "Metal unavailable (skipping)"
patched:   metal backend tests -> ok
```

The unpatched run is the positive control — without it, "the tests pass" would be indistinguishable from "the tests skipped".

With the patch the full suite passes, including the bitwise `serial==parallel` top-8 contracts across the tie/denormal/`E!=256` cases and fused attention at ~5e-6 against the 1e-4 tolerance. End to end on a 744B GLM-5.2 int4 container the backend then reports **29,733 experts dispatched to the GPU with 0 CPU fallbacks**, where before it dispatched none.

`c/tests/test_backend_metal.mm` gets the same fallback, otherwise the suite would keep skipping on machines where the backend now works — which is how this stayed invisible.

## One observation, not part of this PR

While confirming the backend was live I noticed GPU scheduling dominating GPU time on this machine — attention `gpu-wall 27.74s` against `kernel 4.16s` (`gpu-sched 23.20s`), and overall `gpu-wall 57.13s` against `kernel 6.16s`. The kernels are fast; the dispatch pattern appears to be the limiter. Happy to open a separate issue with the full profile if that is useful.

https://claude.ai/code/session_01SYo9dZhniVTKtytD1sehbh


JustVugg:
dev already carries this fallback since 5b9d011a ("metal: fall back to MTLCopyAllDevices when there is no system default device"), with the same shape plus a preference for a non-low-power device on dual-GPU Macs, and the test suite's device lookup was updated with it. So the branch now conflicts with code that does what it does. Closing as superseded; thank you for the report on the M2 Ultra, it is the case that motivated the same fix.

PR #1610, opened 2026-09-18
fix: stage v41_dsml.py during installation

JohnRTitor:
## Summary

Fixes build.

Otherwise it fails to load v41_dsml module.

Updated flake lockfile while I was at it.

## Validation

- [ ] `make -C c check`

## Compatibility


JohnRTitor:
Thanks! I forgot to look at CONTRIBUTING.md.

PR #1611, opened 2026-09-18
fix(olmoe): use the shared RAM probe, add a plausibility floor, warn …

### LINKED ISSUE #1601: coli serve: RAM auto-detect returns 0.0 GB and silently falls back to a tiny expert cache (no warning)
## Summary

`coli serve` (via `openai_server.py`) sometimes fails to auto-detect available RAM and silently falls back to a very small expert-cache budget, without warning the user. On a 128 GB Apple M3 Max, starting `coli serve --model <olmoe_dir>` with no `--ram`/`--cap` flags produced:

```
[cache] 1 slots/layer of 64 experts: 2.0 GB budget (88% of what the OS still offers), 2.0 GB dense resident, ...
```

and `/health` reported:

```json
"hwinfo":{"cores":16,"ram_total_gb":0.0,"ram_avail_gb":0.0,"gpus":0,"vram_total_gb":0.0,"cpu":"unknown","gpu":""}
```

i.e. RAM detection returned 0.0 GB total/available on a machine that actually has 128 GB, and the cache sizing logic used a small fallback (2 GB) instead of anything close to the ~85 GB `coli doctor` calculates as viable for the same model on the same host.

## Impact

For a small model (OLMoE-7B, fully resident at ~7 GB with the right cache size), this silently limited decode throughput: with the auto default, effective decode throughput was well below the residency-based projection; passing explicit `--ram 96 --cap 64` (matching `coli doctor`'s own recommended plan) raised measured decode from a degraded/uncharacterized baseline to **25.8 tok/s cold / 24.9 tok/s warm** on the same hardware and model. There is no warning printed when the RAM auto-detect falls back to 0/small values, so the degradation is invisible unless the user happens to inspect `/health` or compare against `coli doctor`'s numbers.

## Repro

```bash
cd c
make olmoe
python3 tools/convert_olmoe_merged.py --repo allenai/OLMoE-1B-7B-0125-Instruct --out ./olmoe_merged
python3 ./coli doctor --model ./olmoe_merged
#  -> RAM 85.0 GB budget · ... cap 64/layer

python3 ./coli serve --model ./olmoe_merged --port 8123 &
curl -s http://127.0.0.1:8123/health
#  -> hwinfo.ram_total_gb: 0.0, hwinfo.ram_avail_gb: 0.0
#  -> serve log: "[cache] 1 slots/layer of 64 experts: 2.0 GB budget ..."

# workaround: explicit --ram/--cap fixes it
python3 ./coli serve --model ./olmoe_merged --port 8123 --ram 96 --cap 64
#  -> hwinfo populates correctly (ram_total_gb: 137.4, ram_avail_gb: 80.2 when reproduced on a second run)
#  -> cache sizes to the full residency plan, decode throughput matches doctor's projection
```

Reproduced on macOS 26.6.2, Apple M3 Max, 128 GB unified memory, colibri v1.11.0 built from source. It's possible this is specific to how `openai_server.py`'s RAM probe behaves in this environment (no swap configured at the time, `sysctl vm.swapusage` reported 0 initially) rather than a generic macOS issue, but there's no diagnostic output distinguishing "detected 0 GB, using safe fallback" from "everything is fine" — worth at least a warning/log line when the detector returns 0 or an implausibly low value so users don't silently get a degraded cache.

## Suggested fix

- Surface a warning when RAM auto-detection returns 0 (or below some sane floor) instead of silently sizing the cache off it.
- Consider having `coli serve` reuse the same RAM-detection path `coli doctor` already uses successfully, if they currently differ.

Cometbuster4969:
…on degenerate cache

olmoe.c was the only engine still carrying a Linux-only RAM probe. Its mem_available_gb() read /proc/meminfo directly and returned 0.0 on macOS/Windows, collapsing the auto cache budget to the dense resident footprint -- one expert slot per layer -- with no warning (#1601).

- mem_available_gb() now delegates to compat_mem_available_gb() (the cross-platform probe) plus a plausibility floor: values <=0 or <2% of physical RAM are treated as unmeasured, warned once, and replaced with half the physical total -- the same documented fallback as the zero case -- rather than sizing the cache off an implausible number in silence. --ram / --cap stay the explicit override.

- New helper compat_mem_total_gb() in compat.h (macOS hw.memsize / Windows ullTotalPhys / Linux MemTotal) to supply the denominator.

- The derived <1 clamp now warns and names --ram/--cap.

- serve_hwinfo() fills RAM and CPU brand from the Darwin/Windows probes when /proc is absent, so /health stops reporting 0.0 GB.

Fixes #1601.

## Summary

## Validation

- [ ] `make -C c check`

## Compatibility


KhalidAlnujaidi:
Third repro + validation on a macOS host this branch targets — and two gate findings.

## #1601: the silent collapse reproduces on `main` @ a8f2ca6 and this branch fixes it

Apple M5 Pro, 24 GB unified, macOS 26.5.1 (Darwin 25.5.1), Apple clang 21 + Homebrew libomp, arm64. OLMoE-1B-7B int8 merged container, 6.9 GB (`convert_olmoe_merged.py`). Engine built per-worktree with `make -C c olmoe`, default flags. `coli serve --model <dir>`, fresh process per arm, **no `--ram`, no `--cap`** — the exact silent path from #1601.

Baseline `main` @ a8f2ca6:
```
[cache] 1 slots/layer of 64 experts: 1.6 GB budget (88% of what the OS still offers), 1.6 GB dense resident, 1.1 GB projected KV, 6 MB per expert
```
`/health`: `"hwinfo":{"cores":18,"ram_total_gb":0.0,"ram_avail_gb":0.0,"cpu":"unknown",...}`

This branch @ 74aeb28:
```
[cache] 37 slots/layer of 64 experts: 7.3 GB budget (88% of what the OS still offers), 2.0 GB dense resident, 1.1 GB projected KV, 6 MB per expert
```
`/health`: `"hwinfo":{"cores":18,"ram_total_gb":25.8,"ram_avail_gb":6.1,"cpu":"Apple M5 Pro",...}`

Same failure signature as the reporter's M3 Max (1 slot/layer, 0.0 GB, "unknown" cpu) on a second macOS machine class, and every field the commit message claims to fix is measurably fixed. The `[cache]` line's "88% of what the OS still offers" wording is what made this invisible on our side too — it reads plausible while quoting the degenerate fallback.

One scoping note: `dev` @ HEAD already carries the engine half via 7ed6084 (`[cache] 26 slots/layer`, RAM 25.8/4.9), but still reports `cpu: "unknown"` and no plausibility warning. So the PR's remaining delta over dev on macOS is the floor+warn, the `/health` brand/RAM fill, and the #1050 eviction work.

## Gate findings (not the author's design, the branch's state)

1. `make check` fails at tip: `No rule to make target 'expert_ffn.h', needed by 'tests/test_qwen36_cache_index'`. The conflict-resolution commit 74aeb28 brought in dev's new Makefile dependency (c/Makefile:1333) but not the header itself — `c/expert_ffn.h` exists on `dev`, not on this branch. GitHub also reports `mergeable_state: dirty`. A rebase onto current `dev` should clear both.
2. Partial checklist status, for the record: the branch's targeted validation passes on macOS (`tests/test_olmoe_victim_index: ok`; `tests/test_olmoe_differential`: 5000+5000 steps OK). `make check` as a whole cannot run until (1) is fixed.

## #1050 datapoint on this host (underpowered — labeled as such)

Paired cap sweep, main@a8f2ca6 vs this branch, `PPL=1`, 12 scored tokens from the repo's `ref.json`, 3 fresh processes per arm per cap, arms interleaved within each cap block, OMP_NUM_THREADS=12:

| cap | main tok/s | PR tok/s |
|---|---|---|
| 4 | 5.86 | 5.64 |
| 8 | 5.85 | 5.74 |
| 16 | 6.00 | 6.18 |
| 32 | 4.96 | 4.82 |
| 64 | 5.24 | 5.03 |

TF-NLL bit-identical across all 30 runs and both arms: `0.5137` nats/token (string-compared; the acceptance guard). So the eviction change is output-exact here. On the perf half: within-arm spread reaches 26% (cap 4: 4.62–6.12) and the arm-to-arm gap is 2–4%, so this sweep **neither confirms nor refutes** the win — and because the branch conflates the probe fix, the eviction work, and web/UI files in one diff, it is not a one-variable comparison by the manifest rules. Reporting it as parity-not-proven rather than a result. If the #1050 part gets split out, a dev@HEAD-vs-that-branch rerun of this exact protocol is single-concern and I'll publish it properly.

Cometbuster4969:
thanks for the third repro on M5 Pro — confirms the silent collapse signature (1 slot/layer, 0.0 GB, unknown CPU) and that 74aeb28 fixes it to 37 slots/layer / 7.3GB / 25.8/6.1 / Apple M5 Pro.

Gate findings addressed:

make check failing on expert_ffn.h — fixed in fb2a194 / 591c368 (added c/expert_ffn.h from dev, now builds). Will rebase onto latest dev@2a88335 to clear dirty state (base was 5a0b725).
Targeted validation passes: test_olmoe_victim_index: ok, test_olmoe_differential: 5000+5000 OK — matches your macOS run.
On #1050 datapoint: agreed parity-not-proven due to 26% within-arm spread vs 2-4% arm gap + multi-concern diff. Happy to split this PR:

This PR #1611 → minimal #1601 only (9a04a19: compat_mem_total_gb + 2% floor + /health brand/RAM)
Separate PRs for #1050 LRU and #1550 dark mode
Let me know if you want me to push the minimal branch.

Cometbuster4969:
Rebased onto latest dev@2a88335 and pushed minimal fix — now 1 commit +91 -31, 2 files only (compat_mem_total_gb + 2% floor + /health brand/RAM). Ready for re-review. #1050 LRU and #1550 dark mode will be separate PRs.

PR #1618, opened 2026-09-19
docs: align Qwen3.8 overview with implemented vision support

bherald:
## Summary

Correct the Qwen3.8 overview's stale text-only and unimplemented-vision statements to match the implemented vision tower, gateway image path and detailed Vision section. Preserve the optional MTP limitation.

## Validation

- Source review confirms gateway image dispatch and engine IMAGE handling; diff whitespace checks pass.
- Retained September 18 validation on base `cc756a635f05ec88cff42314c285018a18801bf7`: portable CPU build and C suite passed; Python suite ran 970 tests, with 98 skipped. The serial `make check` was continued using `make -C c -j2 test`.
- Upstream compiler warnings and Python ResourceWarnings remain; the model oracle was not run.
- September 19: clean merge check against dev `5a0b72527bb7c23a542dd93a859d31381518d7d0`; the edited document is unchanged since the tested base. Tests were not rerun.

## Compatibility

Documentation only: one file, seven additions and ten deletions. No runtime, dependency or performance changes; no model files or generated artifacts.

This documentation correction does not constitute a security qualification for processing untrusted images.


JustVugg:
Thanks for looking at this page, but the doc cannot say this yet: the engine on `dev` still refuses images.

`c/qwen38.c` line 9 says "Vision and MTP tensors are indexed but never read", and the gateway path answers "this engine has no vision tower; images are not supported" (`c/qwen38.c` line 1178). What exists is the half the current doc describes: the image preprocessing in `tools/qwen38_image.py`, pinned against `Qwen2VLImageProcessor` in `tests/test_qwen38_image.py`. The tower itself is not loaded, so an `image_url` part on qwen38 is refused rather than answered.

If a vision tower for qwen38 lands, this rewrite is exactly the doc change that should come with it, in the same pull request. Until then the page is correct as it stands, so I am not merging this one.

PR #1640, opened 2026-09-20
readme: brio mode, and the dashboard as it is now (1.12.0 docs)

JustVugg:
Docs only, on top of the 1.12.0 release: the README now mentions brio mode and shows the dashboard as it is, in all four languages. Same content as #1639, merged into dev. No version change, no code change.


JustVugg:
Closing: this was meant to carry only the README update, but `dev` has since taken the post-1.12.0 contributor merges, so it would bring them to `main` ahead of a release. The README alone goes to `main` in #1642; the rest ships with 1.12.1.

Issue #1594, opened 2026-09-17
[Bug]: Qwen 3.8 Flash Next is too slow


MichaelFomenko:
### Commit or version

git clone --branch dev --single-branch https://github.com/JustVugg/colibri.git

### Environment

Debian 13, AMD 7950X, 64GB DDR5, 2TB Gen 4 Samsung Pro 990.

### Reproduction steps

COLI_FORCE_CPU=1 Q38_MAXT=190000 Q38_NATIVE_FP8=1 Q38_NATIVE_BF16=1 Q38_PREFILL_BATCH=1 PROF=1 CTX=190000 URING=1 PILOT=1 SPEC_PIN=1 TOPP=1.0 COLI_TOOL_SALVAGE=1 COLI_DEBUG=1 COLI_TIMERS=1 COLI_API_KEY=local-secret ./coli web --model /media/user/6f1e9b86-52a9-4ef6-a5fd-8965ef1db31f/LLM/Qwen3.8-Flash-Next-FP8 --ram 43 --temp 1.0 --topp 1.0 --host 127.0.0.1 --port 8000 --model-id Qwen3.8-Flash-Next-FP8

### Expected behavior

I watched this Video here: https://www.youtube.com/watch?v=_TCRH725hAM and was wondering why he is getting 20 tok/s with llama.cpp by loading only the experts to the ram but the integrated lookup table is not loaded to the ram, it is acceded directly from ssd.

### Actual behavior and logs

```shell
Qwen 3.8 Flash Next have 0.3 tok/s and Kimi K3 have 0.5 tok/s how is this possible? Qwen have only 6B Active parameter and it barely using the SSD to get there Experts. Using only 200 to 500 MB/s instead of the maximal available 3000 MB/s.
```

JustVugg:
Thanks for the detailed report, and for the exact command line. That made this quick to pin down, and the numbers tie together in a way that is worth showing, because the token rate and the disk rate are not two problems here. They are the same one.

## The arithmetic

One Qwen3.8 decode token routes ten of 512 experts in each of 48 layers, 4.7 MiB each. That is 2.2 GiB of expert weights per token with nothing cached, and roughly half that at a typical hit rate. At 0.3 tok/s you are moving about 1.1 GiB every 3.3 seconds, which is 330 MB/s. That is exactly the 200 to 500 MB/s you are seeing on the device.

[...]

PR #1646, opened 2026-09-21
qwen38: explain parallel expert-read fallbacks

Stamina9:
## Summary

Qwen3.8 can decline its parallel expert-read path and silently use serial reads. In #1594 this made a 5–6 GB/s NVMe deliver only a few hundred MB/s, while the shutdown counter merely showed `parallel-batches=0`.

Record the first fallback reason per model and print one diagnostic before continuing on the existing serial path. The message distinguishes an explicit opt-out, an oversized demand set, insufficient cache capacity, an incompatible scale bank, duplicate routes, and a non-native FP8 layout. The cache-capacity case points to the two relevant controls: lower `--ctx`/`Q38_MAXT` or increase `--ram`.

The normal one-expert case stays quiet, and successful parallel batches are unchanged.

Refs #1594.

## Validation

- [x] `git diff --check`
- [ ] `make -C c check` — the local Windows environment has no C compiler, and its WSL installation has neither `make` nor GCC. The targeted regression is included in `test_qwen38_native_weights`; full compilation remains for CI/review.
- [x] CUDA changes were tested with `make -C c cuda-test` (not applicable; CPU-side Qwen3.8 diagnostics only)
- [x] Performance claims include hardware, commands, and repeatable measurements (not applicable; no performance claim or scheduler change)
- [x] Performance claims include a validated experiment manifest with raw evidence (not applicable)

## Compatibility


JustVugg:
Read it, and the explanation is the kind the code needs. It conflicts with dev now: #1686 (merged today) rewrote q38_expert_get_batch, the load-batch array is heap-allocated and every exit frees it, and the top-k ceiling is gone. Please rebase onto current dev and re-apply your comments and the fallback lines on top of that shape; I will merge once it is clean and green. If you would rather I do the rebase, say so and I will push it to your branch.

Stamina9:
Yes, please do the rebase and push it to my branch. Once it is updated, I’ll re-apply the fallback diagnostics to the heap-allocated batch shape, remove the obsolete top-k fallback, update the tests, and verify the resulting diff.

JustVugg:
Done: merged dev into your branch and re-applied the diagnostics on the #1686 shape. The top-k reason is gone with the ceiling itself; a route wider than the per-layer cache now reports CACHE_CAPACITY (your test's TOO_MANY case checks that instead), and every exit that reports a reason frees the job array first. The message is printed once per model on stderr, so stdout and the serve protocol are untouched. test_qwen38_native_weights and test_qwen38_idot pass on the merge. Have a look; if it reads the way you intended, it goes into 1.12.1 once CI is green.

Issue #1652, opened 2026-09-21
[datapoint] Qwen3.6-35B-A3B on Tesla M10


ggeorgovassilis:
## Summary

The CUDA backend works on Maxwell (Tesla M10, `sm_50`, compute capability 5.0). This is the first validation below the `sm_80` compile-verified floor, and as far as I can tell the first Maxwell-era datapoint in the repo. The portable `quant_matmul` kernels are correct on `sm_50` (the full `cuda-test` suite passes on actual hardware) and reasonably fast where they are used. All four GPUs together (note: the M10 is a PCI card which combines 4 GPUs) hold 100 % of the 10,240 experts of Qwen3.6-35B-A3B in VRAM (zero CPU misses). End-to-end decode is 2.45 tok/s, but the bottleneck is not the GPUs: it is the dense/recurrent trunk, which `qwen36` keeps on a Sandy Bridge CPU with no AVX2/FMA. For reference, the same hardware does 7.3 tok/s under llama.cpp (`-ngl 99 -ts 1,1,1,1`) when the whole model is offloaded.

## Environment

- GPU: 4x Tesla M10 (Maxwell, `sm_50`, no tensor cores), 8 GB VRAM each (32 GB combined), driver 580.173.02 (host CUDA 13.0)
- CPU: 2x Intel Xeon E5-2640 @ 2.50 GHz (Sandy Bridge), 12 physical cores / 24 threads; `avx`, `sse4_1`, `sse4_2` present, `avx2`, `fma`, `avx-512` absent
- RAM: 128 GB (rebalanced 64 GB / 64 GB across the two NUMA nodes)
- Storage: NVMe SSD
- OS: Ubuntu 24.04.2 LTS (Noble Numbat), kernel 6.8.0-138-generic
- Docker: 29.7.2 (client and server), with the nvidia container runtime (nvidia-container-toolkit 1.20.0). All runs were made in containers; the GPU runs used `--gpus "device=..."` to expose the devices, the CPU run used none.
- Toolchain: CUDA 11.8 (the last toolkit that still targets `sm_50`; Maxwell was dropped in CUDA 12.0), inside `nvidia/cuda:11.8.0-devel-ubuntu22.04`
- Commit: `9d5d05d` (`v1.12.0-4-g9d5d05d`, `main`)
[...]

PR #1657, opened 2026-09-21
qwen36: offer the rest of the dense trunk to the VRAM placer (dnout, attnproj, shexp)

JustVugg:
Follows the M10 datapoint in #1652, which showed the ceiling on a CPU without AVX2: with every expert VRAM-resident, DeltaNet 169 ms and attention 55 ms of a 384 ms token stayed on the CPU, because `qwen36` offered the placer only `lm_head` and the DeltaNet input projections.

## Where the time is

Measured on a 16-core CPU (ds, 8 threads), the DeltaNet stack per decoded token: **37.5 ms = input projections 23.4 | out_proj and norm 8.3 | convolution 3.3 | recurrence 2.4**. The matmuls are the cost, not the recurrence, so the trunk is what goes to VRAM, and the tier already knew how: the placer prices any component by name, and `qt_dense_init/qt_dense_matmul` are what Qwen3.8 runs its whole trunk on. `qwen36.c` simply never offered these.

## What is offered now

| component | what | per layer (35B) | total |
|---|---|---|---|
| `dnout` | DeltaNet out_proj | 8.4 MB | 252 MB |
| `attnproj` | q, k, v, o of an attention layer, one item | 27.3 MB | 273 MB |
| `shexp` | the shared expert's gate, up, down, every layer | 3.1 MB | 126 MB |

Offer order after `lmhead` and `dnproj`, each in layer order, so a partial placement is whole layers (#1361). A component is offered only when every one of its matrices has a dense-i8 copy. On decode a placed matrix answers from VRAM with one GEMV; a prompt batch keeps the batched CPU matmul. Every matrix falls back on its own: a failing handle is turned off by the tier and `matmul_d` takes over, so the CPU path is exactly the one that runs today when nothing is placed or the build has no CUDA. `COLI_PLACE` accepts the three new names like the old ones (`dnout=0,attnproj=0,shexp=0`, per layer, `+` splits).

## Verified without a GPU

- `tests/test_qwen36_trunk_dense.c` on the fake CUDA backend (in `make check`): the offers carry the right names, layers and bytes; a component with one matrix missing its dense-i8 copy is not offered; eight matrices placed and uploaded once each; every GEMV from VRAM equals `matmul_d` on the CPU (relative gap <= 2e-7).
- CPU path: qwen36 exact on the tiny fixture and byte-identical in chat (ds); decode timings unchanged (36.97 vs 37.5 ms DeltaNet per token, noise); `make check` 1,292 tests green.

## What needs a GPU

The measurement. Same command as today, the placement is automatic:

```sh
COLI_CUDA=1 COLI_GPUS=0,1,2,3 CUDA_EXPERT_GB=auto COLI_TIMERS=1 ./qwen36 256 4 prompt.txt
```

The `[place]` and `[dense]` lines at startup say what landed; the `[timers]` lines give the per-phase split; `COLI_PLACE=off` is the A/B control, and the output must stay byte-identical between the two. Asking the reporter of #1652 (4x M10) and anyone with an 8 GB card to run it before this is merged.


JustVugg:
@kreuzzelg if the 3070 is free: this is the single-card case the tier doc's calibration table was built on (auto placed lm_head + dnproj there, +49% warm). On this branch the placer also gets `dnout`, `attnproj` and `shexp`, about 650 MB more of int8; on an 8 GB card it will displace some experts, and the pricing rule decides. The two numbers worth having are warm tok/s with `auto` against `COLI_PLACE=off` on the same build, plus the `[place]` lines, with byte-identical output between the two. No hurry, and thank you.

PR #1667, opened 2026-09-22
qwen36: streaming CUDA expert tier for cap < n_experts

EthanJ79:
## What

The qwen36 CUDA expert tier required full RAM residency (`cap == n_experts`): it kept raw pointers into RAM slots that were never recycled. On one 8 GB card that is ~32 GB of int8 slots before the first token, so a 32 GB box could not use its GPU for Qwen3.6 at all (the tier refused, or the warmstart ran the machine out of memory).

This PR reuses the **streaming lifecycle Qwen3.8's fp8 mode already had** for qwen36's own int4/int8 formats. With `cap < n_experts` the RAM slots are an LRU, experts stream from disk, and the tier copies each upload at the moment the bytes pass by, so VRAM, RAM, disk and CPU all work in one run.

## How

- **`qwen36_tier.c/h`**: `G_fp8_stream` meant two things. It is now split into `G_stream` (lifecycle: copy at note, keep no slot pointer, promote when the bytes pass by) and the fp8 **format** flag. New `qt_init_stream()`, `qt_streaming()`, and `qt_touch()` (heat for a routed resident the engine did not load). `qt_init_fp8` behaves exactly as before.
- **`qwen36.c`**
  - `cap < n_experts` starts the tier in streaming mode; `cap == n_experts` is unchanged.
  - A VRAM-resident expert is not read into the RAM LRU at all. If a swap evicts it before `qt_issue`, the miss loop loads it, so the result never depends on that shortcut.
  - The streaming warmstart loads only the planned VRAM set, serially (a slot is valid only until the next `expert_get` of its layer), instead of every expert.
  - CPU misses under the streaming tier run on the shared int4 kernel (`xf_moe_run`, with `xf_act_mode()` like the CPU path) instead of the int8 unpack path. The slot keeps the pair-layout int4 the tier uploads next to the planar copy, as three allocations because every teardown frees `g4`/`u4`/`d4` one by one.
- **`tests/test_qwen36_tier_stream.c`** (fake backend, auto-registered by its rule) checks that:
  - `qt_init` still refuses `cap < n_experts`, and only `qt_init_stream` accepts it;
  - the upload is fmt 4 with group scales, XOR-staged;
  - no slot pointer survives `qt_note` or `qt_note_planned`;
  - `qt_touch` keeps routed residents from being the swap victim;
  - streaming does not leak into a later full-residency init.
- **Docs**: `docs/qwen36-cuda-tier.md` describes both modes; manifest in `docs/experiments/qwen36-streaming-tier.json`.

## Measurement

Measured on an RTX 5060 Laptop (8 GB, `CUDA_EXPERT_GB=6`) with a Ryzen 9 8940HX, 32 GB RAM, and a KIOXIA NVMe drive. Settings: gs64 container, `cap 16`, 64 greedy tokens, no `HEAT_FILE`, one warm-up per arm, runs interleaved. Only `COLI_CUDA` changes between arms (manifest validated with `c/experiment_manifest.py`).

| arm | decode tok/s (3 runs) | median | VRAM hit | peak RSS |
|---|---|---|---|---|
| CPU (`COLI_CUDA=0`) | 3.866, 3.707, 3.616 | 3.707 | – | 11.0 GB |
| streaming tier (`COLI_CUDA=1`) | 5.128, 5.116, 4.137 | **5.116 (+38 %)** | 68–73 % | 11.0 GB |

A laptop, so thermals add noise (third GPU run). With a `HEAT_FILE` from an earlier run the hit rate reached 91.7 % on the prompt it was learned on.

## Exactness (please read)

- With f32 activations on both sides (`QWEN_EXPERT_ACT=f32 COLI_DENSE_IDOT=0`), the generated tokens are **byte-identical to the CPU path** on two prompts (Korean prose, English code).
- Under the **default int8 activations** (since dfec3a4) they are **not identical**, and they vary from run to run. The GPU expert kernel takes f32 activations while the CPU quantizes them to int8, so the output depends on which experts happen to be resident, and that follows the asynchronous upload timing.
- Full residency has the same property; streaming only makes the resident set move more.
- An earlier revision of this branch passed mode 0 to the miss path. That mixed the two numerics inside the CPU side too; fixed in 7588452.

Open question for review: should the tier force f32 expert activations while it is on, so that it is exact again, or should the GPU get an int8-activation expert kernel?

## Checks

- `make check`: 0 warnings, C suite all passed, Python suite OK.
- New `test_qwen36_tier_stream` and the existing `test_qwen36_tier_fp8` pass.
- Windows CUDA build (`qwen36.exe CUDA_DLL=1`, MinGW cross) has 0 warnings.
- `make -C c cuda-test` was **not** run: no CUDA-capable Linux host here. The real-GPU validation above is Windows only.


EthanJ79:
Opened by mistake, closing.

Issue #1306, opened 2026-08-31
Qwen3.8-Flash-Next: is a GPU backend (CUDA/HIP) planned for the qwen38 engine? (offer: gfx1151 / ROCm 7.2.4 validation + benchmarks)


simongonzalezdc:
Thanks for colibri — running frontier MoE on hardware you already own is exactly the right idea, and v1.10.1 runs Flash-Next here on day one.

**Setup:** Qwen3.8-Flash-Next-FP8 (official checkpoint), colibri v1.10.1 linux-x86_64 release, Strix Halo (Ryzen AI Max 395, gfx1151 iGPU, 96 GB unified), ROCm 7.2.4, model on external storage.

**Measured** (per the benchmark bar in CONTRIBUTING):
- `coli serve --ram 20 --auto-tier --policy experimental-fast`
- cold: 49 completion tokens in 248.6 s (~0.20 tok/s); warm: 47 tokens in 215.3 s (~0.22 tok/s); single request, n=1 each (smoke-scale, not a benchmark claim)
- resident matrices 10.88 GiB, load 15.3 s, answers correct, doctor reports "GPU detected but the engine is CPU-only" and "RAM budget cannot hold one expert slot per sparse layer"
- the box co-runs other resident services, so the RAM budget above is what was safe to give — expert residency is effectively 0% and decode is disk-bound, which matches the doctor's placement warning exactly

**Question:** `c/Makefile` builds the qwen38 engine with `NOCUDA_CFLAGS` (the CUDA/HIP backend is wired for the glm/qwen36/dsv4 families), so on this class of APU the Flash-Next path is CPU-only today. Is a GPU tier for qwen38 on the roadmap?

**Offer:** we have exactly the rig the HIP path was tested on class-wise (gfx1151, ROCm 7.2.4 — very close to the doc's tested 7.2), a working HIP toolchain, and we're set up to run reproducible measurements (commit, exact commands, hardware, warm-up policy, run count, median). Happy to validate a qwen38 HIP/CUDA build, file structured reports, or help however is useful. No pressure either way — the CPU multitier path alone is already impressive.

JustVugg:
Thanks for the measured numbers and the offer, both are useful. Your read is right: `qwen38` is CPU-only today, and on your box the 0.2 tok/s is the RAM budget (20 GB against a 10.9 GiB resident trunk leaves no expert slots), not the GPU.

[...]

PR #1686, opened 2026-09-22
qwen38: CPU prefill batching — runtime chunk/workspace knobs, wider expert load batch, parallel QSA prefill

DebugSultan:
Requested by @kreuzzelg in #1306: *"If you PR the prefill batching, I would expect it to be the more valuable of the two pieces for everyone on upstream, and the trunk prefill then becomes the long-prompt complement."* This is that piece, and only that piece.

Three commits on `c/qwen38_core.h`. No new files, no new dependency, **no default changed**.

## What is in it

**1. `8745651` — make the prefill chunk and its workspace runtime knobs.**
`Q38_PREFILL_BATCH_ROWS` and `Q38_PREFILL_WORKSPACE_MIB` now read from the environment, falling back to today's `Q38_PREFILL_BATCH_ROWS` (32) and `Q38_PREFILL_WORKSPACE_BYTES` (64 MiB). The workspace knob is in MiB because the byte count is the thing a human gets wrong. It is orthogonal to the existing `Q38_PREFILL_BATCH` on/off switch — same prefix, different question. On upstream's defaults this commit is a no-op; it only makes the two constants reachable without a rebuild.

**2. `9c0c6bc` — lift the top-k ceiling off the expert load batch.**
`q38_expert_get_batch` capped its job array at `Q38_MAX_TOPK` because the array was fixed-size on the stack; `q38_moe_prefill` then clamped `load_limit` to match. With a wider chunk the batch wants more than top-k experts at once, so the array becomes a `malloc` sized by the actual count and the clamp goes away. The head predicate now compares against the cache capacity rather than the old ceiling. Every bail-out in that function frees the array — the original had no allocation to leak, the new one does, so all four exits and the success path release it.

**3. `c7f5b7f` — run QSA per position at prefill, and time it by the clock.**
The K/V/IK write loop and the ranking/attention loop over the `S` prompt positions are independent per position; both get `#pragma omp parallel for … if(S>1)`, so decode (`S == 1`) stays exactly serial and bit-identical. The ranking scratch (`qidx`, `pool`, `selected`) moves inside the loop so each thread owns its own. The two phase timers were summing per-thread CPU time and reporting it as elapsed — with 20 threads the phase sum read 510 s against a 268 s TTFT. They are now rescaled to the measured wall time of the loop, keeping the index/attention ratio the threads observed.

## The numbers, and whose they are

**I measured nothing for this PR.** The rig was out of window for this work, so everything below is already-published data from #1306, not a fresh run.

@kreuzzelg's RTX 3070 (sm_86) comparison, 315-token prompt, cap 224, upstream `dev` + #1552 against my fork, both with the GPU trunk prefill **off** — i.e. this is the CPU prefill path alone:

| prefill phase, trunk prefill off | upstream | fork |
|---|---:|---:|
| TTFT | 178.6 s | 58.1 s |
| resident-mm | 76.4 s | 6.4 s |
| DeltaNet projections | 50.7 s | 7.8 s |
| routed experts | ~50 s | 11.6 s |

His reading of what produces it: *"That is not the GPU, it is your configurable prefill batch rows (upstream fixes 32), the larger workspace and the parallelised attention prefill."*

Being precise about what this PR can and cannot claim from that table:

- **resident-mm and DeltaNet** are the phases these three commits act on, through the wider chunk and the parallel prefill loop. That is where the 3× lives.
- **routed experts** is partly commit 2 (a wider chunk means fewer, larger expert load batches) and partly the fork's int4 gs64 expert container, which is **not** in this PR. I cannot split that line without a run and I have not made one, so treat it as unattributed.
- The **−38 % TTFT** in his other column belongs to `Q38_TRUNK_PREFILL` (the GPU batched prefill), a different change he says he already carries as a commit on top of #1552. Not this PR.

So: the honest claim is that the CPU prefill phases this touches are the ones that went 76.4 → 6.4 s and 50.7 → 7.8 s on his card, and that the knobs are what let a rig pick a chunk that fits it. The right default for `Q38_PREFILL_BATCH_ROWS` on upstream hardware is an open question — this PR deliberately does not answer it, it just stops requiring a rebuild to ask.

## Correctness, and what the greens do not prove

Run here on this branch:

- `make qwen38` — clean under `-Wall -Wextra`, zero warnings.
- `make qwen38-tiny-check` — **PASS: 16, FAIL: 0**, `Matching tokens: 8/8` in all 16 configurations (`trunk` 1024/0 × `batch` 0/1 × `bf16` 0/1 × `cap` 1/4, `OMP_NUM_THREADS=2`). The fixture's 5-token prompt and its two `qwen_sparse_attention` layers mean the `S > 1` path this PR modifies is genuinely executed, not skipped.
- `make qwen38-ple-prefetch-check` — identical tokens with `Q38_PLE_PREFETCH` on and off, 16 configurations.
- Thread sweep 1/2/4/8/16 — identical tokens *and* identical oracle values (cos 0.99998958, max_abs 0.000840738), which is the bit-identity the parallel loop has to preserve.

**What that does not establish.** I reinserted the race on purpose — shared ranking scratch across positions, a data race by construction — and the fixture stayed **green**. So the tiny fixture is a numeric oracle, not a race detector, and none of the greens above are evidence that the loop is free of races; they are evidence that the results are unchanged. I also ran ThreadSanitizer and it is not usable as evidence here: 268 warnings with the bug against 267 without, the same QSA lines in both, i.e. noise from uninstrumented libgomp. The argument for the parallel loop is the loop body itself — per-position scratch, per-position writes, reduction on the two timers — not a test that caught something.

## Not in this PR

The int4 gs64 expert container and the pinned arena, `Q38_TRUNK_PREFILL`, and the 20-thread default (`ffde373` on my fork) — that last one is a separate, smaller offer if you want it.

Happy to reshape any of the three; they are independent and can be taken one at a time.


JustVugg:
Measured on the released Qwen3.8-Flash-Next-FP8, this PR on top of dev d99dc069 (which already carries the int8 trunk and the vector FP8 kernel of #1668), 16-core server shared with a training job, `OMP_NUM_THREADS=8`, RAM LRU cap 96, the 512-token chunk 0 of our perplexity set as the prompt, 20 generated tokens, `COLI_TIMERS=1`:

| | dev | this PR, defaults | `Q38_PREFILL_BATCH_ROWS=128` | `Q38_PREFILL_BATCH_ROWS=512`, workspace 1024 MiB |
|---|---:|---:|---:|---:|
| wall, 512 prompt tokens + 20 | 164.7 s | 163.1 s | 164.4 s | 162.6 s |
| expert-read | 26.1 s | 25.6 s | 26.1 s | 26.2 s |
| routed-expert | 36.1 s | 34.5 s | 34.6 s | 35.7 s |
| resident-mm | 11.7 s | 11.9 s | 11.8 s | 11.7 s |
| deltanet | 15.7 s | 16.3 s | 16.3 s | 15.3 s |
| qsa-attn | 7.9 s | 7.8 s | 7.7 s | 7.8 s |
| md5 of the output | c392be23 | c392be23 | c392be23 | c392be23 |

Output byte-identical in every configuration, as the PR says. No measurable speed change on this machine at any chunk width: with the trunk already int8 the prefill here is expert reads and expert GEMVs, and the QSA phase reads the same 7.8 s with the parallel loop as with the serial one (the timers are wall clock now, so that is a real comparison). The 3070 numbers in the description were taken against a BF16 trunk, which is no longer what dev runs.

Merging it as what it is: two knobs that no longer need a rebuild, a lifted load-batch ceiling, and a parallel loop that keeps the tokens. Not as a speedup, because none was measured here.

